PR #5, opened 2017-04-03
Adding namespace support

StephenSorriaux:
Hello,

I noticed that the namespace was always specified to "default", I just added some code to make it possible to use other namespace.

Namespace can be configured using the .pgo.yaml file or using the flag --namespace. A postgre-operator container is need for each namespace.


jmccormick2001:
looks great, thanks!

PR #18, opened 2017-06-21
Removed dash so PV name matches correctly

vladimirvivien:
The `create-pv.sh` uses name format `crunchy-pv$i`.  This fixes the `delete-pv.sh` to match that name format.


jmccormick2001:
thanks!

PR #24, opened 2017-06-28
Update docker image tag to match correct version

vladimirvivien:
This is to update the `examples/operator/deployment.json` to match the version of the operator image.


jmccormick2001:
just fixed it to substitute that image tag with the CO_IMAGE_TAG env var value instead of hard coding the image tag.

Issue #108, opened 2018-01-25
Destructive commands should prompt for confirmation


blakebarnett:
It's currently extremely easy to blow away all backups or clusters with a single `pgo delete cluster all` command, which has no prompt for confirmation, I'd recommend a `y/N` prompt with no as default. Preferably all destructive commands would do this, with the ability to avoid the prompt with a `--force-yes` option or something similar.

PR #109, opened 2018-01-25
This will add a prompt for all deletes

blakebarnett:
```
$ pgo delete cluster pgo-test2
WARNING - This is destructive: Are you sure? (yes/no):
```

It forces a response if enter hit without a valid one.

fixes https://github.com/CrunchyData/postgres-operator/issues/108


jmccormick2001:
thanks for the contribution!

PR #135, opened 2018-03-11
use <start year>-<end year> in copyright dates

davecramer:


jmccormick2001:
got these in the master finally from another branch.  but thanks!

PR #212, opened 2018-05-09
Fix Makefile to use consistent tags for Docker images

gurjeet:


gurjeet:
It looks like I had accidentally deleted the branch. I have restored it, and opened the PR.

Issue #226, opened 2018-05-20
Postgres operator 3.0 setup fails with CrashLoopBackOff - k8s 1.10.2


ieugen:
Hello, 

I've installed postgresql operator on my bare-metal cluster setup with kubeadm. 
I've installed postgres-operator with helm: 

1. cloned the repository 5779fd9e2d83d6c16d80f66b3190996670bd6dad
2. installed via helm chart https://crunchydata.github.io/postgres-operator/installation/helm-chart/  . I used the command from the chart README.md: `helm install --name postgres-operator --namespace=postgres-operator postgres-operator` 
3. manually installed missing crd resources with `kubectl creat -f $project/deploy/crd.yaml` as operator container was failing but I got into another issue. 

Checking the status yields:
````
postgres-operator   postgres-operator-788d674444-z8b4s         1/2       CrashLoopBackOff
````

The issue is with "operator" container: 

````
time="2018-05-20T21:40:34Z" level=debug msg="debug flag set to true"
time="2018-05-20T21:40:34Z" level=debug msg="setting NAMESPACE to postgres-operator"
time="2018-05-20T21:40:34Z" level=debug msg="CCP_IMAGE_PREFIX set, using crunchydata"
time="2018-05-20T21:40:34Z" level=debug msg="CO_IMAGE_PREFIX set, using crunchydata"
at end of setup, beginning wait...Watch Pgbackup objects
Watch Pgingestcluster objects
Watch Pgreplica objects
Watch Pgcluster objects
Watch Pgupgrade objects
time="2018-05-20T21:40:34Z" level=info msg="Watch pod objects"
time="2018-05-20T21:40:34Z" level=info msg="Watch job objects"
time="2018-05-20T21:40:34Z" level=info msg="Watch Pgtask objects"
Watch Pgpolicy objects
time="2018-05-20T21:40:34Z" level=info msg="MajorUpgradeProcess watch starting..."
E0520 21:40:35.744735       1 reflector.go:205] github.com/crunchydata/postgres-operator/controller/podcontroller.go:76: Failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:postgres-operator:default" cannot list pods in the namespace "postgres-operator"
time="2018-05-20T21:40:36Z" level=error msg="error watching upgrade jobunknown (get jobs.batch)"
````

jmccormick2001:
[...]

Issue #219, opened 2018-05-18
CRDs not created when installing with Helm chart


petvaa01:
Installing version 3.0 with Helm chart does not create CRDs on the first run.  Version 2.6 works as expected.

Logs:
```
postgres-operator-7fcf6f77c4-57xdb operator time="2018-05-18T11:52:18Z" level=debug msg="debug flag set to true"
postgres-operator-7fcf6f77c4-57xdb operator time="2018-05-18T11:52:18Z" level=debug msg="setting NAMESPACE to default"
postgres-operator-7fcf6f77c4-57xdb operator time="2018-05-18T11:52:18Z" level=debug msg="CCP_IMAGE_PREFIX set, using crunchydata"
postgres-operator-7fcf6f77c4-57xdb operator time="2018-05-18T11:52:18Z" level=debug msg="CO_IMAGE_PREFIX set, using crunchydata"
postgres-operator-7fcf6f77c4-57xdb operator at end of setup, beginning wait...Watch Pgbackup objects
postgres-operator-7fcf6f77c4-57xdb operator time="2018-05-18T11:52:18Z" level=info msg="Watch pod objects"
postgres-operator-7fcf6f77c4-57xdb operator time="2018-05-18T11:52:18Z" level=info msg="MajorUpgradeProcess watch starting..."
postgres-operator-7fcf6f77c4-57xdb operator Watch Pgpolicy objects
postgres-operator-7fcf6f77c4-57xdb operator Watch Pgingestcluster objects
postgres-operator-7fcf6f77c4-57xdb operator Watch Pgupgrade objects
postgres-operator-7fcf6f77c4-57xdb operator time="2018-05-18T11:52:18Z" level=info msg="Watch job objects"
postgres-operator-7fcf6f77c4-57xdb operator Watch Pgcluster objects
postgres-operator-7fcf6f77c4-57xdb operator Watch Pgreplica objects
postgres-operator-7fcf6f77c4-57xdb operator time="2018-05-18T11:52:18Z" level=info msg="Watch Pgtask objects"
postgres-operator-7fcf6f77c4-57xdb operator E0518 11:52:18.770208       1 reflector.go:205] github.com/crunchydata/postgres-operator/controller/replicacontroller.go:82: Failed to list *v1.Pgreplica: the server could not find the requested resource (get pgreplicas.cr.client-go.k8s.io)
[...]

PR #227, opened 2018-05-20
Added roles and bindings to chart

ieugen:
* updated chart to create roles and bindings into installation namespace
* does not yet fix #226

PR #228, opened 2018-05-21
[FIX] helm chart for 3.0

blaggacao:
An attempt, in relation with https://github.com/CrunchyData/postgres-operator/issues/219 and https://github.com/CrunchyData/postgres-operator/issues/226
Alternative to: https://github.com/CrunchyData/postgres-operator/pull/227

Closes: #219 
Closes: #226


ieugen:
Hi,

Seems like we are fixing the same thing :). 

I've tried your patch and it still fails. 

````
 reflector.go:205] github.com/crunchydata/postgres-operator/controller/upgradecontroller.go:82: Failed to list *v1.Pgupgrade: pgupgrades.cr.client-go.k8s.io is forbidden: User "system:serviceaccount:postgres-operator:default" cannot list pgupgrades.cr.client-go.k8s.io in the namespace "postgres-operator"
````

**Comment by ieugen** [comment]
I've closed my issue and PR so we can focus on this one. 

@jmccormick2001 : It seems that the CO_NAMESPACE variable to the right namespace as shown in logs: 
````
time="2018-05-21T06:29:22Z" level=debug msg="debug flag set to true"
>>>> time="2018-05-21T06:29:22Z" level=debug msg="setting NAMESPACE to postgres-operator" <<<<<<< 
time="2018-05-21T06:29:22Z" level=debug msg="CCP_IMAGE_PREFIX set, using crunchydata"
time="2018-05-21T06:29:22Z" level=debug msg="CO_IMAGE_PREFIX set, using crunchydata"
time="2018-05-21T06:29:22Z" level=info msg="Watch Pgtask objects"
time="2018-05-21T06:29:22Z" level=info msg="MajorUpgradeProcess watch starting..."
Watch Pgcluster objects
Watch Pgbackup objects
time="2018-05-21T06:29:22Z" level=info msg="Watch pod objects"
time="2018-05-21T06:29:22Z" level=info msg="Watch job objects"
Watch Pgingestcluster objects
at end of setup, beginning wait...Watch Pgreplica objects
Watch Pgupgrade objects
Watch Pgpolicy objects
time="2018-05-21T06:29:22Z" level=error msg="error watching upgrade jobunknown (get jobs.batch)"
E0521 06:29:22.946741       1 reflector.go:205] github.com/crunchydata/postgres-operator/controller/upgradecontroller.go:82: Failed to list *v1.Pgupgrade: pgupgrades.cr.client-go.k8s.io is forbidden: User "system:serviceaccount:postgres-operator:default" cannot list pgupgrades.cr.client-go.k8s.io in the namespace "postgres-operator"
````

**Comment by blaggacao** [comment]
As far as my search resulted, `CO_NAMESPACE` is not present throughout the whole charts folder (before and after this PR), so I don't see how this could possibly be relevant.

**Comment by blaggacao** [comment]
It's simple `User "system:serviceaccount:postgres-operator:default"` , as per the updated `values.yaml` in this PR needs to become `User "system:serviceaccount:postgres-operator:postgres-operator-cluster"`.

But I could not figure out, where the `default` comes from... Any ideas?

**Comment by ieugen** [comment]
@blaggacao I believe it might be a misunderstanding. The script deployment method uses `CO_NAMESPACE` environment variable to populate some values https://github.com/CrunchyData/postgres-operator/blob/5779fd9e2d83d6c16d80f66b3190996670bd6dad/deploy/rbac.yaml#L65 

Helm charts are "real" templates and they have more advanced templating at their disposal. They don't need to use `CO_NAMESPACE` variable name. They just pass the namespace name, which is the goal. 

My guess right now is that maybe `NAMESPACE` environment variable passed to the operator container should be actualy `CO_NAMESPACE` ? https://github.com/CrunchyData/postgres-operator/blob/master/chart/postgres-operator/templates/deployment.yaml#L29 

Maybe @jmccormick2001 can explain a bit.

File: chart/postgres-operator/values.yaml [hunk 1]
```diff
@@ -21,3 +21,12 @@ resources:
   requests:
     cpu: 100m
     memory: 128Mi
+
+serviceAccount:
+  create: "cluster"
```
blaggacao: Shit!

blaggacao: Will be `name: "cluster"` within a few seconds...

chart/postgres-operator/values.yaml
```diff
@@ -21,3 +21,12 @@ resources:
...
     memory: 128Mi
+
+serviceAccount:
+  name: "cluster"
+  create: "true"
+
+rbac:
+  create: "true"
+
+nameOverride: "pgo"
```
ieugen: What is the purpose of `nameOverride` variable? 
If it is supposed to be the install name, then you don't need it as you can supply it via `helm install --name aaaaa` 
.

chart/postgres-operator/templates/service-account.yaml
```diff
@@ -0,0 +1,12 @@
+{{ if .Values.serviceAccount.create }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "serviceAccountName" . }}
```
ieugen: Do we need the service account name to be customisable independenty? 
Can't we use the name of the chart instance from `helm install --name my-name-aaa` . If the name is not supplied, helm generates one automatically.

chart/postgres-operator/values.yaml
blaggacao: `--name` is the release name, `nameOverride` is appended as the package-name, so a fullname would become:
`--name=fishnet` -> `fishnet-pgo` It's inspired by some pro helm-charts repo (guess sapcc)

jmccormick2001:
let me know when this is ready to test/merge and I'll take a look but I do want the file names to match between chart/postgres-operator/files/postgres-operator/cluster/1 and conf/postgres-operator/cluster/1  

in 3.0 I merged the 3 'cluster' json templates into a single one, so they will need to match over in the chart.

chart/postgres-operator/templates/service-account.yaml
blaggacao: That's the idea of the `serviceAccountName` helper does, basically. I'm improving it atm, so that it anloy asks for creation and uses `default` otherwise. So no custom name anymore...

blaggacao: @ieugen Could you have a look again at the helper, I just fixedup-rebased...

blaggacao:
@jmccormick2001 I can add that to the PR stack.

chart/postgres-operator/values.yaml
ieugen: I don't know that tip and maybe it has some merrit. All I know is: remove all things that you don't have a use case for :) .

chart/postgres-operator/templates/service-account.yaml
ieugen: I've tried the cluster again and it did not start, same issue it seems: 
````
github.com/crunchydata/postgres-operator/controller/clustercontroller.go:82: Failed to list *v1.Pgcluster: pgclusters.cr.client-go.k8s.io is forbidden: User "system:serviceaccount:postgres-operator:default" cannot list pgclusters.cr.client-go.k8s.io in the namespace "postgres-operator"
````

**Comment by blaggacao** [comment]
- [ ] affinity.json
- [ ] cluster-deployment-1.json
- [ ] cluster-service-1.json
- [ ] cluster-upgrade-job-1.json
- [ ] collect.json
- [ ] container-resources.json
- [ ] pgpool-template.json
- [ ] pgpool.conf
- [ ] pool_hba.conf
- [ ] pool_passwd

**Comment by blaggacao** [comment]
@jmccormick2001 Some yamls have mixed tab + 4 spaces, would you accept a linting commit or is this indentation on purpose?

**Comment by jmccormick2001** [comment]
thanks, yep, the yaml file could be formatted, I think there is only pgo.yaml, but some of the other json files are special and should not be formatted.

File: chart/postgres-operator/templates/service-account.yaml (hunk 3 above)
Comment (blaggacao) [reply chart/postgres-operator/templates/service-account.yaml]: I was wondering, as I did not found any reference to `serviceaccount` in no .go file...
Either the service account is intelligently detected by some implicit logic in some dependency or it just takes the first available service account in the namespace, which in my (and your case) is default... @jmccormick2001 Are we on the right track?

**Comment by blaggacao** [comment]
Same for `.../apiserver`:
- [ ] pgo.load-template.json
- [ ] pgo.lspvc-template.json
- [ ] pgo.yaml
- [ ] pgorole
- [ ] pgouser
- [x] keys -> unchanged

**Comment by blaggacao** [comment]
@jmccormick2001 Last commit as a bonus, you might just want to drop it...

**Comment by blaggacao** [comment]
@jmccormick2001 Now back to the remaining issue, which I'm not sure if it's related...

Can you inform something on: https://github.com/CrunchyData/postgres-operator/pull/228#discussion_r189619994 ?

Depending on your answer this would be probably merge ready...

File: chart/postgres-operator/values.yaml (hunk 2 above)
Comment (blaggacao) [reply chart/postgres-operator/values.yaml]: Well, you might have more sensible deployment names, like (just as examples):
`--name=staging`
`--name=production`
`--name=europe`
`--name=americas`

Then you still need to identify the package (`pgo`), but if you just do:

`--name=postgres-operator`

You don't want to end up with `postgres-operator-postgres-operator` as the full name...
So now it would be `postgres-operator-pgo` which is a little more useful and covers both cases...

**Comment by blaggacao** [comment]
Btw, as tiller will be dropped for helm 3, people avoid putting an (unknown) component into their cluster, so defaulting on helm deployments becomes less intrusive and we could make this chart just perfect for what it's worth in future..,

**Comment by blaggacao** [comment]
Second [order tweaking pending](https://github.com/xoes/postgres-operator/pull/1) 
Thanks @TomNeyland

**Comment by blaggacao** [comment]
Quoting and crediting @TomNeyland 

>Some tweaks for your pr [#228](https://github.com/CrunchyData/postgres-operator/pull/228) based on the tips in the following links, and on the default service scaffolding helm provides.
https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices
https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/templates.md\#names-of-defined-templates
https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/rbac.md\#using-rbac-resources

**Comment by ieugen** [comment]
Wow @blaggacao , you really did a lot of work. The big question is: Does it work ? Unfortunately I was and still am caught with day work and can't helm much. 

TLDR: We are preparing for GDPR and at the same time migrating our infrastructure to a new Kubernetes cluster. One of the migration goals is to run the database inside K8S and it seems that the operator is the only sane way to do it. Hopefully things will settle down soon. If not I will have to look for a quick fix now and come back to this when we have more time.

**Comment by blaggacao** [comment]
@ieugen @jmccormick2001 I plan to give it another session next weekend, but It somehow occurs to me that not using the appropriate service account might not be a genuine problem of the helm chart (at least, I couldn't see how)...

**Comment by manwegit** [comment]
Helm chart for pgo-operator is still broken? At least for me it failes because of serviceaccount permissions as it tries to use the default.
Chart template for deployment is missing the created serviceaccount field.
```
...
    spec:
      serviceAccountName: RELEASENAME-pgo
      containers:
      -  ...
```

if it is not specifed, default will be used.

jmccormick2001:
should be fixed now, I updated the chart's deployment to reference the service account also changed the values.yaml to specify the service account name.  Seems to work for me now.

PR #247, opened 2018-06-04
If failover target replica has a metrics sidecar, failover will fail

### LINKED ISSUE #241: Failover doesn't work when replica has metrics sidecar
When a cluster replica has a metrics sidecar, trying to execute a failover with that replica as the target will fail with the following error:

```
time="2018-05-31T12:05:29Z" level=info msg="failover task added"
time="2018-05-31T12:05:29Z" level=info msg="cluster name is map[failtest-metrics:failtest-metrics]\n"
time="2018-05-31T12:05:29Z" level=info msg="dbname is failtest-metrics-failover"
time="2018-05-31T12:05:29Z" level=info msg="strategy found"
time="2018-05-31T12:05:29Z" level=info msg="strategy 1 Failover called on failtest-metrics target is failtest-metrics-kxfi"
time="2018-05-31T12:05:29Z" level=error msg="could not find a container in the pod"
```

RBlaine95:
If the failover target replica has a metrics sidecar `len(pod.Spec.Containers)` will return `2` instead of `1`.

To fix this, I added onto the if statement to also check if the `pod.Lables["crunchy_collect"] == false`.

This is to fix #241

### DIFF
--- a/util/failover.go
+++ b/util/failover.go
@@ -86,7 +86,7 @@ func GetPod(clientset *kubernetes.Clientset, deploymentName, namespace string) (
 	for _, v := range pods.Items {
 		pod = &v
 	}
-	if len(pod.Spec.Containers) != 1 {
+	if len(pod.Spec.Containers) != 1 && pod.Labels["crunchy_collect"] == "false" {
 		return pod, errors.New("could not find a container in the pod")
 	}
 


jmccormick2001:
thanks, I've found other issues in this bit of logic and have made a change as to how the failover target is validated, that change is merged into the autofail branch at this time so I'll  close this PR out, the next release and master branch will have this fix in it shortly.  regards and thank you for pointing out this bug and suggesting a fix.

PR #248, opened 2018-06-04
Corrects name of PVC for restoring from backup

guineveresaenger:
In the documentation for restoring a database from backup, the backup PVC needs to be <clustername>-backup, not just <clustername>.
Since there is also a <clustername> PVC, this can be ambiguous.


guineveresaenger:
Unfortunately, from previous experience, it doesn't look like the generate-docs.sh is entirely working, so I may follow up with a docs publication commit if I can get hugo working locally.

guineveresaenger:
I think I found my error from last week (did not handle the adoc correctly). This should keep the formatting and be good to go.

Thank you for letting me contribute!

xenophenes:
Yep - I realized that afterwards as well. So for future reference, you only need to modify the documents under $CCPROOT/hugo/content and you'd completely ignore anything under docs/ - that gets regenerated when you run the generate-docs.sh script. :)

This looks good to me @jmccormick2001 !

PR #249, opened 2018-06-06
Fix golang-examples/

aduong:
The examples were referring to packages that didn't exist and so could not be built.
Notably, there seemed to be confusion between usages of client-go and apimachinery.

Commented imports were also removed and goimports was run on the files.

### DIFF
--- a/golang-examples/addlabel.go
+++ b/golang-examples/addlabel.go
@@ -1,25 +1,16 @@
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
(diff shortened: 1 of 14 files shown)


jmccormick2001:
thanks for cleaning these up and updating them, they are mostly code samples or simple tests dating back 12 months.

aduong:
This change also puts this repo in a position to move from godep to dep if that suits your devs' workflow. I'm working on multiple Go repos at once and using dep has been nice for that.

PR #251, opened 2018-06-07
Use User & UserSecretName in Pgcluster

aduong:
There are User and UserSecretName fields in the Pgclusterspec but they are being ignored for cluster creation.

Admittedly, User is being set to "testuser" by the API server on cluster creation anyway, but the Pgcluster CR should be source of truth for the operator.


jmccormick2001:
thanks for this PR, I'm evaluating exactly what I want to store in the pgcluster CRD and this is useful to consider.

PR #328, opened 2018-08-22
update pgouser file information, update pgo.yaml.dynamic -> storageclass

xenophenes:
Add roles to all instances of pgouser username:password combinations.
Updated pgo.yaml.dynamic reference to pgo.yaml.storageclass.


jmccormick2001:
the user's .pgouser file only has a username and password, where as the server pgouser file includes the user's role, the docs would need to be updated to reflect this.

Issue #338, opened 2018-08-27
integrate with crunchy-cron-scheduler


jmccormick2001:
1. Pgo backup mycluster --schedule=”* * * * 1” --backup-type=
2. Pgo show schedule mycluster
3. Pgo delete schedule mycluster
4. Pgo create schedule --schedule=”* * * 1” --selector=env=research --schedule-type=pgbackrest  --schedule-opts-”a b c”
5. Crunchy-scheduler included in the operator deployment.yaml
6. Pgo delete cluster mycluster “removes any schedules for this cluster”
7. Uses the existing Job templates for backup and backrest backup jobs
8. Uses the existing credential secrets created for the cluster

jasonodonnell:
- [X] Added schedule service to API
  - [X] Added commands
    - [X] `create`
    - [X] `delete`
    - [X] `show`
  - [X] Create schedules on selectors
    - [X] pgBackRest
    - [ ] pgBaseBackup
      - [X] Discover secrets
      - [ ] Create backup PVC
      - [ ] Create job
  - [X] Delete schedules on selectors
  - [X] Show schedules
  - [ ] Add Crunchy Scheduler to deployment
  - [ ] Documentation (both container suite scheduler doc and operator doc)
  - [ ] Operator scheduler examples

PR #370, opened 2018-09-20
Add scheduler integration

jasonodonnell:
This pull request satisfies the requirements in #338 by integrating the `crunchy scheduler` application in the container suite with operator.

New commands:

```bash
pgo create schedule ...
pgo show schedule ...
pgo delete schedule ...
```

Additionally I added a new flag to delete cluster: `--delete-configs`.  This is used for cleaning up all the config maps (including schedules) for a given cluster.


jmccormick2001:
I looked at this one today, looks good except for the apis/cr/v1/scheduler.go, that directory is meant to hold the CRD definitions only...since we were not going to make schedules a CRD at this point, the schedule definition would be located in the apiserver scheduler package.  With that fix, I'm likely to try and merge the code into 3.3 and just disable the CLI calls to it for now until I can get the container suite release done that this is dependent on.

jasonodonnell:
@jmccormick2001 I'll make that change and resolve these conflicts from the recent merges.

jasonodonnell:
@jmccormick2001 Updated

PR #613, opened 2019-03-02
fix example/envs.sh: use CO_NAMESPACE

Sayrus:
**Checklist:**

 
**Type of Changes:**

 
The example `envs.sh` use neither CO_NAMESPACE nor demo.
The example `envs.sh` use an outdated CO_VERSION which does not support the `CreateDump`, `RestoreDump` and `ScaleCluster`. This will cause the installation tutorial to fail because `CreateDump` is not a valid permission in 3.5.0.

The example envs.sh now use CO_NAMESPACE which is by default demo.
The example envs.sh now use 3.5.1 which supports these three permissions.

**Other information**:
As a new user, having a broken example `envs.sh` is confusing. Your deployment will fail even if you follow the installation tutorial. The error is not obvious until you read the code or the changelog.

### DIFF
--- a/examples/envs.sh
+++ b/examples/envs.sh
@@ -6,7 +6,7 @@ export CO_CMD=kubectl
 export COROOT=$GOPATH/src/github.com/crunchydata/postgres-operator
 export CO_IMAGE_PREFIX=crunchydata
 export CO_BASEOS=centos7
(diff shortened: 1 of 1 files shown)


jmccormick2001:
thanks for catching that one.

PR #630, opened 2019-03-10
Replace "kubectl" with variable defined and add env vars for default admin and clustername 

bysnupy:
**Description**
  Even though already defined `CO_CMD=kubectl` but not used it in the script and `Makefile`, so I replace hard coded `kubectl` with the env variable, additonally add `CO_DEFAULT_CLUSTER` and `CO_ADMIN_USER` env variable for enhancement to make change easily the names on each user system. 

  For instance, `OpenShift` use usually `oc` cli instead of `kubectl` and the `default cluster name` is based on the cluster entry `API URL` and default `system admin` name is `system:admin`.

**Checklist:**

**Type of Changes:**

 
 - [] New feature (non-breaking change which adds functionality)

When install on `OpenShift`, I should edit the `cli` command name, `system admin`, `cluster name` manually for installation.

I can start to install as specifying the `cli`, `admin name`, `cluster name` in the `.bashrc` initially.
And it's not feature changes.

**Other information**:

I've verified `OpenShift v3.11`(`Kubernetes v1.11`) about this changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,11 +50,11 @@ installrbac:
 setup:
 	./bin/get-deps.sh
 setupnamespace:
-	kubectl create -f ./examples/demo-namespace.json
-	kubectl config set-context demo --cluster=kubernetes --namespace=demo --user=kubernetes-admin
-	kubectl config use-context demo
+	$(CO_CMD) create -f ./examples/demo-namespace.json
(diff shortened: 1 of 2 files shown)


jmccormick2001:
thanks, I think  I have this fix mostly in the olm branch, that branch is soon to be merged into develop and then will be the basis of the 4.0 release we are working on.  if you want to resubmit this against the olm branch I'll take another look at it.  regards.

PR #631, opened 2019-03-10
"hostPath" is local storage and not considering to share with other pods

bysnupy:
`hostPath` can not share with other pods, and `Postgresql`'s `HA` is working through own replication storage, not sharing storage. Even though it's just suggestion for testing `PV`/`PVC` definition, you had better to be more accurate details for clarification, because `database`'s `storage` is more important section than other software.

Refer [Types of Persistent Volumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#types-of-persistent-volumes) for more details.
~~~
HostPath (Single node testing only – local storage is not supported in any way and WILL NOT WORK in a multi-node cluster)
~~~

**Checklist:**

 
**Type of Changes:**

 
Current `hostPath` `PV`/`PVC`'s `accessMode` is not correct, and it allows to bind on `PV` with multiple `PVC` potentially.
 
It would be more clarification for user's usage of `hostPath`.

**Other information**:

### DIFF
--- a/conf/postgres-operator/pgo.yaml
+++ b/conf/postgres-operator/pgo.yaml
@@ -28,7 +28,7 @@ ReplicaStorage: hostpathstorage
 BackrestStorage: hostpathstorage
 Storage:
   hostpathstorage:
-    AccessMode:  ReadWriteMany
+    AccessMode:  ReadWriteOnce
     Size:  1G
     StorageType:  create
   replicastorage:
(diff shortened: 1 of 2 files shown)


jmccormick2001:
hostPath is ok to run replicas with on a single kube host which is pretty much the only use case for it today, I'm ok with it being ReadWriteMany in that case.  Perhaps the docs need to discuss the usage of hostpath and when its appropriate.  That might be a new PR.

PR #894, opened 2019-07-29
update selector to include pgpool label = true

jmckulk:
**Checklist:**

 
**Type of Changes:**

 
The selector for getPrimaryPodName ignores pods with pgpool enabled so it returns "primary pod is not in Ready state".

Now the function checks that the service_name matches the cluster name

**Other information**:
[ch4753]


flamingdumpster:
Multiple tests indicate the fix is good without any adverse side effects

PR #965, opened 2019-08-29
Modify event timestamp type to ensure ISO 8601 compliance

abrightwell:
**Checklist:**

 
**Type of Changes:**

 
Currently, the event time stamps are represented as a formatted `string`. This representation is missing the timezone component and is not an ISO 8601 compliant format.

The new behavior changes the representation from a `string` to `time.Time`. When this type is marshaled to JSON the value will be a RFC 3339 representation which includes the timezone and is a profile of ISO 8601.

**Other information**:


wilybrace:
>  it actually updates the time the event occurred.

Where do you see it changing the time beyond the function call overhead? I don't think I've missed one, but as far as I saw every replacement invoked a time.Now() in its implementation, not using a cached one that was calculated once.

flamingdumpster:
From: Timestamp: events.GetTimestamp(),	 To: Timestamp: time.Now() in clusterimpl.go for example. Am I missing something about what events.GetTimestamp() does? 

If all that does is get the current time in a certain format, then my mistake. I think initial glance i believed it to be getting an existing timestamp previously recorded.

PR #993, opened 2019-09-11
Update to user valid days flag behavior [ch5122]

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
Valid-days flag must accompany a password update to take effect.
CLI help for pgo update user needs to be updated for current behavior.

Allow for updates to password valid days without password change.
Updated CLI help output to match new behavior and removed old references to obsolete flags.
Added error handling for the --expire flag to indicate that it runs against all the users that have expiration dates and allowed its use it without specifying --username.

**Other information**:


apiserver/userservice/userimpl.go
```diff
@@ -80,9 +80,15 @@ func UpdateUser(request *msgs.UpdateUserRequest, pgouser string) msgs.UpdateUser
...
+	if request.Username == "" && request.Expired == "" {
+		resp.Status.Code = msgs.Error
+		resp.Status.Msg = "Either --expired or --username must be set. See 'pgo update user -h' for usage."
+		return resp
+	}
+
+	if request.Username != "" && request.Expired != "" {
 		resp.Status.Code = msgs.Error
-		resp.Status.Msg = "--username is required flag"
+		resp.Status.Msg = "The --expired flag executes against all expiring users. See 'pgo update user -h' for usage."
```
flamingdumpster: This needs to be more along the lines of "The --expired flag cannot be used when a user is specified"  and the help and docs both indicate that. 

Also any adverse effect if --all and --expired are used - any logic that gets executed that we don't want to in that case?

pgo/cmd/update.go
```diff
@@ -44,14 +44,14 @@ func init() {
...
+	UpdateUserCmd.Flags().StringVarP(&Expired, "expired", "", "", "Updates passwords that will expire in X days using an autogenerated password.")
 	UpdateUserCmd.Flags().BoolVarP(&ExpireUser, "expire-user", "", false, "Performs expiring a user if set to true.")
 	UpdateUserCmd.Flags().IntVarP(&PasswordAgeDays, "valid-days", "", 30, "Sets passwords for new users to X days.")
 	UpdateUserCmd.Flags().StringVarP(&Username, "username", "", "", "Updates the postgres user on selective clusters.")
 	//UpdateUserCmd.Flags().StringVarP(&UserDBAccess, "db", "", "", "Grants the user access to a database.")
 	UpdateUserCmd.Flags().StringVarP(&Password, "password", "", "", "Specifies the user password when updating a user password or creating a new user.")
 	UpdateUserCmd.Flags().BoolVar(&AllFlag, "all", false, "all clusters.")
-	UpdateUserCmd.Flags().IntVarP(&PasswordLength, "password-length", "", 22, "If no password is supplied, this is the length of the auto generated password")
+	UpdateUserCmd.Flags().IntVarP(&PasswordLength, "password-length", "", 22, "If no password is supplied with the expired flag, this is the length of the auto generated password")
```
flamingdumpster: Is this flag only used with the expired flag? the updated wording seems to indicate that.

pgo/cmd/update.go
```diff
@@ -129,19 +129,24 @@ var UpdateUserCmd = &cobra.Command{
 	Short: "Update a postgres user",
 	Long: `UPDATE allows you to update a pgo user. For example:
 		//change a password
-		pgo update user mycluster --username=someuser --password=foo --update-passwords --expired=200
+		pgo update user mycluster --username=someuser --password=foo --valid-days=40
```
flamingdumpster: Change the comment above this for clarity, something like " change a password, set valid days for 40 days from now"

pgo/cmd/update.go
tjmoore4: Yes, from my testing and code review, it only appears to work with the expired flag when using pgo update user.

pgo/cmd/update.go
tjmoore4: Updated.

PR #995, opened 2019-09-13
adds more updates to use pgo.yaml user

jmckulk:
- updates templates for bash/ansible
- adds error prefix for pgoconfig validation
- removes default user as testuser so operator now fails to start if
	user is not set in pgo.yaml

**Checklist:**

 
**Type of Changes:**

 
currently testuser is a hardcoded default in the pgpool templates

-Now the templates pull the db user from the pgo config (pgo.yaml or inventory file)
-adds a error message prefix to the errors in validation function for config. Prefix tells user what file to check
-removes the default user of testuser if user is not defined in the pgo.yaml file.

**Other information**:
[ch4290]
[ch5411]

### DIFF
--- a/conf/postgres-operator/pgpool-template.json
+++ b/conf/postgres-operator/pgpool-template.json
@@ -44,7 +44,7 @@
 		    {{.ContainerResources }}
                     "env": [{
                         "name": "PG_USERNAME",
-                        "value": "{{.PGUser}}"
+                        "value": "{{.User}}"
                     }, {
                         "name": "PG_PASSWORD",
                         "value": "password"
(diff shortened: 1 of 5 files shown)


conf/postgres-operator/pgpool-template.json
```diff
@@ -44,7 +44,7 @@
 		    {{.ContainerResources }}
                     "env": [{
                         "name": "PG_USERNAME",
-                        "value": "testuser"
+                        "value": "{{.PGUser}}"
```
flamingdumpster: Is .PGUser correct here? it is the only place I see it, where I see "User" in several other places.

jmckulk: No its not. Ill fix it

PR #1010, opened 2019-09-19
Secret template

jmckulk:
**Checklist:**

 
**Type of Changes:**

 
username and passwords are stored in plain text when these templates are used

now 

**Other information**:
[ch5389]

Testing method done on both bash/ansible:
- create pgpool cluster
- check kubectl describe to make sure env vars are using secrets
- create pgbouncer cluster
- check env vars
- create normal cluster
- run pgo-load against normal cluster (based off of test-lodge run)
- check env vars
- run `pgo test --all` verify that all the connections work

### DIFF
--- a/ansible/roles/pgo-operator/files/pgo-configs/pgbouncer-template.json
+++ b/ansible/roles/pgo-operator/files/pgo-configs/pgbouncer-template.json
@@ -46,7 +46,7 @@
                         "name": "PG_USERNAME",
                         "valueFrom": {
                             "secretKeyRef": {
-                                "name": "{{.SecretsName}}",
+                                "name": "{{.PGBouncerSecret}}",
                                 "key": "username"
                             }
                         }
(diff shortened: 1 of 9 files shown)


ansible/roles/pgo-operator/files/pgo-configs/pgbouncer-template.json
```diff
@@ -44,10 +44,20 @@
 		    {{.ContainerResources }}
                     "env": [{
                         "name": "PG_USERNAME",
-                        "value": "{{.PgBouncerUser}}"
+                        "valueFrom": {
+                            "secretKeyRef": {
+                                "name": "{{.SecretsName}}",
```
flamingdumpster: Can we take the time to make these names more understandable in context? i.e. 'SecretsName" could be PGSecretsName or XYXSecretsName. Understand it has been this way, but I think the change will help. Comment applies to anywhere "SecretsName" or equally generic naming in the template(s) is/are used.

jmckulk: Done - let me know if those names don't work

PR #1029, opened 2019-09-30
implements authentication bypass endpoints

wilybrace:
Enables the bypassing of authentication modes for URL routes, allowing health check endpoint to be accessible to large-scale monitoring tools.

**Checklist:**

(Kubernetes only)

**Type of Changes:**

 
Currently, all URL routes require mutual TLS authentication and all but the `/health` route requires HTTP Basic authentication. The `/health` route potentially prompts for HTTP Basic, but does not enforce it.

When the `NOAUTH_ROUTES` environment variable is correctly configured, the server's TLS configuration is reduced to verifying presented client certificates if presented vice both requiring and verifying client certificates.
For all protected routes, the existence of verified client certificates is confirmed or a HTTP Forbidden status is returned.
For noauth routes, that verified certificate check is skipped.

At this time, only the `/health` endpoint is allowed to be a noauth route. Enabling future endpoints will require they remove dependencies on the username returned from HTTP Basic authentication.
`/health` already had no dependency and no validation prior to this change.

### DIFF
--- a/hugo/content/Configuration/configuration.md
+++ b/hugo/content/Configuration/configuration.md
@@ -52,6 +52,33 @@ Files within this directory are used specifically when creating PostgreSQL Clust
 
 As with the other Operator templates, administrators can make custom changes to this set of templates to add custom features or metadata into the Resources created by the Operator.
 
+## Operator API Server
+
+The Operator's API server can be configured to allow access to select URL routes
+without requiring TLS authentication from the client and without
+the HTTP Basic authentication used for role-based-access.
+
+This configuration is performed by defining the `NOAUTH_ROUTES` environment
+variable for the apiserver container within the Operator pod. Typically, this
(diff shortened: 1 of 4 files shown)


hugo/content/Configuration/configuration.md
```diff
@@ -52,6 +52,33 @@ Files within this directory are used specifically when creating PostgreSQL Clust
...
 
+## Operator API Server
+
+The Operator's API server can be configured to allow access to select URL routes
+without requiring TLS authentication from the client and without
+the HTTP Basic authentication used for role-based-access.
+
+This configuration is performed by defining the `NOAUTH_ROUTES` environment
+variable for the apiserver container within the Operator pod. Typically, this
+configuration is made within the `deploy/deployment.yaml` file.
```
tjmoore4: Believe this reference should be deployment.json, not yaml. Is that correct?

PR #1037, opened 2019-10-02
Scrape cAdvisor for Container metrics

cbandy:
We should look at `container_*` metrics to monitor individual clusters.

Unfortunately, this has the undesirable effect of showing `cadvisor` in the dropdowns of existing Grafana dashboards.

**Checklist:**

 
   - Validated in GKE and k3s clusters.
   - Does not work in minikube.

**Type of Changes:**

 
### DIFF
--- a/ansible/roles/pgo-metrics/tasks/prometheus.yml
+++ b/ansible/roles/pgo-metrics/tasks/prometheus.yml
@@ -46,6 +46,7 @@
         dest: "{{ prom_output_dir }}/{{ item | replace('.j2', '') }}"
         mode: '0600'
       with_items: 
+      - prometheus-config.yaml.j2
       - prometheus-pvc.json.j2
       - prometheus-service.json.j2
       - prometheus-deployment.json.j2
@@ -54,6 +55,7 @@
     - name: Create Prometheus Objects
       command: "{{ kubectl_or_oc }} create -f {{ prom_output_dir }}/{{ item }} -n {{ metrics_namespace }}"
       with_items:
+      - prometheus-config.yaml
       - prometheus-pvc.json
       - prometheus-service.json
       - prometheus-deployment.json
--- a/ansible/roles/pgo-metrics/templates/prometheus-config.yaml.j2
+++ b/ansible/roles/pgo-metrics/templates/prometheus-config.yaml.j2
@@ -1,0 +1,63 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: crunchy-prometheus
+data:
+  prometheus.yml: |-
+    ---
+    global:
+      scrape_interval: 10s
+      scrape_timeout: 10s
+      evaluation_interval: 5s
+
+    scrape_configs:
+    - job_name: cadvisor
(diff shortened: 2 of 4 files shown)


cbandy:
> I am having trouble getting this to work in a local Kube environment (a 2-node Kube v1.15.3 cluster created using `kubeadm`), specifically due to connectivity issues and certificate/trust issues. At first I was seeing errors such as follows in Prometheus when attempting viewing the `cadvisor` targets:
> 
> ```
> Get https://master:10250/metrics/cadvisor: context deadline exceeded
> Get https://node01:10250/metrics/cadvisor: context deadline exceeded
> ```
> 
> Since based on the above the scrape request appeared to be timing out, it appeared as though the Prometheus pod was unable to communicate directly with the Kubelet on each node (this was a 2 node cluster). I therefore decided to try giving the Prometheus pod access to the host network just to see if it would be able to connect, specifically by adding `"hostNetwork": true` to pod spec in `prometheus-deployment.json.j2` (although I realize this is a change we'd def want to discuss further before implementing since there are some security implications).

I didn't run into this in my testing, but I did notice that k3s did some work to make node hostnames resolvable from inside the cluster.

> However, while this seemed to resolve the connectivity issues, it then resulted in the following certificate errors:
> 
> ```
> Get https://master:10250/metrics/cadvisor: x509: certificate signed by unknown authority
> Get https://node01:10250/metrics/cadvisor: x509: certificate signed by unknown authority
> ```
> 
> I do see that Prometheus is using a certificate to authenticate (`/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`), but while this certificate works for accessing the Kubernetes API server, the Kubelets in my cluster utilize their own certificate (specifically found under `/var/lib/kubelet/pki`). This therefore appears to be why Prometheus is unable to connect to the Kueblet metrics endpoint using the `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` certificate.

Yep. That's what I ran into in minikube as well. I can investigate the certificate chains in more detail tomorrow.

cbandy:
So the kubelet _server_ certificate is a nontrivial topic. [By default][kubelet-default], kubelet will generate and self-sign this certificate. More complete/advanced Kubernetes setups sign the certificate with the Kubernetes CA and rotate it as needed. To be compatible with simpler setups we must `insecure_skip_verify: true`. In my testing:

| Flavor | Network | Server Certificate |
|--------|---------|--------------------|
| GKE      | InternalIP: yes, Hostname: yes | CA-signed, SAN: DNS={Hostname} IP={InternalIP} IP={ExternalIP} |
| k3s      | InternalIP: yes, Hostname: yes | CA-signed, SAN: DNS={Hostname} IP=127.0.0.1 |
| minikube | InternalIP: yes, Hostname: no  | self-signed, SAN: DNS={Hostname} |
| kubeadm  | InternalIP: yes, Hostname: no  | self-signed, SAN: DNS={Hostname} |

[kubelet-default]: https://github.com/kubernetes/kubernetes/blob/v1.15.4/pkg/kubelet/apis/config/types.go#L88-L95

PR #1038, opened 2019-10-03
Improvements to Postgres Operator Ansible Inventory Validation

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
These changes improve the inventory validation to better ensure
the appropriate values are set before completing an Openshift or
Kubernetes Operator install. Most of the checking now happens during
preflight, meaning the problems are identified earlier.
Additionally, there are checks to see that the required values are not
just set but are also not null.

**Other information**:
[ch4267]

### DIFF
--- a/ansible/roles/pgo-preflight/tasks/check_openshift.yml
+++ b/ansible/roles/pgo-preflight/tasks/check_openshift.yml
@@ -1,5 +1,14 @@
 # check_openshift.yml
 ---
+- name: openshift_token should be defined
+  assert:
+    that:
+      - openshift_token != ''
+    msg: "Set the value of 'openshift_token' in the inventory file."
+  when: 
+    - openshift_token is defined
+  tags: always
+
 - name: openshift_user should be defined
   assert:
     that:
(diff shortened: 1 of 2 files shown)


tjmoore4:
> **We should also do:**
> 
> In postgres-operator/ansible/roles/operator/tasks/main.yml
> Remove this code, this is checked in the pre-flight
> 
> ```
>   - fail:
>     msg: "Only set one of kubernetes_context or openshift_host"
>   when: openshift_host is defined and kubernetes_context is defined
>   tags: always
> ```

There is a very specific series of mis-configuration that can result in the installer attempting to use the empty Openshift values to install the operator in place of the properly set Kubernetes context. This can only happen if the Kube context is set but the Openshift values are defined as null. 

Since this case was caught during testing, the additional check was left in.

hitech73:
The check that we modified checks for this now so the one in the main operator is no longer needed. 

```
- name: Check if required Kube or OpenShift variables are defined
  fail: msg="Please specify either OpenShift or Kubernetes variables in inventory"
  when: openshift_user is undefined and openshift_host is undefined and openshift_token is undefined and kubernetes_context is undefined
  when: > 
    (openshift_host is undefined or (openshift_host is defined and openshift_host == ''))
    and (kubernetes_context is undefined or (kubernetes_context is defined and kubernetes_context == ''))
    or ((openshift_host is defined and openshift_host != '') and (kubernetes_context is defined and kubernetes_context != ''))
  tags: always

- include_tasks: check_openshift.yml
```

That said since the pre-flight is ran before the operator leaving it in will not harm, but it will never be triggered because the pre-flight will catch it first.

PR #1047, opened 2019-10-10
add_target_namespace pains

wilybrace:
add-targeted-namespace.sh has a sinister failure case if
PGO_INSTALLATION_NAME is unset in that the pgo-installation-name label
does not get set correctly. This causes later failures in selectors
built upon the assumption of that label's existence.

This change updates add-targeted-namespace.sh to fail if its expected
environment variables do not exist and updates the references to the
script to warn about needing to ensure its definition.

[ch5543]

**Checklist:**

**Type of Changes:**

 
### DIFF
--- a/deploy/add-targeted-namespace.sh
+++ b/deploy/add-targeted-namespace.sh
@@ -14,13 +14,16 @@
 
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-
+# Enforce required environment variables
+test="${PGO_CMD:?Need to set PGO_CMD env variable}"
+test="${PGOROOT:?Need to set PGOROOT env variable}"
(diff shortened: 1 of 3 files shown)


jmckulk:
Also the script requires `expenv` so they would need to run `go get .../blang/expenv` if they are going to run the script.

It seems to me like this script might require more than just a warning. Do you think we should add a new section for it @andrewlecuyer?

andrewlecuyer:
@jmckulk good catch regarding to `expenv`.  This is def turning into a big warning, so I'm on board with creating a new section.  And going that route, a warning pointing to that section would prob be a good idea, i.e. good to have something to jump out at OCP users.

Also, I think we not only need to show what env vars need to be set, but also how to set them.  For instance, specify that `PGO_INSTALLATION_NAME` should match `pgo_installation_name` in the `inventory`, `PGO_CMD` should be `oc` or `kubectl`, etc.

PR #1067, opened 2019-10-23
Rewrite the Makefile for more consistency

scrummyin:
The makefile was previously hard to read and follow. In order to make
future changes easier, a few changes were made for consistency. They are
as follows:

- All targets are marked as .PHONY directly at the top of the file. The
reason for this is all the targets are phony so the intent is to make
that obvious to the user upfront.

- Added only one black line between each target, but there are 2 black
lines before sections denoted by a `#======= * =======`.

- Reordered targets so that similar ones are next to each other.

- Made targets follow the standard of a single space after each colon.
Previously there were sometimes tabs and sometimes spaces.

- Removed calls to make from inside the Makefile. Targets will now just
call intended targets instead of executing a new make command.

- Added comments to most targets that would be used by outside users to
indicate what they will do.

A few features were added to help with consistency and ease of use:

- There is now a help target that will display lines that have a
target defined followed by a `## <message>`. This allows for `$ make
help` to be run and general idea of what commands do will be.

- The new help command has been set as the default command for make if
no target is specified.

- Added an equivalent clean-*-image for each image target.

- Added a global clean-images that calls all the previously mentioned
clean-*-images (NOTE: buildah is not parallel safe)

**Checklist:**

 
  All the previous documentation still applies. I search for calls examples of make in the hugo docs and all the ones I saw still seem relevant.
 - [?] Have you tested your changes on all related environments with successful results, as applicable?
  I have yet to test them on a rhel enviornment, but I do not expect the changes to cause issues there.

**Type of Changes:**

 
These build changes will not change any functionallity within the operator itself so they may be considered non breaking. However, the default command for `make ` had been changed in this. I would be willing to exclude that, but I think there should be discussion around that.

**Other information**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,20 @@
 RELTMPDIR=/tmp/release.$(PGO_VERSION)
 RELFILE=/tmp/postgres-operator.$(PGO_VERSION).tar.gz
 
+.PHONY: help check-go-vars installrbac setupnamespaces cleannamespaces bounce macpgo winpgo setup deployoperator \
+main runmain runapiserver pgo-apiserver pgo-backrest pgo postgres-operator pgo-rmdata clean cli-docs \
+pgo-backrest-image pgo-event-image pgo-backrest-restore-image pgo-backrest-repo-image pgo-apiserver-image \
+postgres-operator-image pgo-lspvc-image pgo-load-image pgo-rmdata-image pgo-sqlrunner-image pgo-scheduler-image \
+all push pull release deepsix default
+
+
+.DEFAULT: help
+
+# The help function searches the Makefile for targets that have a '##' after then prints the result
+help: ## Prints the help message
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+
 #======= Safety checks =======
 check-go-vars:
 ifndef GOPATH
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -10,118 +24,172 @@ ifndef GOBIN
...
+bounce: ## Restarts the postgres-operator pods
+	$(PGO_CMD) --namespace=$(PGO_OPERATOR_NAMESPACE) get pod --selector=name=postgres-operator -o=jsonpath="{.items[0].metadata.name}" | xargs $(PGO_CMD) --namespace=$(PGO_OPERATOR_NAMESPACE) delete pod
+
+
+#======= Platform specific =======
+macpgo:	check-go-vars ## Builds the pgo cli binary for the mac(darwin) platform
 	cd pgo && env GOOS=darwin GOARCH=amd64 go build pgo.go && mv pgo $(GOBIN)/pgo-mac
-	env GOOS=darwin GOARCH=amd64 go build github.com/blang/expenv && mv expenv $(GOBIN)/expenv-mac
-winpgo:	check-go-vars
+	env GOOS=darwin GOARCH=amd64 go build github.com/blang/expen && mv expenv $(GOBIN)/expenv-mac
```
wilybrace: Build target should be `github.com/blang/expenv`

Makefile
```diff
@@ -10,118 +24,172 @@ ifndef GOBIN
...
+winpgo:	check-go-vars ## Builds the the pgo binary for the windows platform
 	cd pgo && env GOOS=windows GOARCH=386 go build pgo.go && mv pgo.exe $(GOBIN)/pgo.exe
 	env GOOS=windows GOARCH=386 go build github.com/blang/expenv && mv expenv.exe $(GOBIN)/expenv.exe
 
-installrbac:
-	cd deploy && ./install-rbac.sh
-setup:
+
+#====== golang =======
+setup: ## Install the golang dependencies
```
wilybrace: These are build dependencies in a general sense - some are golang, but mostly environment and container dependencies - documentation likely should reflect that and it be moved to the appropriate section.

Makefile
```diff
@@ -162,6 +232,12 @@ release:	check-go-vars
...
-	all
 
+
+#======= Misc =======
+deepsix:
+	cd $(PGOROOT)/apis/cr/v1
+	deepcopy-gen --go-header-file=$(PGOROOT)/apis/cr/v1/header.go.txt --input-dirs=.
+
+
+default: help
```
jkatz: My (strong) preference is that the `default` build target is not help, but remains `all`. There will be some upcoming work that continues to improve upon this.

jkatz:
Closed in favor of #1081

PR #1069, opened 2019-10-24
Add disable TLS functionality to the PGO Client

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
The Postgres Operator's apiserver currently supports the
disabling of TLS authentication when receiving client connections.
However, the PGO client did not have a method for disabling
the necessary resource checks that are then unnecessary, meaning
the unneeded certificates and keys were still required to be
present.

This update allows the PGO client to skip the relevant configuration
setup when using an unauthenticated connection. This can be done in
one of two ways. Either by an environment variable:

DISABLE_TLS=true

(As is used by the Postgres Operator to disable TLS when performing
a Bash install)

Or by using the '--disable-tls' flag with the PGO client, for example:

pgo version --disable-tls

In the event that the DISABLE_TLS variable is set to 'false' and the
--disable-tls is used, the flag is authoritative.

Additionally, this commit includes documenationupdates to the main
TLS configuration page to explain how to properly set the new PGO
CLI configuration variables depending on installation method.

Notes with links to the main page were also added to both
the Bash and Ansible installation sections.

This PR was tested using an Ansible installation of the Postgres Operator with Kubernetes
and using a Bash installation of the Postgres Operator with Openshift.

**Other information**:
For testing, as this update does not add any new functions, unit testing is not ideal. The following steps can be followed to validate the appropriate behavior.

First, install the Postgres Operator with TLS enabled (the default setting) as described here: https://access.crunchydata.com/documentation/postgres-operator/4.1.0/installation/

Validate that commands work as expected, for instance

pgo version

returns the client and server versions

Now, either set the environment variable, DISABLE_TLS=true, or use 

pgo version --disable-tls

The command should now fail.

At this point, update the PGO installation to disable TLS authentication (as described here https://access.crunchydata.com/documentation/postgres-operator/4.1.0/gettingstarted/design/tls/) and unset the TLS specific environment variables:

unset PGO_CA_CERT
unset PGO_CLIENT_CERT
unset PGO_CLIENT_KEY

Now running 'pgo version' without setting either the DISABLE_TLS variable or using the --disable-tls flag should result in an error and failure to return the expected values.

Issue: [ch6126]

### DIFF
--- a/hugo/content/gettingstarted/Design/tls.md
+++ b/hugo/content/gettingstarted/Design/tls.md
@@ -8,25 +8,87 @@ weight: 6
 ## TLS Configuration
 
 Should you desire to alter the default TLS settings for the Postgres Operator, you can set the
-following variables in bash:
+following variables as described below.
+
+### Server Settings
 
 To disable TLS and make an unsecured connection on port 8080 instead of connecting securely over
 the default port, 8443, set:
 
 Bash environment variables    
 
-    DISABLE_TLS=true
-    PGO_APISERVER_PORT=8080		
+```bash
+DISABLE_TLS=true
+PGO_APISERVER_PORT=8080		
+```
 
 Or inventory variables if using Ansible
 
-    pgo_disable_tls='true'
-    pgo_apiserver_port=8080
+```yaml
+pgo_disable_tls='true'
+pgo_apiserver_port=8080
+```
 
 To disable TLS verifcation, set the follwing as a Bash environment variable
 
-    export TLS_NO_VERIFY=false
+```bash
+export TLS_NO_VERIFY=false
+```
 
 Or the following in the inventory file is using Ansible
 
-    pgo_tls_no_verify='false'
+```yaml
+pgo_tls_no_verify='false'
+```
+
+### Connection Settings
+
+If TLS authentication has been disabled, or if the Operator's apiserver port is changed, be sure to 
+update the PGO_APISERVER_URL accordingly. 
+
+For example with an Ansible installation, 
+
+```bash
+export PGO_APISERVER_URL='https://<apiserver IP>:8443'
+```
+
+would become
+
+```bash
+export PGO_APISERVER_URL='http://<apiserver IP>:8080'
+```
+
+With a Bash installation,
+
+```bash
+setip()
+{
+   export PGO_APISERVER_URL=https://`$PGO_CMD -n "$PGO_OPERATOR_NAMESPACE" get service postgres-operator -o=jsonpath="{.spec.clusterIP}"`:8443
+}
+```
+
+would become
+
+```bash
+setip()
+{
+   export PGO_APISERVER_URL=http://`$PGO_CMD -n "$PGO_OPERATOR_NAMESPACE" get service postgres-operator -o=jsonpath="{.spec.clusterIP}"`:8080
+}
+```
+
+### Client Settings
+
+Finally, if TLS has been disabled for the Operator's apiserver, the PGO client connection must be set to match
+the given settings.
+
+Two options are available, either the Bash environment variable
+
+```bash
+DISABLE_TLS=true
+```
+
+must be configured, or the --disable-tls flag must be included when using the client, i.e.
+
+```bash
+pgo version --disable-tls
+```
--- a/pgo/cmd/auth.go
+++ b/pgo/cmd/auth.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	"time"
 
 	msgs "github.com/crunchydata/postgres-operator/apiservermsgs"
 	log "github.com/sirupsen/logrus"
@@ -33,6 +34,7 @@ const etcpath = "/etc/pgo/pgouser"
 const pgoUserFileEnvVar = "PGOUSER"
 const pgoUserNameEnvVar = "PGOUSERNAME"
 const pgoUserPasswordEnvVar = "PGOUSERPASS"
+const httpTimeout = 60
 
 // BasicAuthUsername and BasicAuthPassword are for BasicAuth, they are fetched from a file
 
@@ -42,6 +44,7 @@ var caCertPool *x509.CertPool
 var cert tls.Certificate
 var httpclient *http.Client
 var caCertPath, clientCertPath, clientKeyPath string
+var disabletls bool
 
 // StatusCheck ...
 func StatusCheck(resp *http.Response) {
@@ -175,14 +178,24 @@ func GetCredentialsFromEnvironment() msgs.BasicAuthCredentials {
 	}
 	return creds
 }
-func GetCredentials() {
-	log.Debug("GetCredentials called")
+
+// SetSessionUserCredentials gathers the pgouser and password information
+// and stores them for use by the PGO client
+func SetSessionUserCredentials() {
+	log.Debug("GetSessionCredentials called")
 
 	SessionCredentials = GetCredentialsFromEnvironment()
 
 	if !SessionCredentials.HasUsernameAndPassword() {
 		SessionCredentials = GetCredentialsFromFile()
 	}
+}
+
+// GetTLSCredentials instantiate gathers the necessary client connection
+// TLS settings for use by the PGO client.
+// If TLS is disabled, this will not be called.
+func GetTLSCredentials() *http.Transport {
+	log.Debug("GetCredentials called")
 
 	if PGO_CA_CERT != "" {
 		caCertPath = PGO_CA_CERT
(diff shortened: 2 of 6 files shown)


hugo/content/gettingstarted/Design/tls.md
```diff
@@ -29,4 +31,46 @@ To disable TLS verifcation, set the follwing as a Bash environment variable
...
+    {
+        export PGO_APISERVER_URL=https://`$PGO_CMD -n "$PGO_OPERATOR_NAMESPACE" get service postgres-operator -o=jsonpath="{.spec.clusterIP}"`:8443
+    }
+
+would become
+
+    setip()
+    {
+        export PGO_APISERVER_URL=http://`$PGO_CMD -n "$PGO_OPERATOR_NAMESPACE" get service postgres-operator -o=jsonpath="{.spec.clusterIP}"`:8080
+    }
```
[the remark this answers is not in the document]
tjmoore4: Updated to triple-backticks, used bash or yaml where appropriate.

hugo/content/gettingstarted/Design/tls.md
```diff
@@ -29,4 +31,46 @@ To disable TLS verifcation, set the follwing as a Bash environment variable
...
+
+
+### Client Settings
+
+Finally, if TLS has been disabled for the Operator's apiserver, the PGO client connection must be set to match
+the given settings.
+
+Two options are available, either the Bash environment variable
+
+    DISABLE_TLS=true
```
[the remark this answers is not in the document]
tjmoore4: Updated to triple-backticks, used bash or yaml where appropriate.

pgo/cmd/auth.go
```diff
@@ -184,74 +190,97 @@ func GetCredentials() {
...
-	caCertPool.AppendCertsFromPEM(caCert)
 
-	if PGO_CLIENT_CERT != "" {
-		clientCertPath = PGO_CLIENT_CERT
+	// If TLS is disabled, establish connection with
+	// a 60 second timeout.
+	if disabletls {
+		log.Debug("setting up httpclient without TLS")
+		httpclient = &http.Client{
+			Timeout:    60 * time.Second,
```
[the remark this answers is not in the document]
tjmoore4: Updated, set as `const httpTimeout = 60`.

tjmoore4:
Updated following recommendations. 

In auth.go,
GetCredentials is now only called when necessary and returns a http.Transport object.
As such, a new function was created GetSessionCredentials to pull out that functionality to set up the pgouser and password configuration calls.
SetHttpClient was created and instantiates the httpclient depending on whether the Transport object was initialized or not.

Documentation was updated to use triple-backticks and HTTP client timeout is now set in a constant.

pgo/cmd/auth.go
```diff
@@ -242,16 +261,40 @@ func GetCredentials() {
...
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{
+					RootCAs:            caCertPool,
+					InsecureSkipVerify: true,
+					Certificates:       []tls.Certificate{cert},
+					MinVersion:         tls.VersionTLS11,
+				},
+			},
+		}
+	}
```
jkatz: A few things:

1. Please document the function top, using `godoc` format
1. I'd recommend using a more descriptive variable name, e.g. `httpTransport` instead of `tr` (I'll refer to it as `tr` in the rest of the comments here for simplicity)
1. I'd recommend starting with the `tr != nil` case first, that seems to be "go" way of doing things. It will also be more concise.
1. This does not appear to work, the `tr` parameter passed in is not used in the `tr != nil` case. Example that is not gofmt'd:

```go
// If TLS is enabled, use the transport information already provided to
// establish a secure connection provided with the timeout specified
// in "httpTimeout"
//
// Otherwise, setup a HTTP connection with the timeout specified
// in "httpTimeout"
if tr != nil {
    httpclient = &http.Client{
        Timeout: httpTimeout * time.Second,
        Transport: tr
    }
} else {
    httpclient = &http.Client{
        Timeout:    httpTimeout * time.Second,
    }
}
```
(with appropriate debugging lines)

hugo/content/gettingstarted/Design/tls.md
```diff
@@ -8,25 +8,71 @@ weight: 6
...
-following variables in bash:
+following variables as described below.
+
+### Server Settings
 
 To disable TLS and make an unsecured connection on port 8080 instead of connecting securely over
 the default port, 8443, set:
 
 Bash environment variables    
+```bash
```
jkatz: I'd recommend putting some space between the paragraphs + the coding blogs, e.g.:

```
Bash environment variables

``bash
DISABLE_TLS=true
``
```

pgo/cmd/auth.go
```diff
@@ -175,15 +178,31 @@ func GetCredentialsFromEnvironment() msgs.BasicAuthCredentials {
...
+// If TLS is disabled, this will not be called.
+func GetCredentials() *http.Transport{
+	log.Debug("GetCredentials called")
 
+	SessionCredentials = GetCredentialsFromEnvironment()
+
+	if !SessionCredentials.HasUsernameAndPassword() {
+		SessionCredentials = GetCredentialsFromFile()
+	}
+
```
jkatz: I don't believe this is needed here?

pgo/cmd/auth.go
```diff
@@ -175,15 +178,31 @@ func GetCredentialsFromEnvironment() msgs.BasicAuthCredentials {
 	}
 	return creds
 }
-func GetCredentials() {
-	log.Debug("GetCredentials called")
+
+// GetSessionCredentials gathers the pgouser and password information
+// and stores them for use by the PGO client
+func GetSessionCredentials(){
```
jkatz: This is technically not a `Get` this is a `Set` -- I'd recommend we call it `SetSessionUserCredentials` :smile:

pgo/cmd/auth.go
```diff
@@ -175,15 +178,31 @@ func GetCredentialsFromEnvironment() msgs.BasicAuthCredentials {
...
 
 	if !SessionCredentials.HasUsernameAndPassword() {
 		SessionCredentials = GetCredentialsFromFile()
 	}
+}
+
+// GetCredentials instantiate gathers the necessary client connection 
+// TLS settings for use by the PGO client.
+// If TLS is disabled, this will not be called.
+func GetCredentials() *http.Transport{
```
jkatz: Let's rename to `GetTLSCredentials`

hugo/content/gettingstarted/Design/tls.md
tjmoore4: Added the extra spaces. Verified that this did not impact the formatting when built with Hugo.

pgo/cmd/auth.go
tjmoore4: Updated for all items as requested.

hugo/content/gettingstarted/Design/tls.md
tjmoore4: Added space between sections, verified that Hugo renders the page as expected after changes.

pgo/cmd/auth.go
tjmoore4: Correct, this is called earlier in the new function. This instance is now removed.

pgo/cmd/auth.go
tjmoore4: Makes sense to me, updated!

pgo/cmd/auth.go
tjmoore4: Sure thing, updated.

tjmoore4:
In addition to the main PR commit, a second commit has been added for formatting in line with Go formatting guidelines.

pgo/cmd/auth.go
```diff
@@ -242,16 +256,33 @@ func GetCredentials() {
...
-	httpclient = &http.Client{
-		Transport: &http.Transport{
-			TLSClientConfig: &tls.Config{
-				RootCAs:            caCertPool,
-				InsecureSkipVerify: true,
-				Certificates:       []tls.Certificate{cert},
-				MinVersion:         tls.VersionTLS11,
-			},
+	// create a Transport object for use by the HTTP client
+	httpTransport := &http.Transport{
```
jkatz: we can just `return` here!

pgo/cmd/auth.go
```diff
@@ -55,7 +59,7 @@ func StatusCheck(resp *http.Response) {
 	}
 }
 
-func UserHomeDir() string {
+func userHomeDir() string {
```
jkatz: Great change! Please document this function if we're going to be renaming it ;-)

pgo/cmd/auth.go
```diff
@@ -96,9 +100,9 @@ func parseCredentials(dat string) msgs.BasicAuthCredentials {
 	return creds
 }
 
-func GetCredentialsFromFile() msgs.BasicAuthCredentials {
+func getCredentialsFromFile() msgs.BasicAuthCredentials {
```
jkatz: Great change! Please document this function if we're going to be renaming it ;-)

pgo/cmd/auth.go
```diff
@@ -155,7 +159,7 @@ func GetCredentialsFromFile() msgs.BasicAuthCredentials {
 	return creds
 }
 
-func GetCredentialsFromEnvironment() msgs.BasicAuthCredentials {
+func getCredentialsFromEnvironment() msgs.BasicAuthCredentials {
```
jkatz: Great change! Please document this function if we're going to be renaming it ;-)

pgo/cmd/root.go
```diff
@@ -82,7 +84,18 @@ func initConfig() {
...
+
+	// If TLS is disabled, do not gather unneeded credentials
+	if PGO_DISABLE_TLS == true || os.Getenv("DISABLE_TLS") == "true" {
+		SetHTTPClient(nil)
+	} else {
+		// If TLS is enabled, gather credentials and supply them for the
+		// client connection
+		httpTransport := GetTLSCredentials()
+		SetHTTPClient(httpTransport)
+	}
```
jkatz: Suggestion:

```go
func initConfig() {
    var httpTransport http.Transport

    // blah...

    // If TLS is enabled, get the TLS credentials and set the httpTransport
    // object required to initialize the client
    if PGO_DISABLE_TLS != true && os.Getenv("DISABLE_TLS") != "true" {
        httpTransport = GetTLSCredentials()
    }

    SetHTTPClient(httpTransport)
}
```

(assuming I got DeMorgan's Law correct here)

pgo/cmd/auth.go
tjmoore4: We can, and we did!

pgo/cmd/auth.go
tjmoore4: Function is now documented.

pgo/cmd/auth.go
tjmoore4: Function is now documented.

pgo/cmd/auth.go
tjmoore4: Function is now documented.

pgo/cmd/root.go
tjmoore4: Updated to match, testing confirms the client still functions as expected.

PR #1071, opened 2019-10-29
Add disable TLS functionality to the PGO Client

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
The Postgres Operator's apiserver currently supports the
disabling of TLS authentication when receiving client connections.
However, the PGO client did not have a method for disabling
the necessary resource checks that are then unnecessary, meaning
the unneeded certificates and keys were still required to be
present.

This update allows the PGO client to skip the relevant configuration
setup when using an unauthenticated connection. This can be done in
one of two ways. Either by an environment variable:

DISABLE_TLS=true

(As is used by the Postgres Operator to disable TLS when performing
a Bash install)

Or by using the '--disable-tls' flag with the PGO client, for example:

pgo version --disable-tls

In the event that the DISABLE_TLS variable is set to 'false' and the
--disable-tls is used, the flag is authoritative.

Additionally, this commit includes documenationupdates to the main
TLS configuration page to explain how to properly set the new PGO
CLI configuration variables depending on installation method.

Notes with links to the main page were also added to both
the Bash and Ansible installation sections.

This PR was tested using an Ansible installation of the Postgres Operator with Kubernetes
and using a Bash installation of the Postgres Operator with Openshift.

**Other information**:
For testing, as this update does not add any new functions, unit testing is not ideal. The following steps can be followed to validate the appropriate behavior.

First, install the Postgres Operator with TLS enabled (the default setting) as described here: https://access.crunchydata.com/documentation/postgres-operator/4.1.0/installation/

Validate that commands work as expected, for instance

pgo version

returns the client and server versions

Now, either set the environment variable, DISABLE_TLS=true, or use

pgo version --disable-tls

The command should now fail.

At this point, update the PGO installation to disable TLS authentication (as described here https://access.crunchydata.com/documentation/postgres-operator/4.1.0/gettingstarted/design/tls/) and unset the TLS specific environment variables:

unset PGO_CA_CERT
unset PGO_CLIENT_CERT
unset PGO_CLIENT_KEY

Now running 'pgo version' without setting either the DISABLE_TLS variable or using the --disable-tls flag should result in an error and failure to return the expected values.

Issue: [ch6127]

### DIFF
--- a/hugo/content/gettingstarted/Design/tls.md
+++ b/hugo/content/gettingstarted/Design/tls.md
@@ -8,25 +8,87 @@ weight: 6
 ## TLS Configuration
 
 Should you desire to alter the default TLS settings for the Postgres Operator, you can set the
-following variables in bash:
+following variables as described below.
+
+### Server Settings
 
 To disable TLS and make an unsecured connection on port 8080 instead of connecting securely over
 the default port, 8443, set:
 
 Bash environment variables    
 
-    DISABLE_TLS=true
-    PGO_APISERVER_PORT=8080		
+```bash
+DISABLE_TLS=true
+PGO_APISERVER_PORT=8080		
+```
 
 Or inventory variables if using Ansible
 
-    pgo_disable_tls='true'
-    pgo_apiserver_port=8080
+```
+pgo_disable_tls='true'
+pgo_apiserver_port=8080
+```
 
 To disable TLS verifcation, set the follwing as a Bash environment variable
 
-    export TLS_NO_VERIFY=false
+```bash
+export TLS_NO_VERIFY=false
+```
 
 Or the following in the inventory file is using Ansible
 
-    pgo_tls_no_verify='false'
+```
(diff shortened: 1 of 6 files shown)


hugo/content/gettingstarted/Design/tls.md
```diff
@@ -8,25 +8,87 @@ weight: 6
...
-    export TLS_NO_VERIFY=false
+```bash
+export TLS_NO_VERIFY=false
+```
 
 Or the following in the inventory file is using Ansible
 
-    pgo_tls_no_verify='false'
\ No newline at end of file
+```yaml
```
jkatz: This is not YAML...I'd suggest either `python` (as it somewhat follows Python) or just leave the modifier off.

tjmoore4: Good catch, thanks. I removed the modifier.

PR #1076, opened 2019-11-02
Update `pgo test` to check connectivity on instances & endpoints ([ch5712])

jkatz:
**Checklist:**

 
**Type of Changes:**

 
Previously, `pgo test` would check if each individual user available to a PostgreSQL cluster could authenticate with its credentials. While this certainly determines that one can make a connection to a PostgreSQL database managed by the Operator, it also presents several issues:

- If one has many users available to the Operator, the number of connections made can rapidly multiply. Imagine 10 users across 5 replicas
- This also means that the user's credentials must be stored in the Operator. Currently this is a plaintext password. If the connections are _not_ being made over SSL, this creates the risk of password leakage.

Update the `pgo test` to do the following:

- Provide an instance (pod) level readiness check. The pod readiness check already calls `pg_isready`, which checks to see if a PostgreSQL instance is available to accept connections
- Check the service endpoints to see if they are ready to accept connections.

These two checks provide the equivalent behavior of what is currently happening today with `pgo test` but in a more efficient manner with less potential for information leakage.

The API endpoint now also serves content that is in a structured format.

Additionally, the event that was fired when "test event" is called is removed. This does not affect any lifecycle activity of a PostgreSQL cluster and as it is fired ad-hoc by a user, does not need to be propagated downstream.

**Other information**:

Here is a test case:

### Client #1:
```bash
watch pgo test -n jkatz jkatz
```

### Client #2:
```bash
pgo create namespace -n jkatz
# observe Client #1
pgo create cluster -n jkatz jkatz
# observe Client #1
pgo scale cluster -n jkatz jkatz
# observe Client #1
```

### DIFF
--- a/apiserver/clusterservice/clusterimpl.go
+++ b/apiserver/clusterservice/clusterimpl.go
@@ -405,7 +405,7 @@ func TestCluster(name, selector, ns, pgouser string, allFlag bool) msgs.ClusterT
 				Timestamp: time.Now(),
 				EventType: events.EventTestCluster,
 			},
-			Clustername:       c.Name,
+			Clustername: c.Name,
 		}
 
 		err = events.Publish(f)
--- a/apiserver/clusterservice/clusterimpl.go
+++ b/apiserver/clusterservice/clusterimpl.go
@@ -31,7 +31,6 @@ import (
 
 	msgs "github.com/crunchydata/postgres-operator/apiservermsgs"
 	"github.com/crunchydata/postgres-operator/config"
-	"github.com/crunchydata/postgres-operator/events"
 	"github.com/crunchydata/postgres-operator/kubeapi"
 	"github.com/crunchydata/postgres-operator/sshutil"
 	"github.com/crunchydata/postgres-operator/util"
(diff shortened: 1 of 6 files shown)


pgo/cmd/test.go
```diff
@@ -108,16 +110,28 @@ func showTest(args []string, ns string) {
...
-					fmt.Printf("%s\n", GREEN("Working"))
-				} else {
-					fmt.Printf("%s\n", RED("NOT working"))
-				}
-			}
+			fmt.Println(fmt.Sprintf("cluster : %s", result.ClusterName))
+
+			// first, print the test results for the endpoints
+			printTestResults("Endpoints", result.Endpoints)
+			// first, print the test results for the instances
```
andrewlecuyer: I suggest using `Services` for the header here instead of `Endpoints`, e.g.:

```bash
cluster : mycluster2
        Services
                primary (10.104.67.174:5432): UP
```

This is because validating the addresses in the `endpoint` is really validating that the selector defined in the service actually matches a pod, i.e. that the endpoint contains what should be the IPs for the primary, replica, pgbouncer, etc. pods, effectively making them available via the service.  So really this is validating that the service has been properly configured to send traffic to the appropriate pod (which also results in the automatic creation of the proper endpoints).

Also the IP address displayed in the output is the `ClusterIP` of the service and not an endpoint (i.e. Pod) IP, and since the `endpoint` is typically invisible to the end user and the `service` is the resource a user would interact with directly, it seems appropriate to show that the `service` is up vs. the `endpoint`.

apiserver/clusterservice/clusterimpl.go
```diff
@@ -268,153 +267,200 @@ func getServices(cluster *crv1.Pgcluster, ns string) ([]msgs.ShowClusterService,
...
+			result.Instances = append(result.Instances, instance)
+			response.Results = append(response.Results, result)
+			continue
 		}
-		if !primaryReady {
-			response.Status.Code = msgs.Error
-			response.Status.Msg = "cluster not ready yet, try later"
-			return response
+
+		// Check each instance (i.e. pod) to see if its readyness check passes.
```
andrewlecuyer: Should be "readiness" instead of "readyness"

jkatz:
@andrewlecuyer Thanks for the review. I've made modifications per your suggestions.

PR #1079, opened 2019-11-04
Update prometheus-config.yaml.j2

JuozasA:
Source label must be service_name instead of pod_name otherwise after failover (pod name is still *-replica but the service is primary), Grafana is not changing the PostgreSQL Overview so even though the replica pod becomes primary, it shows it as replica and primary as down...

Tested on Openshift 3.10 env.

**Checklist:**

 
**Type of Changes:**

 
After failover Grafana shows new primary as Replica and Primary as being Down

After failover, replica pod becomes primary and refers to primary service which is being monitored by prometheus

**Other information**:


jkatz:
This is superseded by the 4.5.0 release, which redid the monitoring plumbing. Thanks!

PR #1096, opened 2019-11-16
[ch6317] Ensure disabling BasicAuth skips Basic Authentication check.

jkatz:
**Checklist:**

 
**Type of Changes:**

 
Presently, disabling BasicAuth in the pgo.yaml configuration file would
not actually disable the HTTP Basic Authentication from occuring, as
the check against authorization headers provided by the HTTP requests
would still be scanned.

This ensures this check is skipped when BasicAuth is set to `"false"`.

However, skipping Basic Authentication does not skip authorization, as
the Operator heavily leverages RBAC checks, and as such, a valid username
is required at all times even if BasicAuth is skipped. As such, this fix
only solves one type of error, i.e. the case where no HTTP Authorization
headers are sent to the Operator apiserver. And by "fix," I mean it just
moves the failure from the authentication check to the authorization
check.

**Other information**:

Issue: [ch6317]

### DIFF
--- a/apiserver/root.go
+++ b/apiserver/root.go
@@ -313,19 +313,35 @@ func GetNamespace(clientset *kubernetes.Clientset, username, requestedNS string)
 	return requestedNS, errors.New("requested Namespace was not found to be in the list of Namespaces being watched.")
 }
 
+// Authn performs HTTP Basic Authentication against a user if "BasicAuth" is set
+// to "true" (which it is by default).
+//
+// ...it also performs Authorization (Authz) against the user that is attempting
+// to authenticate, and as such, to truly "authenticate/authorize," one needs
+// at least a valid Operator User account.
 func Authn(perm string, w http.ResponseWriter, r *http.Request) (string, error) {
 	var err error
 	w.Header().Set("WWW-Authenticate", `Basic realm="Restricted"`)
 
+	// Need to run the HTTP library `BasicAuth` even if `BasicAuth == false`, as
+	// this function currently encapsulates authorization as well, and this is
+	// the call where we get the username to check the RBAC settings
 	username, password, authOK := r.BasicAuth()
 	if AuditFlag {
 		log.Infof("[audit] %s username=[%s] method=[%s] ip=[%s] ok=[%t] ", perm, username, r.Method, r.RemoteAddr, authOK)
 	}
 
-	log.Debugf("Authentication Attempt %s username=[%s]", perm, username)
-	if authOK == false {
-		http.Error(w, "Not authorized", 401)
-		return "", errors.New("Not Authorized")
+	// Check to see if this user is authenticated
+	// If BasicAuth is "disabled", skip the authentication; o/w/ check if the
+	// authentication passed
+	if !BasicAuth {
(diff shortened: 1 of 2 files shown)


apiserver/root.go
```diff
@@ -313,19 +313,35 @@ func GetNamespace(clientset *kubernetes.Clientset, username, requestedNS string)
...
-		return "", errors.New("Not Authorized")
+	// Check to see if this user is authenticated
+	// If BasicAuth is "disabled", skip the authentication; o/w/ check if the
+	// authentication passed
+	if !BasicAuth {
+		log.Debugf("BasicAuth disabled, Skipping Authentication %s username=[%s]", perm, username)
+	} else {
+		log.Debugf("Authentication Attempt %s username=[%s]", perm, username)
+		if !authOK {
+			http.Error(w, "Not authorized", 401)
```
andrewlecuyer: Nothing urgent for this patch (and realize the same error message from before is being leveraged), but just making a note that a better error message here could be beneficial in the future.  For instance, since `authOK` should essentially only be false if the basic auth credentials could not be properly parsed from the `Authorization` header (e.g. if they are missing), maybe something along the lines of `Not authorized. Basic authentication credentials must be provided according to RFC 2617, Section 2`.

apiserver/root.go
```diff
@@ -313,19 +313,35 @@ func GetNamespace(clientset *kubernetes.Clientset, username, requestedNS string)
...
-		http.Error(w, "Not authorized", 401)
-		return "", errors.New("Not Authorized")
+	// Check to see if this user is authenticated
+	// If BasicAuth is "disabled", skip the authentication; o/w/ check if the
+	// authentication passed
+	if !BasicAuth {
+		log.Debugf("BasicAuth disabled, Skipping Authentication %s username=[%s]", perm, username)
+	} else {
+		log.Debugf("Authentication Attempt %s username=[%s]", perm, username)
+		if !authOK {
```
andrewlecuyer: My only concern here is that by disabling this check when `BasicAuth` is `false`, we are no longer checking if we were able to properly obtain the username from the `Authorization` header, which is still needed even when `BasicAuth` is disabled.  But I also see that returning an error indicating that Basic Authentication credentials are required when `BasicAuth` is `false` is also confusing to the end user, so I do see why the check isn't enabled in that scenario.

So ultimately I think this is really just a side-effect of the current implementation (which is a bit confusing) and not necessarily an issue at this time, but I did want to at least make a note.  And the bottom line is that `BasicAuthzCheck` will eventually fail anyway if the proper credentials have not been provided in the `Authorization` header.

jkatz: @andrewlecuyer If BasicAuth is disabled, then you have to assume that someone might not even pass in a username, and then will fail at a later step.

..and AFAICT a username AND password are required for Basic Authentication per [RFC7617](https://tools.ietf.org/html/rfc7617), so really if it is disabled we should skip the check.

Additionally, we still need to

apiserver/root.go
jkatz: @andrewlecuyer Noting that it's 7617, which obsoleted 2617 :wink:

PR #1104, opened 2019-11-22
Fully Replace Custom PGO HA Solution with crunchy-postgres-ha Solution

andrewlecuyer:
This commit effectively removes the custom PGO autofail solution from the PGO.  In its place, Patroni (contained within the crunchy-postgres-ha and crunchy-postgres-gis-ha containers) will now be responsible for ensuring any PG clusters created by the PGO are highly available.  This includes automatically failing over to (i.e. promoting) a replica in the event of a failure in the primary PG database.  With this change, once bootstrapped, PG clusters are now self-sufficient and able to self-heal without relying directly on the PGO itself and its availability.

In support of the transition to this new high availability solution are the following notable design changes:
- Autofail is now enabled by default for any newly created cluster, and a warning is returned from the API server in the event that the new cluster contains zero replicas (specifically indicating that at least one replica is needed to leverage autofail functionality
- Disabling/enabling autofail means "pausing"/"resuming" Patroni (i.e. attaching/detaching Patroni from the cluster). This is accompished by setting/removing the `{"pause": true/false/null}` key in the cluster's "config" `configMap`.
- With autofail now enabled by default, the "autofail" flag for the `pgo create` and `pgo update` commands has been changed to `disable-autofail`.  When set, this option disables autofail in the newly created cluster (however, it can still later be enabled).  Similarly, the `Autofail` option in `pgo.yaml` has been changed to 'DisableAutofail'
- The `AutofailReplaceReplica` failover option has been removed from the API server and CLI (i.e. the `autofail-replace-replica` flag, since it is no longer applicable to the current design.  Specifically, with the `crunchy-postgres-ha`  container, the old primary will now automatically be turned into a replica that will follow the new primary (with pg_rewind utilized if needed)

Manual failover capabilities are now also implemented in support of the `crunchy-postgres-ha` container.  Specifically, the proper Patroni REST endpoint is now called via the exec endpoint in order to failover to a replica.  Additionally, when any failover occurs, whether automatic or manual, the role label change on the new primary from "replica" to "master" is detected, and the pgBackRest dedicated repository host is updated to point to the proper "PGDATA" directory for the promoted DB node.  And finally, in support of proper replica selection for failover functionality, label selectors have been updated for the `--query` options for `pgo scaledown` and `pgo failover` to return the proper cluster replicas (including after a failover), while `pgo show cluster` also now properly displays primary and replicas following both before and after failover.

**Checklist:**

 
**Type of Changes:**

 
A custom PGO HA solution is included in the PGO solution.

[ch5460]
[ch6325]

The custom PGO HA solution included in the PGO has been fully replaced by the `crunchy-postgres-ha` container and Patroni, and the PGO has been updated as needed to support this new HA soltuion.

**Other information**:


pgo/cmd/create.go
```diff
@@ -264,7 +264,7 @@ func init() {
 	//	createClusterCmd.Flags().BoolVarP(&ArchiveFlag, "archive", "", false, "Enables archive logging for the database cluster.")
 	createClusterCmd.Flags().StringVarP(&PgpoolSecret, "pgpool-secret", "", "", "The name of a pgpool secret to use for the pgpool configuration.")
 	createClusterCmd.Flags().BoolVarP(&MetricsFlag, "metrics", "", false, "Adds the crunchy-collect container to the database pod.")
-	createClusterCmd.Flags().BoolVarP(&AutofailFlag, "autofail", "", false, "If set, will cause autofailover to be enabled on this cluster.")
+	createClusterCmd.Flags().BoolVarP(&AutofailFlag, "disable-autofail", "", false, "If set, will cause autofailover to be enabled on this cluster.")
```
abrightwell: I think you mean the opposite in this description string, right?

pgo/cmd/update.go
```diff
@@ -34,7 +34,7 @@ func init() {
 
 	UpdateClusterCmd.Flags().BoolVar(&NoPrompt, "no-prompt", false, "No command line confirmation.")
 	UpdateClusterCmd.Flags().BoolVar(&AllFlag, "all", false, "all resources.")
-	UpdateClusterCmd.Flags().BoolVar(&AutofailFlag, "autofail", false, "autofail default is false.")
+	UpdateClusterCmd.Flags().BoolVar(&AutofailFlag, "disable-autofail", false, "autofail default is false.")
```
abrightwell: Should it be "disable-autofail default is false" instead?

util/failover.go
```diff
@@ -316,3 +302,69 @@ func GetPreferredNodes(clientset *kubernetes.Clientset, selector, namespace stri
...
+// Patroni to stop responding to failures or other database activities, e.g. it will not attempt to start the
+// database when stopped to perform maintenance
+func ToggleAutoFailover(clientset *kubernetes.Clientset, enable bool, clusterName, namespace string) error  {
+	
+	// find the "config" configMap created by Patroni
+	configMapName := clusterName+"-config"
+	log.Debugf("setting autofailover to %t for cluster %s", enable, clusterName)
+
+	configMap, found := kubeapi.GetConfigMap(clientset, configMapName, namespace)
+	if found {
```
abrightwell: Could we 'short-circuit' here? It would at least reduce on level of nested conditionals.

Something like this perhaps?

```
if !found {
  log(err)  
  return err
}

... other stuff ...
```

operator/backrest/restore.go
```diff
@@ -71,15 +71,11 @@ func Restore(restclient *rest.RESTClient, namespace string, clientset *kubernete
...
-		err = kubeapi.Updatepgcluster(restclient, &cluster, clusterName, namespace)
-		if err != nil {
-			log.Errorf("restore workflow error: could not turn off autofail on %s", clusterName)
-			return
-		}
+	// disable autofail if it is currently enabled
+	err = util.ToggleAutoFailover(clientset, false, clusterName, namespace)
+	if err != nil {
+		log.Error(err)
+		return
```
abrightwell: just another stylistic thing:

```
if err := util.ToggleAutoFailover(...); err != nil {
    log(err)
    return
}
```

controller/podcontroller.go
```diff
@@ -239,6 +229,11 @@ func (c *PodController) checkReadyStatus(oldpod, newpod *apiv1.Pod, cluster *crv
 						found, err := kubeapi.Getpgtask(c.PodClient, &tmptask, clusterName+"-stanza-create", newpod.ObjectMeta.Namespace)
 						if !found && err != nil {
 							backrestoperator.StanzaCreate(newpod.ObjectMeta.Namespace, clusterName, c.PodClientset, c.PodClient)
+							
+							// Disable autofailover if the autofail label is set to "false" on the pgcluster
+							if cluster.ObjectMeta.Labels[config.LABEL_AUTOFAIL] == "false" {
+								 util.ToggleAutoFailover(c.PodClientset, false, cluster.Name, cluster.ObjectMeta.Namespace)
+							}
```
abrightwell: Would this be 'cleaner'? It sounds like the intent it to ensure that the toggle is equivalent to the value of the label. So, why test for it and not just set it to the value? If the label is anything other than ```1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False``` then it will just move on.

```
autofail, err := strconv.ParseBool(cluster.ObjectMeta.Labels[config.LABEL_AUTOFAIL])
if err != nil {
    util.ToggleAutoFailover(c.PodClientset, autofail, cluster.Name, cluster.ObjectMeta.Namespace)
}
```

https://golang.org/pkg/strconv/#ParseBool

controller/clustercontroller.go
```diff
@@ -174,24 +175,10 @@ func (c *PgclusterController) onUpdate(oldObj, newObj interface{}) {
...
-				if err != nil {
-					log.Error(err.Error())
-					return
-				}
-				//call the autofail logic on this cluster
-				clusteroperator.AutofailBase(c.PgclusterClientset, c.PgclusterClient, ready, newcluster.ObjectMeta.Name, newcluster.ObjectMeta.Namespace)
+			if newValue == "true" {
+				util.ToggleAutoFailover(c.PgclusterClientset, true, newcluster.Name, newcluster.ObjectMeta.Namespace)
+			} else if newValue == "false" {
+				util.ToggleAutoFailover(c.PgclusterClientset, false, newcluster.Name, newcluster.ObjectMeta.Namespace)
```
abrightwell: Couldn't this be condensed down to:

```
autofail, err := strconv.ParseBool(newValue)

if err != nil { 
    log(err)
    return
}

util.ToggleAutoFailover(..., autofail, ...)
```

apiserver/failoverservice/failoverimpl.go
```diff
@@ -243,13 +190,16 @@ func validateDeploymentName(deployName, clusterName, ns string) (*v1.Deployment,
 	}
 
 	//make sure the primary is not being selected by the user
```
abrightwell: This is more of a comment about the function over all and not necessarily the changes at hand. The function name is a bit confusing, IMO. Could we add some documentation to clear up what this function is doing?

apiserver/common.go
```diff
@@ -52,10 +52,14 @@ func GetSecrets(cluster *crv1.Pgcluster, ns string) ([]msgs.ShowUserSecret, erro
...
 }
 
-func GetPodStatus(deployName, ns string) (string, string) {
-
-	//get pods with replica-name=deployName
-	pods, err := kubeapi.GetPods(Clientset, config.LABEL_REPLICA_NAME+"="+deployName, ns)
+// GetReplicaPodStatus gets the status of all replica pods in the cluster. Specifically, using
+// the provided cluster name and namespace, it looks up all replica pod in the cluster, and then
+// provides a status for each pod ("Ready" or "Not Ready")
+func GetReplicaPodStatus(clusterName, ns string) (string, string) {
```
abrightwell: Would it be too much involved of a refactor to perhaps turn this signature into something like:

```
func GetReplicaPodStatus(clusterName, ns string) (ReplicaPodStatus, error) { ... }
```

Where,

```
type ReplicaPodStatus struct {
    Name string
    Ready bool
}
```

So, that:

```
if c.Name == "database" {
    return ReplicaPodStatus{Name: nodeName, Status: c.Ready}, nil
}
```

And:

```
var (
    ErrReplicaNotFound = errors.New("replica not found")
    ...
)
...
func GetReplicaPodStatus(...) (ReplicaPodStatus, error) {
return ReplicPodStatus{}, ErrReplicaNotFound  // or other applicable 'error'
}
```

hugo/content/overview/failover-overview.md
```diff
@@ -29,11 +29,6 @@ Manual failover is performed by PostgreSQL Operator API actions involving a *que
...
 
-The PostgreSQL Operator automatic failover logic includes:
-
- * deletion of the failed primary Deployment
- * pick the best replica to become the new primary
- * label change of the targeted Replica to match the primary Service
- * execute the PostgreSQL promote command on the targeted replica
+When a failover does ocurr, the system automatically attempts to turn the old primary into a replica (using `pg_rewind` if needed), ensuring the cluster maintains the same amount of database pods and replicas following a failure.  Additionally, the `role` label on each pod is updated as needed to properly identify the `master` pod and any `replica` pods following a failover event, therefore ensuring the primary and replica services point to the proper database pods.  And finally, the `pgBackRest` dedicated repository host is also automatically reconfigured to point to the `PGDATA` directory of the new `primary` pod following a failover.
```
jkatz: `ocurr` => `occur`

(I'll likely make a wordsmith pass at some point)

pgo/cmd/create.go
jkatz: I would double up on this, and just say let's rename it to `DisableAutofailFlag`.

For the CLI, this seems fair enough to ensure we capture the correct functionality internally and not rely on `!Autofail`

For the API we can leave it as `Autofail` given we are saying `true`/`false` there.

util/failover.go
```diff
@@ -128,33 +129,17 @@ func GetPod(clientset *kubernetes.Clientset, deploymentName, namespace string) (
 	return pod, err
 }
 
-func GetRepStatus(restclient *rest.RESTClient, clientset *kubernetes.Clientset, dep *appsv1.Deployment, namespace, databasePort string) (uint64, uint64, string, error) {
+func GetRepStatus(restclient *rest.RESTClient, clientset *kubernetes.Clientset, pod *v1.Pod, namespace, databasePort string) (uint64, uint64, string, error) {
```
jkatz: Please add godoc for what this function does.

util/failover.go
```diff
@@ -316,3 +302,69 @@ func GetPreferredNodes(clientset *kubernetes.Clientset, selector, namespace stri
...
+				configMap.ObjectMeta.Annotations["config"] = string(configJSONFinalStr)
+				err = kubeapi.UpdateConfigMap(clientset, configMap, namespace)
+				if err != nil {
+					return err
+				}
+			} else {
+				log.Debugf("autofailover already enabled according to the pause key (or lack thereof) in configMap %s", 
+					configMapName)
+			}
+		}
```
jkatz: I'd recommend breaking these two clauses out into `disableAutofail` and `enableAutofail` helper functions.

apiserver/clusterservice/clusterimpl.go
```diff
@@ -739,13 +739,14 @@ func CreateCluster(request *msgs.CreateClusterRequest, ns, pgouser string) msgs.
...
 
-		//verify that for autofail clusters we have a replica requested
+		// include a warning in the response indicating that autofail capabilities will not be able to be utilized
+		// until at least one replica is added to the cluster 
 		if newInstance.Spec.Replicas == "0" {
-			if request.AutofailFlag || apiserver.Pgo.Cluster.Autofail {
-				resp.Status.Code = msgs.Error
-				resp.Status.Msg = "replica-count of 1 or more is required when autofail is specified"
-				return resp
+			noReplicasResult := "Warning: Please add at least one replica to utilize autofail functionality"
```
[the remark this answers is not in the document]
andrewlecuyer: I will go ahead and remove (it was more that a user couldn't create a single node cluster with autofail enabled in the past that I was concerned could cause confusion since they are now able to).

apiserver/clusterservice/clusterimpl.go
```diff
@@ -739,13 +739,14 @@ func CreateCluster(request *msgs.CreateClusterRequest, ns, pgouser string) msgs.
...
+			if !request.AutofailFlag || apiserver.Pgo.Cluster.DisableAutofail {
```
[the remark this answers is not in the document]
andrewlecuyer: Just to clarify, the only thing that changed was the CLI, and not the API Server.  As for the old CLI option, I was hesitant to call it deprecated, since it is really gone/removed at this point.   Do you want me to add the old `autofail` CLI option back in and wire it up (also marking for deprecation), so the user can still use the old option in addition to the new `--disable-autofail` option?

apiserver/clusterservice/clusterimpl.go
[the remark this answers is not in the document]
jkatz: :+1:

apiserver/clusterservice/clusterimpl.go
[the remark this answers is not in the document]
jkatz: @andrewlecuyer I see my confusion. I would just leave this as is and ignore me.

andrewlecuyer:
@jkatz & @abrightwell all review items have been addressed.  Please let me know if any additional questions, and otherwise appreciate the review.

Also, just a quick note that I have been trying to be careful about keeping refactoring within scope of the changes I am currently making, especially to avoid muddying the waters (i.e. introducing regressions) with some of the large-scale changes being implemented to support the new `crunchy-postgres-ha` container (e.g. replacing the entire autofail solution in this instance).  So while I went ahead and updated `GetReplicaPodStatus`, for instance, I realize there are other similar changes required across our API server as well that will be planned for updates in future releases.

And just one more note that I will be following-up with a containers PR to address some strange behavior that can result from backing up and then restoring while autofail is disabled (i.e. Patroni is "paused").  I am going to recommend that we remove `--disable-autofail` from `pgo create` alltogether as a result of this behavior as well, to protect against attempting to operate clusters in this detached/maintenance state.

util/failover.go
```diff
@@ -316,3 +309,85 @@ func GetPreferredNodes(clientset *kubernetes.Clientset, selector, namespace stri
...
+			configMap.Name)
+	}
+	return nil
+}
+
+// If "pause" isn't present in the config then assume autofail is enabled and needs to be disabled by setting "pause
+// to true.  Or if it is present and set to "false", then it also needs to be disabled by setting "pause" to true.
+func disableFailover(clientset *kubernetes.Clientset, configMap *v1.ConfigMap, configJSON map[string]interface{}, 
+	namespace string) error {
+	if _,ok := configJSON["pause"] ; !ok || configJSON["pause"] == false  {
```
abrightwell: Same thoughts here as above.

apiserver/common.go
```diff
@@ -52,28 +65,33 @@ func GetSecrets(cluster *crv1.Pgcluster, ns string) ([]msgs.ShowUserSecret, erro
 	return output, err
 }
 
-func GetPodStatus(deployName, ns string) (string, string) {
+// GetReplicaPodStatus gets the status of all replica pods in the cluster. Specifically, using
+// the provided cluster name and namespace, it looks up all replica pod in the cluster, and then
+// provides a status for each pod ("Ready" or "Not Ready")
+func GetReplicaPodStatus(clusterName, ns string) (*ReplicaPodStatus, error) {
```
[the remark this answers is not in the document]
jkatz: @abrightwell I believe the reasoning is to be able to return a `nil` in an error case, so we don't have to pass the full struct around.

If there is something that is terribly breaking with this approach, we could refactor, but I suggest that we leave it and address it with future API updates.

apiserver/common.go
```diff
@@ -27,7 +28,19 @@ import (
...
+	// ErrDBContainerNotFound is an error that indicates that a "database" container 
+	// could not be found in a specific pod
+	ErrDBContainerNotFound = errors.New("\"database\" container not found in pod")
+)
+
+// ReplicaPodStatus stores the name of the node a replica pod is assigned to, as well
+// as whether or not the pod is considered "Ready" in the Kubernetes cluster
+type ReplicaPodStatus struct {
+    NodeName string
+    ReadyStatus string
```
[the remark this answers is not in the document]
jkatz: @abrightwell I would recommend handling this in a separate PR as it touches a bunch of different points.

util/failover.go
andrewlecuyer: I think the logic holds here, i.e. I want to take action if the `pause` key doesn't exist in the map at all, or if it exists but is set to `false`.

util/failover.go
```diff
@@ -316,3 +309,85 @@ func GetPreferredNodes(clientset *kubernetes.Clientset, selector, namespace stri
...
+
+	return nil
+}
+
+// If "pause" is present in the config and set to "true", then it needs to be removed to enable failver.  
+// Otherwise, if "pause" isn't present in the config or if it has a value other than true, then assume autofail 
+// is enabled and do nothing (when Patroni see's an invalid value for "pause" it sets it to "true")
+func enableFailover(clientset *kubernetes.Clientset, configMap *v1.ConfigMap, configJSON map[string]interface{}, 
+	namespace string) error {
+	if _,ok := configJSON["pause"] ; ok || configJSON["pause"] == true  {
```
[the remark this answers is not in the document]
andrewlecuyer: Good catch, for now I think I can essentially just remove the RHS portion of this, i.e. I simply need to know if the `pause` key exists in order to take action.

apiserver/common.go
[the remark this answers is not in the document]
abrightwell: That's fine by me as that was a concern of mine as well.

util/failover.go
```diff
@@ -128,33 +129,24 @@ func GetPod(clientset *kubernetes.Clientset, deploymentName, namespace string) (
...
 }
 
-func GetRepStatus(restclient *rest.RESTClient, clientset *kubernetes.Clientset, dep *appsv1.Deployment, namespace, databasePort string) (uint64, uint64, string, error) {
+// GetRepStatus is responsible for retrieving and returning the replication status of the pg replica deployed in the 
+// pod specified.  Specifically, the function returns the location of the WAL file that was most recently synced to 
+// disk on the replica (i.e. pg_last_xlog_receive_location() or pg_last_wal_receive_lsn()), and the location of the WAL file that 
+// was most recently replayed on the replica(i.e. pg_last_xlog_replay_location() or pg_last_wal_replay_lsn()), both as unint64. 
+// Additionally, the  name of the Kubernetes node the replica is running on is returned.  This function therefore provides 
+// insight into the replication lag for a replica, as needed to support selection of replicas when performing manual failovers 
+// and scaling down the cluster.
```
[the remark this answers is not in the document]
andrewlecuyer: Hopefully we can establish some standards around this sorta thing to clarify what a "regular" line limit is for the PGO (I typically stick with around 120 chars these days myself).

apiserver/common.go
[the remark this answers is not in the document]
abrightwell: On it's own, I don't believe that is enough justification for keeping this a pointer. While it is not necessarily prescribed in the literature (though [this](https://github.com/golang/go/wiki/CodeReviewComments#pass-values) touches on the essence of the topic), the general guidance is not to worry about saving a few bytes as the cost of maintaining the pointer is far more expensive. Does it 'break' anything? No, but it's not necessarily good practice either.

util/failover.go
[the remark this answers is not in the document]
abrightwell: Yeah, having some conventions established would be nice at some point in the future. When it comes to 'code' I think the rules are a bit more flexible (for obvious reasons), but for comments I think we should strive to keep it to something limited and consistent. Just my opinion.

apiserver/common.go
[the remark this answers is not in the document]
andrewlecuyer: I was simply looking to return `nil` in the event that there was an error, and a `ReplicaPodStatus` was not created.  And while I could simply return an empty struct in an error scenario since the `error` itself should be the true indicator of an error condition, I still tend to like returning `nil` (e.g. for `ReplicaPodStatus`) to clearly indicate that no `ReplicaPodStatus` was created, and a pointer return value allows me to do so.

Also, the guidance is pretty vague around this area, i.e. it doesn't specifically say this is a bad practice or anything.  So I just read the docs as saying there is not really a reason to do so in most cases (i.e. no need to be paranoid about saving a few bytes), but if it does make sense to do so, there is no harm either (and therefore a viable solution).

util/failover.go
abrightwell: I see what you're getting at and don't necessarily disagree. Though, candidly, it still feels grotty to me.

Two reasons:

1) `configJSON` is a `map[string]interface{}`. What happens if the value represented by the key "pause" is ever anything other than a `bool`? Unless you assert (safely) you have no protection for that case. Also a key benefit of the the proposed solution is that you get that safety along with the safe assumption that if it is missing it is assumed to be false.
2) Kind of ties into the above... `value == bool` vs `value` or `!value`. In the first, you are making assumptions about `value` which can be unsafe to do given the nature of the map being `map[string]interface{}`. It would be better to operate on something you actually KNOW to be a `bool`. And unfortunately, that's not guaranteed without the assertion, which should be done safely.

I still think something like the following would be cleaner and 'safer' in the long run.

```
if value, ok := config["pause"].(bool); !ok || !value { ... }
```

Just my two cents. :-)

util/failover.go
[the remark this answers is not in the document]
abrightwell: Hmmm... if the `pause` key existing means that it is always set to `true` then yeah, I'd agree. Is that guaranteed to always be the case?

apiserver/common.go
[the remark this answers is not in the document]
jkatz: [https://golang.org/doc/effective_go.html#pointers_vs_values](https://golang.org/doc/effective_go.html#pointers_vs_values) as the arbiter...which reading through it, one can see it's open to some interpretation for the art of it.

But perhaps the reveal of this is that there is another status, i.e. `Unknown` when we are getting an error and as such `ReadyStatus` would have to remain a `string` :wink:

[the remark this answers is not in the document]
abrightwell: > Also, the guidance is pretty vague around this area, i.e. it doesn't specifically say this is a bad practice or anything. So I just read the docs as saying there is not really a reason to do so in most cases (i.e. no need to be paranoid about saving a few bytes), but if it does make sense to do so, there is no harm either (and therefore a viable solution).

Right, I think we'd be hard pressed to find any official documentation that prescribes what to do in any case. The best we'll get is some 'guidance' as stated, specifically, that it doesn't makes sense in most cases. I personally feel like that this falls under the umbrella of most cases as I don't see anything special that justifies it, but that's just my opinion. Not something I feel so strongly about that I'd insist on holding up these changes.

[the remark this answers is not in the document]
abrightwell: > But perhaps the reveal of this is that there is another status, i.e. Unknown when we are getting an error and as such ReadyStatus would have to remain a string

Or it's own type to ensure valid values. :-)

So, something like:

```
type readyStatus string

const (
  ReadyStatusReady readyStatus = "ready"
  ReadyStatusNotReady readyStatus = "not ready"
  ReadyStatusUnknown readyStatus = "unknown"
)

type ReplicaPodStatus struct {
...
  ReadyStatus readyStatus
}
```

But perhaps that's all a discussion for another day. :-)

util/failover.go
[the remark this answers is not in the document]
andrewlecuyer: To re-enable autofail the `pause` key can either be set to `null` or `false`, or can be deleted altogether.  The plan here is to simply always do the latter when this function is called, i.e. just delete the key regardless.

apiserver/failoverservice/failoverimpl.go
```diff
@@ -235,21 +194,42 @@ func validateClusterName(clusterName, ns string) (*crv1.Pgcluster, error) {
...
-	deployment, found, err := kubeapi.GetDeployment(apiserver.Clientset, deployName, ns)
-	if !found {
-		return deployment, errors.New("no target found named " + deployName)
-	}
-
-	//make sure the primary is not being selected by the user
-	if deployment.ObjectMeta.Labels[config.LABEL_SERVICE_NAME] == clusterName {
-		return deployment, errors.New("deployment primary can not be selected as failover target")
+// isValidFailoverTarget checks to see if the failover target specified in the request is the
+// is valid, i.e. that it represents a valid replica deployment in the cluster specified.  This is
```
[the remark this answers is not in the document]
andrewlecuyer: comment updated

util/failover.go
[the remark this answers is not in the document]
andrewlecuyer: Comment has been reformatted.

util/failover.go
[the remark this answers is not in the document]
andrewlecuyer: It's actually an AND that I'm looking for here, i.e. I want to update the key when it exists AND is set `true` (I was originally thinking I could just remove it regardless - a surefire way to disable autofail - but this saves from an unneeded update if it is already set to `false` or `null`).

util/failover.go
andrewlecuyer: Appreciate the input here, this one got me thinking.  The one things that "bugs" me with this approach (and that is admittedly not currently relevant to this function), is that I loose visibility into whether not the key even existed in the map in the first place, or if it was simply set to an invalid value (i.e. both the key missing altogether and an invalid key [e.g. some random string] result in `ok=false`).

Also, I'm not clearly seeing what protection I'm missing by using `value == bool`.  Specifically, I know in this case that I only care that the result is (or with my recent update _isn't_) a certain boolean value, i.e. that it isn't `true` in the following case (and I know that Go will store a `bool` for a valid JSON boolean, which is what I want):

`configJSON["pause"] != true `

...and therefore I can store any other value of any type (null, string, int, etc.) in the map for the "pause" key and the result will be the same, i.e. the condition will fail.

wilybrace: I think the current construct `_,ok := configJSON["pause"] ; ok && configJSON["pause"] == true` is certainly clearer and achieves what Adam recommended via a quirk of the language. (It seems to perform either an implicit type conversion or immediately understands that a value of type `interface{}` is not equal to `bool`(true))
However, there's a behavioral caution here that wasn't captured in Adam's recommendation and that's concerning the underlying type of the JSON. In deserializing the JSON, we're going from a weakly typed system to a strongly typed one, which presents opportunities for this type of failure:
https://play.golang.com/p/v1Xr42-qA51
That is, the check will fail if the value is not of `bool` type according to javascript. In this case, using `ParseBool` on the value would ensure correct behavior without someone having to get the type correct in the JSON. (Think about someone modifying the CRD directly and including quotes when they shouldn't have)

wilybrace: FWIW, I'm fine either way on deciding to postpone the failure scenario, but do recognize it's leaving a potential trap.

abrightwell: > Appreciate the input here, this one got me thinking. The one things that "bugs" me with this approach (and that is admittedly not currently relevant to this function), is that I loose visibility into whether not the key even existed in the map in the first place, or if it was simply set to an invalid value (i.e. both the key missing altogether and an invalid key [e.g. some random string] result in ok=false).

Sure, but that's a different case altogether though, right? The point is that as soon as you want to operate on the value represented by the key, you want to know it's type. Otherwise, you're at risk of a `panic`, which I do think is worth protecting against.

> Also, I'm not clearly seeing what protection I'm missing by using value == bool. Specifically, I know in this case that I only care that the result is (or with my recent update isn't) a certain boolean value, i.e. that it isn't true in the following case (and I know that Go will store a bool for a valid JSON boolean, which is what I want):

Well, there are two things here:

1) `if bool == bool { ... }` is really just `if bool { ... }` right? perhaps that's just a stylistic preference, but doing an extra 'comparison' to determine the value of what you (should) already have seems unnecessary to me.

2) There is value in actually knowing and guaranteeing the type of the thing you are comparing. Making assumptions about the type, in my view, is inherently dangerous.

Essentially, interface values are considered equal when their underlying concrete types and values are comparable and equal. However, if the underlying concrete types are not comparable any attempt to compare them will cause a runtime panic.

> ...and therefore I can store any other value of any type (null, string, int, etc.) in the map for the "pause" key and the result will be the same, i.e. the condition will fail.

I see your point, but I feel like this is an misuse of `interface{}`. It's nice to have the 'flexibility' but when operating on the underlying type, it's best, in my view, to ensure you're actually operating on that concrete type.

Ultimately, for me, it's about type safety and safe guarding against panics.  Maybe that case may never arise, but to me it still seems worth taking measures to guard against it.

andrewlecuyer: What value for `pause` would result in a panic?

https://play.golang.org/p/iz8pimpcO0G

wilybrace: > What value for `pause` would result in a panic?

My read of this feedback is to be careful about using / relying on a value of type `interface{}` to "behave well" in all cases. In this _specific_ scenario, performing a comparison of a type `interface{}` against a literal, built-in type is guaranteed to be safe, but the pattern is not universally so.

util/failover.go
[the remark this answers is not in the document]
cbandy: > It's actually an AND that I'm looking for here, i.e. I want to update the key when it exists AND is set `true` (I was originally thinking I could just remove it regardless - a surefire way to disable autofail - but this saves from an unneeded update if it is already set to `false` or `null`).

The comment on this function makes it sound like `{"pause": "steve"}` will disable failover. Do we want to enable failover in that case?

If so, the logic should be: "delete when not (missing or nil or false)." This very close to the latter half of your statement above but differs from what is implemented. I think the desired behavior is expressed clearly as the following:

```go
if v, ok := configJSON["pause"]; ok && v != nil && v != false {
  …
  delete(configJSON, "pause")
  …
}
```

util/failover.go
jkatz: I think overall this is a very interesting discussion and certainly makes mention of some things we want to incorporate in the future especially around type safety, which we should certainly incorporate as the codebase evolves. In fact, this discussion could serve as a good case study and be applied to "design patterns" we want to incorporate into our go code.

However, I am satisfied right now with what we have and so long as the code is correct, I advise we move on.

PR #1112, opened 2019-12-02
Housekeeping and extending TLS trust

wilybrace:
This PR does some significant cleanup in the main function of the apiserver while adding support for client certificate validation beyond the existing "Trust me, I'm you" model.

Of the clean-up steps, the likely most contentious change is moving the route configuration into the apiserver/routing package. This implementation is currently a "lesser of evils" between dot imports across all service packages and adding an additional bridging package to break the cyclical dependency.

I strongly recommend looking at individual commits instead of the full diff as they have been structured to be easier to follow the change steps.

 
**Type of Changes:**

 
Currently, the operator only trusts itself, thus the pgo client application must obtain a copy of the server certificate and key to communicate with the server. This forces insecure security practices by exposing the server's key. By extending the configuration to allow more trust sources, clients can be issued certificates independent of the operator API secret.


apiserver/root.go
```diff
@@ -649,8 +651,9 @@ func GetTLS(certPath, keyPath string) error {
 
 }
 
-// generate a self signed cert and the pgo.tls Secret to hold it
-func generateTLS(certPath, keyPath string) error {
+// generateTLSCert generates a self signed cert and the stores it in both
```
abrightwell: `the` -> `then`

apiserver/routing/routes.go
```diff
@@ -0,0 +1,247 @@
...
+	RegisterPGPoolSvcRoutes(r)
+	RegisterPolicySvcRoutes(r)
+	RegisterPVCSvcRoutes(r)
+	RegisterReloadSvcRoutes(r)
+	RegisterScheduleSvcRoutes(r)
+	RegisterStatusSvcRoutes(r)
+	RegisterUpgradeSvcRoutes(r)
+	RegisterUserSvcRoutes(r)
+	RegisterVersionSvcRoutes(r)
+	RegisterWorkflowSvcRoutes(r)
```
abrightwell: Is there a need to have all these functions exported?

wilybrace: The scenario that spawned the change only registered one service and the pattern is to allow only importing the routes of a specific "service", allowing routes to be composed on a per-service basis as might be used for testing or service-specific middleware. As I mentioned in the comments (PR? commit?), ultimately it would make sense for each service to export a registration for its handlers for better cohesion.

apiserver/root.go
wilybrace: thanks, corrected in a local commit - holding for any other requested changes.

wilybrace:
> It seems like we've add a few new ENV/options.
> 
> For example the one's I noticed initially are the following:
> 
>     * `ADD_OS_TRUSTSTORE`
>     * `TLS_CA_TRUST`
>     * `NOAUTH_ROUTES`
> 
> From the code comments, it's clear what these do, however, I would suggest that the appropriate end-user documentation be updated with that information as part of this PR as well.

`NOAUTH_ROUTES` isn't new, it's documentation is located in `hugo/content/Configuration/configuration.md` in the "Operator API Server" section.

Wholly agree on the other two, changes forthcoming after some in-house discussions this morning regarding the scope of the documentation.

apiserver/routing/routes.go
abrightwell: Gotcha, was just a curiosity. But, yeah, that makes sense as to the motivation for the refactor. And I suppose it answers the question of 'exporting' given the goal of testing, etc. Certainly a good change for now. Though, for "later", I think moving towards the concept of 'service' and 'routing/transport' being separate/decoupled would be beneficial to those ends as well. Just a thought.

tlsutil/primitives_test.go
```diff
@@ -0,0 +1,166 @@
...
+		t.Fatalf("error getting response - %s\n", err)
+	}
+
+	body, err := ioutil.ReadAll(res.Body)
+	res.Body.Close()
+	if err != nil {
+		t.Fatalf("error reading response -%s\n", err)
+	}
+
+	if recv := string(bytes.TrimSpace(body)) + "."; recv != expected {
```
jkatz: This assertion appears to be failing due to an extra `.`

wilybrace:
No on the ADD_OS_TRUSTSTORE as trust should be opt-in. The list makes more sense in a client scenario (e.g. browser trusting HTTPS connections on the internet) as many of the major industry CAs are included - (e.g. Verisgin, THAWTE, Secure Trust, Network Solutions).

wilybrace:
Additional scope required:
* Rebase required: #1121 - processing of `NOAUTH_ROUTES` changed
* Trust configuration / automation for pgo client
* Configuration of new env variables and NOAUTH setting in both manual and ansible deployment templates

wilybrace:
Force-push reflects rebasing from #1121

wilybrace:
Force-push reflects rebase triggered by #1117 and #1147

wilybrace:
Force-push reflects rebase triggered by #1117, #1147, and #1148

wilybrace:
Force-push reflects rebase triggered by #1150, #1151

pgo/cmd/auth.go
```diff
@@ -200,96 +198,99 @@ func SetSessionUserCredentials() {
...
+			Certificates:       []tls.Certificate{certPair},
 			MinVersion:         tls.VersionTLS11,
 		},
+	}, nil
+}
+
+// NewAPIClient returns an http client configured with a tls.Config
+// based on environmental settings and a default timeout
+func NewAPIClient() *http.Client {
+	defaultTimeout := 60 * time.Second
```
abrightwell: Is it really worth declaring `defaultTimeout` here if it isn't being used in multiple places?

wilybrace: I debated that a lot, honestly. Ultimately I decided to leave it as a defined variable so that the literal definition following looks less "magical", more self-documenting, and set up to be refactored out if knowledge of the duration is needed elsewhere (or if it needs to be parameterized). Said differently, I settled on it to communicate the 60 was a 'default' choice and not some hard-coded value with undocumented dependency.

At this point, I'll leave it as-is unless someone else has a stronger opinion than you or I.

PR #1120, opened 2019-12-05
Remove PGcluster Configmaps on Delete

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
When a new pgcluster is created, three configmaps are also created. 
They consist of two that are created by Patroni when it initializes a new cluster:
 <cluster-name>-leader (stores data pertinent to the leader election process)
 <cluster-name>-config (stores global/cluster-wide configuration settings)
and one that contains a default Patroni configuration file:
 <cluster-name>-pgha-default-config (stores a Patroni config file in YAML format)
On delete, however, they are not removed.

This commit adds the function removeClusterConfigmaps which deletes the three
configmaps that are now created for each cluster. 
These configmaps are specific to the cluster and, therefore, do not need to
persist after cluster deletion.

**Other information**:
Tested in Kubernetes and Openshift environments.

To test, after creating a cluster with 

pgo create cluster mycluster -n mynamespace

the cluster should be created and the three configmaps mentioned above should be 
visible when you execute

kubectl get cm -n mynamespace

Now, run

pgo delete cluster mycluster -n mynamespace

after the finishes, you should no longer see the three configmaps for that cluster.

Issue: [CH6324]

### DIFF
--- a/pgo-rmdata/rmdata/process.go
+++ b/pgo-rmdata/rmdata/process.go
@@ -120,6 +120,7 @@ func Delete(request Request) {
 		removePVCs(pvcList, request)
 	}
 
+	removeClusterConfigmaps(request)
 }
 
 func removeBackups(request Request) {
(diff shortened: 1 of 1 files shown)


tjmoore4:
@jkatz Unfortunately not. I looked into doing something like this, but it would require reading the content of different labels depending on the configmap. At this time, we would need to look at 'crunchy-pgha-scope' for `<cluster>-config` and `<cluster>-leader` and 'pg-cluster' for `<cluster>--pgha-default-config`.

With that noted, it seemed cleaner to just delete by explicit label name.

jkatz:
@tjmoore4 Thanks for the explanation. As such I am giving my approval. Thanks!

Issue #1119, opened 2019-12-05
4.1.1 events: failed to read protocol version


cubed-it:
I'm running 4.1.1 on AKS 1.14.8 and I observe constant Liveness/Readiness probe failed. Is this known?

events log:
```
[nsqd] 2019/12/05 14:59:22.899954 INFO: TCP: new client(10.244.0.1:54834)
[nsqd] 2019/12/05 14:59:22.900011 ERROR: failed to read protocol version - EOF
[nsqd] 2019/12/05 14:59:24.598864 INFO: TCP: new client(10.244.0.1:54888)
[nsqd] 2019/12/05 14:59:24.598960 ERROR: failed to read protocol version - EOF
[nsqd] 2019/12/05 14:59:29.599034 INFO: TCP: new client(10.244.0.1:55010)
[nsqd] 2019/12/05 14:59:29.599114 ERROR: failed to read protocol version - EOF
[nsqd] 2019/12/05 14:59:32.899860 INFO: TCP: new client(10.244.0.1:55070)
[nsqd] 2019/12/05 14:59:32.899883 ERROR: failed to read protocol version - EOF
[nsqd] 2019/12/05 14:59:34.598817 INFO: TCP: new client(10.244.0.1:55110)
[nsqd] 2019/12/05 14:59:34.599023 ERROR: failed to read protocol version - EOF
[nsqd] 2019/12/05 14:59:44.598897 INFO: TCP: new client(10.244.0.1:55334)
[nsqd] 2019/12/05 14:59:44.598921 ERROR: failed to read protocol version - EOF
Signal trap triggered, beginning shutdown..
++ trap_sigterm
++ echo 'Signal trap triggered, beginning shutdown..'
+++ pidof nsqd
/usr/local/bin/pgo-event.sh: line 18: pidof: command not found
++ kill -9
kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
+++ pidof nsqadmin
/usr/local/bin/pgo-event.sh: line 19: pidof: command not found
++ kill -9
kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
end of pgo-event
+ echo 'end of pgo-event'
```

To mitigate this issue I modified the event deployment. Is this okay?
```
            "livenessProbe": {
              "httpGet": {
                "path": "/ping",
                "port": 4151,
                "scheme": "HTTP"
              },
[...]

Issue #1111, opened 2019-11-30
V4.1.1 apiserver Ready but "http: TLS handshake error X.X.X.X EOF"


ook:
** Which example are you working with? **

Installing  v4.1.1 from instructions here: https://access.crunchydata.com/documentation/postgres-operator/4.1.1/installation/operator-install/
Note: I had previously a fully working  4.0.1 operator on that same k8s cluster.

**What is the current behavior?**

RBAC and NS created, pod ready (4/4) but can't use pgo client against it. When checking pod apiserver:

```
NAME                                READY   STATUS    RESTARTS   AGE
postgres-operator-f87b4ccfb-zsjdm   4/4     Running   22         2d4h

$ kubectl -n pgo logs postgres-operator-f87b4ccfb-zsjdm apiserver
time="2019-11-28 17:14:23 +0000" level=debug msg="debug flag set to true" func="main.main()" file="apiserver.go:80" version=4.1.1
time="2019-11-28 17:14:23 +0000" level=debug msg="TLS_NO_VERIFY set to false" func="main.main()" file="apiserver.go:90" version=4.1.1
time="2019-11-28 17:14:23 +0000" level=debug msg="DISABLE_TLS set to false" func="main.main()" file="apiserver.go:99" version=4.1.1
time="2019-11-28 17:14:23 +0000" level=info msg="postgres-operator apiserver starts" func="main.main()" file="apiserver.go:105" version=4.1.1
time="2019-11-28 17:14:23 +0000" level=info msg="Pgo Namespace is [pgo]" func="apiserver.Initialize()" file="apiserver/root.go:104" version=4.1.1
time="2019-11-28 17:14:24 +0000" level=info msg="InstallationName is [devtest]" func="apiserver.Initialize()" file="apiserver/root.go:114" version=4.1.1
time="2019-11-28 17:14:24 +0000" level=info msg="apiserver starts" func="apiserver.Initialize()" file="apiserver/root.go:126" version=4.1.1
time="2019-11-28 17:14:24 +0000" level=info msg="loading PermMap with 60 Permissions\n" func="apiserver.InitializePerms()" file="apiserver/perms.go:177" version=4.1.1
time="2019-11-28 17:14:24 +0000" level=info msg="Config: pgo-config ConfigMap found, using config files from the configmap" func="config.getRootPath()" file="config/pgoconfig.go:812" version=4.1.1
[...]

PR #1122, opened 2019-12-05
prevents misleading error-level logs (4.1 backport)

wilybrace:
The apiserver and nsqd processes do not gracefully handle the
empty TCP connections generated by kubelet as part of probe
execution, causing misleading error messages to accumulate
in the logs at the probe polling intervals.

For the nsqd process, there exists an HTTP endpoint that can
be used for liveness testing, but it not useful for readiness
testing as the `/ping` endpoint does not reflect readiness
for connections on port 4150. Thus, the liveness test is moved
to an HTTP test of the `/ping` endpoint and the readiness test
is removed.

Furthermore, each access to `/ping` is logged as an INFO level
log in the nsqd log files. While this is not as misleading as
ERROR level logging, it still represents unnecessary logging
and the log level for the nsqd process has been moved up to
WARNING.

For the pgo-apiserver, a new lightweight route is added to
the conventional location of `/healthz` with a simple `ok`
response with the HTTP 200.

As a result, NOAUTH_ROUTES no longer optionally activates
deferred certificate checking - it is required to support
`/healthz` as kubelet does not present a certificate when
performing HTTPS checks.

NOAUTH_ROUTES may still be used to enable unauthenticated
access to the existing whitelist of routes.

Fixes missing probes in Ansible template

[ch6429]
[ch6453]

Related issues:
#1111, #1119

**Type of Changes:**

 
**Backport Differences**
* No swagger
* Ansible deployment template was missing the probes

### DIFF
--- a/ansible/roles/pgo-operator/templates/deployment.json.j2
+++ b/ansible/roles/pgo-operator/templates/deployment.json.j2
@@ -30,6 +30,24 @@
                                 "containerPort": {{ pgo_apiserver_port }}
                             }
                         ],
+                        "readinessProbe": {
+                            "httpGet": {
+                                "path": "/healthz",
+                                "port": {{ pgo_apiserver_port }},
+                                "scheme": "HTTPS"
+                            },
+                            "initialDelaySeconds": 15,
+                            "periodSeconds": 5
+                        },
+                        "livenessProbe": {
+                            "httpGet": {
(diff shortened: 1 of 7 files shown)


jkatz:
LGTM. I will write up something for the release notes to indicate that one needs to reload their deployment with the new configuration to silence the errors.

PR #1123, opened 2019-12-06
Minor Upgrade Updates in Support of the crunchy-postgres-ha Container

andrewlecuyer:
This PR updates the minor upgrade functionality within the PostgreSQL Operator (i.e. an upgrade initiated via `pgo upgrade mycluster`) as needed to support the new `crunchy-postgres-ha` container (and GIS equivalents).  This includes the following changes:

* The mechanism for patching  (i.e. updating) the image name for the database, as well as any sidecar containers (`crunchy-collect`, `crunchy-pgbadger` and/or `crunchyadm`) included in a primary or replica Deployment, has been updated.  Now instead of using the `json` patch type to perform a JSON patch in accordance with RFC 6902 to upgrade the image name  for any/all containers included in a PG primary or replica Deployment, a Kubernetes strategic merge patch is utilized instead. This provides a cleaner mechanism for updating container image names, allowing for easier maintenance of the upgrade process when sidecars are added or remove.  Specifically, sidecars can now be patched based on their name, and not their position in the `containers` array within the Deployment spec, which may vary across PG clusters.  For instance, the new `crunchyadm` can be updated by name without worrying about its position in the `containers` array, which varies depending on  whether or not various other sidecars have been enabled.
* If autofail is disabled when performing a minor upgrade, the primary and all replicas are patched at the same time, and the upgrade task is immediately marked as complete. This is because when performing a minor upgrade with autofail disabled, once the upgrade is complete and new pods have been created for the primary and each replica, any `database` containers will remain in an unready state, i.e. readiness probes will fail for those containers (and will continue to do so indefinitely).  This is because when autofail is disabled (i.e. Patroni is "paused"), Patroni will not attempt to start the PG database on a primary or replica.  Therefore, the user must re-enable autofail following a minor upgrade during which autofail was disabled in order to fully bring the cluster (which includes the primary and all replicas) back online.

Also,  the `crunchyadm` sidecar container is now disabled by default.  This sidecar container can be re-enabled by setting `EnableCrunchyadm` to true in the `pgo.yaml` file, or by setting `enable_crunchyadm` to `true` in the Ansible inventory file.  Please note that at this time setting will not be present in the default `pgo.yaml` file or the default Ansbile inventory file, nor will it be referenced in the documentation, since it is not yet supported across all environments and results may therefore vary.

And finally,  the `PGHA_PRIMARY_HOST` environment variable has been removed from the `cluster-deployment.json` template, which means it will no longer be included in any replicas created
by the PGO using this template. This environment variable, while useful for orchestrating primary and replica creation in a containers environment such as Docker, is not needed when utilizing the `crunchy-postgres-ha` container with the PostgreSQL Operator. This is because the PostgreSQL Operator is able to leverage Kubernetes to orchestrate primary and replica creation.

**Checklist:**

 
**Type of Changes:**

 
PostgreSQL Operator minor upgrade functionality has not been updated in support of the new `crunchy-postgres-ha` container.

[ch5463]
[ch6466]

PostgreSQL Operator minor upgrade functionality has been updated in support of the new `crunchy-postgres-ha` container.

**Other information**:

N/A


operator/cluster/upgrade.go
```diff
@@ -100,9 +102,15 @@ func AddUpgrade(clientset *kubernetes.Clientset, restclient *rest.RESTClient, up
...
-
+	// if autofail is enabled, then patch the replicas, followed by the primary, sequentially.
+	// this process is started by calling ProcessNextUpgradeItem to start the upgrade
+	// otherwise if autofail is disabled, upgrade the primary and all replicas at the same time
+	// by calling UpgradeAutofailDisabled
+	if util.IsAutofailEnabled(&cl) {
+		ProcessNextUpgradeItem(clientset, restclient, cl, currentTask.Name, namespace)
+	} else {
+		UpgradeAutofailDisabled(clientset, restclient, cl, currentTask.Name, namespace)
+	}
```
jkatz: I would suggest some function renames to help with clarity, i.e.:

- `UpgradeWithAutofailEnabled` (I'm cool with "breaking the API" on this one)
- ~UpgradeWithAutofailDisabled`

andrewlecuyer:
@jkatz definitely agree that similar upgrade functionality is needed for `pgbouncer`.  However, my goal for this ticket was to simply maintain existing minor upgrade functionality with the new `crunchy-postgres-ha` container, which currently includes upgrading the PG primary and replica Deployments only (including any sidecars contained within those Deployments).

I do also wonder if we would want to handle the upgrade of `pgbouncer` separately from the primary and replica deployments comprising the PG cluster.  So therefore, just like we create `pgbouncer` in a cluster as follows:

```
pgo create pgbouncer mycluster
```

...we would then upgrade `pgbouncer` using the following command:

```
pgo upgrade pgbouncer mycluster
```

This would allow `pgbouncer` to be created, upgraded, etc. independently of the PG Deployments.  However, its worth noting that our current cluster upgrade functionality does not fully follow this paradigm, i.e. the `upgrade` command is:

```
pgo upgrade mycluster
```

...and not:

```
pgo upgrade cluster mycluster
```

Considering the above, my recommendation is that we create a new ticket to add `pgboucncer` upgrade functionality in a future release.  This will also allow us to put some thought around how we want to handle upgrades in general, while simply allowing us to ensure we maintain existing upgrade functionality with this specific PR.

jkatz:
@andrewlecuyer I agree. And thinking it through, you can have your pgbouncer deployment interface with multiple PostgreSQL clusters anyway (with some configuration), so it can be treated as an independent managed service.

operator/cluster/upgrade.go
andrewlecuyer: Function `UpgradeAutofailDisabled` has been renamed to `UpgradeWithAutofailDisabled`, and function `UpgradeWithAutofailEnabled` has now been created (and `UpgradeWithAutofailEnabled` is also now called instead of `ProcessNextUpgradeItem`).

PR #1130, opened 2019-12-08
[ch6494] Update `pgo scaledown --query` / `pgo failover --query`

jkatz:
**Checklist:**

 
**Type of Changes:**

 
The original genesis of this was to fix a bug that was shown in 053387c to ensure that a replica that was not currently matched with a primary could still have its information displayed without breaking the `pgo scaledown --query` and `pgo failover --query` commands.

However, this turned into an inspection of the information that these commands are returning and it was decided to make it a bit more helpful.

The commands provided information that could help a
user make a decision about which instance in a PostgreSQL cluster they
would want to fail over to, it could be challenging to decipher as not
everyone is versed in PostgreSQL log sequence numbers (LSN).

This updates the output for both `pgo scaledown --query` and `pgo failover --query` to give the user information that can help them make a better decision around these actions.

This changes leverages the information that is being collected by the
new HA system in order to provide better feedback about the status of
the replicas available. This information includes:

- NAME: the name of the replica
- STATUS: the current status of the replica
- NODE: the node that the replica resides on
- REPLICATION LAG: how far behind the replica is behind the primary

Additionally, "TIMELINE" is available in the API but is not currently
displayed to the user.

This also cleans out some now unused code.

Issue: [ch6494]

**Other information**:

The original bug fix is still kept as part of this commit in case the other changes are rejected.

This does make modifications to the API returns for the `--query` commands on `scaledown` + `failover`.

Issue: [ch6494]

### DIFF
--- a/pgo/cmd/failover.go
+++ b/pgo/cmd/failover.go
@@ -93,39 +93,59 @@ func createFailover(args []string, ns string) {
 
 }
 
-// queryFailover ....
+// queryFailover is a helper function to return the user information about the
+// replicas that can be failed over to for this cluster. This is called when the
+// "--query" flag is specified
 func queryFailover(args []string, ns string) {
+	var hasPreferredNode bool
+
 	log.Debugf("queryFailover called %v", args)
 
+	// indicate which cluster this is. Put a newline before to put some
+	// separation between each line
+	fmt.Printf("\nCluster: %s\n", args[0])
+
+	// call the API
 	response, err := api.QueryFailover(httpclient, args[0], &SessionCredentials, ns)
 	if err != nil {
 		fmt.Println("Error: " + err.Error())
 		os.Exit(2)
 	}
 
-	if response.Status.Code == msgs.Ok {
-		fmt.Println(response.Status.Msg)
-		if len(response.Targets) > 0 {
-			fmt.Println("Failover targets include:")
-			for i := 0; i < len(response.Targets); i++ {
-				printTarget(response.Targets[i])
-			}
-		}
-		for k := range response.Results {
-			fmt.Println(response.Results[k])
-		}
-	} else {
+	// If there is a controlled error, output the message here and continue
+	// to iterate through the list
+	if response.Status.Code != msgs.Ok {
 		fmt.Println("Error: " + response.Status.Msg)
-		os.Exit(2)
+		os.Exit(1)
 	}
 
-}
-
-func printTarget(target msgs.FailoverTargetSpec) {
-	var preferredNode string
-
-	if target.PreferredNode {
-		preferredNode = "(preferred node)"
+	// If there are no replicas found for this cluster, indicate so, and
+	// continue to iterate through the list
+	if len(response.Results) == 0 {
+		fmt.Println("No replicas found.")
+		return
 	}
-	fmt.Printf("\t%s (%s) (%s) %s ReceiveLoc (%d) ReplayLoc (%d)\n", target.Name, target.ReadyStatus, target.Node, preferredNode, target.ReceiveLocation, target.ReplayLocation)
+
+	// output the information about each instance
+	fmt.Printf("%-20s\t%-10s\t%-10s\t%s\n", "REPLICA", "STATUS", "NODE", "REPLICATION LAG")
+
+	for i := 0; i < len(response.Results); i++ {
+		instance := response.Results[i]
+
+		node := instance.Node
+		// check if this is a preferred node
+		if instance.PreferredNode {
+			hasPreferredNode = true
+			node = fmt.Sprintf("*%s", node)
+		}
+
+		fmt.Printf("%-20s\t%-10s\t%-10s\t%12d MB\n",
+			instance.Name, instance.Status, node, instance.ReplicationLag)
+	}
+
+	// if a node exists that is a preferred failover target, print an informative
+	// message
+	if hasPreferredNode {
+		fmt.Println("* next to a NODE name indicates a preferred failover target")
+	}
 }
--- a/pgo/cmd/scaledown.go
+++ b/pgo/cmd/scaledown.go
@@ -76,31 +76,48 @@ func init() {
 
 }
 
+// queryCluster is a helper function that returns information about the
+// available replicas that can be scaled down. This is called when the "--query"
+// flag is specified
 func queryCluster(args []string, ns string) {
 
+	// iterate through the clusters and output information about each one
 	for _, arg := range args {
+		// indicate which cluster this is. Put a newline before to put some
+		// separation between each line
+		fmt.Printf("\nCluster: %s\n", arg)
+
+		// call the API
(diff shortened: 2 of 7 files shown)


pgo/cmd/scaledown.go
```diff
@@ -76,31 +76,48 @@ func init() {
 
 }
 
+// queryCluster is a helper function that returns information about the
+// available replicas that can be scaled down. This is called when the "--query"
+// flag is specified
 func queryCluster(args []string, ns string) {
 
+	// iterate through the clusters and output informatio about each one
```
abrightwell: `information`

pgo/cmd/scaledown.go
```diff
@@ -76,31 +76,48 @@ func init() {
...
 	for _, arg := range args {
+		// indicate which cluster this is. Put a newline before to put some
+		// separation between each line
+		fmt.Println(fmt.Sprintf("\nCluster: %s", arg))
```
abrightwell: `fmt.Printf(...)`?

util/failover.go
```diff
@@ -129,169 +116,93 @@ func GetPod(clientset *kubernetes.Clientset, deploymentName, namespace string) (
...
-		return receiveLocation, replayLocation, nodeName, err
+// ReplicationStatus is responsible for retrieving and returning the replication
+// information about the status of the replicas in a PostgreSQL cluster. It
+// executes into a single replica pod and leverages the functionality of Patroni
+// for getting the key metrics that are appropriate to help the user understand
+// the current state of their replicas.
+//
+// Statistics include: the current node the replica is on, if it is up, the
+// replication lag, etc.
+func ReplicationStatus(request ReplicationStatusRequest) (ReplicationStatusResponse, error) {
```
abrightwell: I'm not sure how I feel about wrapping the 'client' handles in the request. I'm understanding that this is simply a util function, but it seems like the request should be 'data' only, IMV. :thinking: Based on what it's doing, seems like this could be it's own or part of a related 'service' and it would then have/be aware of it's own client state?

util/failover.go
```diff
@@ -129,169 +116,93 @@ func GetPod(clientset *kubernetes.Clientset, deploymentName, namespace string) (
...
-	if err != nil {
-		log.Errorf("Could not connect to: %s", target)
-		return nil, err
+	if len(pods.Items) == 0 {
+		return response, err
 	}
 
-	defer conn.Close()
+	// We need to create a quick map of "instance name" => node name
+	// We will iterate through the pod list once to extract the name we refere to
```
abrightwell: `refer`

util/failover.go
```diff
@@ -129,169 +116,93 @@ func GetPod(clientset *kubernetes.Clientset, deploymentName, namespace string) (
...
-	return &ReplicationInfo{
-		ReceiveLocation: recvLocation,
-		ReplayLocation:  replayLocation,
-		Node:            "",
-		DeploymentName:  "",
-	}, nil
-}
 
-func getSecrets(clientset *kubernetes.Clientset, cluster *crv1.Pgcluster, namespace string) ([]msgs.ShowUserSecret, error) {
+		// get the instance name that is recognize by the Operator, which is the
```
abrightwell: `recognized`

util/failover.go
```diff
@@ -129,169 +116,93 @@ func GetPod(clientset *kubernetes.Clientset, deploymentName, namespace string) (
...
+		// first part of the name
+		instance.Name = instanceNamePattern.FindStringSubmatch(rawInstance.PodName)[1]
```
abrightwell: FWIW, it seems like this might be worthy of being it's own utility function, something like `GetInstanceName(...)`?

util/failover.go
```diff
@@ -344,6 +255,27 @@ func ToggleAutoFailover(clientset *kubernetes.Clientset, enable bool, pghaScope,
 	return nil
 }
 
+// createInstanceNodeMap creates an mapping between the names of the PostgreSQL
```
abrightwell: `a`

util/failover.go
```diff
@@ -344,6 +255,27 @@ func ToggleAutoFailover(clientset *kubernetes.Clientset, enable bool, pghaScope,
...
+// instances to the Nodes that they run on, based upon the output from a
+// Kubernetes API query
+func createInstanceNodeMap(pods *v1.PodList) map[string]string {
+	instanceNodeMap := map[string]string{}
+
+	// Iterate through each pod that is return and get the mapping between the
```
abrightwell: `returned` ... `them`

pgo/cmd/failover.go
```diff
@@ -93,39 +93,59 @@ func createFailover(args []string, ns string) {
...
+// replicas that can be failed over to for this cluster. This is called when the
+// "--query" flag is specified
 func queryFailover(args []string, ns string) {
+	var hasPreferredNode bool
+
 	log.Debugf("queryFailover called %v", args)
 
+	// indicate which cluster this is. Put a newline before to put some
+	// separation between each line
+	fmt.Println(fmt.Sprintf("\nCluster: %s", args[0]))
```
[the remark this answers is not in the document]
jkatz: Maybe...but I need a newline anyway. I could consolidate though to make it clearer.

util/failover.go
jkatz: @abrightwell I think this is something to think about for a future release. My goal was to take what we have in the existing infrastructure and ensure it works for the new HA changes. To have it work in that way would require a much larger effort compared to the remainder of the HA work that needs to take place.

That said, I could be convinced that this function could leave in the `apiserver` realm given it's called primarily by two `apiserver` endpoints, which would be a significantly lower effort to take place.

util/failover.go
jkatz: This is very specific in deriving the instance name from the Pods that is being output by Patroni. If we need it some more, we could consider moving it into its own utility function.

jkatz:
@abrightwell Thanks for the review. I addressed most of the comments.

The point about the `util.ReplicationStatus` status mixing the client request information is valid, I believe the question is do we want to do another lift to move it into the `apiserver` realm. It appears that this function is only used there, so it would not be complicated to move it into an area it can be commonly accessed from.

OTOH the goal was to keep it as much of a "drop in replacement" from the previous function, and it is not the only function in there that uses portions of the client request.

Open either way on this one.

util/failover.go
abrightwell: Sounds good.

util/failover.go
abrightwell: > I think this is something to think about for a future release. My goal was to take what we have in the existing infrastructure and ensure it works for the new HA changes. To have it work in that way would require a much larger effort compared to the remainder of the HA work that needs to take place.

:+1:

abrightwell:
If the original intent was to keep it a 'drop in', then I think as it sits is fine. Especially since there is already precedent for it. And FWIW, the boy scout in me would want to clean up the rest along with it and that's just unnecessary scope creep. Keeping the changes relevant and minimal given current goals I think is the right choice for now.

PR #1140, opened 2019-12-10
Add service account to every pod

cbandy:
This attaches a Service Account to every Pod deployed by the operator. A new account, `pgo-default`, is used for pods that don't need access to the Kubernetes API. This account provides a way to configure [image pull secrets](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account) for each namespace.

**Checklist:**

 
**Type of Changes:**

 
### DIFF
--- a/ns/nslogic.go
+++ b/ns/nslogic.go
@@ -33,6 +33,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const PGO_DEFAULT_SERVICE_ACCOUNT = "pgo-default"
+
 const PGO_TARGET_ROLE = "pgo-target-role"
 const PGO_TARGET_ROLE_BINDING = "pgo-target-role-binding"
 const PGO_TARGET_SERVICE_ACCOUNT = "pgo-target"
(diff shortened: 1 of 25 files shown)


ns/nslogic.go
```diff
@@ -610,6 +623,43 @@ func WatchingNamespace(clientset *kubernetes.Clientset, requestedNS, installatio
...
+	if err != nil {
+		log.Error("error unmarshalling " + config.PGODefaultServiceAccountPath + " json ServiceAccount " + err.Error())
+		return err
+	}
+
+	err = kubeapi.CreateServiceAccount(clientset, &rb, targetNamespace)
+	if err != nil {
+		return err
+	}
+	return err
```
abrightwell: Did you intend to log the err or do something else with it here? Otherwise, why not just return `err` or even just simply `return kubeapi.CreateServiceAccount(...)`?

cbandy: Oof. I copy-pasta'd from the other SA create functions. I'll amend this and the others.

cbandy:
I need to do something with the deployments in pgo-metrics as well.

PR #1148, opened 2019-12-16
Add Default Pod Anti-Affinity Rules to PG Cluster Deployments

andrewlecuyer:
Adds default pod ant-affinity rules to all deployment specifications utilized to deploy the various pods that comprise a full PG cluster. This includes the PG primary, each PG replica, the `pgBackRest` dedicated repository host, as well as `pgBouncer` (if enabled).

The default pod anti-affinity rules specifically attempt to schedule each pod created for the various deployments discussed above to a unique pod within the Kubernetes cluster.  By default, the anti affinity rules are applied using the `preferredDuringSchedulingIgnoredDuringExecution` affinity type, which means Kubernetes will attempt to adhere to the anti-affinity rules, but will not guarantee that they will be enforced.  This configuration therefore favors ensuring various pods comprising the PG cluster can always be scheduled, even if not always on the desired node.

It is also possible to use the `requiredDuringSchedulingIgnoredDuringExecution` affinity type when applying default pod anti-affinity.  The specific affinity type for a PG cluster can be specified by setting the`pod-anti-affinity` option available on the `pgo create cluster` command to either `required` (for `requiredDuringSchedulingIgnoredDuringExecution`) or `preferred` (for
`preferredDuringSchedulingIgnoredDuringExecution`). The same option can also be set globally
for all PG deployments by using the `PodAntiAffinity`setting in the `pgo.yaml` configuration file.

When the `required` option is utilized, a unique node will be required for each pod comprising the full PG cluster.  Therefore, as new deployments (e.g. `pgBouncer` and/or additional replicas) are
added, there must be enough nodes available without existing PG pods scheduled and deployed for the cluster, otherwise those deployments will not be successfully scheduled.  Therefore, while `required` supports a highly-available configuration by ensuring node failures only potentially impact
a single deployment in the cluster, it is necessary that the cluster has sufficient nodes
to support this configuration.

And lastly, pod anti-affinity can be disabled all together for all deployments by setting
either the `pod-anti-affinity` CLI option to `disable`, or by setting the `PodAntiAffinity`
setting in pgo.yaml to disabled.

**Checklist:**

 
**Type of Changes:**

 
Pod anti-affinity is not applied to the pods comprising a PG cluster.

[ch6372]

Default pod anti-affinity is applied to the pods comprising a PG cluster in order to schedule the pods comprising the PG cluster of different nodes.

**Other information**:

N/A

### DIFF
--- a/apis/cr/v1/cluster.go
+++ b/apis/cr/v1/cluster.go
@@ -16,6 +16,8 @@ package v1
 */
 
 import (
+	"fmt"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -65,6 +67,7 @@ type PgclusterSpec struct {
 	PswLastUpdate      string               `json:"pswlastupdate"`
 	CustomConfig       string               `json:"customconfig"`
 	UserLabels         map[string]string    `json:"userlabels"`
+	PodAntiAffinity    string               `json:"podPodAntiAffinity"`
 }
 
 // PgclusterList is the CRD that defines a Crunchy PG Cluster List
@@ -88,6 +91,14 @@ type PgclusterStatus struct {
 // swagger:ignore
 type PgclusterState string
 
+// PodAntiAffinityType defines the different types of type of anti-affinity rules applied to pg
+// clusters when utilizing the default pod anti-affinity rules provided by the PostgreSQL Operator,
+// which are enabled for a new pg cluster by default.  Valid Values include "required" for
+// requiredDuringSchedulingIgnoredDuringExecution anti-affinity, "preferred" for
+// preferredDuringSchedulingIgnoredDuringExecution anti-affinity, and "disabled" to disable the
+// default pod anti-affinity rules for the pg cluster all together.
+type PodAntiAffinityType string
+
 const (
 	// PgclusterStateCreated ...
 	PgclusterStateCreated PgclusterState = "pgcluster Created"
@@ -97,4 +108,29 @@ const (
 	PgclusterStateInitialized PgclusterState = "pgcluster Initialized"
 	// PgclusterStateRestore ...
 	PgclusterStateRestore PgclusterState = "pgcluster Restoring"
+
+	// PodAntiAffinityRequired results in requiredDuringSchedulingIgnoredDuringExecution for any
+	// default pod anti-affinity rules applied to pg custers
+	PodAntiAffinityRequired PodAntiAffinityType = "required"
+
+	// PodAntiAffinityPreffered results in preferredDuringSchedulingIgnoredDuringExecution for any
+	// default pod anti-affinity rules applied to pg custers
+	PodAntiAffinityPreffered PodAntiAffinityType = "preferred"
+
+	// PodAntiAffinityDisabled disables any default pod anti-affinity rules applied to pg custers
+	PodAntiAffinityDisabled PodAntiAffinityType = "disabled"
 )
+
+// ValidatePodAntiAffinityType is responsible for validating whether or not the type of pod
+// anti-affinity specified is valid
+func (p PodAntiAffinityType) ValidatePodAntiAffinityType() error {
+	switch p {
+	case
+		PodAntiAffinityRequired,
+		PodAntiAffinityPreffered,
+		PodAntiAffinityDisabled:
+		return nil
+	}
+	return fmt.Errorf("Invalid pod anti-affinity type.  Valid values are '%s', '%s' or '%s'",
+		PodAntiAffinityRequired, PodAntiAffinityPreffered, PodAntiAffinityDisabled)
+}
--- a/apis/cr/v1/cluster.go
+++ b/apis/cr/v1/cluster.go
@@ -123,7 +123,7 @@ const (
 
 // ValidatePodAntiAffinityType is responsible for validating whether or not the type of pod
 // anti-affinity specified is valid
-func (p PodAntiAffinityType) ValidatePodAntiAffinityType() error {
+func (p PodAntiAffinityType) Validate() error {
 	switch p {
 	case
 		PodAntiAffinityRequired,
--- a/apiserver/clusterservice/clusterimpl.go
+++ b/apiserver/clusterservice/clusterimpl.go
@@ -705,6 +705,21 @@ func CreateCluster(request *msgs.CreateClusterRequest, ns, pgouser string) msgs.
 			}
 		}
 
+		// if a value is provided in the request for PodAntiAffinity, then ensure is valid.  If
+		// it is, then set the user label for pod anti-affinity to the request value.  Otherwise,
+		// return the validation error.
+		if request.PodAntiAffinity != "" {
+			podAntiAffinityType := crv1.PodAntiAffinityType(request.PodAntiAffinity)
+			if err := podAntiAffinityType.ValidatePodAntiAffinityType(); err != nil {
+				resp.Status.Code = msgs.Error
+				resp.Status.Msg = err.Error()
+				return resp
+			} 
+			userLabelsMap[config.LABEL_POD_ANTI_AFFINITY] = request.PodAntiAffinity
+		} else {
+			userLabelsMap[config.LABEL_POD_ANTI_AFFINITY] = ""
+		}
+
 		// Create an instance of our CRD
 		newInstance := getClusterParams(request, clusterName, userLabelsMap, ns)
 		newInstance.ObjectMeta.Labels[config.LABEL_PGOUSER] = pgouser
@@ -935,6 +950,8 @@ func getClusterParams(request *msgs.CreateClusterRequest, name string, userLabel
 	}
 
 	spec.CustomConfig = request.CustomConfig
+
+	spec.PodAntiAffinity = request.PodAntiAffinity
 
 	labels := make(map[string]string)
 	labels[config.LABEL_NAME] = name
--- a/apiserver/clusterservice/clusterimpl.go
+++ b/apiserver/clusterservice/clusterimpl.go
@@ -710,7 +710,7 @@ func CreateCluster(request *msgs.CreateClusterRequest, ns, pgouser string) msgs.
 		// return the validation error.
 		if request.PodAntiAffinity != "" {
 			podAntiAffinityType := crv1.PodAntiAffinityType(request.PodAntiAffinity)
-			if err := podAntiAffinityType.ValidatePodAntiAffinityType(); err != nil {
+			if err := podAntiAffinityType.Validate(); err != nil {
 				resp.Status.Code = msgs.Error
 				resp.Status.Msg = err.Error()
 				return resp
--- a/conf/postgres-operator/affinity.json
+++ b/conf/postgres-operator/affinity.json
@@ -1,7 +1,6 @@
-        "affinity": {
             "nodeAffinity": {
             "preferredDuringSchedulingIgnoredDuringExecution": [{
-                "weight": 1,
+                "weight": 10,
                 "preference": {
                 "matchExpressions": [{
                     "key": "{{.NodeLabelKey}}",
@@ -13,4 +12,3 @@
                 }
             }]
             }
-        },
--- a/config/pgoconfig.go
+++ b/config/pgoconfig.go
@@ -498,7 +498,7 @@ func (c *PgoConfig) Validate() error {
 
 	// if provided, ensure that the type of pod anti-affinity specified is valid
 	podAntiAffinityType := crv1.PodAntiAffinityType(c.Cluster.PodAntiAffinity)
-	if err := podAntiAffinityType.ValidatePodAntiAffinityType(); podAntiAffinityType != "" && err != nil {
+	if err := podAntiAffinityType.Validate(); podAntiAffinityType != "" && err != nil {
 		return errors.New(errPrefix + "Invalid value provided for Cluster.PodAntiAffinityType")
 	}
 
--- a/config/pgoconfig.go
+++ b/config/pgoconfig.go
@@ -105,6 +105,10 @@ var AffinityTemplate *template.Template
 
 const affinityTemplatePath = "affinity.json"
 
+var PodAntiAffinityTemplate *template.Template
+
+const podAntiAffinityTemplatePath = "pod-anti-affinity.json"
+
 var PgoBackrestRepoServiceTemplate *template.Template
 
 const pgoBackrestRepoServiceTemplatePath = "pgo-backrest-repo-service-template.json"
@@ -228,6 +232,7 @@ type ClusterStruct struct {
 	PgmonitorPassword             string `yaml:"PgmonitorPassword"`
 	EnableCrunchyadm              bool   `yaml:"EnableCrunchyadm"`
 	DisableReplicaStartFailReinit bool   `yaml:"DisableReplicaStartFailReinit"`
+	PodAntiAffinity               string `yaml:"PodAntiAffinity"`
 }
 
 type StorageStruct struct {
@@ -490,6 +495,13 @@ func (c *PgoConfig) Validate() error {
 			return errors.New(errPrefix + msg)
 		}
 	}
+
+	// if provided, ensure that the type of pod anti-affinity specified is valid
+	podAntiAffinityType := crv1.PodAntiAffinityType(c.Cluster.PodAntiAffinity)
+	if err := podAntiAffinityType.ValidatePodAntiAffinityType(); podAntiAffinityType != "" && err != nil {
+		return errors.New(errPrefix + "Invalid value provided for Cluster.PodAntiAffinityType")
+	}
+
 	return err
 }
 
@@ -783,6 +795,11 @@ func (c *PgoConfig) GetConfig(clientset *kubernetes.Clientset, namespace string)
 	}
 
 	AffinityTemplate, err = c.LoadTemplate(cMap, rootPath, affinityTemplatePath)
+	if err != nil {
+		return err
+	}
+
+	PodAntiAffinityTemplate, err = c.LoadTemplate(cMap, rootPath, podAntiAffinityTemplatePath)
 	if err != nil {
 		return err
 	}
--- a/controller/podcontroller.go
+++ b/controller/podcontroller.go
@@ -157,10 +157,10 @@ func (c *PodController) onUpdate(oldObj, newObj interface{}) {
 			log.Error(err)
 			return
 		}
-		
-		// if the backrest repo pod is ready, then see if is reporting ready following an configuration
-		// update due to a failover over (i.e. to point the pgbackrest repo to the new primary).  if it
-		// is again ready following a failover event, the initiate a backup
+
+		// if the backrest repo pod is ready, then see if it is reporting ready following a
+		// configuration update due to a failover over (i.e. to point the pgbackrest repo to the
+		// new primary).  if it is again ready following a failover event, then initiate a backup
 		if isBackrestRepo && isPrimaryOnRoleChange(c.PodClientset, pgcluster,
 			newpod.ObjectMeta.Namespace) && isBackrestRepoReady(oldpod, newpod) {
 
@@ -547,8 +547,7 @@ func isPrimaryOnRoleChange(clientset *kubernetes.Clientset, pgcluster crv1.Pgclu
 	var err error
 
 	// return false right away if scope isn't set
-	if _, valExists := pgcluster.ObjectMeta.Labels[config.LABEL_PGHA_SCOPE]; 
-		!valExists {
+	if _, valExists := pgcluster.ObjectMeta.Labels[config.LABEL_PGHA_SCOPE]; !valExists {
 		return false
 	}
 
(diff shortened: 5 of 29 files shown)


apis/cr/v1/cluster.go
```diff
@@ -88,6 +91,14 @@ type PgclusterStatus struct {
 // swagger:ignore
 type PgclusterState string
 
+// PodAntiAffinityType defines the different types of type of anti-affinity rules applied to pg
+// clusters when utilizing the default pod anti-affinity rules provided by the PGO, which are
```
jkatz: I'd recommend referring to it as the "PostgreSQL Operator" as "PGO" is more likely to be grepped when looking up things related to the CLI

conf/postgres-operator/affinity.json
```diff
@@ -1,7 +1,6 @@
-        "affinity": {
             "nodeAffinity": {
             "preferredDuringSchedulingIgnoredDuringExecution": [{
-                "weight": 1,
+                "weight": 10,
```
jkatz: This appears to be set to `10` whereas in the Ansible role it's still at `1`

hugo/content/overview/pod-affinitiy-overview.md
```diff
@@ -0,0 +1,76 @@
...
+Or it can also be explicitly enabled globally for all clusters by setting `PodAntiAffinity` to
+`preferred` in the `pgo.yaml` configuration file.
+
+In addition to providing the ability to set default pod anti-affinity using the `preferred`
+option, i.e. an affinity rule that utilizes the `preferredDuringSchedulingIgnoredDuringExecution`
+affinity type, it is also possible to use `requiredDuringSchedulingIgnoredDuringExecution` for
+the default pod anti-affinity, specifically by setting `pod-anti-affinity` to `require`:
+
+```bash
+pgo create cluster mycluster --replica-count=2 --pod-anti-affinity=require
```
jkatz: Should be `required`

hugo/content/overview/pod-affinitiy-overview.md
```diff
@@ -0,0 +1,76 @@
...
+pgo create cluster mycluster --replica-count=2 --pod-anti-affinity=preferred
+```
+
+Or it can also be explicitly enabled globally for all clusters by setting `PodAntiAffinity` to
+`preferred` in the `pgo.yaml` configuration file.
+
+In addition to providing the ability to set default pod anti-affinity using the `preferred`
+option, i.e. an affinity rule that utilizes the `preferredDuringSchedulingIgnoredDuringExecution`
+affinity type, it is also possible to use `requiredDuringSchedulingIgnoredDuringExecution` for
+the default pod anti-affinity, specifically by setting `pod-anti-affinity` to `require`:
```
jkatz: Should be `required`

hugo/content/overview/pod-affinitiy-overview.md
```diff
@@ -0,0 +1,76 @@
...
+highly-available by ensuring that node failures do not impact any other deployments in the cluster.
+However, this does mean that the PG primary, each PG replica, the `pgBackRestRepo` and `pgBouncer`,
+will each require a unique node, meaning the minimum number of nodes required for the Kubernetes 
+cluster will increase as more deployments (e.g. PG replicas) are added to the PG cluster.  Further,
+if an insufficient number of nodes are available to support this configuration, certain deployments 
+will fail, since it will not be possible for Kubernetes to successfully schedule the pods for each 
+deployment.
+
+And finally, it is also possible to disable the default pod anti-affinity settings all together.
+Using the `pgo` CLI this is done by setting `pod-anti-affinity` to `disable`, e.g.:
```
jkatz: `disabled`

hugo/content/overview/pod-affinitiy-overview.md
```diff
@@ -0,0 +1,76 @@
...
+
+```bash
+pgo create cluster mycluster --replica-count=2 --pod-anti-affinity=disable
```
jkatz: `disabled`

apis/cr/v1/cluster.go
andrewlecuyer: Changed to "PostgreSQL Operator".

conf/postgres-operator/affinity.json
andrewlecuyer: Updated the `weight` to `10` in the Ansible installer to align with the setting here.  This is to give the node affinity precedence when node preference is being determined.

hugo/content/overview/pod-affinitiy-overview.md
andrewlecuyer: Changed to `required`.

hugo/content/overview/pod-affinitiy-overview.md
andrewlecuyer: Changed to `required`.

hugo/content/overview/pod-affinitiy-overview.md
andrewlecuyer: Changed to `disabled`.

hugo/content/overview/pod-affinitiy-overview.md
andrewlecuyer: Changed to `disabled`.

hugo/content/Configuration/pgo-yaml-configuration.md
```diff
@@ -53,7 +53,8 @@ The *pgo.yaml* file is broken into major sections as described below:
 |Storage.storage1.StorageType        |supported values are either *dynamic*,  *create*,  if not supplied, *create* is used
 |Fsgroup        | optional, if set, will cause a *SecurityContext* and *fsGroup* attributes to be added to generated Pod and Deployment definitions
 |SupplementalGroups        | optional, if set, will cause a SecurityContext to be added to generated Pod and Deployment definitions
-|MatchLabels        | optional, if set, will cause the PVC to add a *matchlabels* selector in order to match a PV, only useful when the StorageType is *create*, when specified a label of *key=value* is added to the PVC as a match criteria
+|MatchLabels        | optional, if set, will cause the PVC to add a *matchlabels* selector in order to match a PV, only useful when the StorageType is *create*, when specified a label of
+|PodAntiAffinity        | either `preferred`, `require` or `disable` to either specify the type of affinity that should be utilized for the default pod anti-affinity applied to PG clusters, or to disable default pod anti-affinity all together  (default `preferred`)
```
cbandy: Should be:
> `preferred`, `required` or `disabled`

hugo/content/overview/pod-affinitiy-overview.md
```diff
@@ -0,0 +1,76 @@
...
+- The primary PG deployment
+- The deployments for each PG replica
+- The `pgBackrest` dedicated repostiory deployment
+- The `pgBouncer` deployment (if enabled for the cluster)
+
+With default pod anti-affinity enabled, Kubernetes will attempt to schedule the pods created
+by each individual deployment above on a unique node, but at the same time will not guarentee that
+each deployment will be scheduled on a unqiue node.  This therefore ensures that the various pods
+comprising the PG cluster can always be scheduled, even if though they might not end up on the 
+desired node.  This is specifically done using the following:
```
cbandy: Suggest:

```
With default pod anti-affinity enabled, Kubernetes will attempt (but not guarantee) to schedule
pods created by each separate deployment above on a unique node. This ensures that the pods
comprising the PG cluster can always be scheduled, though perhaps not on the desired node.
This is specifically done using the following:
```

hugo/content/overview/pod-affinitiy-overview.md
```diff
@@ -0,0 +1,76 @@
...
+In addition to providing the ability to set default pod anti-affinity using the `preferred`
+option, i.e. an affinity rule that utilizes the `preferredDuringSchedulingIgnoredDuringExecution`
+affinity type, it is also possible to use `requiredDuringSchedulingIgnoredDuringExecution` for
+the default pod anti-affinity, specifically by setting `pod-anti-affinity` to `required`:
+
+```bash
+pgo create cluster mycluster --replica-count=2 --pod-anti-affinity=required
+```
+
+Or similarly the `require` option can be enabled globally for all clusters by setting
```
cbandy: Should be `required` here and in the following paragraph.

hugo/content/overview/pod-affinitiy-overview.md
```diff
@@ -0,0 +1,76 @@
...
+
+And finally, it is also possible to disable the default pod anti-affinity settings all together.
+Using the `pgo` CLI this is done by setting `pod-anti-affinity` to `disabled`, e.g.:
+
+```bash
+pgo create cluster mycluster --replica-count=2 --pod-anti-affinity=disabled
+```
+
+Or the same setting can be applied globally to all clusters created by setting `PodAntiAffinity` 
+to `require` in the `pgo.yaml` configuration file.
```
cbandy: Should be `disabled` here, if I understand correctly.

operator/clusterutilities.go
```diff
@@ -43,12 +43,30 @@ const AFFINITY_NOTINOperator = "NotIn"
...
 const DefaultPgHaConfigMapSuffix = "pgha-default-config"
 
+// affinityType represents the two affinity types provided by Kubernetes, specifically
+// either preferredDuringSchedulingIgnoredDuringExecution or
+// requiredDuringSchedulingIgnoredDuringExecution
+type affinityType string
+
+const (
+	requireScheduleIgnoreExec affinityType = "requiredDuringSchedulingIgnoredDuringExecution"
+	preferScheduleIgnoreExec   affinityType = "preferredDuringSchedulingIgnoredDuringExecution"
```
cbandy: Whitespace looks off. Might need `go fmt` on this file.

pgo/cmd/create.go
```diff
@@ -252,6 +253,9 @@ func init() {
 	createClusterCmd.Flags().IntVarP(&Series, "series", "e", 1, "The number of clusters to create in a series.")
 	createClusterCmd.Flags().IntVarP(&ClusterReplicaCount, "replica-count", "", 0, "The number of replicas to create as part of the cluster.")
 	createClusterCmd.Flags().StringVarP(&ContainerResources, "resources-config", "r", "", "The name of a container resource configuration in pgo.yaml that holds CPU and memory requests and limits.")
+	createClusterCmd.Flags().StringVarP(&PodAntiAffinity, "pod-anti-affinity", "", "",
+		"Specifies the type of anti-affinity that should be utilized when applying  "+
+			"default pod anti-affinity rules to PG clusters (default \"preferred\")")
```
cbandy: The default value is `""` which, if I understand correctly, defers to pgo.yaml which then defaults to `preferred`. Should we refrain from mentioning "default" here?

config/pgoconfig.go
```diff
@@ -490,6 +495,13 @@ func (c *PgoConfig) Validate() error {
 			return errors.New(errPrefix + msg)
 		}
 	}
+
+	// if provided, ensure that the type of pod anti-affinity specified is valid
+	podAntiAffinityType := crv1.PodAntiAffinityType(c.Cluster.PodAntiAffinity)
+	if err := podAntiAffinityType.ValidatePodAntiAffinityType(); podAntiAffinityType != "" && err != nil {
```
abrightwell: Shouldn't the validation have caught the `""` case here?

controller/podcontroller.go
```diff
@@ -151,28 +151,31 @@ func (c *PodController) onUpdate(oldObj, newObj interface{}) {
...
 			log.Error(err)
 			return
 		}
-		_, err = backrest.CreatePostFailoverBackup(c.PodClient,
-			newpod.ObjectMeta.Namespace, clusterName, newpod.Name)
-		if err != nil {
-			log.Error(err)
-			return
+		
+		// if the backrest repo pod is ready, then see if is reporting ready following an configuration
```
abrightwell: `then see if reporting is ready following a configuration`

controller/podcontroller.go
```diff
@@ -151,28 +151,31 @@ func (c *PodController) onUpdate(oldObj, newObj interface{}) {
...
+		// update due to a failover over (i.e. to point the pgbackrest repo to the new primary).  if it
+		// is again ready following a failover event, the initiate a backup
```
abrightwell: `then`

operator/clusterutilities.go
```diff
@@ -422,6 +441,67 @@ func GetReplicaAffinity(clusterLabels, replicaLabels map[string]string) string {
...
+	var affinityTypeParam crv1.PodAntiAffinityType
+	if podAntiAffinityType != "" {
+		affinityTypeParam = crv1.PodAntiAffinityType(podAntiAffinityType)
+	} else if Pgo.Cluster.PodAntiAffinity != "" {
+		affinityTypeParam = crv1.PodAntiAffinityType(Pgo.Cluster.PodAntiAffinity)
+	} 
+
+	// verify that the affinity type provided is valid (i.e. 'required' or 'preffered'), and
+	// log an error and return an empty string if not
+	if err := affinityTypeParam.ValidatePodAntiAffinityType(); affinityTypeParam != "" &&
```
abrightwell: Same here, shouldn't the `""` be handled by validation?

apiserver/clusterservice/clusterimpl.go
```diff
@@ -705,6 +705,21 @@ func CreateCluster(request *msgs.CreateClusterRequest, ns, pgouser string) msgs.
...
+		if request.PodAntiAffinity != "" {
+			podAntiAffinityType := crv1.PodAntiAffinityType(request.PodAntiAffinity)
+			if err := podAntiAffinityType.ValidatePodAntiAffinityType(); err != nil {
+				resp.Status.Code = msgs.Error
+				resp.Status.Msg = err.Error()
+				return resp
+			} 
+			userLabelsMap[config.LABEL_POD_ANTI_AFFINITY] = request.PodAntiAffinity
+		} else {
+			userLabelsMap[config.LABEL_POD_ANTI_AFFINITY] = ""
```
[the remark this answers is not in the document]
andrewlecuyer: Correct, `exists` should simply check for the presence of the label, regardless of its value.  So if a actual anti affinity value is not provided by the user, the label will be `pgha-pod-anti-affinity=""`.

pgo/cmd/create.go
andrewlecuyer: If no value is provided for `pod-anti-affinity`, then the default value of `preferred` will be applied when creating the cluster.  I therefore mentioned the default as being `preferred` here, so that the user will understand that `preferred` will the default form of anti-affinity applied to their cluster once created.

apiserver/clusterservice/clusterimpl.go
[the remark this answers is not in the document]
cbandy: I thought this was trying to trigger the `disabled` behavior, but I see now that we don't emit any `podAntiAffinity` in the pod template in that case.

What's the intention of this rule in the selector?
https://github.com/CrunchyData/postgres-operator/blob/9b20095b9ac85cd65341a0306f31d1b67cda2306/conf/postgres-operator/pod-anti-affinity.json#L17-L20

config/pgoconfig.go
andrewlecuyer: If an empty affinity type is provided, then it will default to `preferred` once processed by the operator.  Therefore, since a empty value is not intended to be a affinity type, it does not need to be validated.  Instead, it simply indicates that the default should be applied (specifically by the Operator).

operator/clusterutilities.go
andrewlecuyer: If an empty affinity type is provided, then it will default to preferred once processed by the operator. Therefore, since a empty value is not intended to be a affinity type, it does not need to be validated. Instead, it simply indicates that the default should be applied (specifically by the Operator).

apis/cr/v1/cluster.go
```diff
@@ -97,4 +108,29 @@ const (
...
+	// default pod anti-affinity rules applied to pg custers
+	PodAntiAffinityPreffered PodAntiAffinityType = "preferred"
+
+	// PodAntiAffinityDisabled disables any default pod anti-affinity rules applied to pg custers
+	PodAntiAffinityDisabled PodAntiAffinityType = "disabled"
 )
+
+// ValidatePodAntiAffinityType is responsible for validating whether or not the type of pod
+// anti-affinity specified is valid
+func (p PodAntiAffinityType) ValidatePodAntiAffinityType() error {
```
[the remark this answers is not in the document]
andrewlecuyer: I will go ahead and change to `Validate()` to align with some other similar validation functions we have, e.g. those in the benchmark API.

As for what is returned (i.e. `error` vs `bool`) I would like to defer any changes and discuss further, especially in regards to how we perform validation across the PostgreSQL Operator in general moving forward.  My problem with simply returning `bool` is that all context for why it failed validation, or if more than one failure occurred (although not applicable in the above use case), is lost.  And since an error here will often result in an error message being logged somewhere in the operator logs, or in the case of the apiserver will be returned to the caller (e.g. as an error response to the `pgo create cluster`command), I want to ensure there is more than simply "invalid value", e.g. context around why it was invalid and what value it instead should/could have been.  

I could require the caller to display the pertinent error information, but now the caller, and not the model where the various types actually live, is responsible for figuring out why it failed and displaying the proper error (and error messages can always be ignored by the caller anyway, allowing them to display any error they wish, if at all).  With this approach I have a clear error message that can easily be identified in the logs, or by any UI end user, in a consistent way, allowing them to easily correct the validation issue (and without having to re-implement and continuously update logic for why something failed everywhere this is called, e.g. all UI layers, the operater itself, apiserver, etc.).

In general (and while not currently implemented in our code), when it comes to designing validation I tend to think of Matrin Fowler's notification pattern, which results in validation error information being returned to the user in the form of collections of errors, which each provide context as to what validation failed and why.  Anyway, considering the above, I'd like to defer any further changes here pending further discussion around our overall strategy and design for PostgreSQL Operator validation.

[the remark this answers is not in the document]
jkatz: WFM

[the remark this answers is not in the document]
abrightwell: > As for what is returned (i.e. error vs bool) I would like to defer any changes and discuss further, especially in regards to how we perform validation across the PostgreSQL Operator in general moving forward.

I can accept that.

> And since an error here will often result in an error message being logged somewhere in the operator logs, or in the case of the apiserver will be returned to the caller (e.g. as an error response to the pgo create clustercommand), I want to ensure there is more than simply "invalid value", e.g. context around why it was invalid and what value it instead should/could have been.

This is what I mean by mixing 'presentation'. I see value in communicating back what is 'wrong' about the particular thing being validated, however, I feel that `error` is the wrong place to do this. I would obviously agree that an error is meant to convey something/what went wrong. But I believe there are better ways for us to do that than simply `error`. For example, I'd rather see something implement the `Error` interface, that allows for 1) allows it to be handled as an `error` but 2) provides a mechanism to get more information about what went wrong without encoding it into a string.

At any rate, I'm fine with deferring for now. But it is certainly something we need to revisit and address in a more idiomatic way.

apiserver/clusterservice/clusterimpl.go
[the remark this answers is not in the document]
andrewlecuyer: We only want the pod affinity rule to apply to pods that have at have the `pod-anti-affinity` label, so the idea here is that we match pods where this label exists, but the specific value of that label is irrelevant.

apis/cr/v1/cluster.go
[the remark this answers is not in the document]
abrightwell: And FWIW, the new Wrap/Unwrap feature that was introduced in 1.13 would be an acceptable approach. It would enable a reusable type across all validation, while allowing for the composition/collection of errors for what specifically went wrong as a result.

Food for thought.

https://blog.golang.org/go1.13-errors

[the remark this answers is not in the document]
andrewlecuyer: Yea, great feature - prior to 1.13 would reach for the following to get wrap/unwrap functionality:

hugo/content/Configuration/pgo-yaml-configuration.md
andrewlecuyer: Updated

hugo/content/overview/pod-affinitiy-overview.md
andrewlecuyer: Updated to `required`

operator/clusterutilities.go
andrewlecuyer: Formatting has been fixed

controller/podcontroller.go
andrewlecuyer: Updated to `then`

controller/podcontroller.go
andrewlecuyer: Updated wording

hugo/content/overview/pod-affinitiy-overview.md
andrewlecuyer: Changed to `disabled`

hugo/content/overview/pod-affinitiy-overview.md
andrewlecuyer: Updated wording

PR #1164, opened 2019-12-19
[ch6673] Fix for cluster names with "-" in scaledown/failover query

jkatz:
**Checklist:**

 
**Type of Changes:**

 
Cluster names that had dashes (e.g. "broken-cluster") would not properly
display replica names in the `pgo scaledown --query` and
`pgo failover --query` commands. This is due to us trying to leverage
the information in Patroni to calculate values that are made available
in the output of these commands (e.g. replication lag) so we don't
reinvent the wheel, but have to contend with Patroni get the names
from the pods themselves.

The fix uses the information we've collected in the instance-node map
to look for the instance name prefix, which will be less brittle than
the regular expression I had concocted in 4fbcd2239.

**Other information**:

Issue: [ch6673]

### DIFF
--- a/util/failover.go
+++ b/util/failover.go
@@ -19,7 +19,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"regexp"
+	"strings"
 
 	"github.com/crunchydata/postgres-operator/config"
 	"github.com/crunchydata/postgres-operator/kubeapi"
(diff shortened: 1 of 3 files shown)


util/failover.go
```diff
@@ -190,12 +189,17 @@ func ReplicationStatus(request ReplicationStatusRequest) (ReplicationStatusRespo
...
-		instance.Node = instanceNodeMap[instance.Name]
+		// first part of the name and is kept on a deployment label. We have these
+		// available in our instanceNodeMap, and because we skip over the primary,
+		// this will not lead to false positive
+		//
+		// This is not the cleanest way of doing it, but it works
+		for name, node := range instanceNodeMap {
+			if strings.HasPrefix(rawInstance.PodName, name) {
+				instance.Name = name
+				instance.Node = node
```
abrightwell: Do you need to `break` out of the loop here given that you've found the instance?

jkatz: Yes.

PR #1169, opened 2019-12-20
Allow separate image pull secrets for container registry

cbandy:
**Checklist:**

 
**Type of Changes:**

 
This adds `ccp_image_pull_secret` to complement `ccp_image_prefix` the same way `pgo_image_pull_secret` is available for `pgo_image_prefix`.

The two image prefixes could refer to different registries with different credentials. This also provides a way to specify image pull secrets for the `pgo-metrics` Ansible role.

Issue: [ch5511]

### DIFF
--- a/ansible/roles/pgo-metrics/templates/grafana-rbac.json.j2
+++ b/ansible/roles/pgo-metrics/templates/grafana-rbac.json.j2
@@ -5,5 +5,10 @@
         "name": "grafana",
         "namespace": "{{ metrics_namespace }}"
     },
-    "automountServiceAccountToken": false
+    "automountServiceAccountToken": false,
+    "imagePullSecrets": [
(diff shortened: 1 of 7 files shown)


ansible/roles/pgo-metrics/templates/grafana-rbac.json.j2
```diff
@@ -5,5 +5,10 @@
         "name": "grafana",
         "namespace": "{{ metrics_namespace }}"
     },
-    "automountServiceAccountToken": false
+    "automountServiceAccountToken": false,
+    "imagePullSecrets": [
+{% if ccp_image_pull_secret %}
+        { "name": "{{ ccp_image_pull_secret }}" }
```
cbandy: This works for a single element in the array, but we might want to switch to YAML if this gets any longer.

PR #1171, opened 2019-12-21
pgo-client deployment file missing fields

jmckulk:
The deployment file was updated to `apps/v1` instead of `extensions/v1beta` in a
previous commit. This change required a new field in the deployment json file
under `spec.selector`. The field has been added to the ansible deployment
template to user `spec.selector.matchLabels` which is set to match the
`spec.template.labels`. This is based on Kubernetes API requirements that the two must match.

**Checklist:**

 
**Type of Changes:**

 
Adds label selectors to pods. 

**Other information**:
[ch6738]

### DIFF
--- a/ansible/roles/pgo-operator/templates/pgo-client.json.j2
+++ b/ansible/roles/pgo-operator/templates/pgo-client.json.j2
@@ -9,12 +9,12 @@
     },
     "spec": {
         "replicas": 1,
-	"selector": {
-		"matchLabels": {
-			"name": "pgo-client",
-			"vendor": "crunchydata"
-		}	
-	},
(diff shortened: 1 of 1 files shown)


ansible/roles/pgo-operator/templates/pgo-client.json.j2
```diff
@@ -9,6 +9,12 @@
     },
     "spec": {
         "replicas": 1,
+	"selector": {
+		"matchLabels": {
+			"name": "pgo-client",
+			"vendor": "crunchydata"
+		}	
+	},
```
jkatz: Please make the indentation match.

PR #1181, opened 2019-12-24
Do not panic if PGHA_PGBACKREST_LOCAL_S3_STORAGE is not set

jkatz:
**Checklist:**

 
**Type of Changes:**

 
`PGHA_PGBACKREST_LOCAL_S3_STORAGE` causes `pgo-backrest` to PANIC and exit if it is unset.

This could happen for a variety of reasons, i.e. it is simply just
not set in one's environment. The default is to set this value to
"false" as it is using strconv.ParseBool, therefore if it errors,
it's okay.

### DIFF
--- a/pgo-backrest/pgo-backrest.go
+++ b/pgo-backrest/pgo-backrest.go
@@ -79,10 +79,10 @@ func main() {
 	REPO_TYPE := os.Getenv("PGBACKREST_REPO_TYPE")
 	log.Debugf("setting REPO_TYPE to %s", REPO_TYPE)
 
-	PGHA_PGBACKREST_LOCAL_S3_STORAGE, err := strconv.ParseBool(os.Getenv("PGHA_PGBACKREST_LOCAL_S3_STORAGE"))
-	if err != nil {
-		panic(err)
-	}
+	// determine the setting of PGHA_PGBACKREST_LOCAL_S3_STORAGE
+	// we will discard the error and treat the value as "false" if it is not
+	// explicitly set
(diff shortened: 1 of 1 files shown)


pgo-backrest/pgo-backrest.go
```diff
@@ -79,9 +79,11 @@ func main() {
 	REPO_TYPE := os.Getenv("PGBACKREST_REPO_TYPE")
 	log.Debugf("setting REPO_TYPE to %s", REPO_TYPE)
 
+	// determine the setting of PGHA_PGBACKREST_LOCAL_S3_STORAGE
+	// if there is an error, we will treat this setting as false
 	PGHA_PGBACKREST_LOCAL_S3_STORAGE, err := strconv.ParseBool(os.Getenv("PGHA_PGBACKREST_LOCAL_S3_STORAGE"))
 	if err != nil {
-		panic(err)
+		log.Warnf("%s, PGHA_PGBACKREST_LOCAL_S3_STORAGE is set to false", err)
```
wilybrace: Is the warning intended to replace a perceived need for alarm (from the prior panic)? That is, does the program expect the variable to be set to something and not interpreted?

Based on the later usage of the variable, it seems that would be the place to issue a warning about no local and no S3 being a (bad?) configuration. If this warning isn't something for someone looking at logs to address, I'd recommend leaving it quiet or reduced down to INFO level.

For this use case, it seems reasonable to ignore the ParseBool error if the operating model is that boolean switches default to false unless someone affirmatively sets it to true.

Issue #1180, opened 2019-12-24
pgo scaledown does not work if postgresql is configured with tls/ssl encryption


yuanlinios:
**Describe the bug**
`pgo scaledown` does not work if pg cluster is configured with tls/ssl encryption (encryption only, no tls/ssl authN)

**To Reproduce**
Customized `pg_hba.conf`
```
local    all          all                               trust
host     all          all           127.0.0.1/32        md5
hostssl  all          all           0.0.0.0/0           md5
hostssl  replication  primaryuser   0.0.0.0/0           md5
```

Create pg cluster with custom config
```
oc create -n pg-proj configmap pg-custom-config \
    --from-file=configs/server.crt \
    --from-file=configs/server.key \
    --from-file=configs/pg_hba.conf \
    --from-file=configs/postgresql.conf

pgo create cluster devcluster --replica-count=1 --custom-config pg-custom-config -n pg-proj
```

Try to run `pgo scaldown` coomand
```
pgo scaledown devcluster -n pg-proj --query
Error: pq: no pg_hba.conf entry for host "10.9.3.4", user "postgres", database "postgres", SSL off
```

**Expected behavior**
`pgo scaledown` command should be executed successfully

**Please tell us about your environment:**

* Container Image Tag: crunchydata/crunchy-postgres:centos7-12.1-4.1.1
* PostgreSQL Version: 12.1
* Platform (Docker, Kubernetes, OpenShift): Openshift
* Platform Version: OCP 3.11

**Additional context**
The documentation mentions only following commands do not work with ssl enabled in pg cluster
```
pgo test
pgo load
pgo apply
```

jkatz:
Hi @yuanlinios,

Thanks for the report!

The behavior of `pgo scaledown` has changed in the upcoming v4.2.0 release (which you can read about here: https://github.com/CrunchyData/postgres-operator/releases/tag/v4.2.0-beta.1) where it does not perform a query directly to the PostgreSQL database in order to get the information as part of the `--query` flag. As such, this issue is resolved in that release as it does not matter if PostgreSQL has TLS enabled

Similarly, in `v4.2.0`, `pgo test` no longer connects directly to the PostgreSQL database and as such the issue is resolved there.

As the nature of these changes qualify as new features, they will not be backpatched to `REL_4_1`.

[...]

PR #1182, opened 2019-12-24
 Execute policies directly on the PostgreSQL primary Pod

jkatz:
**Checklist:**

 
**Type of Changes:**

 
Polices that were being applied (via `pgo apply` and `pgo create cluster --policies`) were executing a separate script that
would connect to a PostgreSQL primary over an unencrypted connection
as a PostgreSQL superuser to execute the policy. The issue is further
exacerbated if one choose to enable TLS connections over all remote
connections to the database.

To ameliorate several problems here, any SQL policy is now applied by
directly connection to the Pod and executing the command. This
eliminates both the insecure connection and allows for users to operate
correctly within a full TLS environment.

**Other information**:

This also:

- Removes the `runpsql.sh` script and its references from containers
- Code around policy management that was no longer being referenced

Reported in Issue #1180


util/policy.go
```diff
@@ -17,51 +17,86 @@ package util
...
+	// NOTE: this executes as the "postgres" user on the "postgres" database,
+	// because that is what the existing functionality does
+	//
+	// However, unlike the previous implementation, this will connect over a UNIX
+	// socket. There are certainly additional improvements that can be made, but
+	// this gets us closer to what we want to do
+	command := []string{
+		"psql",
+		"postgres",
+		"postgres",
```
cbandy: I recall there being subtle differences between running a script through psql stdin and through a file. The latter gives line numbers in error messages, for example. Can we use `-f -` here?

```
# psql postgres postgres <<< 'select 1/0'
ERROR:  division by zero
# psql -f - postgres postgres <<< 'select 1/0'
psql:<stdin>:1: ERROR:  division by zero
```

jkatz: @cbandy I would like to: however the issue as I see it is that the file is not actually on the Pod: we would need to put it into a volume/secret/configmap and mount it.

For the nature and timing of this fix, I'm not prepared to do a full scale refactor of how the policies work. While I do agree the format of the error message is nicer, this would technically be new functionality based on how the policy system works.

cbandy: `-f -` tells psql to read from stdin as though it were a file.

jkatz: @cbandy I missed the `-` in the `-f -`. Agreed with the analysis. Will make the change

jkatz: @cbandy OK I changed to `-f -` and tested. All appears to work.

PR #1184, opened 2019-12-29
Update Custom Config Examples & Fix Label Used to Identify Cluster Scope when Scaling

andrewlecuyer:
This PR brings the `custom-config` and `custom-config-ssl` examples inline with the latest methods for customizing a PG cluster using the `crunchy-postgres-ha` container (e.g. enabling SSL), while also ensuring the proper Patroni scope is defined for a new replica when scaling the cluster.

Additionally, it provides an example of how client certificates can be utilized by the replication user for streaming replication instead of a password.

**Checklist:**

 
**Type of Changes:**

 
- The `custom-config` and `custom-config-ssl` examples have not been updated with the latest methods for applying custom configuration using the `crunchy-postgres-ha` container (includes configuring SSL)
- Only passwords can be utilized by the replication user when connecting to the DB for streaming replication
- The name of the current primary is utilized to define the cluster scope for the new replica when scaling, resulting in the incorrect scope and therefore errors when initializing a replica following a failover

[ch6755]
[ch6788]
[ch6752]

- The `custom-config` and `custom-config-ssl` examples have been updated with the latest methods for applying custom configuration using the `crunchy-postgres-ha` container (includes configuring SSL)
- Client certificates can be utilized by the replication user when connecting to the DB for streaming replication
- The proper PGHA scope is obtained from the `pgcluster` when initializing a new replica after scaling, ensuring the replica properly joins the existing cluster and successfully initializes

**Other information**:

N/A


examples/custom-config-ssl/configs/postgres-ha.yaml
```diff
@@ -0,0 +1,19 @@
...
+        log_directory: pglogs
+        log_min_duration_statement: 0
+        log_statement: none
+        max_wal_senders: 6
+        shared_preload_libraries: pgaudit.so
+postgresql:  
+  pg_hba:
+    - local all postgres peer
+    - local all crunchyadm peer
+    - hostssl replication primaryuser 0.0.0.0/0 cert clientcert=1
```
jkatz: As of PostgreSQL 12, `clientcert` is now a loaded parameter.

In PostgreSQL 12, `clientcert` now accepts two values: `client=verify-ca` and `clientcert=verify-full`. This are similar to the `sslmode` versions of the parameter, i.e. 

- `verify-ca`: The PostgreSQL server will verify that the certificate that the client presents was signed by certificate in the trusted CA chain
- `verify-full`: Like `verify-ca`, but the additional check that the `CN` in the certificate matches the username (which could be aided by a map file). This is also pretty cool :smile:

`clientcert=1` is equivalent to `clientcert=verify-ca` in PostgreSQL 12...I believe `clientcert=1` is still honored in PostgreSQL 12, but would have to check.

examples/custom-config-ssl/configs/postgres-ha.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - hostssl all primaryuser 0.0.0.0/0 reject
+    - hostssl all all 0.0.0.0/0 cert clientcert=1 map=ssl-testuser
+  pg_ident:
+    - ssl-testuser testuser@crunchydata.com testuser
```
jkatz: Is the example showing how to make a certificate with `cn=testuser@crunchydata.com`? If so, this works, otherwise, it's kind of silly to include the map.

If it's not showing how to do that, I would suggest we circle back up to make a map example that's more helpful, but we may do that anyway as we continue to advance our TLS enabled integration.

examples/custom-config/postgres-ha.yaml
```diff
@@ -0,0 +1,18 @@
...
+        max_wal_senders: 6
+        shared_preload_libraries: pgaudit.so
+postgresql:  
+  pg_hba:
+    - local all postgres peer
+    - local all crunchyadm peer
+    - host replication primaryuser 0.0.0.0/0 md5
+    - host all primaryuser 0.0.0.0/0 reject
+    - host all testuser1 0.0.0.0/0 md5
+    - host all testuser2 0.0.0.0/0 md5
```
jkatz: In our current setup, it would be fine to do

```yaml
- host all postgres 0.0.0.0/0md5
```

also because without that, we block the `postgres` user from being able to log in and take the actions that are available in `pgo [create|update|delete] user` user.

examples/custom-config-ssl/configs/postgres-ha.yaml
```diff
@@ -0,0 +1,19 @@
...
+        log_statement: none
+        max_wal_senders: 6
+        shared_preload_libraries: pgaudit.so
+postgresql:  
+  pg_hba:
+    - local all postgres peer
+    - local all crunchyadm peer
+    - hostssl replication primaryuser 0.0.0.0/0 cert clientcert=1
+    - hostssl all primaryuser 0.0.0.0/0 reject
+    - hostssl all all 0.0.0.0/0 cert clientcert=1 map=ssl-testuser
```
jkatz: This might be accidentally blocking `pgo [create|update|delete] user` for this iteration, unless we modify those commands to use peer auth. For the latter (peer auth) method, what would hurt as at this moment is that we allow for selectors to do that in bulk and the command is synchronous, so it would be a very costly operation

examples/custom-config/postgres-ha.yaml
```diff
@@ -0,0 +1,18 @@
+---
+bootstrap:
+  dcs:
+    postgresql:
+      parameters:
+        log_directory: pglogs
+        log_min_duration_statement: 0
+        log_statement: none
+        max_wal_senders: 6
+        shared_preload_libraries: pgaudit.so
```
jkatz: If we add in `pg_stat_statements` (which we do in https://github.com/CrunchyData/postgres-operator/pull/1184/commits/aaf69811b6e361ecf1b1d1b7168d19aacd9fa9a5#diff-64d4c5af3fcff84b3768e377f9654aa6R4-R5) we need to add it to `shared_preload_libraries`:

```yaml
shared_preload_libraries: "pgaudit.so,pg_stat_statements"
```

operator/cluster/clusterlogic.go
```diff
@@ -152,7 +153,7 @@ func AddCluster(clientset *kubernetes.Clientset, client *rest.RESTClient, cl *cr
 		log.Debugf("Custom postgres-ha configmap not found, creating configMap with default postgres-ha config file")
 		operator.AddDefaultPostgresHaConfigMap(clientset, cl, deploymentFields.IsInit, true, namespace)
 	} else {
-		configMap, found := kubeapi.GetConfigMap(clientset, config.GLOBAL_CUSTOM_CONFIGMAP, namespace)
+		configMap, found := kubeapi.GetConfigMap(clientset, strings.Trim(deploymentFields.ConfVolume, "\""), namespace)
```
jkatz: I'd recommend putting the Trimmed configmap in its own variable to make it easier to debug in the future.

operator/clusterutilities.go
```diff
@@ -258,25 +258,24 @@ func GetConfVolume(clientset *kubernetes.Clientset, cl *crv1.Pgcluster, namespac
...
-			_, found = kubeapi.GetConfigMap(clientset, cl.Spec.CustomConfig, namespace)
-			if !found {
-				//you should NOT get this error because of apiserver validation of this value!
-				log.Errorf("%s was not found, error, skipping user provided configMap", cl.Spec.CustomConfig)
-			} else {
-				log.Debugf("user provided configmap %s was used for this cluster", cl.Spec.CustomConfig)
-				configMapStr = "\"" + cl.Spec.CustomConfig + "\""
-			}
-		}
+		return "\"pgo-custom-pg-config\""
```
jkatz: Perhaps

```go
return `"pgo-custom-pg-config"`
```

saves on escapes :smile:

examples/custom-config-ssl/configs/postgres-ha.yaml
andrewlecuyer: The common name for the client certificate (`client.crt`) generated in the example is `testuser@crunchydata.com`, so the map is utilized when authenticating via that certificate:

https://github.com/CrunchyData/postgres-operator/blob/045820b21eab0a8b44b6515a4c856694fc8ecdeb/examples/custom-config-ssl/create.sh#L57

examples/custom-config/postgres-ha.yaml
andrewlecuyer: I will go ahead and update the `setup.sql` to no longer create the `pg_stat_statements` extension.  We typically enable both of these by default, and the intention here was to simply show the ability to customize that behavior.

examples/custom-config-ssl/configs/postgres-ha.yaml
andrewlecuyer: The PG 12 source confirms that `clientcert=1` is equivalent to `clientcert=verify-ca`:

https://github.com/postgres/postgres/blob/b4507a22fed9c2398895d1d4b94815ad1a053bb0/src/backend/libpq/hba.c#L1709-L1713

examples/custom-config/postgres-ha.yaml
andrewlecuyer: This was added in a subsequent commit: https://github.com/CrunchyData/postgres-operator/pull/1184/commits/7549466771ffdaff8d6f644d46332049da504d0f

examples/custom-config-ssl/configs/postgres-ha.yaml
andrewlecuyer: The following line was added in a subsequent commit: 7549466

```
- host all postgres 0.0.0.0/0 md5
```

operator/cluster/clusterlogic.go
andrewlecuyer: The trimmed string has now been separated into its own variable.

operator/clusterutilities.go
andrewlecuyer: Syntax has been updated.

examples/custom-config-ssl/configs/postgres-ha.yaml
jkatz: Great. In the future we should update this example to be a bit more general purpose (the PostgreSQL docs have a good example). For now it's okay.

PR #1196, opened 2020-01-08
changes default CA exclusion for Win pgo clients

wilybrace:
x509.SystemCertPool has an exclusion for Windows environments
that results in an error being returned for those.
https://golang.org/src/crypto/x509/cert_pool.go?s=1351:1391#L45

This change makes the default behavior more pleasant than
immediately erroring out on Windows clients.

 
 - [ ] Have you tested your changes on all related environments with successful results, as applicable? (Don't have access to a Win terminal)

 
### DIFF
--- a/hugo/content/Configuration/tls.md
+++ b/hugo/content/Configuration/tls.md
@@ -117,7 +117,8 @@ setip()
 By default, the pgo client will trust certificates issued by one of the
 Certificate Authorities listed in the operating system's default CA trust
 store. To prevent their inclusion, two configuation methods are available:
-either the Bash environment variable
+either the Bash environment variable. Windows clients have no default CA
+trust and thus have this setting configured to `true` by default.
 
 ```bash
 EXCLUDE_OS_TRUST=true
--- a/hugo/content/Configuration/tls.md
+++ b/hugo/content/Configuration/tls.md
@@ -116,15 +116,13 @@ setip()
 
 By default, the pgo client will trust certificates issued by one of the
 Certificate Authorities listed in the operating system's default CA trust
-store. To prevent their inclusion, two configuation methods are available:
-either the Bash environment variable. Windows clients have no default CA
-trust and thus have this setting configured to `true` by default.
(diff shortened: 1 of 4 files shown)


pgo/cmd/root.go
```diff
@@ -54,13 +55,17 @@ func init() {
 	GREEN = color.New(color.FgGreen).SprintFunc()
 	RED = color.New(color.FgRed).SprintFunc()
 
+	// Go currently guarantees an error when attempting to load OS_TRUST for
+	// windows-based systems (see crypto/x509/cert_pool.go)
```
cbandy: I'd like a link to `https://golang.org/issue/16736` rather than the `crypto` files here.

hugo/content/Configuration/tls.md
```diff
@@ -117,7 +117,8 @@ setip()
 By default, the pgo client will trust certificates issued by one of the
 Certificate Authorities listed in the operating system's default CA trust
 store. To prevent their inclusion, two configuation methods are available:
-either the Bash environment variable
+either the Bash environment variable. Windows clients have no default CA
+trust and thus have this setting configured to `true` by default.
```
cbandy: This is an awkward placement for this sentence. `either the Bash environment variable` is the beginning of a list that continues `or the --exclude-os-trust can be included during client use`.

My inclination is to get vague, drop this sentence, and add `if any` to the first sentence in this paragraph:

> By default, the pgo client will trust certificates issued by one of the Certificate Authorities listed in the operating system's default CA trust store, if any.

pgo/cmd/root.go
wilybrace: Agreed, updated

hugo/content/Configuration/tls.md
wilybrace: 1) Yes, that placement was terrible.
2) My reservation on being too vague was in response to needing to explicitly disable this in the 4.2.0 release for windows clients and no longer needing it going forward. That said, I agree with your suggestions and have updated.

PR #1212, opened 2020-01-15
Add support for `RELATED_IMAGE_*` specified containers

jkatz:
This comes a specification for OLM that allows one to inject information
about container images into the Operator based on the image repository
and sources they are using.

While the PostgreSQL Operator has supported this for awhile vis-a-vis
the `*_IMAGE_PREFIX` and `*_TAG` variables, this ensures compatibility
with OLM based deployments.

See: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/contributors/design-proposals/related-images.md


operator/common.go
```diff
@@ -35,6 +37,11 @@ var EventTCPAddress = "localhost:4150"
 
 var Pgo config.PgoConfig
 
+// ContainerImageOverrides contains a list of container images that are
+// overriden by the RELATED_IMAGE_* environmental variables that can be set by
+// people deploaying the Operator
```
andrewlecuyer: Just a typo here, should be `deploying`

config/images.go
```diff
@@ -0,0 +1,68 @@
...
+	"RELATED_IMAGE_CRUNCHY_COLLECT":          CONTAINER_IMAGE_CRUNCHY_COLLECT,
+	"RELATED_IMAGE_CRUNCHY_GRAFANA":          CONTAINER_IMAGE_CRUNCHY_GRAFANA,
+	"RELATED_IMAGE_CRUNCHY_PGBADGER":         CONTAINER_IMAGE_CRUNCHY_PGBADGER,
+	"RELATED_IMAGE_CRUNCHY_PGBENCH":          CONTAINER_IMAGE_CRUNCHY_PGBENCH,
+	"RELATED_IMAGE_CRUNCHY_PGBOUNCER":        CONTAINER_IMAGE_CRUNCHY_PGBOUNCER,
+	"RELATED_IMAGE_CRUNCHY_PGDUMP":           CONTAINER_IMAGE_CRUNCHY_PGDUMP,
+	"RELATED_IMAGE_CRUNCHY_PGRESTORE":        CONTAINER_IMAGE_CRUNCHY_PGRESTORE,
+	"RELATED_IMAGE_CRUNCHY_POSTGRES_HA":      CONTAINER_IMAGE_CRUNCHY_POSTGRES_HA,
+	"RELATED_IMAGE_CRUNCHY_POSTGRES_GIS_HA":  CONTAINER_IMAGE_CRUNCHY_POSTGRES_GIS_HA,
+	"RELATED_IMAGE_CRUNCHY_PROMETHEUS":       CONTAINER_IMAGE_CRUNCHY_PROMETHEUS,
```
andrewlecuyer: Just a note that the Grafana and Prometheus images are deployed via Ansible, and therefore will not be utilized directly by the Operator.

operator/backrest/restore.go
```diff
@@ -377,6 +381,10 @@ func CreateRestoredDeployment(restclient *rest.RESTClient, cluster *crv1.Pgclust
 		log.Error("error unmarshalling primary json into Deployment " + err.Error())
 		return err
 	}
+
+	// determine if any of the container images need to be overriden
```
andrewlecuyer: Typo, should be `overridden` (this is in various other places as well)

operator/cluster/upgrade.go
```diff
@@ -334,6 +334,15 @@ func createImageNamePatch(cluster crv1.Pgcluster, ccpImagePrefix, ccpImageTag st
 		Name:  "database",
 		Image: ccpImagePrefix + "/" + cluster.Spec.CCPImage + ":" + ccpImageTag,
 	}
+
```
andrewlecuyer: Wondering if we should add a note in our docs or OLM letting OLM users letting them know that the image tag they specify via `pgo upgrade` (which requires a new tag to be specified) will be overridden by whatever is specified in the new `RELATED_IMAGE_` env vars defined in the Operator deployment.

jkatz:
@andrewlecuyer Thanks for the review, feedback applied.

operator/common.go
```diff
@@ -35,6 +37,11 @@ var EventTCPAddress = "localhost:4150"
 
 var Pgo config.PgoConfig
 
+// ContainerImageOverrides contains a list of container images that are
+// overridden by the RELATED_IMAGE_* environmental variables that can be set by
+// people deploying the Operator
+var ContainerImageOverrides map[string]string
```
cbandy: It looks to me like every reference to this is in this package; other packages use `SetContainerImageOverride()`. Should this variable be unexported?

jkatz: Nope: https://github.com/CrunchyData/postgres-operator/pull/1212/files#diff-5720bacd93133eaff22b85a98af3c162R358

PR #1237, opened 2020-01-21
Updates add-targeted-namespace script

jmckulk:
- Removes dependency on expenv by using sed to replace environment variables
- If a namespace exists in the environment we need to confirm that it has
the correct labels. This is done by blindly adding the labels to the existing
namespace. These commands are moved inside of the if statement so that they
are only called if the namespace exists. If the namespace doesn't exist it
is created with the labels so adding them again adds unnecessary errors.
Moving the lines also makes the script and the purpose of these commands
more clear.

**Checklist:**

 
 - [x] Have you tested your changes on all related environments with successful results, as applicable? tested both scripts using either the ansible installer or `make setupnamespaces` target

**Type of Changes:**

 
**Other information**:
[ch5232]

### DIFF
--- a/ansible/roles/pgo-operator/templates/add-targeted-namespace.sh.j2
+++ b/ansible/roles/pgo-operator/templates/add-targeted-namespace.sh.j2
@@ -29,15 +29,16 @@ PG_SA="pgo-pg"
 # create the namespace if necessary
 {{ kubectl_or_oc }} get ns {{ item }}  > /dev/null
 if [ $? -eq 0 ]; then
-	echo "namespace" {{ item }} "already exists, adding labels"
-	# set the labels so that existing namespace is owned by this installation
-	{{ kubectl_or_oc }} label namespace/{{ item }} pgo-created-by=add-script
-	{{ kubectl_or_oc }} label namespace/{{ item }} vendor=crunchydata
-	{{ kubectl_or_oc }} label namespace/{{ item }} pgo-installation-name={{ pgo_installation_name }}
+	echo "namespace" {{ item }} "already exists"
 else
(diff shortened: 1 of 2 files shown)


ansible/roles/pgo-operator/templates/add-targeted-namespace.sh.j2
```diff
@@ -29,17 +29,16 @@ PG_SA="pgo-pg"
...
 if [ $? -eq 0 ]; then
-	echo "namespace" {{ item }} "already exists"
+	echo "namespace" {{ item }} "already exists, adding labels"
+	# set the labels so that existing namespace is owned by this installation
+	{{ kubectl_or_oc }} label namespace/{{ item }} pgo-created-by=add-script
+	{{ kubectl_or_oc }} label namespace/{{ item }} vendor=crunchydata
+	{{ kubectl_or_oc }} label namespace/{{ item }} pgo-installation-name={{ pgo_installation_name }}
 else
 	echo "namespace" {{ item }} "is new"
 	{{ kubectl_or_oc }} create ns {{ item }}
```
cbandy: It looks like this command will create a namespace without any labels.

jmckulk: Yeah it does look like that. The bash script uses a `target-namespace.yaml` that includes the labels and i assumed they did the same thing. Would it be better to update the ansible script to use a yaml file or just revert this change for ansible?

jmckulk: Reverted this file change. Ansible doesn't suffer from the same false positive error that bash does so changing the ansible version seems unnecessary.

Issue #1236, opened 2020-01-21
Documentation for pgo update cluster inconsistent


GeorgFleig:
**Describe the bug**

The `pgo update cluster` command and the documentation contain references to the `--autofail` flag while this does not exist anymore.

- https://access.crunchydata.com/documentation/postgres-operator/4.2.0/pgo-client/common-tasks/
- https://access.crunchydata.com/documentation/postgres-operator/4.2.0/architecture/disaster-recovery/
- https://access.crunchydata.com/documentation/postgres-operator/4.2.0/pgo-client/reference/pgo_update/
- https://access.crunchydata.com/documentation/postgres-operator/4.2.0/pgo-client/reference/pgo_update_cluster/

```
pgo update cluster --help
Update a PostgreSQL cluster. For example:

    pgo update cluster mycluster --autofail=false
    pgo update cluster mycluster myothercluster --autofail=false
    pgo update cluster --selector=name=mycluster --autofail=false
    pgo update cluster --all --autofail=true

Usage:
  pgo update cluster [flags]

Flags:
      --all                all resources.
      --disable-autofail   Disables autofail capabitilies in the cluster.
  -h, --help               help for cluster
      --no-prompt          No command line confirmation.
  -s, --selector string    The selector to use for cluster filtering.
```

I stumbled upon this when trying to enable autofail again after a restore. Can `--disable-autofail=false` be used as a replacement for `--autofail=true`?

**Please tell us about your environment:**

* Deployment method: ansible
* Operating System: Ubuntu 18.04
* Where is this running: Private cloud
* Storage being used: Ceph
* Container Image Tag: 4.2.0
* PostgreSQL Version: 11.6
* Platform (Docker, Kubernetes, OpenShift): k8s 1.15the problem here.

GeorgFleig:
Thanks @jkatz for providing the `--enable-autofail` flag, this clears things up.

The main reason for this ticket remains however. All the traces of the deprecated `--autofail` flag are still present in the documentation:


- https://access.crunchydata.com/documentation/postgres-operator/4.2.2/pgo-client/common-tasks/
- https://access.crunchydata.com/documentation/postgres-operator/4.2.2/architecture/disaster-recovery/
- https://access.crunchydata.com/documentation/postgres-operator/4.2.2/pgo-client/reference/pgo_update/
- https://access.crunchydata.com/documentation/postgres-operator/4.2.2/pgo-client/reference/pgo_update_cluster/

PR #1254, opened 2020-01-29
Add an `--enable-autofail` flag to `pgo update cluster`

jkatz:
**Checklist:**

 
**Type of Changes:**

 
While `--disable-autofail=false` would work in enabling autofail,
it was generally confusing to users. Having the `--enable-autofail`
flag makes it clear that one can return a PostreSQL cluster to using
HA.

This also introduces an API change to the update cluster ressource,
as the "Autofail" request parameter is turned into an integer. This
avoids having to deal with a "3-value boolean", which could get
messy.

Issue: [ch7023]
Issue: #1236


apiservermsgs/clustermsgs.go
```diff
@@ -182,6 +182,13 @@ type DeleteClusterResponse struct {
 	Status
 }
 
+// set the types for updating the Autofail status
+type UpdateClusterAutofailStatus int
+
+const UpdateClusterAutofailDoNothing UpdateClusterAutofailStatus = 0
+const UpdateClusterAutofailEnable UpdateClusterAutofailStatus = 1
+const UpdateClusterAutofailDisable UpdateClusterAutofailStatus = 2
```
abrightwell: Two things,

1. I would recommend:

```
const (
  UpdateClusterAutofailDoNothing UpdateClusteAutofailStatus = iota
  UpdateClusterAutofailEnable
  UpdateClusterAutofailDisable
)

https://golang.org/ref/spec#Iota

2. What's the difference between 'autofail do nothing' and 'autofail disable'?

File: pgo/cmd/update.go [hunk 2]
```diff
@@ -104,23 +107,31 @@ var UpdateClusterCmd = &cobra.Command{
...
-				fmt.Println("Aborting...")
-			}
+			os.Exit(1)
+		}
+
+		// if both --enable-autofail and --disable-autofail are true, then abort
+		if (EnableAutofailFlag && DisableAutofailFlag) || !(EnableAutofailFlag || DisableAutofailFlag) {
+			fmt.Println("Error: Cannot set --enable-autofail and --disable-autofail simultaneously")
+			os.Exit(1)
+		}
```
abrightwell: Does this mean that at least one of these flags is required? Seems like if that were the case, then wouldn't it be better to have a single flag that defaults to the opposite value? In my view, I'm not sure I'm just not sure it makes sense to have two separate flags to control the same thing. :thinking:

pgo/cmd/create.go
```diff
@@ -28,7 +28,7 @@ var AllNamespaces bool
 var ContainerResources string
 var ReplicaStorageConfig, StorageConfig string
 var CustomConfig string
-var ArchiveFlag, DisableAutofailFlag, PgbouncerFlag, MetricsFlag, BadgerFlag bool
+var ArchiveFlag, DisableAutofailFlag, EnableAutofailFlag, PgbouncerFlag, MetricsFlag, BadgerFlag bool
```
abrightwell: If you need to make this change here, then don't you also need to address it in `createCluster(...)` in `pgo/cmd/cluster.go`?

This would also bring me back to my original concern about two flags for the same thing on `create`. :thinking:

jkatz: Autofail is enabled by default on create. We only need to disable it there.

pgo/cmd/update.go
jkatz: Let me check, there is a chance I screwed that up.

apiservermsgs/clustermsgs.go
jkatz: That is magic ✨

pgo/cmd/create.go
jkatz: I just put it there for consistency where all the other flags were defined.

pgo/cmd/update.go
jkatz: Yup, I did screw that one up. Good catch.

apiservermsgs/clustermsgs.go
```diff
@@ -182,6 +182,16 @@ type DeleteClusterResponse struct {
 	Status
 }
 
+// set the types for updating the Autofail status
+type UpdateClusterAutofailStatus int
+
+// set the different values around updating the autofail configuration
+const (
+	UpdateClusterAutofailDoNothing UpdateClusterAutofailStatus = iota
```
abrightwell: You only need `iota` once.

jkatz: @abrightwell How do you initialize it then?

abrightwell:
```
const (
	UpdateClusterAutofailDoNothing UpdateClusterAutofailStatus = iota
	UpdateClusterAutofailEnable
	UpdateClusterAutofailDisable
)

jkatz:
> I guess my only concern is that the AutoFail field of the API interface (apiserver messages) for Update is a different type than for Create. It's not something I have a strong opinion on at this point, but something to consider, I think.

Let's a legitimate concern, but then it can't really be backpatched and would have to wait until 4.3. This can at least be in 4.2.2

apiservermsgs/clustermsgs.go
jkatz: I was testing it and thought it was that way, but got an error. The error turned out to be at a different line fo code

jkatz:
@abrightwell I'd consider updating the apiserver API in #1245 as that will be in 4.3, but not in this one.

PR #1257, opened 2020-01-30
Consider environment when configuring k8s clients

cbandy:
**Checklist:**

 
**Type of Changes:**

 
OLM scanners expect to be able to redirect Kubernetes API calls by setting environment variables. We use the most basic kube client configuration functions which don't consider env vars.

The higher-level configuration functions of k8s.io/client-go check `KUBECONFIG` env var and `~/.kube/config` directory and fall back to in-cluster config, if available.

**Other information**:

Issue: [ch7130]


kubeapi/client_config.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Match the settings applied by sigs.k8s.io/controller-runtime@v0.4.0;
+	// see https://github.com/kubernetes-sigs/controller-runtime/issues/365.
+	if config.QPS == 0.0 {
+		config.QPS = 20.0
+		config.Burst = 30.0
+	}
```
cbandy: I came across this logic in the referenced package, but I'm not _certain_ we should change them at this time. [The defaults](https://godoc.org/k8s.io/client-go/rest#Config) are `QPS: 5, Burst: 10`. Thoughts?

cbandy: I came across this logic in the referenced package, but I'm not _certain_ we should change them at this time. [The defaults](https://godoc.org/k8s.io/client-go/rest#Config) are `QPS: 5, Burst: 10`. Thoughts?

jkatz: @cbandy My preference is if it ain't broke, don't fix it.

jkatz: Which is another way of saying that if this fixes something that is broke, sure...

jkatz: Let's leave it.

kubeapi/client_config.go
```diff
@@ -0,0 +1,64 @@
...
+	loader.ExplicitPath = kubeconfig
+
+	// The deferred loader tries an in-cluster config if the default loading
+	// rules produce no results.
+	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		loader, &clientcmd.ConfigOverrides{},
+	).ClientConfig()
+}
+
+func NewClientConsideringFlag(kubeconfig string) (*rest.Config, *kubernetes.Clientset, error) {
```
jkatz: Pleaes add documentation on this function

kubeapi/client_config.go
```diff
@@ -0,0 +1,64 @@
...
+	config, err := loadClientConfig(kubeconfig)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	return config, clientset, err
+}
+
+func NewControllerClientConsideringFlag(kubeconfig string) (*rest.Config, *kubernetes.Clientset, error) {
```
jkatz: Please add documentation on this function

PR #1264, opened 2020-02-03
Updates backrest ssh key to use ed25519

jmckulk:
- Removes ssh key generation from bash and ansible install
- Updates vendor packages
- Updates ssh key generation to ed25519
- Updates backrest scripts for ed25519
- Updates operator pgo-backrest config files

Keys are currently generated in the operator code and in both the bash and
ansible installers. The ansible installer gerenates not only the id_rsa key but
id_ecdsa and id_25519 keys as well. The bash installer only generates the id_rsa
key. The backrest-repo-config secret is created using the ssh key gereration in
the golang code instead of the keys generated by the installers. This commit
removes the key generation from the installers.

The backrest-repo-config secret is updated to used an ed25519 ssh key instead of
rsa. The fucntion to gereate a public/private ssh keypair is updated to generate
the new key. This code change required an update to the vendor packaging for the
crypto package. This package had not been updated in the `GOPKG.lock` file since
since Jun 14, 2018 and needed newer functionality to generate the key. The new
keys are used in place of the original rsa key in the relevant places. The files
that require the ssh key are updated to use the new file names based on ed25519.
This includes the sshd_config and ssh config.

**Checklist:**

 
**Type of Changes:**

 
**Other information**:
Related crunchy-containers pr [#1191](https://github.com/CrunchyData/crunchy-containers/pull/1191)


sshutil/sshutil.go
```diff
@@ -1,59 +1,113 @@
...
 
 import (
-	"bytes"
-	"crypto/rand"
-	"crypto/rsa"
-	"crypto/x509"
+	"math/rand"
+	"crypto/ed25519"
 	"encoding/pem"
+	"bytes"
```
jkatz: I would suggest keeping these in alphabetical order.

sshutil/sshutil.go
```diff
@@ -1,59 +1,113 @@
...
-		return keys, err
+		return SSHKey{}, err
 	}
 
 	return keys, nil
 }
 
-func NewPrivateKey(bits int) (*rsa.PrivateKey, error) {
-	pk, err := rsa.GenerateKey(rand.Reader, bits)
+func newPublicKey(key ed25519.PublicKey) ([]byte, error) {
```
jkatz: Please document what this function does.

sshutil/sshutil.go
```diff
@@ -1,59 +1,113 @@
...
-		return pk, err
+		return nil, err
 	}
+	return ssh.MarshalAuthorizedKey(pubKey), nil
+}
 
-	err = pk.Validate()
-	if err != nil {
-		return pk, err
+func newPrivateKey(key ed25519.PrivateKey) ([]byte, error) {
```
jkatz: Please document what this function does

sshutil/sshutil.go
```diff
@@ -1,59 +1,113 @@
...
+	// https://anongit.mindrot.org/openssh.git/tree/PROTOCOL.key?h=V_8_1_P1
```
jkatz: What's the purpose of this comment? Why are we linking to this? Do we need to link to this?

Please provide more context around this code.

jmckulk: This link was in the piece of code that chris shared to generate the keys. I found it helpful in understanding why the scripts were setup that way so i left it in. It gives some context to why the structs are setup the way there are. I can be removed but if that context is needed it would need to be added as other comments

sshutil/sshutil.go
```diff
@@ -1,59 +1,113 @@
...
 )
 
+// SSHKey stores byte slices that represent private and public ssh keys
 type SSHKey struct {
 	Private []byte
 	Public  []byte
 }
 
+// NewPrivatePublicKeyPair generates a an ed25519 ssh private and public key
 func NewPrivatePublicKeyPair(bits int) (SSHKey, error) {
```
jmckulk: This `bits` value is no longer used so it should be removed

jkatz: Even better 😄 !

sshutil/sshutil.go
jkatz: Great. Please explain that in the code in terms of what it explains, and then use the URL as your reference 😉

PR #1265, opened 2020-02-04
Ensure Successful Cluster Creation when Restoring After Scale Down

andrewlecuyer:
**Checklist:**

 
**Type of Changes:**

 
If a cluster is scaled down prior to a restore (and without then taking a backup), then the initial cluster backup taken immediately following the restore will fail, and cluster replicas will not be bootstrapped.

Additionally, following the successful completion of any `rmdata` job, all `pgtasks` for the cluster are deleted.

[ch6991]

A restored cluster will now properly bootstrap and deploy regardless of whether or not it has been scaled down immediately prior to the restore.

**Other information**:

N/A

### DIFF
--- a/controller/jobcontroller.go
+++ b/controller/jobcontroller.go
@@ -562,15 +562,7 @@ func handleRmdata(job *apiv1.Job, restClient *rest.RESTClient, clientset *kubern
 	labels := job.GetObjectMeta().GetLabels()
 	clusterName := labels[config.LABEL_PG_CLUSTER]
 
-	//delete any pgtask for this cluster
-	log.Debugf("deleting pgtask for rmdata job name is %s", job.ObjectMeta.Name)
-	err = kubeapi.Deletepgtasks(restClient, config.LABEL_PG_CLUSTER+"="+clusterName, namespace)
-	if err != nil {
-		return err
-	}
-
-	err = kubeapi.DeleteJob(clientset, job.Name, namespace)
-	if err != nil {
+	if err = kubeapi.DeleteJob(clientset, job.Name, namespace); err != nil {
 		log.Error(err)
 	}
 
--- a/operator/backrest/backup.go
+++ b/operator/backrest/backup.go
@@ -225,22 +225,27 @@ func CreateBackup(restclient *rest.RESTClient, namespace, clusterName, podName s
 func CleanBackupResources(restclient *rest.RESTClient, clientset *kubernetes.Clientset, namespace,
 	clusterName string) error {
 
-	pgtask := crv1.Pgtask{}
 	taskName := "backrest-backup-" + clusterName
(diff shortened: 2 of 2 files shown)


operator/backrest/backup.go
```diff
@@ -225,22 +225,27 @@ func CreateBackup(restclient *rest.RESTClient, namespace, clusterName, podName s
...
+		log.Debugf("pgtask %s was found when cleaning backup resources prior to creating a new "+
+			"backrest backup pgtask and will be deleted", taskName)
+		// delete the existing pgBackRest backup pgtask
+		err = kubeapi.Deletepgtask(restclient, taskName, namespace)
+		if err != nil {
+			return err
+		}
+	} else {
+		log.Debugf("pgtask %s was not found when cleaning backup resources prior to creating a "+
+			"new backrest backup pgtask", taskName)
```
abrightwell: Just a suggestion.

Refactoring these debug statements to a single statement prior to the `if` would make this a bit cleaner. Perhaps something like:

```
log.Debugf("pgtask: '%s' found: %t:  cleaning up backup resources", taskName, found)
if found {
  log.Debugf("deleting pgtask: '%s', taskName)
  ...
}
```

Or perhaps something similar?

The `else` for the sole purpose of a debug statement just seems unnecessary, IMV.

jkatz: Or!

```go
if !found {
    log.Debugf("pgtask %s was not found when cleaning backup resources prior to creating a "+
			"new backrest backup pgtask", taskName)
    return
}

// ...
```

Less nested if statements 😄

andrewlecuyer: Recommendations have been incorporated

controller/jobcontroller.go
```diff
@@ -562,15 +562,7 @@ func handleRmdata(job *apiv1.Job, restClient *rest.RESTClient, clientset *kubern
...
-	//delete any pgtask for this cluster
-	log.Debugf("deleting pgtask for rmdata job name is %s", job.ObjectMeta.Name)
-	err = kubeapi.Deletepgtasks(restClient, config.LABEL_PG_CLUSTER+"="+clusterName, namespace)
-	if err != nil {
-		return err
-	}
-
-	err = kubeapi.DeleteJob(clientset, job.Name, namespace)
-	if err != nil {
+	if err = kubeapi.DeleteJob(clientset, job.Name, namespace); err != nil {
```
abrightwell: Unless you care about the error outside of this scope, do `:=` to assign it here. Which would allow you to get rid of declaring it above.

jkatz: Agree with the comment, looking through the code, this would take further work and testing given the meaning of `err` appears to be overloaded further down in this block. Likely a change/refactor we need to come back to later.

PR #1286, opened 2020-02-13
Updates for HA Container bin Reorg & Conf Changes and Removal of PG Settings from pgo.yaml

andrewlecuyer:
This Pull Request includes changes to update the PostgreSQL Operator according to the the `bin` directory reorganization recently implemented in the `crunchy-postgres-ha` container, as well as the inclusion of `archive_timeout` as a default PG setting with the same container (these changes can be seen in PR https://github.com/CrunchyData/crunchy-containers/pull/1200)  

Additionally, as a result of the changes above, the `postgres-ha.yaml` file previously created by the Operator for all PG clusters (unless a custom file was provided via custom config) is no longer needed, and therefore has been removed.

**Checklist:**

 
**Type of Changes:**

 
- The liveness and readniness probes for the PG cluster reference directory `/opt/cpm/bin` within the `crunchy-postgres-ha` container, and not the new `/opt/cpm/bin/health` directory
- A default `postgres-ha.yaml` file is created by the Operator for every PG cluster in order to provide `archive_timeout`, `log_statement` and `log_min_duration_statement` settings
- `ArchiveTimeout`, `LogStatement` and `LogMinDurationStatement` settings can be defined in the `pgo.yaml` configuration file in order to set the `archive_timeout`, `log_statement` and `log_min_duration_statement` settings in the `postgresql.conf` file

[ch7040]

- The liveness and readniness probes for the PG cluster now reference the new `/opt/cpm/bin/health` directory within the `crunchy-postgres-ha` container
- A default `postgres-ha.yaml` file is not created by the Operator for every PG cluster
- Custom configuration (e.g. a custom `postgresql.conf` and/or `postgres-ha.yaml` file) must be utilized to set the `archive_timeout`, `log_statement` and `log_min_duration_statement` settings in the `postgresql.conf` file (i.e. these settings can no longer be configured via the `pgo.yaml` file)

**Other information**:

N/A


controller/podcontroller.go
```diff
@@ -221,7 +221,12 @@ func (c *PodController) checkReadyStatus(oldpod, newpod *apiv1.Pod, cluster *crv
 							cluster.ObjectMeta.Namespace)
 					}
 
-					operator.UpdatePghaDefaultConfigInitFlag(c.PodClientset, false, clusterName, newpod.ObjectMeta.Namespace)
+					err = operator.UpdatePGHAConfigInitFlag(c.PodClientset, false, clusterName,
+						newpod.ObjectMeta.Namespace)
+					if err != nil {
+						log.Error(err)
+						return
+					}
```
abrightwell: Suggest condensing this to `if err := ... ; err != nI { ... }`

operator/cluster/clusterlogic.go
```diff
@@ -151,30 +149,11 @@ func AddCluster(clientset *kubernetes.Clientset, client *rest.RESTClient, cl *cr
...
-			log.Debugf("Custom postgres-ha config file found in custom configMap, " +
-				"creating default configMap without default postgres-ha config file")
-			operator.AddDefaultPostgresHaConfigMap(clientset, cl, deploymentFields.IsInit, false, namespace)
-		}
-	}
+	// Create a configMap for the cluster that will be utilized to configure whether or not
+	// initialization logic should be executed when the postgres-ha container is run.  This
+	// ensures that the original primary in a PG cluster does not attempt to run any initialization
+	// logic following a restart of the container.
+	operator.CreatePGHAConfigMap(clientset, cl, namespace)
```
abrightwell: Do you want to handle the error that could be returned by this call?

andrewlecuyer: Thanks, definitely wanted to handle any errors.  Code has been updated.

controller/podcontroller.go
andrewlecuyer: I tend to keep them separate (as seen here) when the expression for calling the function is long, e.g. do to a long function name, lots of params, etc., so this was actually a conscious decision here.  IMO it just feels cleaner/easier to read in certain situations.

Definitely some grey area with these things though and I'm not too passionate one way or another, i.e. if we want to establish that the initialization statement approach should _always_ be utilized.

operator/cluster/clusterlogic.go
andrewlecuyer: Thanks, definitely wanted to handle any errors.  Code has been updated.

controller/podcontroller.go
jkatz: IMV even if it's long, it does control scope,and when it's too long, I have the `if` statement take up two lines :smile:

abrightwell: Yeah, I'm not necessarily a proponent of _always_ (I don't think), as I'm sure there are times that it doesn't make sense. But, FWIW, when it's long I might do something like:

```
if err := someOperation(
  with,
  a,
  long,
  parameter,
  list,
); err != nil {
  // handle error
}
```

The benefit, IMO, is that:

1. it's easier to determine and document what's being passed to the function. For instance: what does `false` mean or set? Or what if `nil` or `""` is passed to a function? Sure, there are docs for that (right? :thinking: :wink:) but, bringing that context into the documentation/comment for such a call, helps for long term/overall maintenance, IMO.
2. It's also easier to modify the parameter list if it changes (for some reason) as well as identify what changed in a `diff`.

Granted, I'm not so opinionated in this case as to insist one way or the other. It's just my preference. :-)

PR #1331, opened 2020-03-10
Allow pod anti-affinity rules to be overriden for pgBackRest & pgBouncer

jkatz:
**Checklist:**

 
**Type of Changes:**

 
v4.2.0 introduced the ability to set pod anti-affinity rules for pods
deployed as part of a PostgreSQL cluster, which include the pgBackRest
repository and the optional pgBouncer deployment. However, there may be
cases where a user may not want these rules to be universally applied,
e.g. where PostgreSQL instances should be "required" to be on different
nodes, but pgBackRest only has "preferred" anti-affinity.

This patch introduces the ability to set the pod anti-affinity rules
specifically for pgBackRest and pgbouncer as well. This is exposed to
the end user with the `--pod-anti-affinity-pgbackrest` and
`--pod-anti-affinity-pgbouncer` on `pgo create cluster`, which set the
anti-affinity rules for pgBackRest and pgBouncer respectively.

This can also be set as part of the "pgo.yaml" configuration file, using
the `PodAntiAffinityPgBackRest` and `PodAntiAffinityPgBouncer` keys for
pgBackRest and pgBouncer respectively.

This commit also creates a change in the behavior of the "required"
pod anti-affinity setting. Deployments that leverage the "required" mode
of anti-affinity set their pods to look for other pods with the
"required" label in their group. This is to ensure that pods that have
the lesser "preferred" label don't necessarily cause a "required" pod
to fail to be scheduled to a node.

**Other information**:

Issue: [ch7136]


operator/clusterutilities.go
```diff
@@ -580,6 +574,56 @@ func GetPodAntiAffinity(podAntiAffinityType string, clusterName string) string {
...
+// based on the deployment type (cluster, pgBackRest, pgBouncer), the value
+// in the cluster spec, and the defaults available in pgo.yaml.
+//
+// In other words, the pod anti-affinity is determined by this heuristic, in
+// priority order:
+//
+// 1. If it's pgBackRest/pgBouncer the value set by the user (available in the
+//    cluster spec)
+// 2. If it's pgBackRest/pgBouncer the value set in pgo.yaml
+// 3. The value set in "Default" in the cluster spec)
```
andrewlecuyer: Erroneous closed parenthesis

docs/content/Installation/install-with-ansible/prerequisites.md
```diff
@@ -212,9 +212,12 @@ sets of variables cannot be used at the same time.
...
-| `pgo_client_container_install` | false | | Installs the pgo-client deployment along with ansible isnstall |
+| `pgo_client_container_install` | false | | Installs the pgo-client deployment along with ansible install |
 | `pgo_apiserver_url` | `https://postgres-operator` | | Sets the `pgo_apiserver_url` in the pgo-client deployment |
 | `pgo_client_cert_secret` | `pgo.tls` | | Sets the secret that the pgo-client will use when connecting to the operator. Secret should hold the TLS certs |
+| `pod_anti_affinity`               | preferred   |           | Sets the default pod anti-affinity for the deployed PostgreSQL clusters, which is applied to the PostgreSQL instances, the pgBackRest repository, and any pgBouncer instances |
+| `pod_anti_affinity_pgbackrest`    |             |           | If set, overrides the value of `pod_anti_affinity` for just the pgBackRest repository and sets the anti-affinity for the pgBackRest repository compared to its pod anti-affinity group |
```
andrewlecuyer: Wording here is a bit confusing, specifically where is says:
`...sets the anti-affinity for the pgBackRest repository compared to its pod anti-affinity group`.

Maybe just leaving the first part is sufficient, i.e. just including the following without any implementation details:
`If set, overrides the value of `pod_anti_affinity` for just the pgBackRest repository.`

(we could do the same for `pod_anti_affinity_pgbouncer` below as well)

PR #1335, opened 2020-03-11
Installer image

jmckulk:
**Checklist:**
 
**Type of Changes:**

 
Currently install is managed with multiple methods (bash, ansible, OLM, etc.)

This change packages the ansible installer into a kubernetes job that runs in a cluster. This simplifies the install methods by providing an image that does most of the install work. This job simplifies the installer management by setting up a few steps that will need to be created for a specific installer. A set of ansible roles is provided as a base way to setup the installer namespace and run the job. 

**Other information**:
[ch6969]


installers/pgo-installer/ansible/inventory
```diff
@@ -0,0 +1,307 @@
...
+#   kubectl config current-context
+# ==================
+# kubernetes_context=''
+
+# Deploy with PGO-Installer
+# ==================
+# Note: Enable kubernetes_in_cluster if you are using
+# the pgo-installer image to install the operator
+# ==================
+kubernetes_in_cluster=true
```
jmckulk: need to pull inventory changes out

installers/pgo-installer/ansible/roles/pgo-installer/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+pgo_installer_namespace: "pgo-install"
+pgo_install_sa: "pgo-installer-sa"
+pgo_install_crb: "pgo-installer-crb"
+job_timeout: -1s
+installer_img_pull_policy: "Always"
```
jmckulk: needs to be set to `IfNotPresent`

installers/pgo-installer/ansible/roles/pgo-installer/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+pgo_installer_namespace: "pgo-install"
+pgo_install_sa: "pgo-installer-sa"
+pgo_install_crb: "pgo-installer-crb"
+job_timeout: -1s
```
jmckulk: need to change to a reasonable timeout for the install job to complete. I'm thinking 120s-180s but feedback welcomed

installers/pgo-installer/ansible/roles/pgo-installer/tasks/main.yml
```diff
@@ -0,0 +1,87 @@
...
+  tags: 
+    - always
+
+- name: Setup Install Container
+  tags:
+    - always
+  block:
+  - name: Create Install Namespace
+    shell: "kubectl create namespace {{ pgo_installer_namespace }}"
+    ignore_errors: yes
```
jmckulk: Not sure if the ignore errors are needed here. How the ansible roles are currently setup it shouldn't ever exist unless the job is being run. After every run the ns and clusterrolebinding are cleaned up

installers/pgo-installer/ansible/roles/pgo-installer/tasks/main.yml
```diff
@@ -0,0 +1,87 @@
...
+        src: pgo-update-job.json.j2
+        dest: "{{ output_dir }}/pgo-update-job.json"
+        mode: '0600'
+    - name: Create Update Job
+      shell: "kubectl create -n {{ pgo_installer_namespace }} -f {{ output_dir }}/pgo-update-job.json"
+      register: job_started
+    - name: Wait for Update Job
+      when: job_started != 0
+      shell: "kubectl wait -n {{ pgo_installer_namespace }} --timeout={{ job_timeout }} --for=condition=Complete jobs.batch/pgo-updater"
+  always:
```
jmckulk: need to move debug statement to a `rescue` block

installers/pgo-installer/ansible/roles/pgo-installer/tasks/main.yml
```diff
@@ -0,0 +1,87 @@
...
+        dest: "{{ output_dir }}/pgo-uninstaller-job.json"
+        mode: '0600'
+    - name: Create Uninstall Job
+      shell: "kubectl create -n {{ pgo_installer_namespace }} -f {{ output_dir }}/pgo-uninstaller-job.json"
+      register: job_started
+    - name: Wait for Uninstall Job
+      when: job_started != 0
+      shell: "kubectl wait -n {{ pgo_installer_namespace }} --timeout={{ job_timeout }} --for=condition=Complete jobs.batch/pgo-uninstaller"
+  always:
+    - debug:
```
jmckulk: move to rescue block

ansible/roles/pgo-operator/vars/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
 pgo_client_url: "https://github.com/CrunchyData/postgres-operator/releases/download/{{ pgo_client_version }}"
-pgo_keys_dir: "~/.pgo/{{ pgo_operator_namespace }}"
+pgo_keys_dir: "{{ ansible_env.HOME}}/.pgo/{{ pgo_operator_namespace }}"
```
abrightwell: spacing

installers/pgo-installer/bin/pgo-install.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+ansible-playbook -i /ansible-inventory/inventory --tags=$1 /ansible/main.yml
```
abrightwell: Do we want to assume that `ansible-playbook` is on the `$PATH`?

installers/pgo-installer/bin/install.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+INSTALLNS="pgo-install"
+INSTALLDIR="$PGOROOT/installers/pgo-installer"
+
+kubectl create ns $INSTALLNS
```
abrightwell: Do  we want to assume that `kubectl` is on the `$PATH`?

jmckulk: I may end up removing this as well. I was using for initial testing before i added the roles.

installers/pgo-installer/bin/pgo-install.sh
jmckulk: Possibly not, but we are creating the image so I would imagine we know where it is.

I'm honestly not sold on keeping this script anyway. @abrightwell what are your thoughts on that? I could pass the whole ansible line into the job instead of using this.

centos7/Dockerfile.pgo-installer.centos7
```diff
@@ -0,0 +1,36 @@
...
+
+RUN yum -y install epel-release
+
+# which is used in the ansible roles to verify
+# used to install ssk-keyen and gen ssh keys
+RUN yum -y install \
+    --setopt=skip_missing_names_on_install=False \
+    kubectl \
+    ansible \ 
+    which
```
jkatz: Can these be combined into one `RUN` statement? I think potentially so can the other one too

centos7/Dockerfile.pgo-installer.centos7
```diff
@@ -0,0 +1,36 @@
...
+
+
+# Kubernetes repository
+RUN echo $'[kubernetes] \n\
+name=Kubernetes \n\
+baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64 \n\
+enabled=1 \n\
+gpgcheck=1 \n\
+repo_gpgcheck=1 \n\
+gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg' > /etc/yum.repos.d/kubernetes.repo
```
jkatz: I'd suggested

```shell
RUN cat >> /etc/yum.repos.d/kubernetes.repo <<-EOF
[kubernetes]
name=Kubernetes
baseurl=https://packages.cloud.google.com/yum/repos/kubernetes-el7-x86_64
enabled=1
gpgcheck=1
repo_gpgcheck=1
gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
EOF
```

if possible, though I presume that needs a bit of testing.

centos7/Dockerfile.pgo-installer.centos7
```diff
@@ -0,0 +1,36 @@
...
+    which
+
+
+USER daemon
+
+ADD ansible /ansible
+ADD installers/pgo-installer/bin/pgo-install.sh pgo-install.sh
+
+RUN chmod +x pgo-install.sh
+
```
jkatz: Remove extra whitespace

ansible/ansible.cfg
```diff
@@ -1,5 +1,6 @@
 [defaults]
 retry_files_enabled = False
+remote_tmp=/tmp
```
jkatz: If everything is moved under the `installers/pgo-installer` folder, should this go there?

installers/pgo-installer/bin/pgo-install.sh
jkatz: I mean this is encapsulated in the container?

It would not hurt to have a `/usr/bin/env ansible-playbook`

rhel7/Dockerfile.pgo-installer.rhel7
```diff
@@ -0,0 +1,36 @@
...
+# REVIEW IN PR **** confirm with build team ***
+RUN yum install -y https://download.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+
+# which is used in the ansible roles to verify
+# used to install ssk-keyen and gen ssh keys
+RUN yum -y install \
+    --setopt=skip_missing_names_on_install=False \
+    kubectl \
+    ansible \
+    which
```
jkatz: Similar comments to the CentOS file

ansible/ansible.cfg
jmckulk: This is required in the `$PGOROOT/ansible` roles to allow installing the operator from within a kubecluster. The new roles `$PGOROOT/installers/pgo-installers/ansible` setup and run the install job. The original roles still install the operator. Would renaming one of the ansible directories help make this clear?

rhel7/Dockerfile.pgo-installer.rhel7
jmckulk: missed this one - will fix

installers/pgo-installer/bin/install.sh
jmckulk: looking back - we were assuming that that `kubectl` and `oc` are on the `$PATH` for the current ansible install. If that needs to be changed it shouldn't be difficult

installers/pgo-installer/ansible/roles/pgo-preflight/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
+---
+- fail:
+    msg: "Please specify the a tag: uninstall, uninstall-metrics, install, install-metrics or update"
```
jmckulk: May need to pull out the metrics options and update the install/uninstall/upgrade tags

installers/pgo-installer/ansible/roles/pgo-preflight/tasks/main.yml
```diff
@@ -0,0 +1,53 @@
...
+  tags: always
+  when: ansible_run_tags[0] == "all"
+
+- name: Check Operating System
```
jmckulk: Is this needed since we will know what os we are creating the image

ansible/ansible.cfg
jkatz: I guess my question was more fundamental, i.e. should we move the `ansible` folder out of `$PGOROOT` into a different directory?

installers/pgo-installer/bin/install.sh
jkatz: Similarly, one option for `oc` is to first see if the `kubectl` commands work correctly on a OCP install.

They should be compatible: [https://docs.openshift.com/container-platform/3.10/cli_reference/differences_oc_kubectl.html(https://docs.openshift.com/container-platform/3.10/cli_reference/differences_oc_kubectl.html)

ansible/ansible.cfg
jmckulk: That would make sense to me. Possibly move `$PGOROOT/ansible` to `$PGOROOT/installers/ansible`

installers/pgo-installer/bin/install.sh
jmckulk: I tested the installer with only `kubectl` installed before I added `oc`. I didn't notice any issues with the install when using `kubectl` with openshift.

Looking at the differences shown in the link, most things shouldn't be necessary. The login feature won't be an issue because the pod will be using the `kubernetes_in_cluster` option and won't need to login. As long as we don't need access to the OpehShift resources (DeploymentConfigs, BuildConfigs, Routes, ImageStreams, and ImageStreamTags) I don't see an issue only using `kubectl`.

jkatz: The only one I could see us potentially needing are `ImageStreams`, but I'd advise waiting until that becomes a request. So for now let's ship `kubectl` and simplify the dependency management.

ansible/ansible.cfg
jkatz: :+1:

installers/pgo-installer/ansible/roles/pgo-preflight/tasks/main.yml
jkatz: Nope. Remove it.

installers/pgo-installer/bin/install.sh
jmckulk: I've removed the `oc` steps from the centos7 image. For rhel7/ubi7 i'm installing with a common package that installs both so i may just leave that unless there is a reason to switch back to adding the repo file for kubernetes.

rhel7/Dockerfile.pgo-installer.rhel7
jmckulk: @jkatz I tested adding the download into the longer yum install script and the build failed - Seems like those two installs need to be separate

installers/ansible/inventory
```diff
@@ -7,6 +7,10 @@ crunchy_debug='false'
 # ==================
 # Kubernetes and OpenShift Settings
 # Note: Either Kubernetes or OpenShift must be used
+# One of the three following blocks must be updated:
+# - Deploy into Openshift
+# - Deploy into Kubernetes
+# - Deploy with PGO-Installer
```
jkatz: `# - Deploy with the PostgreSQL Operator Installer (pgo-installer)`

and make this one the first one.

Provide a bit more explanation about the differences, i.e. the Installer is a container that runs within your Kubernetes/OpenShift environment, whereas the other ones allow you to run the installer from your local environment into the remote Kube/OpenShift ones

installers/ansible/inventory
```diff
@@ -26,6 +30,14 @@ crunchy_debug='false'
 # ==================
 # kubernetes_context=''
 
+# Deploy with PGO-Installer
+# ==================
+# Note: Enable kubernetes_in_cluster if you are using
+# the pgo-installer image to install the operator
+# ==================
+# kubernetes_in_cluster=false
+# use_oc=false
```
jkatz: Remove this attribute, but perhaps replace with

`install_into_openshift`

or some other name.

Or perhaps this is a different flag, like

`pgo_installer_environment="openshift"`

and the default is to presume Kubernetes

installers/ansible/inventory
```diff
@@ -26,6 +30,14 @@ crunchy_debug='false'
 # ==================
 # kubernetes_context=''
 
+# Deploy with PGO-Installer
+# ==================
+# Note: Enable kubernetes_in_cluster if you are using
+# the pgo-installer image to install the operator
+# ==================
+# kubernetes_in_cluster=false
```
jkatz: This name is a bit confusing. Why not

`use_pgo_installer`

installers/ansible/inventory
```diff
@@ -26,6 +30,14 @@ crunchy_debug='false'
 # ==================
 # kubernetes_context=''
 
+# Deploy with PGO-Installer
```
jkatz: ` # Deploy with PostgreSQL Operator Installer (pgo-installer)`

Let's also put this above the other two methods, try to encourage people to use this method

installers/pgo-installer/README.md
```diff
@@ -0,0 +1,5 @@
+# Ansible installer for pgo-installer image
+### Ansible command
+```
+ansible-playbook -i $PGOROOT/ansible/inventory --tags=install main.yml
+```
```
jkatz: This needs to be written for directions that actually pertain to running the `pgo-installer` container

installers/pgo-installer/ansible/roles/pgo-installer/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+pgo_installer_namespace: "pgo-install"
```
jkatz: We need to ensure the user can overwrite this. The user may not have control over what namespace they can deploy the installer into.

The most likely case is they are deploying the Operator into the same namespace they are running the installer.

installers/pgo-installer/ansible/roles/pgo-installer/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+    path: "{{ output_dir }}"
+    state: directory
+    mode: 0700
+  tags:
+    - always
+
+- include_vars: openshift.yml
+  when: use_oc | default(false) | bool
+  tags:
+    - always
```
jkatz: I think we need a different flag or the like, given we'll use `kubectl` instead of oc.

The suggestion I had in an earlier comment:

`pgo_installer_environment="openshift"` so we check here if `pgo_installer_environment` containers the value `openshift`

installers/pgo-installer/ansible/roles/pgo-installer/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+    shell: "{{ kubectl_or_oc }} create namespace {{ pgo_installer_namespace }}"
+    ignore_errors: yes
+      
+  - name: Create PGO-Install Service Account
+    shell: "{{ kubectl_or_oc }} create serviceaccount {{ pgo_install_sa }} -n {{ pgo_installer_namespace }}"
+    ignore_errors: yes
+
+  - name: Create PGO-Installer Cluster Role Binding
+    shell: "{{ kubectl_or_oc }} create clusterrolebinding {{ pgo_install_crb }} -n {{ pgo_installer_namespace }} --clusterrole=cluster-admin --serviceaccount={{ pgo_installer_namespace }}:{{ pgo_install_sa }}"
+    ignore_errors: yes
```
jkatz: I think we have to see if we can eliminate the `clusterrolebinding` for the install. That may prevent people from using it, or at least have a way to accomplish the non-multinamespace steps without a clusterrole binding.

I realize this may be a big ask for the first iteration, but wanted to mention it

installers/pgo-installer/ansible/roles/pgo-preflight/tasks/check_inventory.yml
```diff
@@ -0,0 +1,47 @@
...
+
+- assert:
+    msg: Cannot Install PGO Client when Installing in kubernetes 
+    that:
+      - not (pgo_client_install | default(true) | bool)
+    quiet: yes
+  when: kubernetes_in_cluster | default(false) | bool
+  tags:
+    - install-container
+    - update-container
```
jkatz: Why is this a restriction?

rhel7/Dockerfile.pgo-installer.rhel7
```diff
@@ -0,0 +1,24 @@
...
+RUN yum -y install \
+    --setopt=skip_missing_names_on_install=False \
+    --enablerepo='rhel-7-server-ose-3.11-rpms' \
+    atomic-openshift-clients \
+    ansible \
+    which
+
+USER daemon
+
+ADD ansible /ansible
```
jkatz: Same comment as in the centos7 one

cbandy:
It's somewhat unrelated, but ... with the move into the `installers` folder, can/should we ditch this copy of the logo? There are two in `installers` already that we should probably standardize on.

jkatz:
> It's somewhat unrelated, but ... with the move into the `installers` folder, can/should we ditch this copy of the logo? There are two in `installers` already that we should probably standardize on.

Hah, I was going to suggest that, but didn't know how to properly do so :smile: I agree.

I think for the installer container, we likely could remove it altogether.

installers/pgo-installer/ansible/roles/pgo-installer/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+
+  - name: Create Inventory Config Map
+    shell: "{{ kubectl_or_oc }} create configmap pgo-install-inventory -n {{ pgo_installer_namespace }} --from-file=inventory={{ inventory_dir }}/inventory"
```
cbandy: It seems like this is the point where the caller gets to decide how the operator is configured. Where does `inventory_dir` come from?

centos7/Dockerfile.pgo-installer.centos7
```diff
@@ -0,0 +1,24 @@
...
+RUN yum -y install \
+    --setopt=skip_missing_names_on_install=False \
+    epel-release \
+    kubectl \
+    ansible \ 
+    which
+
+USER daemon
+
+ADD ansible /ansible
```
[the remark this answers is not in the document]
jmckulk: yes - found this in testing

installers/ansible/inventory
jmckulk: This variable existed before my changes so i didn't name it. AIUI it disables some of the checks that are in place when installing with ansible locally.  It might make sense to have a `use_pgo_installer` that will go in and enable this option. Or we could rename it but that might be something to run by chris since he created it.

installers/pgo-installer/ansible/roles/pgo-installer/defaults/main.yml
jmckulk: That was on thing that i was thinking about. I can definitely add this to the inventory file so that they can specify.

@jkatz currently the installer deletes the install namespace after the install completes. That would not work if they are installing in the same namespace as the operator. Is this something that would need to work?

installers/pgo-installer/ansible/roles/pgo-installer/tasks/main.yml
jmckulk: This will be what they are using locally so we won't have control over it. Beyond telling users which to use. I'm not sure `pgo_installer_environemtn` would do a good enough job at letting the user know this is what is used locally.

installers/pgo-installer/ansible/roles/pgo-installer/tasks/main.yml
jmckulk: AIUI cluster-admin is required to install. I can look at how to get around it but that might be a design question that would require some discussion.

installers/pgo-installer/ansible/roles/pgo-preflight/tasks/check_inventory.yml
jmckulk: The ansible installer uses sudo to as part of the download/setup of the pgo client binary. Sudo doesn't exist in containers so the command and install job fails.

This would also be installing the client inside of the installer pod/job which in turn gets deleted.

rhel7/Dockerfile.pgo-installer.rhel7
jmckulk: found this in testing - will fix

centos7/Dockerfile.pgo-installer.centos7
```diff
@@ -0,0 +1,24 @@
...
+ADD installers/pgo-installer/bin/pgo-install.sh pgo-install.sh
+
+RUN chmod +x /usr/bin/oc
+RUN chmod +x pgo-install.sh
```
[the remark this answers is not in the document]
jmckulk: Ah I didn't realize - i can make that change

installers/pgo-installer/ansible/roles/pgo-installer/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
...
+pgo_install_sa: "pgo-installer-sa"
+pgo_install_crb: "pgo-installer-crb"
+job_wait_timeout: 180s
```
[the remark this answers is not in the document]
jmckulk: Can/will update to 5m but i didn't have any issues running the install in `3m`

installers/pgo-installer/ansible/roles/pgo-installer/tasks/main.yml
jmckulk: `(( inventory_dir ))` points to the directory that holds the inventory file that the user passed in. Seems like this could be updated to `inventory_file`.

installers/ansible/inventory
cbandy: `kubernetes_in_cluster=true` effectively disables the checks and actions we take around `kubernetes_context` and `openshift_host`. It indicates that you _want_ to use `kubectl` as-is. In the GCP installer there is _no_ kubeconfig file at all. There is no correct setting for context.

installers/pgo-installer/ansible/roles/pgo-installer/defaults/main.yml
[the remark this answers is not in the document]
cbandy: Over there, the longest installs happened when images weren't yet pulled.

installers/pgo-installer/ansible/roles/pgo-installer/tasks/main.yml
cbandy: The same inventory is used by the person invoking the installer and the installer itself? If so, what switches from "install/create remote k8s objects" to "install/create in-cluster k8s objects"?

rhel7/Dockerfile.pgo-installer.rhel7
jmckulk: These changes break the order of operations in the container build. This ends up installing an incorrect version of ansible that is incompatible with the current ansible install.  Reverting the change to the centos7 docker file to have 

```
yum install epel-release

yum install \
    other-stuff
```

cbandy: Each `RUN` instruction is a new layer in the container image. You can combine multiple commands into one `RUN` using `&&`:
```dockerfile
RUN yum install epel-release \
 && yum install \
    other-stuff
```

jmckulk: ah nice - ill update that

installers/pgo-installer/ansible/roles/pgo-installer/tasks/main.yml
jmckulk: >The same inventory is used by the person invoking the installer and the installer itself?

Yes, I'm definitely open to better ideas on how to handle this.

>If so, what switches from "install/create remote k8s objects" to "install/create in-cluster k8s objects"?

I'm not following.

cbandy: Perhaps I'll understand better with some docs. Right now, my assumption is that a person who is not inside the cluster will fill out their inventory file with `kubernetes_context='staging-or-so'` then run `ansible-playbook installers/pgo-installer/ansible/main.yml` to create the install container in the remote (staging) cluster. If so, when the install container runs, it will also have `kubernetes_context='staging-or-so'` (which isn't defined?)

jmckulk: I do need to get some docs going and hopefully that would explain better. Right now, I'm not accounting for contexts for the installer. I assume that the user is logged into their kube/ocp cluster has the ability to run kubectl/oc commands to create resources in that cluster. For example, the user has already run `oc login` and could run `oc create ns` on the environment in which they are attempting to install.

`kubectl_or_oc` in this case would be which binary do they want the ansible roles for the ansible installer to use. I'm thinking about setting this up with the `pgo_installer_environment=kubernetes/oc` variable that @jkatz mentioned so that the "correct" binary is used.

jmckulk: The user would fill out the inventory file in `$PGOROOT/installers/ansible` then use the following to run the install.
```
ansible-playbook -i $PGOROOT/installers/ansible/inventory --tags=install-container $PGOROOT/installers/pgo-installer/ansible/main.yml
```

installers/pgo-installer/README.md
jmckulk: This was very quick and dirty as a way to show how the installer would be used for purposes of review. Documentation will be more detailed

docs/content/Installation/postgres-operator-installer/_index.md
```diff
@@ -0,0 +1,53 @@
...
+### Installer Namespace
+The job template allows you to specify the namespace in
+which to run the install job. This does not specify which namespace the
+postgres-operator will be installed but they can both be in the same namespace.
+The namespace should be defined in place of the `{{ pgo_installer_namespace }}`
+variable.
+
+### Cluster Resources
+#### Service Account
+The postgres-operator-instaler
```
cbandy: `installer`

docs/content/Installation/postgres-operator-installer/_index.md
```diff
@@ -0,0 +1,53 @@
...
+requires a service account with cluster-admin privledges. You can create a
```
cbandy: `privileges`

docs/content/Installation/postgres-operator-installer/_index.md
```diff
@@ -0,0 +1,53 @@
...
+
+#### Config Map
+The ansible installer used by the `pgo-installer` image requires
+an inventory file to be created as a configmap in your environment. This
+configmap will be used to install the  PostgreSQL Operator and should meet all
+of the requirements outlined in the ansible install instructions.
+
+### Job Varibles
+#### Command
+The commanded defined in the installer job uses
```
cbandy: `command`

docs/content/Installation/postgres-operator-installer/_index.md
```diff
@@ -0,0 +1,53 @@
...
+the `pgo-install.sh` script to pass in the ansible tag will be run. The command
```
cbandy: `the ansible tag that will be run` or `the ansible tag to be run` maybe?

Issue #1306, opened 2020-02-27
Remove S3 secrets from logs/environment variables


dvianello:
**What is the motivation or use case for the change? **
S3 secrets for Backrest backups are visible in:
- environment variables
- logs

**Describe the solution you'd like**
- Secrets not visible in logs/environment variables
- Secrets mounted via K8S secrets

**Please tell us about your environment:**
- Operating System: CentOS
- Where is this running ( Local, Cloud Provider): VMware
- Storage being used (NFS, Hostpath, Gluster, etc): VMware vSan
- Container Image Tag: `v1.16.3`
- PostgreSQL Version: `centos7-12.1-4.2.1`
- Platform (Docker, Kubernetes, OpenShift): Kubernetes
- Platform Version: `v1.16.3`

jkatz:
@dvianello I agree 100% with this. Do you have redacted log examples with line info? That will help with tracing it. Thanks!

jkatz:
Also please provide the information that is requested in the bug reporting template:

* Operating System:
* Where is this running ( Local, Cloud Provider)
* Storage being used (NFS, Hostpath, Gluster, etc):
* Container Image Tag:
* PostgreSQL Version:
* Platform (Docker, Kubernetes, OpenShift):
* Platform Version:

dvianello:
@jkatz, updated the initial post above including the info you required.

Going back to where secrets are visible, if I configure the Operator (deployed via Ansible) with S3 secrets, and then spin up a cluster with
```pgo create cluster test  --pgbackrest-storage-type=local,s3 -n postgres-dbs```

I can see secrets in clear-text in the master description:
```
Name:         test-6874bb948b-7hntp
Namespace:    postgres-dbs
Priority:     0
Node:         node5/xxx.xxx.xxx.xxx
Start Time:   Mon, 02 Mar 2020 15:05:28 +0000
Labels:       archive-timeout=60
              crunchy-pgha-scope=test
              crunchy_collect=true
              deployment-name=test
              name=test
              pg-cluster=test
              pg-cluster-id=250b36ab-8310-44f5-87f2-bae9de7832fd
              pg-pod-anti-affinity=
              pgo-pg-database=true
              pgo-version=4.2.1
              pgouser=admin
              pod-template-hash=6874bb948b
              role=master
              service-name=test
              vendor=crunchydata
              workflowid=0ab6d073-39b6-4359-967a-85bf5df77ded
Annotations:  status:
[...]

PR #1352, opened 2020-03-17
Eliminate Logging of pgBackRest Environmental Variables with Sensitive information

jkatz:
**Checklist:**

 
**Type of Changes:**

 
Though the pgBackRest S3 credentials are stored in Kubernetes secrets, there were several cases of these credentials being logged in various Kubernetes resources or being made available in Kubernetes resources (e.g. `kubectl describe pod`), which ideally should not happen.

- The environmental variables are not printed out when certain containers load, which appeared to be legacy debugging code 
- The pgBackRest S3 credentials are now loaded into containers via "[environmental variables secrets](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables)" and are hidden by Kubernetes in its various resources.

There is a breaking change, in that the S3 credentials are no longer stored in a YAML file in the Secret, but rather get their own keys and key-value pairs.

**Other information**:

Issue: [ch7638]
Issue: #1306

### DIFF
--- a/operator/clusterutilities.go
+++ b/operator/clusterutilities.go
@@ -23,17 +23,15 @@ import (
 	"strconv"
 	"strings"
 
-	v1 "k8s.io/api/core/v1"
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/validation"
-
-	"gopkg.in/yaml.v2"
-
 	crv1 "github.com/crunchydata/postgres-operator/apis/cr/v1"
 	"github.com/crunchydata/postgres-operator/config"
 	"github.com/crunchydata/postgres-operator/kubeapi"
 	"github.com/crunchydata/postgres-operator/util"
+
 	log "github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/kubernetes"
 )
 
@@ -115,11 +113,12 @@ type PgbackrestEnvVarsTemplateFields struct {
 }
 
 type PgbackrestS3EnvVarsTemplateFields struct {
-	PgbackrestS3Bucket    string
-	PgbackrestS3Endpoint  string
-	PgbackrestS3Region    string
-	PgbackrestS3Key       string
-	PgbackrestS3KeySecret string
+	PgbackrestS3Bucket     string
+	PgbackrestS3Endpoint   string
+	PgbackrestS3Region     string
+	PgbackrestS3Key        string
+	PgbackrestS3KeySecret  string
+	PgbackrestS3SecretName string
 }
 
 type PgmonitorEnvVarsTemplateFields struct {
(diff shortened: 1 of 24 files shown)


util/cluster.go
```diff
@@ -132,15 +144,17 @@ func CreateBackrestRepoSecrets(clientset *kubernetes.Clientset,
...
-			"aws-s3-ca.crt":           configs.Data["aws-s3-ca.crt"],
-			"aws-s3-credentials.yaml": s3KeySecretData,
-			"config":                  configs.Data["config"],
-			"id_ed25519":              keys.Private,
-			"sshd_config":             configs.Data["sshd_config"],
-			"ssh_host_ed25519_key":    keys.Private,
+			BackRestRepoSecretKeyAWSS3KeyAWSS3Key:       backrestS3Key,
+			BackRestRepoSecretKeyAWSS3KeyAWSS3KeySecret: backrestS3KeySecret,
+			backRestRepoSecretKeyAuthorizedKeys:         keys.Public,
+			backRestRepoSecretKeyAWSS3KeyAWSS3CACert:    configs.Data["aws-s3-ca.crt"],
```
andrewlecuyer: Can we use the `const` defined above here instead of the hardcoded string? E.g.:
`configs.Data[backRestRepoSecretKeyAWSS3KeyAWSS3CACert]`

Same would apply to the next two as well.

util/cluster.go
```diff
@@ -104,23 +117,22 @@ func CreateBackrestRepoSecrets(clientset *kubernetes.Clientset,
...
-		s3KeySecretData, err = yaml.Marshal(s3Secret)
-		if err != nil {
-			return err
-		}
-	} else {
-		s3KeySecretData = configs.Data["aws-s3-credentials.yaml"]
+	// if an S3 key has been provided via the request, then use key and key secret
+	// inlcuded in the request instead of the default credentials that are
+	// available in the Operator pgBackRest secret
+	backrestS3Key := []byte(backrestRepoConfig.BackrestS3Key)
```
andrewlecuyer: Just wondering if the syntax and logic might be a bit simpler/terse here if we stick with strings until adding the value to the map, e.g. sticking with string comparisons here:

```
if backrestRepoConfig.BackrestS3Key == "" {
    backrestS3Key = configs.Data[BackRestRepoSecretKeyAWSS3KeyAWSS3Key]
}

if backrestRepoConfig.backrestS3KeySecret == "" {
    backrestS3KeySecret = configs.Data[BackRestRepoSecretKeyAWSS3KeyAWSS3KeySecret]
}
```

and then just convert to a byte array when setting in the map, e.g.:

```
BackRestRepoSecretKeyAWSS3KeyAWSS3Key:       []byte(backrestS3Key)
BackRestRepoSecretKeyAWSS3KeyAWSS3KeySecret: []byte(backrestS3KeySecret),
```

Anyway, just a thought to remove a few lines and do a string comparison vs. using `len` on the byte array.  Ultimately the same result either way, so def no need to change this one.

operator/clusterutilities.go
```diff
@@ -683,96 +682,51 @@ func GetPgmonitorEnvVars(metricsEnabled string) string {
...
-		} else {
-			s3EnvVars.PgbackrestS3Endpoint = Pgo.Cluster.BackrestS3Endpoint
-		}
-
-		if cluster.Spec.BackrestS3Region != "" {
-			s3EnvVars.PgbackrestS3Region = cluster.Spec.BackrestS3Region
-		} else {
-			s3EnvVars.PgbackrestS3Region = Pgo.Cluster.BackrestS3Region
-		}
+	if !(cluster.Labels[config.LABEL_BACKREST] == "true" &&
```
andrewlecuyer: I realize this is just carried over from the previous code, but at this point we really don't need the check of the backrest label since backrest is essentially always enabled as of 4.2.  And since this is one of those labels that needs to ultimately be removed holistically across the entire codebase, I'm def fine with leaving until we do so.

util/cluster.go
jkatz: How'd I miss that... 😐 Yes, I'll fix.

util/cluster.go
jkatz: I settled on this, so it's already casted, so no need to cast further down.

```go
	// if an S3 key has been provided via the request, then use key and key secret
	// inlcuded in the request instead of the default credentials that are
	// available in the Operator pgBackRest secret
	backrestS3Key := []byte(backrestRepoConfig.BackrestS3Key)

	if backrestRepoConfig.BackrestS3Key == "" {
		backrestS3Key = configs.Data[BackRestRepoSecretKeyAWSS3KeyAWSS3Key]
	}

	backrestS3KeySecret := []byte(backrestRepoConfig.BackrestS3KeySecret)

	if backrestRepoConfig.BackrestS3KeySecret == "" {
		backrestS3KeySecret = configs.Data[BackRestRepoSecretKeyAWSS3KeyAWSS3KeySecret]
	}
```

PR #1357, opened 2020-03-24
Allow tablespaces to be added to existing PostgreSQL clusters

jkatz:
**Checklist:**

 
**Type of Changes:**

 
This commit introduces the "pgo update cluster --tablespace"
flag, which allows for users to add additional tablespaces to a
PostgreSQL cluster.

The Operator takes care of creating PVCs for the additional PVCs
for the primary and all replicas in the PostgreSQL cluster, as well
as updating the Deployment template for each instance. This does
have the side effect of causing the PostgreSQL Pods to restart,
as the new volumes need to be mounted.

**Other information**:

Issue: [ch7232]


controller/pgcluster/pgclustercontroller.go
```diff
@@ -314,3 +324,131 @@ func addIdentifier(clusterCopy *crv1.Pgcluster) {
...
+		if _, ok := oldCluster.Spec.TablespaceMounts[tablespaceName]; !ok {
+			log.Debugf("new tablespace found: [%s]", tablespaceName)
+
+			newTablespaces[tablespaceName] = storageSpec
+		}
+	}
+
+	// now we can start creating the new tablespaces! First, create the new
+	// PVCs. The PVCs are created for each **instance** in the cluster, as every
+	// instnace needs to have a distinct PVC for each tablespace
```
abrightwell: `instance`

controller/pgcluster/pgclustercontroller.go
```diff
@@ -314,3 +324,131 @@ func addIdentifier(clusterCopy *crv1.Pgcluster) {
...
+			}
+		}
+	}
+
+	// now the fun step: update each deployment with the new volumes
+	for _, deployment := range deployments.Items {
+		labels := deployment.ObjectMeta.GetLabels()
+
+		// same deal as before: if this is not a PostgreSQL instance, skip it
+		instanceName, ok := labels[config.LABEL_DEPLOYMENT_NAME]
```
abrightwell: Should we filter these once up top and reuse in this step? No strong opinions either way, just seems like a potential optimization that may be worth the effort. :man_shrugging:

jkatz: Yeah, I had thought about that. There was a reason I had not, but I can't recall what it was and it may be moot at this point.

controller/pgcluster/pgclustercontroller.go
```diff
@@ -314,3 +324,131 @@ func addIdentifier(clusterCopy *crv1.Pgcluster) {
...
+	instanceNames := []string{}
+
+	for _, deployment := range deployments.Items {
+		labels := deployment.ObjectMeta.GetLabels()
+
+		// get the name of the PostgreSQL instance. If the "deployment-name"
+		// label is not present, then we know it's not a PostgreSQL cluster.
+		// Otherwise, the "deployment-name" label doubles as the name of the
+		// instance
+		if instanceName, ok := labels[config.LABEL_DEPLOYMENT_NAME]; ok {
```
andrewlecuyer: I don't think an issue here since I believe the deployment label is currently only on PG pods as stated here, but did want to mention that a `pgo-pg-database` label was added to the cluster deployment template to positively identify a pod or deployment that is a PG pod (i.e. a primary or replica), in the event that the more generic `deployment-name` label is used elsewhere.

controller/pgcluster/pgclustercontroller.go
jkatz: Ah ok. So we had needed the instance names in the step before. I thought the tradeoff (adding in the apps dependency, storing the deployments in an additional array) vs. the actual difference in size was not incrementally worth it.

PR #1365, opened 2020-03-25
Update pgBackRest Supported Flags

jmckulk:
**Checklist:**

 
**Type of Changes:**

 
Currently the apiserver runs validation on the backrest backup/restore options that can be passed in when one of these jobs are created. The flags that can be passed are checked in two ways, if it is a flag that is known to the operator and if that flag is on the "blacklist". If the flag is unknown the backup will be denied and the apiserver will provide a response saying that which flag is unknown. If this unknown flag is on the "blacklist" then it is valid but not supported by PGO, a response will be given by the apiserver saying which flag is unsupported. Validation is also run on the known flags to ensure that their values are allowed and provide fast feedback to the user if they try to pass in an invalid value for a valid flag.

The overall goal of this ticket is to allow users to access more of the pgbackrest functions and new features without adding the overhead of having to add each new pgBackRest flag. This will be done by removing some of the validation login and allowing users to pass in a flag to pgBackRest even if it is "unknown" to PGO.

**Other information**:
[ch7665]

### DIFF
--- a/apiserver/backupoptions/backupoptionsutil.go
+++ b/apiserver/backupoptions/backupoptionsutil.go
@@ -98,7 +98,10 @@ func convertBackupOptsToStruct(backupOpts string, request interface{}) (backupOp
 
 	err = commandLine.Parse(parsedBackupOpts)
 	if err != nil {
-		return nil, nil, handleCustomParseErrors(err, usage, optsStruct)
+		err = handleCustomParseErrors(err, usage, optsStruct)
+		if err != nil {
+			return nil, nil, err
+		}
 	}
 
 	setFlagFieldNames := obtainSetFlagFieldNames(commandLine, structType)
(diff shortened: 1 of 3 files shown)


apiserver/backupoptions/backupoptionsutil.go
```diff
@@ -98,7 +98,10 @@ func convertBackupOptsToStruct(backupOpts string, request interface{}) (backupOp
 
 	err = commandLine.Parse(parsedBackupOpts)
 	if err != nil {
-		return nil, nil, handleCustomParseErrors(err, usage, optsStruct)
+		err = handleCustomParseErrors(err, usage, optsStruct)
+		if err != nil {
```
jkatz: The use of `err` twice here is confusing. Which error is being passed to the function below?

jmckulk: Sure i'll update it to be more clear

jmckulk: Ill plan on updating to `customErr` unless i hear otherwise

jkatz: This can also use the

```go
if err := ...; err != nil {
}
``` syntax

jmckulk: alright ill update with that

PR #1380, opened 2020-03-31
Scale down Orig Primary, Fix rmdata After Failed Init & Remove rm -rf, Proper ConfigMap Cleanup

andrewlecuyer:
This PR back-patches various defect corrections (described below) to the `v4.2` baseline.

**Checklist:**

 
**Type of Changes:**

 
- It is not possible to scale down the original primary following a failover
- The `rmdata` job errors if the cluster was not successfully (and fully) initialized
- `rmdata` jobs run `rm -rf` when deleting cluster data during cluster deletion
- The `failover` config map (if it exists) is not deleted when the cluster is deleted
- The replica service for the cluster is not properly cleaned up when scaling down
- A `pgtask` already exists` error is shown when deleting a cluster if the original primary had previously been scaled down, and the resulting `pgtask` still remains.

[ch7285]
[ch6917]
[ch7657]
[ch7746]
[ch7741]
[ch7751]

- It is now possible to scale down the original primary following a failover
- The `rmdata` job completes successfully, even if the cluster wasn't successfully (and fully) initialized
- `rmdata` jobs no longer run `rm -rf` when deleting cluster data during cluster deletion
- The `failover` config map (if it exists) is now deleted when the cluster is deleted
- The replica service for the cluster is now properly cleaned up when scaling up and down (i.e. it is removed when the last replica is scaled down, and it is added once a replica is added to the cluster)
- When a new rmdata `pgtask` is being created to delete a cluster, an attempt is made to first delete and `pgtask` with the same name.  This ensures a `pgtask already exists` error will not be encountered when deleting a cluster

**Other information**:

N/A


apiserver/clusterservice/scaleimpl.go
```diff
@@ -292,49 +291,48 @@ func ScaleDown(deleteData bool, clusterName, replicaName, ns string) msgs.ScaleD
...
 	}
 
-	//if this was the last replica then remove the replica service
-	var replicaList *v1.DeploymentList
-	selector := config.LABEL_PG_CLUSTER + "=" + clusterName + "," + config.LABEL_SERVICE_NAME + "=" + clusterName + "-replica"
-	replicaList, err = kubeapi.GetDeployments(apiserver.Clientset, selector, ns)
+	// selector in the format "pg-cluster=<cluster-name>,role=replica"
+	// which will grab any/all replicas
+	selector := fmt.Sprintf("%s=%s,%s=%s", config.LABEL_PG_CLUSTER, clusterName,
+		config.LABEL_PGHA_ROLE, "replica")
```
jkatz: We probably should make `"replica"` a constant in one of the global config files, perhaps the catch-all `config/labels.go`

apiserver/clusterservice/scaleimpl.go
```diff
@@ -292,49 +291,48 @@ func ScaleDown(deleteData bool, clusterName, replicaName, ns string) msgs.ScaleD
...
 	if err != nil {
 		response.Status.Code = msgs.Error
 		response.Status.Msg = err.Error()
 		return response
 	}
-	if len(replicaList.Items) == 0 {
-		response.Status.Code = msgs.Error
-		response.Status.Msg = "no replicas found for this cluster"
-		return response
-	}
```
jkatz: We should not remove this on 4.2. This would technically be a breaking API change. It's ok to remove in 4.3

apiserver/clusterservice/scaleimpl.go
```diff
@@ -292,49 +291,48 @@ func ScaleDown(deleteData bool, clusterName, replicaName, ns string) msgs.ScaleD
...
-	clusterPGHAScope := cluster.ObjectMeta.Labels[config.LABEL_PGHA_SCOPE]
 	err = apiserver.CreateRMDataTask(clusterName, replicaName, taskName, deleteBackups, deleteData, isReplica, isBackup, ns, clusterPGHAScope)
 	if err != nil {
 		response.Status.Code = msgs.Error
 		response.Status.Msg = err.Error()
 		return response
 	}
 
-	response.Results = append(response.Results, "deleted Pgreplica "+replicaName)
+	response.Results = append(response.Results, "deleted replica "+replicaName)
```
jkatz: OK the change on 4.3, not on 4.2, as this would be a breaking API change.

andrewlecuyer: Just a note that without this change, this message doesn't always make sense anymore.  Specifically, when scaling down the original primary there is no `pgreplica` being deleted, which is why this subtle change was needed.

Anyway, I'll go ahead and flip it back to `Pgreplica` for the 4.2 line.

apiserver/clusterservice/scaleimpl.go
andrewlecuyer: I'll go ahead and add this back.

apiserver/clusterservice/scaleimpl.go
andrewlecuyer: Since this was a back-patch, I just grabbed this change as it exists on the Master branch.  Agree a constant would be better though, so I can add that.  We'll prob just want to circle back and do the same on Master as some point.

jkatz: Yup that makes sense.

apiserver/clusterservice/scaleimpl.go
jkatz: Unfortunately that is the pain of a published API :smile: I do agree, but we'll get that fixed up on `master`

andrewlecuyer:
@jkatz thanks for the feedback.

All requested changes have been made.

Issue #1337, opened 2020-03-12
S3 certificate parameter


aurelienmarie:
I am using Minio and it has the ability to enable TLS. However, i need to pass the certificate so the operator is allowed to communicate with Minio. 

I've found a way to provide it during [operator deployment](https://github.com/CrunchyData/postgres-operator/blob/master/deploy/deploy.sh#L35), but there is no documentation around that subject.

Could there be an update about it please ?

jkatz:
Hi @aurelienmarie,

We are already planning to do this for the next release. Thanks for reporting!

jkatz:
This is handled by #1387

Issue #1228, opened 2020-01-17
pgBackrest stanza create fails with URI path style s3 backend


GeorgFleig:
**Describe the bug**
I am experiencing the same issue as described in https://github.com/CrunchyData/postgres-operator/issues/1155 with PGO version 4.2.0.

Created a fresh pg cluster with S3 (Minio) as pgBackrest backend. Primary pg pod and pgbackrest pods are created, the stanza creation fails.

```
# kubectl logs pgs3only-stanza-create-qwn5k
time="2020-01-17T10:36:53Z" level=info msg="pgo-backrest starts"
time="2020-01-17T10:36:53Z" level=info msg="debug flag set to false"
time="2020-01-17T10:36:53Z" level=info msg="backrest stanza-create command requested"
time="2020-01-17T10:36:53Z" level=info msg="s3 flag enabled for backrest command"
time="2020-01-17T10:36:53Z" level=info msg="command to execute is [pgbackrest stanza-create  --db-host=192.168.151.191 --db-path=/pgdata/pgs3only --repo-type=s3]"
time="2020-01-17T10:36:53Z" level=info msg="command is pgbackrest stanza-create  --db-host=192.168.151.191 --db-path=/pgdata/pgs3only --repo-type=s3 "
time="2020-01-17T10:36:54Z" level=error msg="command terminated with exit code 50"
time="2020-01-17T10:36:54Z" level=info msg="output=[]"
time="2020-01-17T10:36:54Z" level=info msg="stderr=[ERROR: [050]: unable to acquire lock on file '/tmp/pgbackrest/db-archive.lock': Resource temporarily unavailable\n       HINT: is another pgBackRest process running?\n]"
time="2020-01-17T10:36:54Z" level=error msg="command terminated with exit code 50"
```

Executing the same command inside the pgbackrest-repo pod produces the same error. The lock file is present and used by a running pgbackrest process:

```
[pgbackrest@pgs3only-backrest-shared-repo-865d79d7f4-dtpg6 /]$ pgbackrest stanza-create  --db-host=192.168.151.191 --db-path=/pgdata/pgs3only --repo-type=s3
ERROR: [050]: unable to acquire lock on file '/tmp/pgbackrest/db-archive.lock': Resource temporarily unavailable
       HINT: is another pgBackRest process running?

[pgbackrest@pgs3only-backrest-shared-repo-865d79d7f4-dtpg6 /]$ cat /tmp/pgbackrest/db-archive.lock
69

[...]

PR #1387, opened 2020-04-01
Specify a value for a S3 CA during cluster creation

jkatz:
**Checklist:**

 
**Type of Changes:**

 
The PostgreSQL Operator by default ships with the trusted CA for AWS
S3 for ease of integration, though people may use other S3 compatible
interfaces. While one can relatively easily update the primary
"pgo-backrest-repo-config" Secret and set the value of "aws-s3-ca.crt"
to the value of the CA that they wish to use, there may be cases where
one must specify this value on a per-cluster basis.

This introduces the "--pgbackrest-s3-ca-secret" flag on the
"pgo create cluster" command. One must first create a Kubernetes Secret
and put the value of the CA in a key called "aws-s3-ca.crt". The name
of this Secret should be passed as the value to "--pgbackrest-s3-ca-secret",
and the PostgreSQL Operator will use this value as the CA that pgBackRest
will use for pushing to a S3 interface.

Any clones of this cluster will also use this CA for their pgBackRest
S3 interactions.

**Other information**:

Issue: [ch6918]
Issue: #1228
Issue: #1337

### DIFF
--- a/util/cluster.go
+++ b/util/cluster.go
@@ -36,6 +36,9 @@ import (
 
 // BackrestRepoConfig represents the configuration required to created backrest repo secrets
 type BackrestRepoConfig struct {
+	// BackrestS3CA is the byte string value of the CA that should be used for the
+	// S3 inerfacd pgBackRest repository
+	BackrestS3CA        []byte
 	BackrestS3Key       string
 	BackrestS3KeySecret string
 	ClusterName         string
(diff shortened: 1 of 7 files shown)


util/cluster.go
```diff
@@ -63,17 +67,17 @@ const (
 
 // values for the keys used to access the pgBackRest repository Secret
 const (
-	// two are these are exported, as they are used to help add the information
-	// into the templates. Say the second one 10 times fast
+	// three are these are exported, as they are used to help add the information
```
abrightwell, line 70: `three are these`?

jkatz: Yes. I'm hijacking English grammar.

PR #1396, opened 2020-04-09
Make Cluster Roles Optional & Introduce Namespace Operating Modes

andrewlecuyer:
This commit makes Cluster Roles optional within this PostgreSQL Operator.  This is specifically accomplished by introducing the concept of namespace operating modes, which represent various modes of operation of the Postgres Operator that dictate which namespace capabilities are enabled (if any at all).  The specific namespace mode assigned to an Operator is determined based on the specific Cluster Roles assigned to the postgres-operator Service Account.  Therefore, users can choose to install the Cluster Roles required for a desired operating mode, and the Operator will automatically detect those roles and enable or disable the proper namespace capabilities.  This includes allowing users to avoid the installation of any Cluster Roles altogether, fully disabling namespace capabilities.

When the PostgreSQL Operator and all of its various components (i.e. the apisever, scheduler, etc.) are run, the Kubernetes environment is inspected to determine what cluster roles are currently assigned to the `pgo-operator` Service Account (i.e. the Service Account running the Pod the PostgreSQL Operator is running within).  Based on the Cluster Roles identified, one of the following namespace operating modes will be enabled for the installation:

- `dynamic`: Enables full dynamic namespace capabilities, in which the Operator can create, delete and update any namespaces within the Kubernetes cluster, while then also having the ability to create the roles, role bindings and service accounts within those namespaces as required for the Operator to create PG clusters.  Additionally, while in this mode the Operator can listen for namespace events (e.g. namespace additions, updates and deletions), and then create or remove controllers for various namespaces as those namespaces are added or removed from the Kubernetes cluster and/or Operator install.
- `readonly`: In this mode the Operator is still able to listen for namespace events within the  Kubernetetes cluster, and then create and run and/or remove controllers as namespaces are added, updated and deleted.  However, while in this mode the Operator is unable to create, delete or update namespaces itself, nor can it create the RBAC it requires in any of those namespaces to create PG clusters.  Therefore, while in a "readonly" mode namespaces must be pre-configured with the proper RBAC, since the Operator cannot create the RBAC itself.
- `disabled`: Disables namespace capabilities within the Operator altogether.  While in this mode the Operator will simply attempt to work with the target namespaces specified during installation.  If no target namespaces are specified, then the Operator will be configured to work within the namespace in which it is deployed.  As with `readonly`, while in this mode namespaces must be pre-configured with the proper RBAC, since the Operator cannot create the RBAC itself.

In support of this change, both the Ansible and Bash installers have been updated to provide the user with a way of easily installing the RBAC corresponding to each namespace operating mode.  The specific variables used to install the proper Cluster Roles and therefore enable the desired namespace operating mode are as follows:

- **Bash**: `PGO_NAMESPACE_MODE`
- **Ansible**: `namespace_mode`

When using the above variables with either installer, the valid options available are:

- `dynamic`
- `readonly`
- `disabled`

When any of the values are selected during installation, the proper Cluster roles will be installed to enable the corresponding namespace operating mode as described above. The default value for either installer is `dynamic`, which maintains the current default behavior when installing the Operator.

And finally this commit also updates various controller logic.  All namespace informers now only listen for namespaces with the proper `crunchydata` vendor label, as well as those labeled for the current Operator install.  Additionally, all namespace controllers how have a resync interval of 10 minutes.  Also, all controller manager functions are now thread safe, while the Scheduler and Operator now share the create the same type of namespace controller.  Otherwise various other controller cleanup and/or reorganization is included.

**Checklist:**

 
**Type of Changes:**

 
- Cluster Roles must be installed in order to use the Operator
- Clean mechanisms/strategies for enabling full/dynamic namespace capabilities and read-only namespace capabilities do not exist

[ch7578]

- Cluster Roles are no longer required to use the Operator, and can be fully disabled
- Namespace Operating Modes now exist that provide clear/concise mechanisms for enbabling full/dyanmic namespace capabilities or read-only namespace capabilities, including mechanisms to install the proper RBAC to enable various operating modes

**Other information**:

N/A


docs/content/architecture/namespace.md
```diff
@@ -75,28 +70,145 @@ variables would be as follows:
...
+determining whether or not certain namespaces capabilities are enabled for the Operator
+installation. When the PosgreSQL Operator is run, the Kubernetes environment is inspected to 
+determine what cluster roles are currently assigned to the `pgo-operator` `ServiceAccount` 
+(i.e. the `ServiceAccount` running the Pod the PostgreSQL Operator is deployed within).  Based
+on the `ClusterRoles` identified, one of the namespace operating modes described below will be
+enabled for the Operator installation.  Please consult the installation guides for the various
+installation methods available to determine the settings required to install the `ClusterRoles`
+required for each mode.
+
+### dynamic
```
jkatz, line 110:
```markdown
### `dynamic`
```

docs/content/architecture/namespace.md
```diff
@@ -75,28 +70,145 @@ variables would be as follows:
...
+      - rbac.authorization.k8s.io
+    resources:
+      - rolebindings
+    verbs:
+      - get
+      - create
+      - delete
+```
+
+### readonly
```
jkatz, line 170:
```markdown
### `readonly`
```

docs/content/architecture/namespace.md
```diff
@@ -75,28 +70,145 @@ variables would be as follows:
...
+      - ''
+    resources:
+      - namespaces
+    verbs:
+      - get
+      - list
+      - watch
+```
+
+### disabled
```
jkatz, line 197:
```markdown
### `disabled`
```

deploy/deploy.sh
```diff
@@ -29,12 +29,15 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 $DIR/cleanup.sh
 
-$PGO_CMD get clusterrole pgo-cluster-role 2> /dev/null > /dev/null
-if [ $? -ne 0 ]
+if [[ "${PGO_NAMESPACE_MODE:-dynamic}" != "disabled" ]]
```
jkatz, line 32: Hm, I feel like the default here contrasts with the [ansible rule](https://github.com/CrunchyData/postgres-operator/pull/1396/files#diff-8b078c64bd0ddf6cc8b7e1e3d60c147dR98) which is `disabled`.

I think we need to default to one or the other.

I'm torn between `dynamic` and `disabled` -- I think `disabled` is a lower barrier to entry though esp. if one does not have permission to create `ClusterRole`

installers/ansible/roles/pgo-operator/defaults/main.yml
```diff
@@ -40,6 +40,7 @@ pod_anti_affinity_pgbackrest: ""
 pod_anti_affinity_pgbouncer: ""
 
 namespace: ""
+namespace_mode: "dynamic"
```
jkatz, line 43: Noting for the defaults between `dynamic` and `disabled`

ns/nslogic.go
```diff
@@ -101,21 +104,70 @@ type PgoPgRoleBinding struct {
...
+	// NamespaceOperatingModeDisabled causes namespace capabilities to be disabled altogether.  In
+	// this mode the Operator will simply attempt to work with the target namespaces specified
+	// during installation.  If no target namespaces are specified, then it will be configured to
+	// work within the namespace in which the Operator is deployed.
+	NamespaceOperatingModeDisabled NamespaceOperatingMode = "disabled"
+
+	// DNS-1123 formatting and error message for validating namespace names
+	dns1123Fmt    string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?"
+	dns1123ErrMsg string = "A namespace name must consist of lower case" +
+		"alphanumeric characters or '-', and must start and end with an alphanumeric character"
```
jkatz, line 135: I would recommend using the standard Kubernetes validation library:

- [https://godoc.org/k8s.io/kubernetes/pkg/apis/core/validation](https://godoc.org/k8s.io/kubernetes/pkg/apis/core/validation)
- [https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go)

jkatz: Per discussion, ignore.

deploy/deploy.sh
andrewlecuyer: The default for Ansible is also `dynamic` (see file `installers/ansible/roles/pgo-operator/defaults/main.yml`).

I do see the potential confusion however, since the `inventory` has commented out variable that specifies `disabled`, e.g.:
```
# namespace_mode='disabled'  
```
I did this to have an easy solution for disabling (i.e. just remove the `#`), but I can def change either of these (the default and `inventory` var) as needed.

jkatz:
Per discussion, defaulting to `dynamic`

deploy/deploy.sh
jkatz: I would just do

```ini
# namespace_mode='dynamic'
```

Really the commented out ones should mirror the defaults.

docs/content/architecture/namespace.md
andrewlecuyer: Updated

docs/content/architecture/namespace.md
andrewlecuyer: Updated

docs/content/architecture/namespace.md
andrewlecuyer: Updated

deploy/deploy.sh
andrewlecuyer: `dynamic` is now specified.

installers/ansible/roles/pgo-operator/defaults/main.yml
andrewlecuyer: Sticking with default of `dynamic` per discussion.

PR #1409, opened 2020-04-14
Update the Pgcluster CRD to Include a CCPImagePrefix Value

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
**Other information**:


operator/clusterutilities.go
```diff
@@ -866,3 +866,12 @@ func writeTablespaceJSON(w *bytes.Buffer, jsonFields interface{}) error {
...
 }
+
+// GetImagePrefix checks whether the CCPImagePrefix value given from the pgcluster
+// CRD has been set. If so, this value is returned, otherwise the default value is used.
+func GetImagePrefix(crdCCPImagePrefix string) string {
+	if crdCCPImagePrefix != "" {
+		return crdCCPImagePrefix
+	}
+	return Pgo.Cluster.CCPImagePrefix
+}
```
jkatz, line 877: Place this in alphabetical order in the file. (or as best as you can as the whole file is not in alpha order)

apiserver/clusterservice/clusterimpl.go
```diff
@@ -1133,6 +1133,16 @@ func getClusterParams(request *msgs.CreateClusterRequest, name string, userLabel
...
 	}
+
+	// update the CRD spec to use the custom CCPImagePrefix, if given
+	// otherwise, set the value from the global configuration
+	if request.CCPImagePrefix != "" {
+		spec.CCPImagePrefix = request.CCPImagePrefix
+		log.Debugf("user is overriding CCPImagePrefix from command line %s", request.CCPImagePrefix)
+	} else {
+		spec.CCPImagePrefix = apiserver.Pgo.Cluster.CCPImagePrefix
+	}
```
jkatz, line 1144: Why not user that operator `GetImagePrefix` function here?

Or make the `GetImagePrefix` function a utility function used across apiserver/operator?

Not a requirement, but seems like a way to not duplicate logic.

jkatz: ...adding on to this thought, based on how `GetImagePrefix` works, do we even need to have the `else` block here? That function automatically sets the image prefix.

We should also have a fallback if neither the user specifies it one off in `pgo create cluster` nor the `pgo.yaml` setting is set. I would propose that we have a `const DefaultImagePrefix` set in the cluster.go CRD file that just sets it to `crunchydata`

operator/pgdump/dump.go
```diff
@@ -81,6 +81,13 @@ func Dump(namespace string, clientset *kubernetes.Clientset, client *rest.RESTCl
 		}
 	}
 
+	// get the pgcluster CRD for cases where a CCPImagePrefix is specified
+	cluster := crv1.Pgcluster{}
+	_, err = kubeapi.Getpgcluster(client, &cluster, task.Spec.Parameters[config.LABEL_PG_CLUSTER], namespace)
+	if err != nil {
+		log.Error(err)
+	}
```
jkatz, line 89: There are a few things here in which if they are not set, this will totally fail.

Recommendation:

```go
clusterName := task.Spec.Parameters[config.LABEL_PG_CLUSTER]

if clusterName == "" {
    log.Error(err)
    return
}

if _, err := kubeapi.Getpgcluster(client, &cluster, clusterName, namespace); err != nil {
  log.Error(err)
  return
}
```

tjmoore4: Agreed, I'll implement that change.

operator/clusterutilities.go
tjmoore4: sure thing. I put it directly above GetInstanceDeployments

apiserver/clusterservice/clusterimpl.go
tjmoore4: My initial thought was not to use `GetImagePrefix` here because it will require importing the `operator` package where it hasn't been previously. Is that worth using the function here? If so, happy to make the change.

tjmoore4:
This PR is being closed as it is being replaced by 
https://github.com/CrunchyData/postgres-operator/pull/1427

PR #1427, opened 2020-04-17
CCPImagePrefix and PGOImagePrefix Added to Pgcluster CRD

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
The CCPImagePrefix and PGOImagePrefix values are currently only set in a global configuration,
set during deployment in either the pgo.yaml, Ansible inventory or in the related environment
variable.

This PR adds a 'CCPImagePrefix' and a 'PGOImagePrefix' value to the pgcluster CRD
which will be used in place of the global configuration values, when set. If not set, the 
existing behavior will be maintained and the global configuration values will be used.

This update also adds two new flag options for the PGO client, 'ccp-image-prefix' and 'pgo-image-prefix' that will override the image prefixes from the global configuration when used and store the provided value in the pgcluster CRD. Note: the flags can be used independently

**Other information**:
Testing:

For testing, a good plan is to set the configured prefix values in your pgo.yaml or inventory file to a
'wrong' value, for instance, if your images are tagged 'crunchydata', configure the Operator
deployment to anything else, such as 'newprefix'. Then, re-tag the four initially deployed Operator containers (pgo-apiserver, postgres-operator, pgo-scheduler and pgo-event) to use the new prefix.

At this point, any new clusters created will only find their container images if the relevant prefix flags are used during creation. An example is provided below:
```
pgo create cluster mycluster --pgo-image-prefix=crunchydata --ccp-image-prefix=crunchydata
```
Since these flags, as well as the underlying CRD values, can be used separately, different prefixes can be used for the two sets of containers.
Issue [CH4918]


apiserver/backrestservice/backrestimpl.go
```diff
@@ -212,9 +214,8 @@ func CreateBackup(request *msgs.CreateBackrestBackupRequest, ns, pgouser string)
 	return resp
 }
 
-func getBackupParams(identifier, clusterName, taskName, action, podName, containerName, backupOpts, backrestStorageType, jobName, ns, pgouser string) *crv1.Pgtask {
+func getBackupParams(identifier, clusterName, taskName, action, podName, containerName, imagePrefix, backupOpts, backrestStorageType, jobName, ns, pgouser string) *crv1.Pgtask {
```
abrightwell, line 217: Perhaps out of scope for this changeset, however, is it possible to perhaps create some kind of non-exported struct for these params? It just seems cumbersome to work with as it sits.

perhaps something like:

```
type getBackupParamsInfo struct {
  identifier string
  clusterName string,
 ...
}
```

Just a thought to make it a bit easier to maintain/use.

operator/clusterutilities.go
```diff
@@ -392,6 +392,16 @@ func GetTablespaceNamePVCMap(clusterName string, tablespaceStorageTypeMap map[st
...
+// GetImagePrefix checks whether the imagePrefix value provided from the pgcluster
+// CRD has been set. If so, this value is returned, otherwise the default global value
+// is used. This supports both the CCPImagePrefix and the PGOImagePrefix
+func GetImagePrefix(crdImagePrefix, globalImagePrefix string) string {
+	if crdImagePrefix != "" {
+		return crdImagePrefix
+	}
+	return globalImagePrefix
+}
+
```
abrightwell, line 404: Slightly picking a nit here, but wouldn't it be more appropriate to call it 'default' over 'global' since the value is being provided to the function?  Global, IMV, means that it's a static value that is accessible  'globally' and doesn't need to be passed as a parameter for such a function.

If it is truly 'global', then perhaps the following would be more appropriate?

```
func GetImagePrefix(imagePrefix string) string {
    if imagePrefix != "" {
        return imagePrefix
    }
    return GlobalimagePrefix
}
```

Otherwise, what you have but with...

```
func GetImagePrefix(imagePrefix string, defaultPrefix string) {
    ...
    return defaultPrefix
}
```

Just a thought.

pgo-scheduler/scheduler/policy.go
```diff
@@ -191,3 +191,12 @@ func (p PolicyJob) Run() {
...
 }
+
+// getCCPImagePrefix checks whether the CCPImagePrefix value is set in the pgcluster CRD.
+// If so, this value is returned to be used, otherwise the value from the job is used.
+func getCCPImagePrefix(cluster crv1.Pgcluster, jobPrefix string) string {
+	if cluster.Spec.CCPImagePrefix != "" {
+		return cluster.Spec.CCPImagePrefix
+	}
+	return jobPrefix
+}
```
abrightwell, line 202: Isn't this effectively the same logic as the `GetImagePrefix`function above? I feel like this could be DRY'ed up a little.

apiserver/backrestservice/backrestimpl.go
jkatz: Agreed for the purposes of documenting doing these kinds of changes, but I will certainly `echo 'out of scope for this changeset'`

operator/clusterutilities.go
jkatz: I agree with Adam on the first one.

pgo-scheduler/scheduler/policy.go
tjmoore4: absolutely, didn't catch this during my rewrite of `GetImagePrefix` on Friday. It's corrected now, thanks.

apis/crunchydata.com/v1/cluster.go
```diff
@@ -105,6 +106,7 @@ type PgclusterSpec struct {
 	BackrestS3Region   string                   `json:"backrestS3Region"`
 	BackrestS3Endpoint string                   `json:"backrestS3Endpoint"`
 	BackrestRepoPath   string                   `json:"backrestRepoPath"`
+	PGOImagePrefix     string                   `json:"pgoimageprefix"`
```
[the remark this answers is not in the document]
tjmoore4: Works for me, updated.

operator/clusterutilities.go
tjmoore4: Understood, it is not a global value from an implementation perspective. I agree defaultPrefix is less confusing. I've updated it as such, thanks!

apiserver/backrestservice/backrestimpl.go
tjmoore4: I'll plan to leave this as is for the time being, but I am interested in a broader backup design discussion. One of the reasons the `GetImagePrefix` function is written to accept two string values (instead of, for instance, accepting a `pgcluster` CRD is due to the variation in how the values are passed, particularly related to the backups.

operator/clusterutilities.go
abrightwell: So, giving this one more thought, it almost seems like this could/should be a `util` like function? It kind of seems like something that could/would be useful elsewhere perhaps? So why not make it available as such?

```
package util

func GetValueOrDefault(value string, default string) string {
    if value != "" {
        return value
    }

    return default
}
```

Thoughts?

tjmoore4: I think it makes sense, it looks like it results in less new package imports, and `util` is a bit more logical place at this point. I'll implement this change, thanks.

util/util.go
```diff
@@ -202,6 +202,15 @@ func PatchClusterCRD(restclient *rest.RESTClient, labelMap map[string]string, ol
 
 }
 
+// GetValueOrDefault checks whether the first value given is set. If it is,
+// that value is returned. If not, the second, default value is returned instead
+func GetValueOrDefault(value, defaultValue string) string {
```
jkatz, line 207: Let's call this `GetImagePrefix`

abrightwell: I have no strong opinions on renaming, but the intent was to provide a more generic utility function. I could agree that better naming could be had... but, FWIW, my initial suggestion for this naming was more for conveying the purpose/intent than the actual implementation naming.

jkatz: I missed that conversation and just saw it in the email thread.

I understand the genericness of it, but it seems like this is ad-hoc to the imageprefix problem. If other such problems exist, we can certainly apply this to it.

I'll leave it up to @tjmoore4 on what he wants to do. I can go either way.

tjmoore4: **Short answer**: I'll leave it as it is now, as I think the general case makes sense at this point.

**Longer answer**: While part of me would like to have it named for it's sole current use, the function itself is _much_ more generic than I originally conceived it would be because of the differences in how the prefixes are handled throughout the code. 
If there's any chance that it could be useful for other purposes (and I believe there is), it's now in the correct package (`util`) and named correctly for more general use.
The other option is to rename it now, and change it later if it's used for other purposes, but that adds needless complication in my opinion.
The last, bad, not really an option, option is someone may want to use it, despite it being named `GetImagePrefix` or similar, in which case it would be very confusing and bad.
So, I say best to leave it as is :smile:

jkatz: I'm at peace with this.

PR #1443, opened 2020-04-23
Relax the default PostgreSQL instance memory to avoid crashes in test

jkatz:
**Checklist:**

 
**Type of Changes:**

 
By default, modern versions of PostgreSQL set the `shared_buffers`
configuration parameter to 128MB, which which under most circumstances
means that the host machine where PostgreSQL is running should have at
least 512MB (or in Kubernetes: 512Mi).

However, in test environments, this causes people to run out of memory
very, very quickly, and in turn, "bad things happen."

As a compromise
between production best practices and getting quickly up and running,
the following compromise is offered:

- The install configuration files set "default instance memory" to 128Mi
- The documentation notes this and adamantly pleads that production
environments change this setting
- The "default of last resort" remains at 512Mi

I also corrected some alphabetization and one misspelling of Kubernetes.

**Other information**:

Issue: [ch7881]


installers/ansible/inventory
```diff
@@ -210,8 +210,10 @@ db_user='testuser'
 # Operator defaults are listed with each variable
 #
 # "default_instance_memory" represents the memory request for a PostgreSQL
-# instance
-# default_instance_memory='512Mi'
+# instance. For production instances, this should be at least 512Mi, but we
+# model a number that is OK for just basic deployment testing. However, you
+# really should make this value bigger
+# default_instance_memory='128Mi'
```
jmckulk, line 216: Should this be uncommented? If not the default value will be different for the ansible install and the installer. Because it is uncommented in the template, and the value in `postgres-operator.yml` is set.

jkatz: Yes, you're corect.

jmckulk: cool, other than that lgtm. Tested it out with the ansible install

PR #1451, opened 2020-04-27
Installer cleanups

jkatz:
**Checklist:**

 
**Type of Changes:**

 
Client Installer:

- Better quoting around the environmental variables
- Allow for adjustment of the "pgouser-admin" user for grabbing
credentials
- Ensure the credentials are safely stored in the home directory
with tighter permissions

Default installer to `hostpathstorage`

This is used by many of the test enviroments (minikube, CRC) and
so this will make it easier for people to get the PostgreSQL Operator
up and running. Production installations typically tweak these
settings.

### DIFF
--- a/installers/kubectl/client-setup.sh
+++ b/installers/kubectl/client-setup.sh
@@ -12,21 +12,22 @@
 #  limitations under the License.
 
 # This script should be run after the operator has been deployed
-export PGO_OPERATOR_NAMESPACE=${PGO_OPERATOR_NAMESPACE:-pgo}
+export PGO_OPERATOR_NAMESPACE="${PGO_OPERATOR_NAMESPACE:-pgo}"
+PGO_USER_ADMIN="${PGO_USER_ADMIN:-pgouser-admin}"
 
 # Check that the pgouser-admin secret exists
-if [ -z "$(kubectl get secret -n $PGO_OPERATOR_NAMESPACE pgouser-admin)" ]
+if [ -z "$(kubectl get secret -n ${PGO_OPERATOR_NAMESPACE} ${PGO_USER_ADMIN})" ]
 then
-    echo "pgouser-admin Secret not found in namespace: $PGO_OPERATOR_NAMESPACE"
+    echo "${PGO_USER_ADMIN} Secret not found in namespace: ${PGO_OPERATOR_NAMESPACE}"
     echo "Please ensure that the PostgreSQL Operator has been installed."
     echo "Exiting..."
     exit 1
 fi
 
 # Check that the pgo.tls secret exists
-if [ -z "$(kubectl get secret -n $PGO_OPERATOR_NAMESPACE pgo.tls)" ]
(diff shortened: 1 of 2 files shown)


installers/kubectl/client-setup.sh
```diff
@@ -12,41 +12,49 @@
 #  limitations under the License.
 
 # This script should be run after the operator has been deployed
-export PGO_OPERATOR_NAMESPACE=${PGO_OPERATOR_NAMESPACE:-pgo}
+export PGO_OPERATOR_NAMESPACE="${PGO_OPERATOR_NAMESPACE:-pgo}"
+PGO_USER_ADMIN="${PGO_USER_ADMIN:-pgouser-admin}"
```
jmckulk, line 16: should there be an export here?

jmckulk: or alternatively remove the export on `PGO_OPERATOR_NAMESPACE`

installers/kubectl/client-setup.sh
```diff
@@ -12,41 +12,49 @@
...
-    echo "pgouser-admin Secret not found in namespace: $PGO_OPERATOR_NAMESPACE"
+    echo "${PGO_USER_ADMIN} Secret not found in namespace: ${PGO_OPERATOR_NAMESPACE}"
     echo "Please ensure that the PostgreSQL Operator has been installed."
     echo "Exiting..."
     exit 1
 fi
 
 # Check that the pgo.tls secret exists
-if [ -z "$(kubectl get secret -n $PGO_OPERATOR_NAMESPACE pgo.tls)" ]
+if [ -z "$(kubectl get secret -n ${PGO_OPERATOR_NAMESPACE} pgo.tls)" ]
```
jmckulk, line 28: If we are adding a variable for the user secret why not add it for the pgo.tls secret as well?

jkatz: `pgo.tls` is unchanging, whereas there may be multiple users accessing an Operator :wink:

jmckulk: gotcha

installers/kubectl/client-setup.sh
jkatz: `PGO_USER_ADMIN` does not need to be exported because it's never used anywhere else but here. Perhaps it should be lowercased, but I leave that to the better Bashers than me.

Looking through the code, we don't need to export `PGO_OPERATOR_NAMESPACE` unless one is running the tests. I would suggest we leave it given it doesn't hurt.

jkatz:
Merged in 9dc755983 and  baa5749e

installers/kubectl/client-setup.sh
jmckulk: Is it the scripts job to setup a users environment? If so, should we also automatically export the variables at the bottom?

jkatz: Yeah...I don't think `PGO_OPERATOR_NAMESPACE` is used unles you're testing (which is a developer task). That's why I think it doesn' thurt to have it available. Should we put it into the environment in general? 🤔 tough one. Will think on it.

PR #1456, opened 2020-04-27
Update Namespace Validation & Orchestrate pgBackRest Repo & Primary DB Init

andrewlecuyer:
This PR includes the following changes:

- Namespace validation has been cleaned up within the API server.  Within the `ns` package, a `ValidateNamespacesWatched` func now provides a single method of validating whether one or more namespaces are watched by the current Operator installation. Additionally, function `GetNamespaceList` has been renamed to `GetInitialNamespaceList` to indicate that it is responsible for obtaining an initial list of watched namespaces during Operator initialization, while function `GetNamespaces` has been renamed to `GetCurrentNamespaceList` to indicate that it obtains to current list of watched namespaces post-Operator initialization.  Otherwise, various other cleanup is included.

  In support of these changes to the `ns` package, the API server functions for validating namespaces, i.e. that a namespace is a watched namespace, have also been updated.  This ensures a single, consolidated method for validating watched namespaces is utilized by the API server, which is able to properly account for the Operator's current namespace Operating mode when determining whether or not a namespace is watched.

- Cluster initialization logic has been updated to ensure the pgBackRest repository is online and ready prior to initializing the primary database container.  This ensures that the pgBackRest repository host is always available and online as needed to fully initialize the database cluster, e.g. 
to create the stanza, initial backup, replicas, etc.  This is especially beneficial in environments where latency can lead to delays in pulling and initializing the pgBackRest repository image, therefore leading to database initialization (and therefore attempts to create the stanza, etc.) prior to the repo coming online.

  To ensure the repo is always initialized prior to initializing the database, the primary deployment is now initially created with 0 replicas.  Then, once the Operator detects that the repo host is ready (specifically via a new repo init handler in the Pod controller), the primary database is scaled to 1. 
 From here cluster creation then proceeds in the same manner as before.

  It should be noted that with this change the repo host also initially started with 0 replicas, but then is immediately scaled up.  This is to ensure that all Deployments and supporting resources (services, configMaps, PVCs, etc.) have been created and are ready to go prior to creating any Pods. This prevent potential race conditions, e.g. the repo host being created and becoming ready prior the creation of the primary DB deployment.  With this solution we can be sure the primary deployment always exists by the time the repo host becomes ready and attempts to scale it to continue initialization of the cluster.

**Checklist:**

 
**Type of Changes:**

 
- The API server does not properly validate watched namespaces when the Operator is using the `disabled` namespace operating mode
- The pgBackRest repo and and primary database are initialized at the same time

[ch7949]
[ch7909]

- The API server properly validates watched namespaces when the Operator is using the "disabled" namespace operating mode
- The pgBackRest repo is initialized prior to initialization of the primary database

**Other information**:

N/A


controller/pod/inithandler.go
```diff
@@ -39,7 +39,18 @@ import (
 // primary PG pod for a new or restored PG cluster reaches a ready status
 func (c *Controller) handleClusterInit(newPod *apiv1.Pod, cluster *crv1.Pgcluster) error {
 
-	clusterName := cluster.Name
+	clusterName := cluster.GetName()
```
jkatz, line 42: Oh that's so smart

PR #1498, opened 2020-05-07
Initial Support for SCRAM-SHA-256

jkatz:
Allow for SCRAM hashed passwords to be set on pgo create/update user

This commit introduces the `--password-type` flag on the
`pgo create user` and `pgo update user` command that allows a user to
specify the following passwords types:

- `""`, `"md5"` => md5
- `"scram"`, `"scram-sha-256"` => SCRAM-SHA-256

In turn, the PostgreSQL Operator will hash the passwords based on the
chosen method and store the computed hash in PostgreSQL.

When using SCRAM support, it is important to note the following
observations and limitations:

- When using one of the password modifications commands on
`pgo update user` (e.g. `--password`, `--rotate-password`, `--expires`)
with the desire to keep the persisted password using SCRAM, it is
necessary to specify the "--password-type=scram-sha-256" directive.

- SCRAM does not work with the current pgBouncer integration with the
PostgreSQL Operator. pgBouncer presently supports only one
password-based authentication type at a time. Additionally, to enable
support for SCRAM, pgBouncer would require a list of plaintext passwords
to be stored in a file that is accessible to it. Future work can
evaluate how to leverage SCRAM support with pgBouncer.

Additional, nonuser facing changes include:

- Moves PostgreSQL password code to `internal` directory
- Creates `PostgresPassword` interface to allow for multiple password generation methods for PostgreSQL
- Move any of the MD5 oriented elements to use the new method
- Add a method for generating the SCRAM verifier
- Add unit tests for said methods

Issue: [ch8179]


internal/pg/postgres/password/scram.go
```diff
@@ -0,0 +1,259 @@
...
+	scramDefaultIterations = 4096
+	// scramDefaultSaltLength is the length of the generated salt used in creating the
+	// hashed password
+	scramDefaultSaltLength = 16
+)
+
+// scramDefaultHash is the hashing algorithm to use
+var scramDefaultHash = sha256.New
+
+// the followig are used as part of the SCRAM verifier generation
```
davecramer, line 61: s/followig/following

internal/pg/postgres/password/scram.go
```diff
@@ -0,0 +1,259 @@
...
+var (
+	scramClientKeyMessage = []byte("Client Key")
+	scramServerKeyMessage = []byte("Server Key")
+)
+
+var (
+	// ErrSCRAMPasswordInvalid is returned when the password attributes are invalid
+	ErrSCRAMPasswordInvalid = errors.New(`invalid password attributes. must "password"`)
```
davecramer, line 69: I don't have the context but this error message doesn't seem to be grammatically correct?

internal/pg/postgres/password/scram.go
```diff
@@ -0,0 +1,259 @@
...
+	// ErrSCRAMSaltLengthInvalid is returned when the salt length is less than 1
+	ErrSCRAMSaltLengthInvalid = errors.New(`salt length must be at least 1`)
+)
+
+// SCRAMPassword contains the building blogs to build a PostgreSQL SCRAM
```
davecramer, line 74: s/blogs/blocks ?

internal/pg/postgres/password/scram.go
```diff
@@ -0,0 +1,259 @@
...
+// verifier. Implements the PostgresPassword interface
+type SCRAMPassword struct {
+	// Hash is the hash algorithm to use. Currently the only thing support in
```
davecramer, line 77: s/support/supported ?

operator/cluster/upgrade.go
```diff
@@ -129,7 +129,7 @@ func AddUpgrade(clientset *kubernetes.Clientset, restclient *rest.RESTClient, up
 	if err != nil {
 		log.Errorf("error submitting upgraded pgcluster CRD for cluster recreation of cluster %s, Error: %v", pgcluster.Name, err)
 	} else {
-		log.Debugf("upgraded cluster %s submitted for recreation, workflowid: %s", pgcluster.Name)
+		log.Debugf("upgraded cluster %s submitted for recreation", pgcluster.Name)
```
davecramer, line 132: s/recreation/re-creation

internal/pg/postgres/password/password.go
```diff
@@ -0,0 +1,77 @@
...
+	var postgresPassword PostgresPassword
+
+	switch passwordType {
+	case MD5:
+		postgresPassword = &MD5Password{}
+		postgresPassword.setAttributes(username, password)
+	case SCRAM:
+		postgresPassword = &SCRAMPassword{}
+		postgresPassword.setAttributes(password)
+	}
```
davecramer, line 69: My preference when coding would be use a default and return immediately out of the switch statement. it's easier to read and explicit, it also avoids the second check for nil. However I'm not sure what our coding convention is.

jkatz: Yeah, that's a good suggestion. This was my first foray into using Go interfaces in this way, so mostly was tryin gto ensure it worked. This would certianly make it more readable and "quicker to fail" so to speak.

operator/cluster/upgrade.go
jkatz: This was part of the original description, not going to change it for now.

apiservermsgs/usermsgs.go
```diff
@@ -25,6 +31,21 @@ const (
...
 
+var (
+	// ErrPasswordTypeInvalid is used when a string that's not included in
+	// PasswordTypeStrings is used
+	ErrPasswordTypeInvalid = errors.New("invalid password type. choices are (md5, scram-sha-256)")
+)
+
+// PasswordTypeStrings is a mapping of strings of password types to their
+// corresponding value of the structured password type
+var PasswordTypeStrings = map[string]pgpassword.PasswordType{
```
cbandy, line 42: This could be unexported, maybe. All the callers go through `GetPasswordType()`.

internal/pg/postgres/password/md5.go
```diff
@@ -0,0 +1,72 @@
...
+	if len(attr) != 2 {
+		return ErrMD5PasswordInvalid
+	}
+
+	// otherwise, set the username and password
+	m.username = attr[0]
+	m.password = attr[1]
+
+	return nil
+}
```
cbandy, line 72: This seems like it wants to be a constructor but it's also unexported. The caller can instead:

```go
MD5Password{username: "x", password: "y"}
```

internal/pg/postgres/password/scram.go
```diff
@@ -0,0 +1,259 @@
...
+	storedKey := s.hash(scramDefaultHash, clientKey)
+
+	// finally, we can build the scram verified!
+	verifier := fmt.Sprintf(scramVerifierFormat,
+		s.Iterations, s.encode(salt), s.encode(storedKey), s.encode(serverKey))
+
+	return verifier, nil
+}
+
+// encode creates a baseb64 encoding of a value that's returned as a string
```
abrightwell, line 131: `base64`

internal/pg/postgres/password/password.go
```diff
@@ -0,0 +1,75 @@
...
+	// Build transforms any plaintext or other attributes that are provided by
+	// the interface implementor and returns a password. If an error occurs in the
+	// building process, it is returned.
+	//
+	// If the build does error, return an empty string
+	Build() (string, error)
+	//
+	// setAttributes sets the key credential attributes (e.g. username, password)
+	// that are used to compute the password
+	setAttributes(attr ...string) error
```
cbandy, line 51: This is never called. That is, there are calls to something named `setAttributes()` but those are not on this interface type.

internal/pg/postgres/password/scram.go
```diff
@@ -0,0 +1,259 @@
...
+		return s.password
+	} else {
+		return cleanedPassword
+	}
+}
+
+// setAttributes accepts just a plaintext password, and uses that to set the
+// password attribute. This also sets the default number of PBKDF2 iterations
+// and generated salt length
+func (s *SCRAMPassword) setAttributes(attr ...string) error {
```
cbandy, line 201: This also seems like it should be a constructor.

abrightwell, line 201: Is there a reason for this to be a variadic function?

internal/pg/postgres/password/scram.go
```diff
@@ -0,0 +1,259 @@
...
+	Iterations int
+	// SaltLength is the length of the generated salt that is used as part of the
+	// PBKDF2 algorithm
+	SaltLength int
+	// password is the plaintext password. This is really the most important
+	// attribute
+	password string
+	// saltGenerator is a function that is used to generate a salt. This can be
+	// mocked for testing purposes
+	saltGenerator saltGenerator
```
cbandy, line 92: It's probably simpler to use `saltGenerator func(length int) ([]byte, error)` here.

The mocks can be inline or variables:
```go
happy := func (int) ([]byte, error) {
  return []byte("h1pp0p4rty2020"), nil
}

sad := func (int) ([]byte, error) {
  return nil, ErrSCRAMSaltLengthInvalid
}
```

The constructor can inline the current crypto implementation:
```go
s.saltGenerator = func(length int) ([]byte, error) {
	if length < 1 {
		return []byte{}, ErrSCRAMSaltLengthInvalid
	}

	// create a salt of size length. The slice needs to be allocated as the
	// crypto random number generator copies the byte values into it
	salt := make([]byte, length)
	_, err := rand.Read(salt)
	return salt, err
}
```

internal/pg/doc.go
```diff
@@ -0,0 +1,20 @@
+// package pg is a collection of resources that interact with the various
+// pieces of software in the PostgreSQL ecosystem (PostgreSQL itself, pgBouncer,
+// pgBackRest etc.)
+
+package pg
```
cbandy, line 5: I don't like this package. `internal/postgres` seems fine to me.

internal/pg/postgres/password/scram.go
jkatz: To match the interface. MD5 takes `username` + `password` whereas SCRAM only needs `password`.

I'm open to a better suggestion.

internal/pg/postgres/password/md5.go
```diff
@@ -0,0 +1,72 @@
...
+func (m *MD5Password) Build() (string, error) {
+	// create the plaintext password/salt that PostgreSQL expects as a byte string
+	plaintext := []byte(fmt.Sprintf("%s%s", m.password, m.username))
+	// set up the password hasher
+	hasher := md5.New()
+	// add the above plaintext to the hash
+	hasher.Write(plaintext)
+	// finish the transformation by getting the string value of the MD5 hash and
+	// encoding it in hexadecimal for PostgreSQL, appending "md5" to the front
+	return fmt.Sprintf("md5%s", hex.EncodeToString(hasher.Sum(nil))), nil
```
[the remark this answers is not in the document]
jkatz: That's magic.

internal/pg/postgres/password/scram.go
jkatz: I was trying to do something like this, but I kept getting panics, likely because I was setting it up incorrectly.

internal/pg/doc.go
jkatz: @cbandy Hm. Per the discussion we had, this is going to be used to group the things that require interacting amongst PG-related services, e.g.:

- postgres
- pgadmin
- pgbouncer
- pgbackrest

et al. Specifically, functionality that requires a direct interaction.

The issue at hand is that there are some things that may be mixed, e.g. pgBouncer leverages some of the pg/password stuff.

I think the grouping shows that there is a potential coupling between them, vs. having it at a higher level.

internal/pg/postgres/password/scram.go
jkatz: Per @cbandy comment, I'm going to remove this and just use a constructor.

internal/pg/postgres/password/scram.go
jkatz: Alright got it correct now. I may have been messing up the initialization.

internal/pg/doc.go
jkatz: Conceded this and moved the directory up a level.

internal/postgres/password/md5.go
```diff
@@ -0,0 +1,53 @@
...
+
+// Build creates the MD5 password format for PostgreSQL which resembles
+// "md5" + md5("password" + "username")
+func (m *MD5Password) Build() (string, error) {
+	// create the plaintext password/salt that PostgreSQL expects as a byte string
+	plaintext := []byte(m.password + m.username)
+	// set up the password hasher
+	hasher := md5.New()
+	// add the above plaintext to the hash
+	hasher.Write(plaintext)
```
cbandy, line 49: `hasher` isn't read from anymore.

internal/postgres/password/password_test.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+		})
+
+		t.Run("invalid", func(t *testing.T) {
+			passwordType := PasswordType(-1)
+
+			if _, err := NewPostgresPassword(passwordType, username, password); err != ErrPasswordType {
+				t.Errorf("expected error: %q", err.Error())
+			}
+		})
```
cbandy, line 87: This is the same test as L42.

It looks like everything can shift up a level and the `valid` tests can become the password type:

```go
func TestNewPostgresPassword(t *testing.T) {
	username := "hippo"
	password := "datalake"

	t.Run("scram",   func(t *testing.T) { passwordType := SCRAM; ... }
	t.Run("md5",     func(t *testing.T) { passwordType := MD5; ... }
	t.Run("invalid", func(t *testing.T) { passwordType := PasswordType(-1); ... }
}
```

internal/postgres/password/scram.go
```diff
@@ -0,0 +1,225 @@
...
+	// ErrSCRAMSaltLengthInvalid is returned when the salt length is less than 1
+	ErrSCRAMSaltLengthInvalid = errors.New(`salt length must be at least 1`)
+)
+
+// SCRAMPassword contains the building blocks to build a PostgreSQL SCRAM
+// verifier. Implements the PostgresPassword interface
+type SCRAMPassword struct {
+	// Hash is the hash algorithm to use. Currently the only thing supported in
+	// PostgreSQL is SHA-256
+	Hash func() hash.Hash
```
cbandy, line 79: This looks unused. `SCRAMPassword.hash` and `SCRAMPassword.hmac` take an argument now.

jkatz: I think this was an attempt to "future proof" for new hashing methods, but in the end, it did not work out all too well. Perhaps in the future when it's needed.

PR #1511, opened 2020-05-12
Introduce the Restart Endpoint & Reload Cleanup

andrewlecuyer:
This commit introduces the `restart` endpoint into the Operator REST API, along with an associated `pgo restart cluster` command in the pgo client.  This allows users to perform a PostgreSQL restart 
on one or more instances (i.e. primary or replicas) within a PG cluster.  Specifically, users can either restart all instances at the same time using the following command:

```bash
pgo restart mycluster
```

Or one or more targets can be specified using the --target flag
(following similar semantics to the 'failover' and 'scaledown'
commands):

```bash
pgo restart mycluster --target=mycluster-abcd
```

The restart itself is performed by calling the Patroni `restart` REST endpoint on the specific instance (primary or replica) being restarted.

As with `failover` and `scaledown` commands it is also possible to specify the `--query` flag to query instances available for restart.

```bash
pgo restart mycluster --query
```

And the output will display whether or not a restart is needed via
a `PENDING RESTART` column:

```bash
Cluster: mycluster2
INSTANCE                ROLE     ...  PENDING RESTART
mycluster2              primary  ...             true
mycluster2-ambq         replica  ...             true
````

It should be noted that the `--query` flag for the `failover` and `scaledown` commands now also include the `PENDING RESTART` information, which is now returned with any replication information (the primary can now also be obtained when retrieving replication information in support of the 'restart' use case).

Also included in this commit are automated end-to-end tests that test the following restart capabilities:

- Querying for restart targets
- Restart all instances in a cluster
- Restarting a specific instance within a cluster using the `--target` flag

And finally, this commit also updates the `reload` endpoint.  Specifically, running the `pgo reload` command now properly reloads all instances (i.e. the primary and all replicas) within the cluster. Also, if more than one cluster is specified for a reload attempt, and an error is encountered while 
attempting a reload for any of those cluster, the endpoint will simply capture the error and move onto the restart attempt for the next cluster, instead of abruptly halting all processing and immediately returning an error.  The error message returned then includes one or more errors depending on the number clusters that encountered errors during the reload attempt.  While ideally multiple errors would be returned in this scenario (instead of multiple errors combined in a single error message, each separated by a new line as seen here), this allows the 'reload' endpoint to react properly when errors are encountered (i.e. not abruptly halt processing on the first error) 
 without breaking the current API contract for the `reload` endpoint (which defines a single error returned in an error scenario).

**Checklist:**

 
**Type of Changes:**

 
- A PostgreSQL restart must be performed by using `kubectl exec` to access a PG pod (primary or replica) and run the `patronictl restart` command.
- The reload endpoint abruptly exits after an error is encountered when reloading a specific cluster, and does not attempt to reload any other clusters specified.
- Only the primary is reloaded when  `pgo reload` is utilized.

[ch8181]

- The Operator REST API and pgo CLI can be utilized to perform a PostgreSQL restart command, e.g. `pgo restart mycluster`.
- The reload endpoint no longer abruptly exits after an error is encountered when reloading a specific cluster, and instead attempts to reload any other cluster specified.
- All instances in a cluster are reloaded when  `pgo reload` is utilized.

**Other information**:
N/A

### CODE REVIEW

File: operator/patroni/patroni.go [hunk 1]
```diff
@@ -0,0 +1,225 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	"github.com/crunchydata/postgres-operator/config"
+	"github.com/crunchydata/postgres-operator/kubeapi"
+)
+
+// dbContainerName is the name of the container containing the PG database in a PG (primary or
+// replica) pod
+const dbContainerName = "database"
```
jkatz, line 34: This probably needs to be a bit more of a global constant, knowing that I wrote some code that references this and like am doing a poor job at it.

Also does not need to be addressed now.

pgo/cmd/restart.go
```diff
@@ -0,0 +1,176 @@
...
+		} else {
+			if Query {
+				queryRestart(args, Namespace)
+			} else if len(args) > 1 {
+				fmt.Println("Error: a single cluster must be specified when performing a restart")
+			} else if util.AskForConfirmation(NoPrompt, "") {
+				restart(args[0], Namespace)
+			} else {
+				fmt.Println("Aborting...")
+			}
```
jkatz, line 68: I think a switch statement would be nicer here.

```suggestion
			switch {
			default:
				fmt.Println("Aborting...")
			case Query:
				queryRestart(args, Namespace)
			case len(args) > 1:
				fmt.Println("Error: a single cluster must be specified when performing a restart")
			case util.AskForConfirmation(NoPrompt, ""):
				restart(args[0], Namespace)
			}
```

apiserver/restartservice/restartimpl.go
```diff
@@ -0,0 +1,161 @@
...
+			return resp
+		}
+	} else {
+		restartResults, err = patroniClient.RestartCluster()
+		if err != nil {
+			resp.Status.Code = msgs.Error
+			resp.Status.Msg = err.Error()
+			return resp
+		}
+	}
```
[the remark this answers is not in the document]
andrewlecuyer: Shortened per recommendation.

pgo/cmd/restart.go
andrewlecuyer: Switched to a switch.

operator/patroni/patroni.go
andrewlecuyer: Per our conversation, will leave as is for now, knowing constants in general need an overhaul and consolidation.

In the meantime I left a note on our reorg doc to get a conversation going around this.

operator/patroni/patroni.go
```diff
@@ -0,0 +1,225 @@
+package patroni
```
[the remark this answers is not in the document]
andrewlecuyer: Went ahead and moved to `internal/patroni/patroni.go`.

PR #1519, opened 2020-05-13
Allow for fsGroup hardcode to be disabled

jkatz:


jkatz:
@NiharDudam This is already handled in #1519:

https://github.com/CrunchyData/postgres-operator/pull/1519/files#diff-55044247848994beb55493f834bfcf95

ruairi-hayes:
Sweet @jkatz , we'll give this a sniff test to make sure it allows and works with the Openshift `restricted` SCC

PR #1534, opened 2020-05-15
Updates command to use pgo-client

jmckulk:
The command to exec into the pgo-client pod was incorrect and this update corrects it

### DIFF
--- a/docs/content/installation/pgo-client.md
+++ b/docs/content/installation/pgo-client.md
@@ -161,11 +161,11 @@ pod. You can run single commands with the kubectl or oc command line tools
 or multiple commands by exec'ing into the pod with bash.
 
 ```
-kubectl exec -it -n pgo <pgo-client-deployment-name> -c "pgo version"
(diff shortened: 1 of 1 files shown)


docs/content/installation/pgo-client.md
```diff
@@ -161,11 +161,11 @@ pod. You can run single commands with the kubectl or oc command line tools
 or multiple commands by exec'ing into the pod with bash.
 
 ```
-kubectl exec -it -n pgo <pgo-client-deployment-name> -c "pgo version"
+kubectl exec -it -n pgo pgo-client-<pod-id> -- pgo version
```
cbandy, line 164: This can actually work by referring to the deploy:

```sh
kubectl exec -it -n pgo deploy/pgo-client -- pgo version
```

jmckulk: I was not aware. thats awesome! I made that change but it ended up being two commits. I can go in and squash them in the cli unless there is a way to do that through GitHub.

cbandy: I can squash at merge.

PR #1536, opened 2020-05-15
Attach image pull secrets to client deployment

cbandy:
**Checklist:**

 
**Type of Changes:**

 
The client pod will not run when its image registry requires credentials.

Use the pull secret configured for that image prefix.

### DIFF
--- a/installers/ansible/roles/pgo-operator/templates/pgo-client.json.j2
+++ b/installers/ansible/roles/pgo-operator/templates/pgo-client.json.j2
@@ -24,6 +24,11 @@
                 }
             },
             "spec": {
+                {% if pgo_image_pull_secret %}
+                "imagePullSecrets": [
+                    { "name": "{{ pgo_image_pull_secret }}" }
+                ],
+                {% endif %}
                 "containers": [
                     {
                         "name": "pgo",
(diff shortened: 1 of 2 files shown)


installers/ansible/roles/pgo-operator/templates/pgo-client.json.j2
```diff
@@ -24,6 +24,11 @@
                 }
             },
             "spec": {
+                "imagePullSecrets": [
+{% if pgo_image_pull_secret %}
+                    { "name": "{{ pgo_image_pull_secret }}" }
+{% endif %}
+                ],
```
jkatz, line 31: Two things:

1. Why not wrap the whole `{% if %}` block around the `"imagePullSecrets"` section? It seems like it's either we have it, or we don't.

1. I'd recommend aligning the `{% if %}` statement with that of the text block.  I think it's easier to read, and typically how I've written/seen Jinja2 style templates.

cbandy: > Two things:
> 
> 1. Why not wrap the whole `{% if %}` block around the `"imagePullSecrets"` section? It seems like it's either we have it, or we don't.

It's like this in other places, probably because we have more than one image/secret sometimes. Not here, though. Done.

> 2. I'd recommend aligning the `{% if %}` statement with that of the text block.  I think it's easier to read, and typically how I've written/seen Jinja2 style templates.

Done.

PR #1540, opened 2020-05-18
Dynamic RBAC for "disabled" & "readonly" Namespace Modes & RBAC Auto-Healing

andrewlecuyer:
With this change, those using `disabled` and `readonly` namespace modes can now have the Operator create the required RBAC in a target namespace.  Specifically, if during initialization the Operator detects that it has the privileges required to create the RBAC it requires (ServiceAccounts, Roles and RoleBindings) within a target namespace to allow it to work properly with that namespace, it will go ahead and reconcile that RBAC.  In other words, it will create the required RBAC if it doesn't exist, or it will recreate it if it exists but is invalid.  And now when a new ControllerGroup is created and started for or a specific namespace, e.g. when a new target namespace is added post-installation, the Operator will also reconcile any required RBAC within that namespace (assuming it has the privileges to do so).

Additionally, on each refresh interval for the namespace controller, the required RBAC in all targeted namespaces will now also be reconciled.  This allows the Operator to self-heal it's required RBAC when it has the ability to do so, therefore ensuring it has the RBAC required to function properly.

When using the `disabled` namespace mode, a namespace controller is now also created in support of RBAC reconciliation, providing the ability to reconcile RBAC on each refresh interval even when the Operator does not have namespace privileges. This is done by creating a fake `clientset`, populated with only those namespaces specified via the NAMESPACE environment variable during Operator installation.  Therefore, while the namespace controller does not truly have access to namespace resources within the Kubernetes cluster while in this mode because it lacks the proper privileges (and therefore cannot react to namespace creations, updates and deletions), this allows it to leverage the same reconciliation logic as when using `dynamic` or `readonly` modes, and therefore reconcile RBAC while in a `disabled` mode (assuming it has the privileges to do so).  This also allows the  Operator to properly respond to the loss of any RBAC required to run various informers used by the Operator's various controllers, even when in a "disabled" mode (this capability is discussed further below).

To leverage the ability to reconcile RBAC when using the `disabled` or `readonly` operating modes, the Operator ServiceAccount must be assigned the privileges within any targeted namespace required to create it's  required RBAC resources within that namespace.  A new
`add-targeted-namespace-dynamic-rbac.sh` script has been provided as a means to create a new namespace that has these privileges.  However, unlike the current `add-targeted-namespace.sh` script, which creates the actual ServiceAccounts, Roles and RoleBindings within a namespace so that the Operator can utilize that namespace to create PostgreSQL clusters, this script creates a Role and RoleBinding within the namespace that allows the Operator itself to create and reconcile the ServiceAccounts, Roles and RoleBindings it requires.

Both the Bash and Ansible installers have also been updated in support of this feature.  When using Bash, if `PGO_DYNAMIC_RBAC=true`, then the `add-targeted-namespace-dynamic-rbac.sh` script is run to create targeted namespaces (instead of `add-targeted-namespace.sh`), ensuring the Operator then has permissions to create it's required RBAC within those namespaces per the details discussed above.  This same principle applies to the Ansible installer to, specifically producing the same results if `pgo_dynamic_rbac='true'` in the Ansible `inventory` file.

Finally, with this commit error handling for the various informers utilized  across the Operator's various controllers has also been updated. Specifically, now when starting a new controller group containing the various controllers required for the Operator to create PG clusters within a target namespace, a check is first done to ensure the Operator has the proper privileges as required to run the various informers in those controllers (specifically the required `list` privileges).  If this check fails, starting of the controller group is aborted and then retried on the next refresh interval for the namespace controller.

Additionally, now on each refresh interval if the Operator detects that it no longer has the privileges to properly run all controllers, the controller group for that namespace will be stopped and removed, after which an attempt will be made to start it again on the next namespace refresh interval (where it will again fail, and the controller group will not be started, if it still does not have the proper privileges). The changes prevent the informers from overwhelming the Operator logs with a huge volume of privilege error messages in the event any required privileges are lost.

Is should also be noted that as of this commit the  Ansible and Bash installers no longer create namespaces when using the `dynamic` namespace Operating mode.  Instead, the Operator itself will handle namespace creation and initialization when in this mode.  At this point users on Kube 1.12 or less (e.g. OpenShift 3.11 users) should utilize either the "readonly" or "disabled" namespace 
Operating modes, since the `dynamic` is not fully/properly supported with these Kube versions.

And finally, this commit also includes the following:
- Worker queues have been added to the Namespace controller, and both the Operator and Scheduler have been updated accordingly
- Requeuing has been removed from both the ConfigMap controller to prevent a large volume of error under certain circumstances, i.e. if required RBAC is missing.  Instead the action will simply be retried on the next refresh interval.

**Checklist:**

 
**Type of Changes:**

 
- All RBAC (ServiceAccounts, Roles and RoleBindings) required for the Operator to create PostgreSQL clusters within a target namespace must be explicitly created by a user within that namespace.
- If assigned the proper privileges, the Operator does not reconcile RBAC within a target namespace (i.e. create it if missing, or updates it if invalid).
- When controllers are started for a specific namespace, it is assumed the controller has the proper privileges within that namespace.
- If a running controller looses the RBAC it requires, no action is taken.
- The namespace contollers do not utilize worker queues.

[ch8257]

- The Operator can be assigned the privileges within a target namespace required for the Operator itself to create the RBAC (ServiceAccounts, Roles and RoleBindings) it requires to create PostgreSQL clusters within that namespace.
- If assigned the proper privileges, the Operator now reconciles RBAC within a target namespace (i.e. creates it if missing, or updates it if invalid).
- The Operator will verify it has the `lister` privileges needed to start a controller, only starting it if all required privileges can be requried.
- If a running controller looses the RBAC it requires, the controller is stopped (and will attempt to restart on the next namespace controller refresh interval).
- The namespace contollers utilize worker queues, allowing for multiple workers.

**Other information**:

N/A


controller/manager/controllermanager.go
```diff
@@ -315,16 +372,133 @@ func (c *ControllerManager) addControllerGroup(namespace string) error {
...
+
+	for _, listerResource := range listerResourcesCrunchy {
+		hasCrunchyPrivs, err = ns.CheckAccessPrivs(controllerGroup.kubeClientset,
+			map[string][]string{listerResource: []string{"list"}},
+			crv1.GroupName, namespace)
+		if err != nil {
+			log.Errorf(err.Error())
+		} else if !hasCrunchyPrivs {
+			log.Errorf("Controller Manager: Controller Group for namespace %s does not have the "+
+				"required list privileges for resource %s in the Crunchy API",
```
jkatz, line 429: The proper name would be `crunchydata.com` API if we're going for that.

jkatz: Or `crunchydata` per the `vendor` label

pgo-scheduler/pgo-scheduler.go
```diff
@@ -170,6 +172,31 @@ func main() {
...
+	// that is able to resond to namespace events in the Kube cluster.  If using the "disabled"
+	// operating mode, then create a fake client containing all namespaces defined for the install
+	// (i.e. via the NAMESPACE environment variable) and use that to create the namespace
+	// controller.  This allows for namespace and RBAC reconciliation logic to be run in a
+	// consistent manner regardless of the namespace operating mode being utilized.
+	if namespaceOperatingMode != ns.NamespaceOperatingModeDisabled {
+		if err := createAndStartNamespaceController(kubeClient, controllerManager, scheduler,
+			stop); err != nil {
+			log.Error(err)
+			os.Exit(2)
```
jkatz, line 185: I'd rather we do a `log.Fatal` which would trigger an `ox.Exit(1)`

postgres-operator.go
```diff
@@ -97,14 +99,29 @@ func main() {
...
 		if err := createAndStartNamespaceController(kubeClientset, controllerManager,
 			stopCh); err != nil {
 			log.Error(err)
 			os.Exit(2)
 		}
+	} else {
+		fakeClient, err := ns.CreateFakeNamespaceClient(operator.InstallationName)
+		if err != nil {
+			log.Error(err)
+			os.Exit(2)
```
jkatz, line 118: Similar to `log.Fatal` comment.

controller/manager/controllermanager.go
andrewlecuyer: Replaced with the `GroupName` for our API, i.e. `crunchydata.com`.

pgo-scheduler/pgo-scheduler.go
andrewlecuyer: Agreed, will start using `log.Fatal` instead of of `os.Exit(2)`.

postgres-operator.go
andrewlecuyer: Switched to `log.Fatal`.

PR #1561, opened 2020-05-20
Fix RELATED_IMAGE overrides for containers in cluster deployment

jkatz:
The RELATED_IMAGE variables for containers in the cluster deployment
where not being respected. This commit fixes it to ensure that
they can be overriden.

Issue: [ch8302]

### DIFF
--- a/operator/clusterutilities_test.go
+++ b/operator/clusterutilities_test.go
@@ -1,0 +1,170 @@
+package operator
+
+/*
+ Copyright 2020 Crunchy Data Solutions, Inc.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


operator/clusterutilities_test.go
```diff
@@ -0,0 +1,174 @@
...
+		"database":   {name: "database", image: config.CONTAINER_IMAGE_CRUNCHY_POSTGRES_HA},
+		"crunchyadm": {name: "crunchyadm", image: config.CONTAINER_IMAGE_CRUNCHY_ADMIN},
+		"collect":    {name: "collect", image: config.CONTAINER_IMAGE_CRUNCHY_COLLECT},
+		"pgbadger":   {name: "pgbadger", image: config.CONTAINER_IMAGE_CRUNCHY_PGBADGER},
+		"future":     {name: "future", image: "crunchy-future"},
+	}
+
+	// {name: "database", image: "crunchy-postgis-ha"},
+
+	// ContainerImageOverrides["foo"] = "bar"
```
andrewlecuyer: Looks like a couple of commented out lines that just need to be cleaned up.

operator/clusterutilities_test.go
```diff
@@ -0,0 +1,170 @@
...
+					return
+				}
+
+				if containerDefault.name == name && containerDefault.image == container.Image {
+					t.Errorf("container %q not overwritten. image name is %q",
+						containerDefault.name, container.Image)
+				}
+			}
+			// unoverride at the end of the test
+			ContainerImageOverrides[defaults.image] = ""
```
andrewlecuyer, line 107: Would just deleting the entry from the map at the end of the test make sense here, e.g.:
```
delete(ContainerImageOverrides, ContainerImageOverrides[defaults.image])
```

PR #1580, opened 2020-05-28
Validate Limit >= Request in some API calls

jkatz:
**Checklist:**

 
**Type of Changes:**

 
There are no validations to check that the limit >= request.

This can be particularly noticeable on a create where one can request
a limit that is less than the server defaults. This provides some
sanity checking.

There are limitations on update: attempting to see if the
request/limit combos are accurate is a bit tedious. The main goal
is to provide a few sanity checks before applying the request/limits
to Deployments.

### DIFF
--- a/apiserver/common.go
+++ b/apiserver/common.go
@@ -17,6 +17,7 @@ limitations under the License.
 
 import (
 	"errors"
+	"fmt"
 	"strconv"
 
 	crv1 "github.com/crunchydata/postgres-operator/apis/crunchydata.com/v1"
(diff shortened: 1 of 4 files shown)


apiserver/common.go
```diff
@@ -125,6 +122,42 @@ func IsValidPVC(pvcName, ns string) bool {
...
+	if !limitQuantity.IsZero() && !requestQuantity.IsZero() && limitQuantity.Cmp(requestQuantity) == -1 {
+		return fmt.Errorf(ErrMessageLimitInvalid, limitQuantity.String(), requestQuantity.String())
+	}
+
+	// if limit and the default are nonzero, but the request is zero, and the
+	// limit is less than the default, return an error
+	if !limitQuantity.IsZero() && requestQuantity.IsZero() && !defaultQuantity.IsZero() &&
+		limitQuantity.Cmp(defaultQuantity) == -1 {
+		return fmt.Errorf(ErrMessageLimitInvalid, limitQuantity.String(), defaultQuantity.String())
+	}
```
cbandy, line 156: This condition is a lot to take in and the comment doesn't explain why it's here. Just from reading, I think these lines can be eliminated by setting request to the default before the check on L147:

```go
	// parse the quantities so we can compare
	requestQuantity, _ := resource.ParseQuantity(request)
	limitQuantity, _ := resource.ParseQuantity(limit)

	if requestQuantity.IsZero() {
		requestQuantity = defaultQuantity
	}

	// if limit and request are nonzero and the limit is less than the request,
	// error
	if !limitQuantity.IsZero() && !requestQuantity.IsZero() && limitQuantity.Cmp(requestQuantity) == -1 {
		return fmt.Errorf(ErrMessageLimitInvalid, limitQuantity.String(), requestQuantity.String())
	}
```

1. My reading could be wrong. Would the existing tests pass with this change? Is this the intended behavior?
2. If so, using the default this way makes sense to me and seems more clear. Is there another reason to keep the current arrangement?

PR #1596, opened 2020-06-02
Pin Ansible version to 2.9.0+

jmckulk:
This update standardizes the Ansible version across the different deployer
images and the documentation. The docs have been updated so that Ansible 2.9.0+
is a prerequisites for the Ansible installer (documented under other install
methods). The gcp-marketplace deployer image has been updated from Ansible 2.8.*
to 2.9.* and the pgo-deployer image has been pinned to 2.9.*.

The image specific environment variables are moved to the Dockerfile. The HOME
and ANSIBLE_CONFIG variables are currently set in the pgo-deploy script. This
causes issues when trying to use ansible without running the script or manually
setting these variables.

**Checklist:**

 
 - [x] Have you tested your changes on all related environments with successful results, as applicable? Built and tested 

**Type of Changes:**

 
**Other information**:

### DIFF
--- a/installers/gcp-marketplace/Dockerfile
+++ b/installers/gcp-marketplace/Dockerfile
@@ -13,11 +13,11 @@ RUN install -D /bin/create_manifests.sh /opt/postgres-operator/cloud-marketplace
 RUN if [ -f /etc/os-release ] && [ debian = "$(. /etc/os-release; echo $ID)" ] && [ 10 -ge "$(. /etc/os-release; echo $VERSION_ID)" ]; then \
       apt-get update && apt-get install -y --no-install-recommends gnupg && rm -rf /var/lib/apt/lists/* && \
       wget -qO- 'https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x93C4A3FD7BB9C367' | apt-key add && \
-      echo > /etc/apt/sources.list.d/ansible.list deb http://ppa.launchpad.net/ansible/ansible-2.8/ubuntu trusty main ; \
+      echo > /etc/apt/sources.list.d/ansible.list deb http://ppa.launchpad.net/ansible/ansible-2.9/ubuntu trusty main ; \
     fi
 
 RUN apt-get update \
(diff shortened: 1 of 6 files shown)


installers/gcp-marketplace/Dockerfile
```diff
@@ -13,11 +13,11 @@ RUN install -D /bin/create_manifests.sh /opt/postgres-operator/cloud-marketplace
 RUN if [ -f /etc/os-release ] && [ debian = "$(. /etc/os-release; echo $ID)" ] && [ 10 -ge "$(. /etc/os-release; echo $VERSION_ID)" ]; then \
```
cbandy, line 13: The link above here is partially broken. It looks like the anchor has changed to:

```
https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html#installing-ansible-on-debian
```

installers/image/bin/pgo-deploy.sh
```diff
@@ -35,9 +35,7 @@ export SERVICE_TYPE=${SERVICE_TYPE:-ClusterIP}
 export METRICS_NAMESPACE=${METRICS_NAMESPACE:-pgo}
 
 # Installer Settings
-export ANSIBLE_CONFIG="/ansible/ansible.cfg"
```
andrewlecuyer, line 38: Just curious why we moved these out of the script into the Dockerfiles (therefore moving from a single place to set, modify, etc. these vars if need be to two).

jmckulk: I noticed that if you exec into the pod without running the script ansible fails because these are not set. I thought that since ansible doesn't work without these being set it would make sense to define the in the image instead of the script. This can definitely be removed considering people should only be running the script but this would get rid of some confusion if anyone is trying to run ansible commands manually.

andrewlecuyer: If it helps with development, testing, etc. of then seems like a fair trade-off (also, can't imagine these vars will change often).  Thanks for explanation.

PR #1643, opened 2020-06-19
Add documentation on how to remove a PostgreSQL tablespace

### LINKED ISSUE #1640: Support to remove tablespace
**What is the motivation or use case for the change? **
I have a system for managing binary documents. There is a need to move documents between partitions/tablespaces. For example I need to move document to new tablespace on faster storage class and then remove old tablespace to remove volume/reduce costs. Adding tablespaces works perfect, but I also need to remove tablespaces. I tried to remove entry from pgCluster CRD but It does nothing - PVC, volumeMount and tablespace still exists. Maybe Is there any mechanism to remove tablespaces / volumes ?

* Where is this running ( Local, Cloud Provider) AWS
* Storage being used (NFS, Hostpath, Gluster, etc): EFS, EBS
* PostgreSQL Version: 4.3.0
* Platform (Docker, Kubernetes, OpenShift): Kubernetes
* Platform Version: v1.15

jkatz:
As deleting a tablespace from a PostgreSQL cluster is nontrivial, this
provides guidance on how to do so, in light of the fact the likelihood
of the Operator automating such a method is fairly slim.

fixes #1640

### DIFF
--- a/docs/content/architecture/tablespaces.md
+++ b/docs/content/architecture/tablespaces.md
@@ -127,6 +127,58 @@ PVC to a Kubernetes Deployment causes the Pods in the deployment to restart.
 When the operation completes, the tablespace will be set up and accessible to
 use within the PostgreSQL cluster.
 
+## Removing Tablespaces
+
+Removing a tablespace is a nontrivial operation. PostgreSQL does not provide a
+`DROP TABLESPACE .. CASCADE` command that would drop any associated objects with
+a tablespace. Additionally, the PostgreSQL documentation covering the
+[`DROP TABLESPACE`](https://www.postgresql.org/docs/current/sql-droptablespace.html)
+command goes on to note:
+
+> A tablespace can only be dropped by its owner or a superuser. The tablespace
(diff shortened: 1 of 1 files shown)


docs/content/architecture/tablespaces.md
```diff
@@ -127,6 +127,58 @@ PVC to a Kubernetes Deployment causes the Pods in the deployment to restart.
...
+> the DROP might fail due to temporary files residing in the tablespace.
+
+Because of this, and to avoid a situation where a PostgreSQL cluster is left in
+an inconsistent state due to trying to remove a tablespace, the PostgreSQL
+Operator does not provide any means to remove tablespaces automatically. If you
+do need to remove a tablespace from a PostgreSQL deployment, we recommend
+following this procedure:
+
+1. As a database administrator:
+  1. Log into the primary instance of your cluster.
```
cbandy, line 152: GitHub doesn't render this as a new list/ordering. Perhaps indent more?

jkatz: But hugo does.

PR #1649, opened 2020-06-21
Remove 'bind' & 'escalate' Verb Requirements for RBAC Reconciliation

andrewlecuyer:
This commit removes the requirement to assign the PostgreSQL Operator ServiceAccount `bind` and `escalate` permissions in order to have the PostgreSQL Operator reconcile RBAC (i.e. create, update and or delete its own ServiceAccounts, Roles and RoleBindings) within any targeted namespaces.  More specifically, the Operator is no longer concerned with exactly how it has been assigned the permissions needed to reconcile RBAC, and will now now always attempt to reconcile RBAC in any targeted namespaces regardless of how any required permissions were assigned.

For users that do not wish to have the Operator reconcile RBAC in target namespaces, this capability can be disabled via the 'DisableReconcileRBAC' setting in the `pgo.yaml` configuration file.  When RBAC reconciliation is disabled, all RBAC must be created manually by a Kuberentes administrator (since the Operator will not attempt to do so itself).

In support of this change, the various installation methods supported by the PostgreSQL Operator have been updated to enable RBAC reconciliation by default.  Additionally, the various configuration settings, descriptions, etc. for the "Dynamic RBAC" capability have been renamed to "Reconcile RBAC" to better align with similar functionality and terminology within Kubernetes (e.g. the `kubectl auth reconcile` command, which provides similar functionality).  The updated settings for each installer are as follows (with each defaulting to `true`):

- kubectl: `RECONCILE_RBAC` (previsouly `DYNAMIC_RBAC`)
- Ansible: `reconcile_rbac` (previously `dynamic_rbac`)
- Bash: `PGO_RECONCILE_RBAC` (previously `PGO_DYNAMIC_RBAC`)

When the various "Reconcile RBAC" installer settings are set to true (as they are by default), the installer will configure the PostgreSQL Operator with the proper permissions needed to reconcile RBAC based on the Namespace Operating Mode configured for the installation.  For instance, when a `dynamic` Namespace Operating Mode is configured, a ClusterRole is installed that allows the Operator to reconcile RBAC in any namespace (present or future) within the Kubernetes cluster. 
 When a `readonly` or `disabled` Namespace Operating Mode is configured, the Roles are installed in each target namespace as needed to manage RBAC in those specific namespaces (therefore, while in `readonly` or `disabled` modes the RBAC required by the Operator to reconcile RBAC in any new namespaces must be manually created by an administrator).

In support of the above changes, the 'Namespace' section of the documentation has been updated, along with the applicable settings in the various installer and configuration documentation.  This includes defining RBAC required to allow the Operator to successfully reconcile RBAC in targeted namespaces.  Additionally, this commit removes all remaining wildcards (`*`) from PostgreSQL Operator Role definitions.

And finally, this commit includes various refactoring for the namespace controller and the associated Operator Controller Manager, specifically to reduce duplication of various RBAC logic, improve error handling and logging, and overall simplify and better structure/organize these portions of the Operator code.

**Checklist:**

 
**Type of Changes:**

 
The `bind` and `escalate` permissions are required to reconcile RBAC in target namespaces.

[ch8366]

The `bind` and `escalate` permissions are no longer required to reconcile RBAC in target namespaces.

**Other information**:

N/A


docs/content/architecture/namespace.md
```diff
@@ -21,28 +21,26 @@ deploy the PostgreSQL Operator.
...
-namespace events (e.g. namespace additions, updates and deletions), and then create or remove
-controllers for various namespaces as those namespaces are added or removed from the Kubernetes
-cluster and/or Operator install.  The mode therefore allows the Operator to dynamically respond
-to namespace events in the cluster, and then interact with those namespaces as required to manage
-PostgreSQL clusters within them.
+any namespaces within the Kubernetes cluster.  Additionally, while  in this mode the Operator can
+listen for namespace events (e.g. namespace additions, updates and deletions), and then create or 
+remove controllers for various namespaces as those namespaces are added or removed from the 
+Kubernetes cluster and/or Operator install.  This mode therefore allows the Operator to dynamically
+respond to namespace events in the cluster.
```
jkatz, line 41:
```markdown
any namespaces within a Kubernetes cluster.  With `dynamic` mode enabled, the PostgreSQL Operator can respond to namespace events in a Kubernetes cluster, such as when a namespace is created, and take an appropriate action, such as adding the PostgreSQL Operator controllers to a newly created namespace.
```

docs/content/architecture/namespace.md
```diff
@@ -62,146 +60,302 @@ rules:
...
+```
+
+### `readonly`
+
+While in this mode the PostgreSQL Operator is still able to listen for  namespace events within the
+Kubernetetes cluster, and then create and run and/or remove controllers as namespaces are added,
+updated and deleted.  Therefore, the Operator is able to dynamically respond to namespace events in 
+the Kubernetes cluster.  However, while in this mode the Operator is unable to create, delete or
+update namespaces itself, which means these actions must be performed externally by an
+administrator.
```
jkatz, line 72: In `readonly` mode, the PostgreSQL Operator is still able to listen to namespace events within a Kubernetes cluster, but it can no longer modify (create, update, delete) namespaces. For example, if a Kubernetes administrator creates a namespace, the PostgreSQL Operator can respond and add its controllers to that namespace.

docs/content/architecture/namespace.md
```diff
@@ -62,146 +60,302 @@ rules:
...
+```
+
+### `disabled`
+
+Disables namespace capabilities within the PostgreSQL Operator altogether.  While in this mode the
+PostgreSQL Operator will simply attempt to work with the target namespaces specified during
+installation.  If no target namespaces are specified, then the Operator will be configured to work
+within the namespace in which it is deployed.  Since the Operator is unable to dynamically respond
+to namespace events in  the cluster, in the event that target namespaces are deleted or new target
+namespaces need to be added, the PostgreSQL Operator will need to be re-deployed.
```
jkatz, line 99:
```markdown
`disabled` mode disables namespace capabilities...
```

installers/image/inventory_template
```diff
@@ -34,21 +34,22 @@ create_rbac='$CREATE_RBAC'
...
+# then a ClusterRole and ClusterRoleBinding is created allowing the
+# Operator to reconcile its own RBAC in any namespace.  If set to 'true' 
+# and 'namespace_mode' is 'readonly' or 'disabled', then local Role and
+# RoleBindings are created in each target namespace that allow the Operator
+# to reconcile it's RBAC in those specific namespaces.  And finally, if set
+# to 'false', the Operator will not be assigned any permissions to 
+# reconcile its own RBAC, and will therefore not attempt to do so.
 # ==================
-dynamic_rbac='$DYNAMIC_RBAC'
+# reconcile_rbac='false'
```
jkatz, line 52: Wouldn't this be:

`reconcile_rbac='$RECONCILE_RBAC'`

internal/controller/manager/rbac.go
```diff
@@ -0,0 +1,139 @@
...
+func (c *ControllerManager) reconcileRBAC(targetNamespace string) {
+
+	log.Debugf("Controller Manager: Now reconciling RBAC in namespace %s", targetNamespace)
+
+	// Use the image pull secrets of the operator service account in the new namespace.
+	operator, err := c.controllers[targetNamespace].kubeClientset.CoreV1().
+		ServiceAccounts(c.pgoNamespace).Get(ns.OPERATOR_SERVICE_ACCOUNT, metav1.GetOptions{})
+	if err != nil {
+		log.Errorf("%s: %v", ErrReconcileRBAC, err)
+	}
```
jkatz, line 48: Do we want to return here?

internal/controller/manager/rbac.go
```diff
@@ -0,0 +1,139 @@
...
+			if kerrors.IsNotFound(err) {
+				doesNotExist = true
+			} else {
+				log.Errorf("%s: %v", ErrReconcileRBAC, err)
+				continue
+			}
+		}
+
+		if doesNotExist || saCreatedOrUpdated {
+			if err = ns.CopySecret(c.controllers[targetNamespace].kubeClientset, reference.Name,
```
jkatz, line 73: `err :=`

internal/controller/manager/rbac.go
```diff
@@ -0,0 +1,139 @@
...
+		ns.PGO_BACKREST_SERVICE_ACCOUNT: config.PgoBackrestServiceAccountTemplate,
+		ns.PGO_PG_SERVICE_ACCOUNT:       config.PgoPgServiceAccountTemplate,
+	}
+
+	for serviceAccount, template := range reconcileServiceAccounts {
+		createdOrUpdated, err := ns.ReconcileServiceAccount(c.controllers[targetNamespace].kubeClientset,
+			serviceAccount, targetNamespace, template, imagePullSecrets)
+		if err != nil {
+			log.Errorf("%s: %v", ErrReconcileRBAC, err)
+		}
```
jkatz, line 133: Should we `continue` here?

installers/ansible/inventory
```diff
@@ -51,21 +51,22 @@ create_rbac='true'
...
+# then a ClusterRole and ClusterRoleBinding is created allowing the
+# Operator to reconcile its own RBAC in any namespace.  If set to 'true' 
+# and 'namespace_mode' is 'readonly' or 'disabled', then local Role and
+# RoleBindings are created in each target namespace that allow the Operator
+# to reconcile it's RBAC in those specific namespaces.  And finally, if set
+# to 'false', the Operator will not be assigned any permissions to 
+# reconcile its own RBAC, and will therefore not attempt to do so.
 # ==================
-# dynamic_rbac='false'
+# reconcile_rbac='false'
```
jkatz, line 69: We default this to `true` so we should probably model this to `true`

internal/controller/manager/rbac.go
andrewlecuyer: Not returning here so that we can still try to reconcile RBAC that isn't dependent on the operator ServiceAccount, e.g. the Roles and RoleBindings.

jkatz: @andrewlecuyer Cool. Can you please add a comment around that?

installers/ansible/inventory
andrewlecuyer: Since it defaults to `true`, I typically think someone explicitly setting this setting  will be disabling it, which is why I used `false`.  However, I'm good either way so will go ahead and set to the `true`.

internal/controller/manager/rbac.go
andrewlecuyer: Sure thing!

deploy/install-rbac.sh
```diff
@@ -46,27 +46,32 @@ fi
...
-	envsubst < $DIR/cluster-roles.yaml | $PGO_CMD create -f -
+	
+	if [[ "${PGO_RECONCILE_RBAC:-true}" == "true" ]]
+	then
+		# create the full ClusterRole with namespace and RBAC permissions if RBAC reconciliation 
+		# is enabled
+		envsubst < $DIR/cluster-roles.yaml | $PGO_CMD create -f -
+	else
+		# create a ClusterRole with only namespace permissions if RBAC reconciliation is disabled
+		sed '/- delete/q' cluster-roles.yaml | $PGO_CMD create -f -
```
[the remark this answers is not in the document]
andrewlecuyer: Added `$DIR`.

docs/content/architecture/namespace.md
```diff
@@ -21,28 +21,26 @@ deploy the PostgreSQL Operator.
...
 determine what cluster roles are currently assigned to the `pgo-operator` `ServiceAccount`
 (i.e. the `ServiceAccount` running the Pod the PostgreSQL Operator is deployed within).  Based
 on the `ClusterRoles` identified, one of the namespace operating modes described below will be
-enabled for the Operator installation.  Please consult the installation guides for the various
-installation methods available to determine the settings required to install the `ClusterRoles`
-required for each mode.
+enabled for the PostgreSQL Operator installation.  Please consult the
+[installation guides]({{< relref "installation" >}}) for the various installation methods
+available to determine proper setting for enabling each of the Namespace Operating Modes discussed
+below.
```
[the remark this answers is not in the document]
andrewlecuyer: Updated per recommendation.

docs/content/architecture/namespace.md
andrewlecuyer: Updated per recommendation.

docs/content/architecture/namespace.md
andrewlecuyer: Updated per recommendation.

installers/image/inventory_template
andrewlecuyer: Thanks!

internal/controller/manager/rbac.go
andrewlecuyer: Updated.

internal/controller/manager/rbac.go
andrewlecuyer: `continue` added.

installers/ansible/inventory
andrewlecuyer: Changed to `true`.

docs/content/architecture/namespace.md
andrewlecuyer: Updated per recommendation.

andrewlecuyer:
Thanks for the review @jkatz, all feedback has been addressed.

installers/image/inventory_template
```diff
@@ -34,21 +34,22 @@ create_rbac='$CREATE_RBAC'
...
+# then a ClusterRole and ClusterRoleBinding is created allowing the
+# Operator to reconcile its own RBAC in any namespace.  If set to 'true' 
+# and 'namespace_mode' is 'readonly' or 'disabled', then local Role and
+# RoleBindings are created in each target namespace that allow the Operator
+# to reconcile it's RBAC in those specific namespaces.  And finally, if set
+# to 'false', the Operator will not be assigned any permissions to 
+# reconcile its own RBAC, and will therefore not attempt to do so.
 # ==================
-dynamic_rbac='$DYNAMIC_RBAC'
+reconcile_rbac='$RECONCILE_RBAC'
```
jkatz, line 52: @jmckulk *NOTE* Please account for the variable name change in your PR

installers/kubectl/postgres-operator.yml
```diff
@@ -206,6 +204,8 @@ spec:
                     value: "devtest"
                   - name: PGO_OPERATOR_NAMESPACE
                     value: "pgo"
+                  - name: RECONCILE_RBAC
+                    value: "true"
```
jkatz, line 208: @jmckulk Note the variable name change (`DYNAMIC_RBAC` => `RECONCILE_RBAC`) -- I don't think this affects your PR, but I wanted to make you aware.

internal/ns/nslogic.go
```diff
@@ -259,287 +243,175 @@ func copySecret(clientset *kubernetes.Clientset, secretName, operatorNamespace,
...
-		OPERATOR_SERVICE_ACCOUNT, metav1.GetOptions{})
+	currRole, err := clientset.RbacV1().Roles(targetNamespace).Get(
+		role, metav1.GetOptions{})
 	if err != nil {
-		log.Errorf("expected the operator account to exist: %v", err)
-		errs = append(errs, err.Error())
+		if kerrors.IsNotFound(err) {
+			log.Debugf("Role %s in namespace %s does not exist and will be created",
+				role, targetNamespace)
+			createRole = true
```
cbandy, line 258: Is it possible to treat both create and update the same in these functions? My understanding is that clientset update becomes REST PUT, which I would expect to mean create-or-replace.

If that is so, then the not-found case can fallthrough to the diff which then triggers the update call, reducing some of the branching/complexity here. 🤞

installers/ansible/roles/pgo-operator/tasks/namespace.yml
```diff
@@ -14,13 +14,12 @@
     target_ns_script: "{{ lookup('template', 'add-targeted-namespace.sh.j2') }}"
   with_items: "{{ nslist.split(',') | map('trim') | list }}"
   when:
-    - namespace_mode != 'dynamic'
-    - dynamic_rbac == 'false'
+    - reconcile_rbac == 'false'
```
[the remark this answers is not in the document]
andrewlecuyer: The idea here is that anytime RBAC reconciliation is disabled (i.e. `reconcile_rbac == 'false'`) regardless of namespace mode, then all RBAC resources required by the Operator will need to be manually created in each target namespace, so we go ahead and run `add-targeted-namespace.sh` in each space when installing.

And if `reconcile_rbac` is set to `true` and reconciliation is enabled, then we do not create all of the required RBAC using `add-targeted-namespace.sh`, but rather execute the tasks below to assign the Operator the RBAC it needs into those target namespaces to reconcile its required RBAC (and more specifically, these tasks are executed when namespace mode is `readonly` or `disabled` - otherwise a ClusterRole is created when the namespace mode is `dynamic`).

internal/controller/manager/rbac.go
```diff
@@ -0,0 +1,142 @@
...
+		ns.PGO_BACKREST_ROLE: config.PgoBackrestRoleTemplate,
+		ns.PGO_PG_ROLE:       config.PgoPgRoleTemplate,
+	}
+
+	for role, template := range reconcileRoles {
+		if err := ns.ReconcileRole(c.controllers[targetNamespace].kubeClientset, role,
+			targetNamespace, template); err != nil {
+			log.Errorf("%s: %v", ErrReconcileRBAC, err)
+		}
+	}
```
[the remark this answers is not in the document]
andrewlecuyer: Since its inception, the namespace package has encapsulated no only namespace functionality, but also the functionality for creating various RBAC resources in the.  So traditionally it has taken that holistic view of a namespace definition, which includes not only the namespace itself, but also the RBAC required by the Operator to operate within that namespace.  So that's essentially what you're still seeing here, i.e. the relationship between namespace and RBAC reconciliation.

I can honestly see it from both perspectives.  I feel as though this patch takes some big strides in further decoupling RBAC stuff from namespace stuff, which implies further separating these things.  But RBAC reconciliation is handled/called by the namespace controller, so there is still a relationship with namespace and RBAC at a technical level there.  

As for whether it should be in the controller, I tend to think of the the controller being the caller some other package containing the business logic for whatever action needs to be performed, and then handling the coordination of all actions that need to be performed based on a specific event.  So in this example the `ns` package knows how to reconcile a RBAC resource, and the controller has the domain specific knowledge to tell it what resources to reconcile and coordinate the reconciliation of all RBAC resources.  So maybe eventually we break it out into an `rbac` package?

Honestly, I think there are various ways to look at this things, and as we continue to refactor this code (which has been a bit of a journey) we can continue to evolve some of these pieces as we see fit.  But since we're already making some big changes here in general, I want to be sure to tackle that incrementally.

internal/controller/namespace/namespacecontroller.go
```diff
@@ -133,6 +134,21 @@ func (c *Controller) processNextWorkItem() bool {
 		return true
 	}
 
+	// remove the controller group for the namespace if the namespace no longer exists or is
+	// termininating
+	ns, err := c.namespaceLister.Get(namespace)
+	if (err == nil && ns.Status.Phase == corev1.NamespaceTerminating) ||
+		(err != nil && kerrors.IsNotFound(err)) {
```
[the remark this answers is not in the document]
andrewlecuyer: I'll admit I'm in the habit of always checking for an error before doing anything with that error, even if just calling something like `IsNotFound`.

internal/controller/manager/controllermanager.go
```diff
@@ -136,41 +139,12 @@ func (c *ControllerManager) AddGroup(namespace string) error {
...
-		// now reconcile RBAC in the namespace if allowed
-		if canCreateRBACInNamespace {
-			if err := ns.ReconcileTargetRBAC(c.controllers[namespace].kubeClientset, c.pgoNamespace,
-				namespace); err != nil {
-				return err
-			}
-		}
+	if c.controllers[namespace] != nil && !c.pgoConfig.Pgo.DisableReconcileRBAC {
+		// first reconcile RBAC in the target namespace if RBAC reconciliation is enabled
+		c.reconcileRBAC(namespace)
```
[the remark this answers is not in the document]
andrewlecuyer: This will be called on each namespace refresh interval to ensure all RBAC is properly reconciled.

internal/controller/manager/controllermanager.go
```diff
@@ -502,6 +424,14 @@ func (c *ControllerManager) runControllerGroup(namespace string) error {
 			"refresh interval", namespace)
 	}
 
+	// before starting, first successfully check the versions of all pgcluster's in the namespace
+	if err := operatorupgrade.CheckVersion(c.controllers[namespace].pgoRESTClient,
```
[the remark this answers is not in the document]
andrewlecuyer: `reconcileRBAC` is called in two places:

- When a new controller groups is added
- On each namespace refresh interval

This means when we first detect a new namespace we attempt to reconcile RBAC for it.  But in the event that we can't yet (because say some other Kube Admin needs to create the RBAC needed to allow it), we'll then just try again on the next refresh interval.

As for the line referenced here with the `CheckVersion` function, this is simply used to check whether or not any `pgclusters` in the namespace are from an older version of the Operator, and therefore need to be upgraded.

postgres-operator.go
```diff
@@ -20,21 +20,19 @@ import (
...
 	crunchylog "github.com/crunchydata/postgres-operator/internal/logging"
 	"github.com/crunchydata/postgres-operator/internal/ns"
-	"github.com/crunchydata/postgres-operator/internal/operator/operatorupgrade"
 	log "github.com/sirupsen/logrus"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
```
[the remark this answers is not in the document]
andrewlecuyer: Thanks, updated `toml` and `lock` files have now been committed.

deploy/install-rbac.sh
```diff
@@ -46,27 +46,32 @@ fi
...
-	envsubst < $DIR/cluster-roles.yaml | $PGO_CMD create -f -
+	
+	if [[ "${PGO_RECONCILE_RBAC:-true}" == "true" ]]
+	then
+		# create the full ClusterRole with namespace and RBAC permissions if RBAC reconciliation 
+		# is enabled
+		envsubst < $DIR/cluster-roles.yaml | $PGO_CMD create -f -
+	else
+		# create a ClusterRole with only namespace permissions if RBAC reconciliation is disabled
+		sed '/- delete/q' $DIR/cluster-roles.yaml | $PGO_CMD create -f -
```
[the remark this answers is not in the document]
andrewlecuyer: Thanks, the `envsubst` isn't needed and has been removed.

internal/controller/manager/rbac.go
```diff
@@ -0,0 +1,142 @@
...
+
+const (
+	// ErrReconcileRBAC defines the error string that is displayed when RBAC reconciliation is
+	// enabled for the current PostgreSQL Operator installation but the Operator is unable to
+	// to properly/fully reconcile its own RBAC in a target namespace.
+	ErrReconcileRBAC = "operator is unable to reconcile RBAC resource"
+)
+
+// ReconcileRBAC is responsible for reconciling  a configMap resource that has obtained from
+// the ConfigMap controller's worker queue
```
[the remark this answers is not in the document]
andrewlecuyer: Thanks, the comment has been updated.

internal/ns/nslogic.go
andrewlecuyer: Like the idea here, and would like to try it out.  However, per our discussion will defer this to a future commit.

internal/controller/manager/controllermanager.go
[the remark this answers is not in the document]
cbandy: When `c.controllers[namespace] != nil`, `c.reconcileRBAC(namespace)` is called twice. Why is that?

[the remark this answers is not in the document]
cbandy: > When `c.controllers[namespace] != nil`, `c.reconcileRBAC(namespace)` is called twice. Why is that?

No, I see... this runs even if `addControllerGroup()` returns `ErrControllerGroupExists`...

Something about this being in two places still seems off. Shouldn't there be one place in the control loop where we make this choice?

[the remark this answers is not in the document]
andrewlecuyer: It's really just an effort to reconcile RBAC as quickly as possible.  So while we don't necessarily need to do this when adding the controller group since on the next refresh interval it will reconcile the RBAC anyway (which is essentially what always happens in `readonly` or `disabled` modes since the create namespace event comes in before the RBAC resources exist within it), it just provides a means to reconcile these resources right away when we can (e.g. when using `dynamic` and we have the ClusterRole needed to do so).

Also worth noting that on each refresh interval I do want the RBAC reconcile attempt to happen before we check to ensure the controller is running, so that each worker can reconcile the current namespace concurrently with any other workers prior to running into the mutex below.

PR #1665, opened 2020-06-24
Kubectl install cm

jmckulk:
There are two parts to this change. One includes updating the Ansible install to use a combination of a yaml inventory file and values file in place of a single ini inventory file. The `inventory.yaml` will include information that is either required by or only used by Ansible. This includes host information and variables that are used to connect to a Kubernetes or OpenShift cluster. The `values.yaml` file, which is now included the the `$PGOROOT/installers/ansible` dir, will contain all configuration options for the PostgreSQL Operator. This file should be at the path defined by `config_path` in the `inventory.yaml`. The default for this path is `$PGOROOT/installers/ansible/values.yaml` but can be updated in the `inventory.yaml file` or at the command line.

The `kubectl` install method is now updated to use a ConfigMap to pass in this `values.yaml` file instead of setting environment variables in the deployer image. The `postgres-operator.yml` and `postgres-operator-311.yml` will create the required ConfigMap and ensure that it is mounted as a volume to the `pgo-deploy` job. In order to configure the PostgreSQL Operator deployment you will need to update the `values.yaml` section that is defined in your manifest file. This file will then be used to run the Ansible installer in the `pgo-deployer` image. The `DEPLOY_ACTION` is still used to define which Ansible tag will be used and should be defined in the `postgres-operator.yml` (or `postgres-operator-311.yml`).

The `pgo-deployer` image has been updated to work with the updates from the Ansible install and now makes use of the ConfigMap defined in the `kubectl` job manifest. The image now uses an updated `ansible-playbook` command that overwrites the `config_path` and enables `kubernetes_in_cluster`.  This set of changes updates is an attempt to standardize the input for the `pgo-deployer`. The image now takes a generic yaml configuration file (`values.yaml`) that can be use across different install methods without having to maintain multiple files.

**Checklist:**

 
Tested by running the automated tests on a deployment using the updated `kubectl` install method.  These tests were run with `centos7` images. Automated tests still need to be run for `rhel7/ubi7` and the base Ansible install.

**Type of Changes:**

 
**Other information**:


installers/ansible/roles/pgo-preflight/tasks/vars.yml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Include values.yml
```
jmckulk, line 2: I think this will always use `ansible/values.yaml` considering it will be defined in the image. Need to add a check so that it picks one `config_path` or the other and doesn't always set it to `{{ playbook_dir }}/values.yaml`

installers/image/bin/pgo-deploy.sh
```diff
@@ -11,34 +11,11 @@
...
-# Installer Settings
 export ANSIBLE_CONFIG="/ansible/ansible.cfg"
 export DEPLOY_ACTION=${DEPLOY_ACTION:-install}
 export HOME="/tmp"
-export KUBERNETES_IN_CLUSTER=${KUBERNETES_IN_CLUSTER:-true}
 
-cat /inventory_template | envsubst > /tmp/inventory
-/usr/bin/env ansible-playbook -i /tmp/inventory --tags=$DEPLOY_ACTION /ansible/main.yml
+/usr/bin/env ansible-playbook \
+    -i /inventory.yaml \
```
jmckulk, line 19: This could also work. With this change there would only be one `inventory.yaml` file, instead of having one for the `pgo-deployer` and one for ansible. The only difference is that `kubernetes_in_cluster` needs to be set somewhere for ansible to work in the pgo-deployer image. The `inventory.yaml` file for ansible is baked into the image along with the other roles. This can be used and we can set `kubernetes_in_cluster` at the command line.

```suggestion
    -i /ansible/inventory.yaml \
    --extra-vars "kubernetes_in_cluster=true" \
```

jkatz: @jmckulk As long as that extra variable is encapsulated so that the user does not have to do anything I'm ok with that approach.

installers/ansible/inventory.yaml
```diff
@@ -0,0 +1,30 @@
...
+            # openshift_user: ''
+            # openshift_password: ''
+            # openshift_token: ''
+
+            # Deploy into Kubernetes
+            # ==================
+            # Note: Context name can be found using:
+            #   kubectl config current-context
+            # ==================
+            # kubernetes_context: ''
```
jkatz, line 30: Does this need to be in here?

installers/image/bin/pgo-deploy.sh
jmckulk: Yes this flag would be encapsulated in the `pgo-deployer` and it wouldn't change how users interact with `kubectl`

installers/ansible/inventory.yaml
jmckulk: It needs to be somewhere. This is how ansible connects to the kube/ocp cluster when you aren't using the deployer. It doesn't have to be in that file but since it isn't used to configure the operator i thought it would be good to separate it into this file.

If someone is installing with ansible they will need to update the connection info here and update configuration parameters in the values.yaml file. Both of these files will then be used to install

installers/image/bin/pgo-deploy.sh
jkatz: Then I'm ok with it :wink:

installers/ansible/inventory.yaml
jkatz: OK. Please put the Kubernetes block before OpenShift.

installers/ansible/roles/pgo-preflight/tasks/vars.yml
```diff
@@ -0,0 +1,32 @@
...
+  tags: always
+  block:
+    - name: Check for /conf/values.yaml
+      stat:
+          path: /conf/values.yaml
+      register: conf_dir_result
+
+    - when: conf_dir_result.stat.exists
+      set_fact:
+        config_path: "/conf/values.yaml"
```
cbandy, line 12: I like this approach, but I wonder if we should skip checking this directory in the normal ansible case. Something about checking root-path files makes me uneasy.

One of the benefits of the direct inventory file is that `ansible-playbook` accepts it as an argument. 🤔 

It _is_ possible (though a bit cumbersome) to specify a variable on the command line. What do you think of treating this path as a variable? The default could be `{{ playbook_dir }}/values.yaml` and a person/container can override when necessary. I imagine the container with `--extra-vars "something=/conf/values.yaml"`.

jmckulk: huh so the pgo-deploy script would have?

```
/usr/bin/env ansible-playbook \
    -i /ansible/inventory.yaml \
    --extra-vars "kubernetes_in_cluster=true" \
    --extra-vars "config_path=/conf/values.yaml" \
    --tags=$DEPLOY_ACTION \
    /ansible/main.yml
```
If we have a default of `config_path={{ playbook_dir }}/values.yaml` we could get rid of that logic. Then we would just have to check that the `{{ config_path }}` exists and include it.

jmckulk: I'm not sure this is necessary with only two `extra-vars` but we could have the pgo-deployer specific vars in a file and include it at the command line with `--extra-vars "@/path/to/deployer-specific.yml"`

docs/content/installation/configuration.md
```diff
@@ -276,3 +294,92 @@ for pgBackRest and pgBouncer respectively
 | `POD_ANTI_AFFINITY` | preferred |  | This will set the default pod anti-affinity for the deployed PostgreSQL clusters. |
 | `POD_ANTI_AFFINITY_PGBACKREST` |  |  | This will set the default pod anti-affinity for the pgBackRest pods. |
 | `POD_ANTI_AFFINITY_PGBOUNCER` |  |  | This will set the default pod anti-affinity for the pgBouncer pods. |
+
+## Resource Configuration
```
jmckulk, line 298: I'm not sure this section is still valid. I don't see references to these resource configuration variables anywhere else in the code. Does anyone know where its used or can I remove it?

jkatz: @jmckulk Yup, please remove.

installers/ansible/inventory.yaml
```diff
@@ -0,0 +1,31 @@
+---
+    all:
+        hosts:
+            localhost:
+        vars:
+            ansible_connection: local
+            ansible_python_interpreter: "/usr/bin/env python"
```
andrewlecuyer, line 7: Just curious, does the deployer container require `ansible_python_interpreter`?  If not, wondering if this is a good chance to remove this, and just let it auto-detect (or just explicitly define for the deployer if it needs it).  I know I always have to delete this setting and just let it auto-detect python 3 on my machine.

Anyway, not a big deal right now either way (not a blocker or anything), just figured I'd ask while we're touching this part of the `inventory`.

docs/content/installation/configuration.md
```diff
@@ -158,7 +152,32 @@ other storage classes in production deployments.
 | `REPLICA_STORAGE` | hostpathstorage | **Required** | Set to configure which storage definition to use when creating volumes used by PostgreSQL replicas on all newly created clusters. |
 | `WAL_STORAGE` |  |  | Set to configure which storage definition to use when creating volumes used for PostgreSQL Write-Ahead Log |
```
andrewlecuyer, line 153: I assume all vars defined in here will now be updated to their lowercase equivalent?

jmckulk: Yeah the will need to be eventually. I may hold off until I get the helm stuff up

installers/ansible/inventory.yaml
jmckulk: I can test that out and see

andrewlecuyer: Per our discussions this is OBE at this time, and we can therefore defer looking into this.  I will mark this conversation as resolved.

docs/content/installation/other/ansible/prerequisites.md
```diff
@@ -49,187 +49,35 @@ However, if the user wishes to add a new watched namespace after installation, w
...
-PostgreSQL Operator with any required storage configuration.  Guidance for
-defining `storage` variables can be found further in this documentation.
+The `inventory.yaml` file included with the PostgreSQL Operator Playbooks allows installers
+to configure how Ansible will connect to your Kubernetes cluster.  This file
+should contain the following connection variables:
 
 {{% notice tip %}}
 You should remove or comment out variables either either the  `kubernetes` or
 `openshift` variables if you are not being using them for your environment. Both
 sets of variables cannot be used at the same time.
```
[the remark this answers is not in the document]
jmckulk: Hows that read? @cbandy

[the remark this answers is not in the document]
cbandy: lgtm

PR #1676, opened 2020-06-29
Release resources as workqueue items are processed

cbandy:
**Checklist:**

 
**Type of Changes:**

 
### DIFF
--- a/internal/controller/pgcluster/pgclustercontroller.go
+++ b/internal/controller/pgcluster/pgclustercontroller.go
@@ -114,6 +114,7 @@ func (c *Controller) processNextItem() bool {
 
 	if err == nil {
 		log.Debugf("cluster add - dep already found, not creating again")
+		c.Queue.Forget(key)
 		return true
 	}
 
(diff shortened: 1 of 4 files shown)


internal/controller/pgcluster/pgclustercontroller.go
```diff
@@ -122,7 +122,7 @@ func (c *Controller) processNextItem() bool {
 	found, err := kubeapi.Getpgcluster(c.PgclusterClient, &cluster, keyResourceName, keyNamespace)
 	if !found {
 		log.Debugf("cluster add - pgcluster not found, this is invalid")
-		return false
+		return true
```
andrewlecuyer, line 125: I think we're going to want to call `Forget` on these before returning `true`.  Otherwise I'm thinking the item will be tracked indefinitely, e.g. by the rate limiter.

cbandy: Very true. There were a few other places that needed the same treatment.

internal/controller/pgcluster/pgclustercontroller.go
```diff
@@ -156,6 +159,7 @@ func (c *Controller) processNextItem() bool {
 	log.Debugf("Scaled pgBackRest repo deployment %s to 1 to proceed with initializing "+
 		"cluster %s", clusterInfo.PrimaryDeployment, cluster.Name)
 
+	c.Queue.Forget(key)
```
cbandy, line 162: Looks like this contradicts an `Add` call a few lines up.

cbandy: Fixed in 1100bde2c.

PR #1691, opened 2020-07-07
Ensure pgBackRest Settings Work When Local & S3 Storage is Enabled

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
Currently, pgBackRest backups that utilize the 'local,s3' setting do not
work as expected with the 's3-verify-tls' & 's3-uri-style' configuration 
settings.

This update ensures that the 's3-verify-tls' & 's3-uri-style' settings
work as expected when both local & S3 storage are enabled for pgBackRest
when creating a pgcluster. Currently, when a new cluster is created using
Local and AWS S3 storage (e.g. by specifiying
--pgbackrest-storage-type=local,s3 with the pgo create command),
stanza creation fails and the cluster is not initialized.

Since this is failing because the 'non S3' pgBackRest command runs when
the repo1-s3-verify-tls flag is detected in the environment. This fix updates
the pgBackRest flag implementation so that the repo1-s3-verify-tls setting
is applied only for the second pgbackrest command only, i.e. the command using
--repo-type-s3, and not apply it when running the first pgbackrest command for
the local repository. This is done by using the --no-repo1-s3-verify-tls
command flag rather than the relevant environment variable or the
related configuration file setting. This method allows both commands to
execute as expected.

**Other information**:
[ch8634]


internal/operator/clusterutilities.go
```diff
@@ -778,20 +778,36 @@ func GetPgbackrestS3EnvVars(cluster crv1.Pgcluster, clientset kubernetes.Interfa
...
+		return ""
+	}
+
+	return doc.String()
+}
+
+// GetS3VerifyTLSSetting parses the configured value as a boolean to ensure a valid
+// option is used, then returns the pgBackRest S3 configuration value to either enable
+// or disable TLS verification as the expected string value.
+func GetS3VerifyTLSSetting(cluster *crv1.Pgcluster) string {
```
jkatz, line 809: I believe very call to this passes in a pointer reference, so unsure why we are having this accept a pointer and not just a `crv1.Pgcluster` itself.

PR #1694, opened 2020-07-08
Remove PrimaryHost attribute from pgcluster CRD

jkatz:
This is no longer used.


internal/operator/clusterutilities.go
```diff
@@ -166,7 +166,6 @@ type DeploymentTemplateFields struct {
 	ScopeLabel          string
 	//next 2 are for the replica deployment only
 	Replicas                 string
-	PrimaryHost              string
```
cbandy, line 169: Is this comment still accurate/necessary?

jkatz: the `Replicas` one is still accurate.

Issue #1305, opened 2020-02-27
Restore deleted cluster


dvianello:
** Which example are you working with? **
Created a cluster called `test`, backed it up on S3 via `pgbackrest`, deleted cluster via pgo, tried a restore from s3

**What is the current behavior?**
```
bash-4.2$ pgo restore test -n postgres-dbs --pitr-target="2020-02-27 09:40:00.000000+00" --backup-opts="--type=time" --pgbackrest-storage-type s3
Warning:  If currently running, the primary database in this cluster will be stopped and recreated as part of this workflow!
WARNING: Are you sure? (yes/no): yes
Error: test was not found, verify cluster name
```
**What is the expected behavior?**
PGO recovers lost database from S3.

**Other information** (e.g. detailed explanation, related issues,  etc)
The above seems a relevant use-case to cover in case of: 
- a severe disaster recovery scenario where a K8S cluster is lost. 
- a clumsy human operator deletes the wrong cluster

jkatz:
Did you call `pgo delete cluster` with the `--keep-backups` flag?

I will note even with that flag on, while it does keep the the pgBackRest repository, there are some hoops to jump through at the moment, and we would have to make a restore from that backup repository easier. You could be able to restore from a logical backup (if you took them) but I would have to check.

dvianello:
Nope, no `--keep-backups` flag, as the intent was simulating the loss of a K8S cluster and recovery onto a new one, which failed.

Are we saying that `pgo` relies on the `backrest-shared-repo` pod being around _even_ for S3 backups? I thought backing off to S3 was a way to ensure that, even if everything went wrong with the cluster, one could recover from just the data stored in S3?

jkatz:
H @dvianello ,

If you have your data in a S3 repository, you can still restore it, regardless of having the Operator around. There are some manual steps for that process, though we are working on making some improvements to ensure it's more automatic.

Thanks.

dvianello:
Hi @jkatz,

glad to hear there are manual steps - are these documented somewhere?

Best,
Dario

dejwsz:
I am interested in the manual steps too.

philmcc:
I would like to know the process for this. It would be very useful for migrating data into the cluster from outside as well.

asoltesz:
I am struggling with this exact same test and would also like to know the manual process needed for this.

@dvianello May I recommend extending the issue title with "from s3 storage" or something similar to make the issue description more specific?

[...]

PR #1711, opened 2020-07-11
Rewrite docs to promote "pgo create cluster --restore-from" & deprecate clone.

jkatz:
"pgo create cluster --restore-from" is equivalent to "pgo clone"
but is much more robust, as it provides the ability to use the
full set of "pgo create cluster" options. As such, techniques
that use "pgo clone" are rewritten to showcase
"pgo create cluster --restore-from".

This also makes it much clearer how the request of #1305 can
be leveraged, as both the architecture of the feature and several
use cases are further fleshed out.

### DIFF
--- a/docs/content/pgo-client/common-tasks.md
+++ b/docs/content/pgo-client/common-tasks.md
@@ -638,42 +638,96 @@ pgo create schedule hacluster --schedule="0 0 * * *" \
 ### Restore a Cluster
 
 The PostgreSQL Operator supports the ability to perform a full restore on a
-PostgreSQL cluster as well as a point-in-time-recovery using the `pgo restore`
-command. Note that both of these options are **destructive** to the existing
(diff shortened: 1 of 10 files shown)


docs/content/pgo-client/common-tasks.md
```diff
@@ -914,16 +966,27 @@ section of the documentation.
...
 pgo clone hacluster newhacluster
 ```
 
+By leveraging `pgo create cluster`, you are able to copy the data from a
+PostgreSQL cluster while creating the topology of a new cluster the way you want
+to. For instance, if you want to copy data from an existing cluster that does
+not have metrics to a new cluster that does, you can accomplish that with the
+following command:
+
+`pgo create cluster newcluster --restore-from=oldcluster --metrics`
```
cbandy, line 987: This should probably use <code>```</code> (triple-ticks) to be consistent with the other examples.

PR #1713, opened 2020-07-11
4.3: Have pgBouncer come up with specified # of replicas after restore

jkatz:
**Checklist:**

 
**Type of Changes:**

 
This was defaulting to 1 instead of taking the value that was in
the pgcluster CRD.

**Other information**:

This not an issue on `master`, only `REL_4_3`.

Issue: [ch8675]

### DIFF
--- a/operator/cluster/clusterlogic.go
+++ b/operator/cluster/clusterlogic.go
@@ -632,6 +632,11 @@ func ScaleClusterDeployments(clientset *kubernetes.Clientset, cluster crv1.Pgclu
 			if !scalePGBouncer {
 				continue
 			}
+			// if the replica total is greater than 0, set number of pgBouncer
+			// replicas to the number that is specified in the cluster entry
+			if replicas > 0 {
+				replicas = int(cluster.Spec.PgBouncer.Replicas)
+			}
 		case deployment.Labels[config.LABEL_PGO_BACKREST_REPO] == "true":
 			clusterInfo.PGBackRestRepoDeployment = deployment.Name
 			// if not scaling services simply move on to the next deployment


operator/cluster/clusterlogic.go
```diff
@@ -632,6 +632,11 @@ func ScaleClusterDeployments(clientset *kubernetes.Clientset, cluster crv1.Pgclu
 			if !scalePGBouncer {
 				continue
 			}
+			// if the replica total is greater than 0, set number of pgBouncer
+			// replicas to the number that is specified in the cluster entry
+			if replicas > 0 {
+				replicas = int(cluster.Spec.PgBouncer.Replicas)
```
andrewlecuyer, line 638: Since the caller to `ScaleClusterDeployments` should typically specify the proper replica count, I'm thinking there is a call to this function that needs to be modified, e.g. `pgbouncer` would be scaled to 2 by calling the function as follows:

```bash
ScaleClusterDeployments(c.PodClientset, *cluster, int(cluster.Spec.PgBouncer.Replicas), false, false, false, true)
```

More specifically I'm thinking the scaling of the other Deployments (primary, replicas and pgBackRest repository) via this function needs to occur independently from `pgbouncer`, since `pgbouncer` can now have multiple replicas (while the other currently have no more than one replica at any given time).

I can take a look to see what I can find.

jkatz: It makes me wonder if this function is trying to do too much. It seems it can get called in a bunch of places, some wehre we scale pgBouncer, some where don't, and given that I'm not sure if we want to be passing in the # of replicas, given pgBouncer could be scaling at the same time as the cluster itself.

PR #1717, opened 2020-07-14
Adds uri path variables to config files

jmckulk:
These uri path variables for backrest were missed when updating the new
files. This change adds them to relevant config files and templates.

### DIFF
--- a/installers/helm/postgres-operator/templates/values_configmap.yaml
+++ b/installers/helm/postgres-operator/templates/values_configmap.yaml
@@ -8,7 +8,7 @@ metadata:
 data:
   values.yaml: |
     ---
-    archive_mode: "{{ .Values.archive_mode}}"
+    archive_mode: {{ .Values.archive_mode}}
     archive_timeout: "{{ .Values.archive_timeout}}"
     backrest_aws_s3_bucket: "{{ .Values.backrest_aws_s3_bucket}}"
     backrest_aws_s3_endpoint: "{{ .Values.backrest_aws_s3_endpoint}}"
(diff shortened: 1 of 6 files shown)


installers/helm/postgres-operator/templates/values_configmap.yaml
```diff
@@ -15,6 +15,8 @@ data:
     backrest_aws_s3_key: "{{ .Values.backrest_aws_s3_key}}"
     backrest_aws_s3_region: "{{ .Values.backrest_aws_s3_region}}"
     backrest_aws_s3_secret: "{{ .Values.backrest_aws_s3_secret}}"
+    backrest_aws_s3_uri_style: "{{ .Values.backrest_aws_s3_uri_style }}"
+    backrest_aws_s3_verify_tls: "{{ .Values.backrest_aws_s3_verify_tls }}"
```
cbandy, line 19: This is a YAML string, but in the other installers, it is a YAML boolean. 🤔 ​ I'm not sure there's any change necessary but thought I should point it out.

jmckulk: Good point - I left it in quotes since it was a templated value.. Not sure if those are necessary or not.

jkatz: They are not necessary, as the template is being transformed into a YAML file.

jmckulk: alright i remove double quotes from around the bool in these two templates. I used helm to template the values and the yaml no longer has double quotes

PR #1722, opened 2020-07-15
Update ansible to handle bool values

jmckulk:
Some conditional checks in the ansible installer were setup to only check for
strings "true" or "false". When the configuration options are set to bool values
the checks were not being handled like they should. This update uses the bool
filter to ensure that bools are handled properly.

**Checklist:**

 
Tested both ansible and kubectl w/ deployer. Each task was run based on the provided bool. 

I tested `reconcile_rbac` and saw that the ` DisableReconcileRBAC: True` setting was updated in `pgo-config`. Was also able to create a namespace using `namespace: pgouser1` var in values.yaml with `reconcile_rbac: false` and `namespace_mode: readonly`. Namespace had expected labels and `pgo-local-role` was created. I think this is good but i haven't personally tested `reconcile_rbac` without these changes. Would be good to look at with next tag.

**Type of Changes:**

 
**Other information**:
[ch8713]

### DIFF
--- a/installers/ansible/roles/pgo-operator/tasks/cleanup.yml
+++ b/installers/ansible/roles/pgo-operator/tasks/cleanup.yml
@@ -276,7 +276,7 @@
   file:
     state: absent
     path: "/usr/local/bin/pgo"
-  when: pgo_client_install == "true"
+  when: pgo_client_install | default(False) | bool
   ignore_errors: yes
   no_log: false
   tags: uninstall
--- a/installers/ansible/roles/pgo-operator/tasks/cleanup.yml
+++ b/installers/ansible/roles/pgo-operator/tasks/cleanup.yml
@@ -276,7 +276,7 @@
   file:
     state: absent
     path: "/usr/local/bin/pgo"
-  when: pgo_client_install | default(False) | bool
+  when: pgo_client_install | bool
   ignore_errors: yes
   no_log: false
   tags: uninstall
--- a/installers/ansible/roles/pgo-operator/tasks/namespace.yml
+++ b/installers/ansible/roles/pgo-operator/tasks/namespace.yml
@@ -28,7 +28,7 @@
   with_items: "{{ nslist.split(',') | map('trim') | list }}"
   when:
     - namespace_mode != 'dynamic'
-    - not (reconcile_rbac | bool)
+    - reconcile_rbac | bool
   tags:
     - install
     - update
@@ -40,7 +40,7 @@
   with_items: "{{ nslist.split(',') | map('trim') | list }}"
   when:
     - namespace_mode != 'dynamic'
-    - not (reconcile_rbac | bool)
+    - reconcile_rbac | bool
   tags:
     - install
     - update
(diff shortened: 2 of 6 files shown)


installers/ansible/roles/pgo-operator/tasks/namespace.yml
```diff
@@ -14,7 +14,7 @@
     target_ns_script: "{{ lookup('template', 'add-targeted-namespace.sh.j2') }}"
   with_items: "{{ nslist.split(',') | map('trim') | list }}"
   when:
-    - reconcile_rbac == 'false'
+    - not (reconcile_rbac | default(False) | bool)
```
jkatz, line 17: Isn't this defaulted to `true`? cc @andrewlecuyer

jmckulk: yeah it is

jmckulk: It is defaulted to `true` in the `pgo-operator/defaults`, do we need to have a default here too?

jkatz: @jmckulk I don't know -- what's the purpose of the `default(False)` here?

andrewlecuyer: @jkatz correct, `reconcile_rbac` should default to `true`.

jmckulk: to set a default value when `reconcile_rbac` is not set. In this case it is always set because of the default defaults.yaml

jkatz: So do we need a `default` block?

jmckulk: no it isn't getting used here. Ill remove it

jkatz: Thanks.

installers/ansible/roles/pgo-operator/tasks/namespace.yml
```diff
@@ -28,7 +28,7 @@
   with_items: "{{ nslist.split(',') | map('trim') | list }}"
   when:
     - namespace_mode != 'dynamic'
-    - reconcile_rbac == 'true'
+    - not (reconcile_rbac | bool)
```
cbandy, line 31: Seems like here and below should lack `not`. Anything that used to be `reconcile_rbac == 'true'`

```suggestion
    - reconcile_rbac | bool
```

installers/ansible/roles/pgo-operator/tasks/cleanup.yml
```diff
@@ -276,7 +276,7 @@
   file:
     state: absent
     path: "/usr/local/bin/pgo"
-  when: pgo_client_install == "true"
+  when: pgo_client_install | default(False) | bool
```
jkatz, line 279: The documentation says it's defaulting `true`? Do we need the `default(False)` here?

jmckulk: No i don't think we do. Like the other places, this is set in a defaults yaml so the block wouldn't be used. Fixing.

jkatz:
Superseded by #1726

PR #1738, opened 2020-07-28
Main Config Update

jmckulk:
This update adds a master config that is used for the Ansible, Helm, and Kubectl
installers. The config_sync.sh script is added to the hack directory, that will use 
the ansible/values.yaml file as the master config to update each of the installer config files.

The values.yaml for each installer is updated to exactly match the master
config. This is a small change that includes adding a link to the docs and
syncing the pgo_client options.

The master config and the helm installer values.yaml will be out of sync until
the installers are updated with coming metrics changes. The master config still
contains metrics options but those options were removed from the helm
values.yaml.

**Checklist:**

 
**Type of Changes:**

 
Currently installer configs must be updated in multiple places, this leads to difficulties maintaining the files and the files getting out of sync.

The installer files are now based off of a single main config, located in the ansible installer, that is then moved into place for the other installers by a script.

**Other information**:
[ch8712]


installers/conf/config_sync.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+# Copyright 2019 - 2020 Crunchy Data Solutions, Inc.
```
jkatz, line 3:
```suggestion
# Copyright 2020 Crunchy Data Solutions, Inc.
```

installers/ansible/values.yaml
```diff
@@ -1,4 +1,8 @@
----
+# =====================
+# Configuration Options
+# More info for these options can be found in the docs
+# https://crunchydata.github.io/postgres-operator/latest/installation/configuration/
```
jkatz, line 4:
```suggestion
# https://access.crunchydata.com/documentation/postgres-operator/latest/installation/configuration/
```

installers/conf/values.yaml
```diff
@@ -0,0 +1,133 @@
+# =====================
+# Configuration Options
+# More info for these options can be found in the docs
+# https://crunchydata.github.io/postgres-operator/latest/installation/configuration/
```
jkatz, line 4:
```suggestion
# https://access.crunchydata.com/documentation/postgres-operator/latest/installation/configuration/
```

installers/helm/postgres-operator/values.yaml
```diff
@@ -69,7 +69,8 @@ pgo_apiserver_port: "8443"
 pgo_apiserver_url: "https://postgres-operator"
 pgo_client_cert_secret: "pgo.tls"
 pgo_client_container_install: "false"
-pgo_client_version: "4.4.0"
+pgo_client_install: "true"
+pgo_client_version: "v4.4.0"
```
jkatz, line 73: I think it'd be good if we can leave the `v` off of this, and handle that for the user. and/or be able to accept `4.4.0` and `v4.4.0`

installers/kubectl/postgres-operator-ocp311.yml
```diff
@@ -1,31 +1,35 @@
...
   name: pgo-deployer-cm
   namespace: pgo
 data:
-  values.yaml: |+
-    ---
+  values.yaml: |-
+    # =====================
+    # Configuration Options
+    # More info for these options can be found in the docs
+    # https://crunchydata.github.io/postgres-operator/latest/installation/configuration/
```
jkatz, line 31:
```suggestion
    # https://access.crunchydata.com/documentation/postgres-operator/latest/installation/configuration/
```

installers/conf/config_sync.sh
```diff
@@ -0,0 +1,39 @@
...
+then
+	echo "error: \$PGOROOT must be set"
+	exit 1
+fi
+
+INSTALLER_DIR="$PGOROOT/installers"
+CONF_DIR="$INSTALLER_DIR/conf"
+TEMPLATE_DIR="$CONF_DIR/templates"
+
+if [[ -d ${INSTALLER_DIR} || -d ${CONF_DIR} || -d ${TEMPLATE_DIR} ]]
```
jmckulk, line 26: Is this sort of check necessary? This would be a developer script but could also be used with CI. 

I'm thinking that you'd have to be doing something really weird to have those files not exist in `$PGOROOT`

andrewlecuyer: Since these directories are part of the actual repo, I think we can safely assume they'll always be there (i.e. no user action is really needed to be able to run this script, and they would need to explicitly delete the referenced dirs to break it).

installers/helm/postgres-operator/values.yaml
jmckulk: Sure. This will require a minor change to the ansible

installers/conf/config_sync.sh
jmckulk: Thats what I was thinking, alright I'm gonna remove the checks then.

installers/conf/templates/helm_template.yaml
```diff
@@ -0,0 +1,20 @@
+---
```
andrewlecuyer, line 1: It would be great to be able to keep the Helm specific stuff down in the Helm install directories (and therefore keep the `conf` dir for the config that applies to all installers).  Can we therefore move this file down into `installers/helm/postgres-operator`?

I'm thinking we can then leverage `.helmignore` to keep this file from being picked up in the event the chart is ever packaged into an archive file (e.g. via `helm package`).

installers/conf/config_sync.sh
andrewlecuyer, line 26: Thinking we should move this script to the `hack` directory under the project root, alongside the `update-codegen.sh` and `verify-codegen.sh` scripts.  This is a pattern that emerged within the Kube world for storing various developer-facing scripts, and this appears to fall into that category:

https://github.com/kubernetes/kubernetes/tree/master/hack
https://github.com/kubernetes/kops/blob/master/docs/development/hack.md
https://github.com/kubernetes/kops/issues/444

This will allow us to continue consolidating our various dev scripts into a single, central location.

jmckulk: Yeah that sounds like a good idea, I can move that script. With that change, and moving the `helm_template.yaml` into the helm directory, the only thing that is in the `installers/conf` directory is the master `values.yaml`. We can move that file and not have to add a new directory. What would yall think about one of these:

- using the values.yaml in the ansible directory as the master
- moving the master values.yaml to the root installers directory, so `installers/values.yaml`

andrewlecuyer: I personally like the idea of just using Ansible as the master.  If Ansible is essentially always the "master" config anyway (i.e. they are always identical), then seems to make sense to just use it as the master, instead of having a duplicate copy.

And this also further consolidates around Ansible essentially being the master installer for everything (i.e. master install engine as used by `kubectl` and Helm, the master config location, etc.).

jmckulk: Sounds good to me, I'll move forward with that change

PR #1747, opened 2020-07-30
Json templates

jmckulk:
Bash and Ansible installers are currently pulling json templates from two
different places. `$PGOROOT/conf` and
`$PGOROOT/installers/ansible/roles/pgo-operator/files` both contain the same
json templates. By setting both installers to pull the templates from the same
directory we can simplify maintenance when changing these files.

We are standardizing on Ansible as the main installer that is a base for each
other installer. This commit includes changes to make the Bash installer use the
Ansible config directory. Bash is now setup to use an environment variable to determine the config
directory. This environment variable is updated to point to the Ansible config
directory. The config files in the `conf` directory were removed, the only files that are left are either not used by ansible or generated through an ansible template.

Users of the bash install will still need to update the `conf/pgo.yaml` file.

**Checklist:**

 
**Type of Changes:**

 
**Other information**:
[ch8712]


deploy/upgrade-creds.sh
```diff
@@ -15,10 +15,6 @@
 
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
jmckulk, line 17: Should these lines be removed? The files no longer exist in the repo but the script is used to upgrade from a previous version where they did.

Makefile
```diff
@@ -154,6 +154,7 @@ $(PGOROOT)/$(DFSET)/Dockerfile.%.$(DFSET):
 		--build-arg PGVERSION=$(PGO_PG_VERSION) \
 		--build-arg BACKREST_VERSION=$(PGO_BACKREST_VERSION) \
 		--build-arg ANSIBLE_VERSION=$(ANSIBLE_VERSION) \
+		--build-arg PGO_CONF_DIR=$(PGO_CONF_DIR) \
```
jmckulk, line 157: needs to be pulled out - I was going to try to add a variable for the config directory but that didn't work out

deploy/upgrade-creds.sh
andrewlecuyer: I see no problem with removing these lines.  Those defaults really no longer make any sense since as you said, those files no longer exist at those locations (or at all, for that matter).  Therefore, any pre-v4.1.0 holdouts (do they even still exist :thinking:) can just pass in the location of these files via params, as supported by the script.

deploy/install-rbac.sh
```diff
@@ -17,11 +17,16 @@
...
 
 # see if CRDs need to be created
 $PGO_CMD get crd pgclusters.crunchydata.com > /dev/null
 if [ $? -eq 1 ]; then
-	$PGO_CMD create -f $DIR/crd.yaml
+	$PGO_CMD create \
+		-f ${PGO_CONF_DIR}/crds/pgclusters-crd.yaml \
+		-f ${PGO_CONF_DIR}/crds/pgpolicies-crd.yaml \
+		-f ${PGO_CONF_DIR}/crds/pgreplicas-crd.yaml \
+		-f ${PGO_CONF_DIR}/crds/pgtasks-crd.yaml
```
cbandy, line 29: `-f` accepts a directory, if you want: `$PGO_CMD create -f "${PGO_CONF_DIR}/crds"`

PR #1763, opened 2020-08-06
Update sane default for pgbadger resources

jmckulk:
pgbadger was running into the OOM Killer because of the memory resource limits.
This commit ups the memory limit as a workaround.

**Checklist:**

 
**Type of Changes:**

 
**Other information**:
[ch8727]

### DIFF
--- a/installers/ansible/roles/pgo-operator/files/pgo-configs/pgbadger.json
+++ b/installers/ansible/roles/pgo-operator/files/pgo-configs/pgbadger.json
@@ -23,7 +23,7 @@
                 "resources": {
                   "limits": {
                     "cpu": "500m",
-                    "memory": "24Mi"
+                    "memory": "64Mi"
                   }
                 },
                 "volumeMounts": [


installers/ansible/roles/pgo-operator/files/pgo-configs/pgbadger.json
```diff
@@ -23,7 +23,7 @@
                 "resources": {
                   "limits": {
                     "cpu": "500m",
-                    "memory": "24Mi"
+                    "memory": "36Mi"
```
jkatz, line 26: Did you get this number from profiling?

If the reason the OOM killer is being activated is because we're running out of memory, my guess is that this number needs to be even higher. It's OK to set a higher limit for pgBadger. The limit is the high water mark - it does not mean we'll use, say 36Mi, but rather that we _won't_ let it use 36Mi. I think even something like 128Mi is fine.

jmckulk: I updated the memory limit and was able to generate the report multiple times without error. Without updating the limit it will sometimes work a few times be eventually throws an error. I was trying to keep the value generally the same considering I didn't know why it was set that way in the first place. If updating to 128Mi is fine with you I think that would leave plenty of room for error.

jmckulk: After talking with @andrewlecuyer I updated with 64Mi. This should provide plenty of working memory while also not taking more than we need.

PR #1767, opened 2020-08-07
Accept custom pgBackRest configuration files

cbandy:
**Checklist:**

 
**Type of Changes:**

 
A new argument to `pgo create`, `--pgbackrest-custom-config`, allows one to specify a ConfigMap that is loaded into every PostgreSQL and pgBackRest container.

In addition, configurations are automatically loaded from a ConfigMap and/or Secret named `{cluster}-backrest-config` when they exist. These objects may be populated *after* a PgCluster is created.

**Other information**:

- [ ] TODO: regenerate command docs


internal/apiserver/clusterservice/clusterimpl.go
```diff
@@ -1356,6 +1356,12 @@ func getClusterParams(request *msgs.CreateClusterRequest, name string, userLabel
 	spec.CustomConfig = request.CustomConfig
 	spec.SyncReplication = request.SyncReplication
 
+	if request.BackrestConfig != "" {
```
andrewlecuyer, line 1359: Thinking we should validate that the ConfigMap provided is valid, like we do for custom PG/Patroni config:

https://github.com/CrunchyData/postgres-operator/blob/6fbd804b57ab68cd4286473ce22b2b1ccf58c25e/internal/apiserver/clusterservice/clusterimpl.go#L682-L683

cbandy: Done!

andrewlecuyer:
Thinking we should add something to the [`Custom Configuration` section](https://access.crunchydata.com/documentation/postgres-operator/4.4.1/advanced/custom-configuration/) of the docs for this one.   This includes details for using a custom pgBackRest ConfigMap, as well some high-level information regarding the various settings we control via env var (and therefore can be set via a custom `conf` file).

PR #1770, opened 2020-08-09
Default URI Style for S3 Bootstrap

andrewlecuyer:
The proper default S3 URI style is now set when generating the pgBackRest S3 environment variables for a bootstrap Job (i.e. a Job created to bootstrap a cluster from an existing backup, such as when using the `--restore-from` flag with the `pgo` client). 
Specifically, if an empty string is detected in the `s3-uri-style` annotation for the pgBackRest repo secret for the cluster being bootstrapped/restored from, a proper default value of `host` is now set.  This ensures there is never an empty value for the `PGBACKREST_REPO1_S3_URI_STYLE` environment variable in a boostrap Pod, which can lead to problems when running various pgBackRest commands.

Additionally, tests have been created to verify the output of the `GetPgbackrestBootstrapS3EnvVars` function that is utilized to populate the proper S3 environment variables for a bootstrap Job (specifically using the pgBackRest repo secret from the cluster being bootstrapped from).  This includes ensuring the proper default URI style is set when the `s3-uri-style` annotation is empty, as well as ensuring all annotation values are properly reflected in the final S3 environment variable output.  In support of these tests, a `NewFakePGOClient` function has also been created in the `fake` package (underneath `kubeapi`).  This is needed to create a fake client containing the proper resources (i.e. a mock `pgo-config` ConfigMap) as need to initialize the Operator and therefore the various templates needed to properly call the `GetPgbackrestBootstrapS3EnvVars` function for these tests (`PGOROOT` must be properly set in the environment when running tests so that the proper default template files and `pgo.yaml` file can be found as needed to create the mock `pgo-config` ConfigMap).

**Checklist:**

 
**Type of Changes:**

 
The `PGBACKREST_REPO1_S3_URI_STYLE` environment variable does not contain the proper default (`host`) when the `s3-uri-style` annotation is empty for the secret corresponding to the pgBackRest repo being utilized to bootstrap a new cluster.

[ch8915]

The `PGBACKREST_REPO1_S3_URI_STYLE` environment variable defaults to `host` when the `s3-uri-style` annotation is empty for the secret corresponding to the pgBackRest repo being utilized to bootstrap a new cluster.

**Other information**:

N/A

### DIFF
--- a/internal/kubeapi/fake/fakeclients.go
+++ b/internal/kubeapi/fake/fakeclients.go
@@ -1,0 +1,123 @@
+package fake
+
+/*
+ Copyright 2020 Crunchy Data Solutions, Inc.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+import (
+	"errors"
+	"io/ioutil"
+	"os"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	fakekube "k8s.io/client-go/kubernetes/fake"
+
+	"github.com/crunchydata/postgres-operator/internal/config"
+	"github.com/crunchydata/postgres-operator/internal/kubeapi"
+	fakecrunchy "github.com/crunchydata/postgres-operator/pkg/generated/clientset/versioned/fake"
+)
+
+const (
+	// defaultPGOInstallationName is the default installation name for a fake PGO client
+	defaultPGOInstallationName = "test"
+	// pgoNamespace is the default operator namespace for a fake PGO client
+	defaultPGONamespace = "pgo"
+	// defaultTargetNamespaces are the default target namespaces for a fake PGO client
+	defaultTargetNamespaces = "pgouser1,pgouser2"
+)
+
+var (
+	// pgoRoot represents the root of the PostgreSQL Operator project repository
+	pgoRoot = os.Getenv("PGOROOT")
+	// templatePath defines the default location for the PostgreSQL Operator templates relative to
+	// env var PGOROOT
+	templatePath = pgoRoot + "/installers/ansible/roles/pgo-operator/files/pgo-configs/"
+	// pgoYAMLPath defines the default location for the default pgo.yaml configuration file
+	// relative to env var PGOROOT
+	pgoYAMLPath = pgoRoot + "/conf/postgres-operator/pgo.yaml"
+)
+
+// NewFakePGOClient creates a fake PostgreSQL Operator client.  Specifically, it creates
+// a fake client containing a 'pgo-config' ConfigMap as needed to initialize the Operator
+// (i.e. call the 'operator' packages 'Initialize()' function).  This allows for the proper
+// initialization of the Operator in various unit tests where the various resources loaded
+// during intialization (e.g. templates, config and/or global variables) are required.
+func NewFakePGOClient() (kubeapi.Interface, error) {
+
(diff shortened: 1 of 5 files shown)


internal/kubeapi/fake/fakeclients.go
```diff
@@ -0,0 +1,123 @@
...
+	// relative to env var PGOROORT
+	pgoYAMLPath = pgoRoot + "/conf/postgres-operator/pgo.yaml"
+)
+
+// NewFakePGOClient creates a fake PostgreSQL Operator client.  Specifically, it creates
+// a fake client containing a 'pgo-config' ConfigMap as needed to initialize the Operator
+// (i.e. call the 'operator' packages 'Initialize()' function).  This allows for the proper
+// initialization of the Operator in various unit tests where the various resources loaded
+// during intialization (e.g. templates, config and/or global variables) are required.
+func NewFakePGOClient() (kubernetes.Interface, error) {
```
cbandy, line 57: It looks like this could be a `kubeapi.Interface`?

internal/kubeapi/fake/fakeclients.go
```diff
@@ -0,0 +1,123 @@
...
+	"io/ioutil"
+	"os"
+
+	"github.com/crunchydata/postgres-operator/internal/config"
+	fakecrunchy "github.com/crunchydata/postgres-operator/pkg/generated/clientset/versioned/fake"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	fakekube "k8s.io/client-go/kubernetes/fake"
```
cbandy, line 29: Not a blocker for this PR, but I've noticed we're placing our imports between stdlib and third-party. The convention is to group imports as stdlib, third-party, then "local". `goimports` even has a flag to enforce it:

```
  -local string
        put imports beginning with this string after 3rd-party packages; comma-separated list
```

internal/kubeapi/fake/fakeclients.go
```diff
@@ -0,0 +1,123 @@
...
+	pgoConfig, err := createMockPGOConfigMap(defaultPGONamespace)
+	if err != nil {
+		return nil, err
+	}
+
+	// now create and return a fake client containing the ConfigMap
+	return Clientset{
+		Clientset:    fakekube.NewSimpleClientset(pgoConfig),
+		PGOClientset: fakecrunchy.NewSimpleClientset(),
+	}, nil
```
cbandy, line 79: I'm glad to see this type works at all; I think this is the first instance of it. If you find anything about it that can improve, please do.

andrewlecuyer: Will do!  Seems to be working well so far :)

internal/kubeapi/fake/fakeclients.go
andrewlecuyer: Went ahead and re-arranged per this convention.

internal/kubeapi/fake/fakeclients.go
andrewlecuyer: Switched to `kubeapi.Interface`

internal/operator/clusterutilities_test.go
```diff
@@ -168,3 +174,117 @@ func TestOverrideClusterContainerImages(t *testing.T) {
...
+	// repo secret
+	t.Run("populate from secret", func(t *testing.T) {
+
+		backRestRepoSecret := mockBackRestRepoSecret.DeepCopy()
+		s3EnvVars := GetPgbackrestBootstrapS3EnvVars(defaultRestoreFromCluster, backRestRepoSecret)
+		// massage the results a bit so that we can parse as proper JSON to validate contents
+		s3EnvVarsJSON := strings.TrimSuffix(`{"EnvVars": [`+s3EnvVars, ",\n") + "]}"
+
+		s3Env := &Env{}
+		if err := json.Unmarshal([]byte(s3EnvVarsJSON), s3Env); err != nil {
```
cbandy, line 219: It looks like the local type might not be necessary.

```suggestion
		s3EnvVarsJSON := `[`+strings.TrimSuffix(s3EnvVars, ",\n") + `]`

		s3EnvVars := []corev1.EnvVar{}
		if err := json.Unmarshal([]byte(s3EnvVarsJSON), &s3EnvVars); err != nil {
```

PR #1772, opened 2020-08-11
Conditionally Enable 'timeout' Flag for the Rollout Commands Used in the Metrics Installer

andrewlecuyer:
The `--timeout` flag is now conditionally enabled for the `rollout status` command depending on whether or not is it supported by the version of `oc` or `kubectl` being used in conjunction with the metrics Ansible installer.  This is to enable compatibility with older versions of Kubernetes and OpenShift, specifically OpenShift v3.11, which does not support the  `timeout` flag.

Now when the metrics installer is run, a check is done to determine whether or not the `oc` or `kubectl` client supports the `timeout` flag.  If it does, a timeout is set for the `rollout status` commands, and the installer proceeds as normal to check the status of the various metrics deployments.  And with the timeout set, the rollout status checks for all deployments will timeout at roughly the same time if none are able to successfully deploy, resulting in consistent/clear timeout behavior.

However, if the `timeout` flag is not supported, it is not set on the`rollout status` command, which means a timeout is not applied directly to the `rollout` commands.  However, the subsequent `async_status` task will still fail after a certain number of retries (60 by default, with a delay of 10 seconds between each check), ensuring expected timeout behavior will still occur.  The only downside is that once one deployment "times out" after 60 retries, the counter will again start at 60 to check the status of the other deployment.  This means that the total timeout time could be double in the event that both deployments are unsuccessful (unlike when using the `--timeout` flag, which results in all deployments timing out at approximately the same time).

The benefit of this approach is that compatibility is maintained with older versions of Kubernetes and OpenShift, while still maintaining very similar functionality and minimal differences in logic within the installer.  This also prevents the need to detect specific versions of `oc` or `kubectl` by instead simply relying on whether or not the `--timeout` flag is supported, regardless of specific Kubernetes or OpenShift version.

**Checklist:**

 
**Type of Changes:**

 
The Metrics installer's `rollout status` commands used to check the deployment status of Grafana and Prometheus do no work with an OpenShift 3.11 client.

The Metrics installer's `rollout status` commands used to check the deployment status of Grafana and Prometheus now work properly with an OpenShift 3.11 client.

**Other information**:

N/A


installers/metrics/ansible/roles/pgo-metrics/tasks/main.yml
```diff
@@ -79,6 +79,13 @@
     - include_tasks: grafana.yml
       when: grafana_install | bool
 
+    - name: Check if Timeout Flag Supported
+      command: |
+        {{ kubectl_or_oc }} rollout status deployment/flagcheck -n {{ metrics_namespace }} \
+          --watch=false --timeout=1s
```
cbandy, line 85: Instead of actually hitting the API, can we use the `--help` flag? On recent versions of `kubectl` and `oc` it goes to stdout:

```
$ kubectl rollout status --help | wc -l
20
```

jkatz:
@andrewlecuyer @cbandy Is this ready for merge?

installers/metrics/ansible/roles/pgo-metrics/tasks/main.yml
andrewlecuyer: Good thought, the task has been updated to use `--help`.

andrewlecuyer:
@jkatz & @cbandy think this one should be good to go now.  I went ahead and pushed an update per the comment above for the `--timeout` flag check, as well as another commit to explicitly set some namespaces when creating ConfigMaps due to an issue with the OCP 3.11 client in which the namespace isn't otherwise set properly.

PR #1794, opened 2020-08-18
WIP: Custom Annotations for Deployments

jkatz:
**Checklist:**

 
**Type of Changes:**

 
This allows one to add custom annotations to the managed
deployments, including the PostgreSQL Operator, pgBackRest, and
pgBouncer. Annotations can be applied to all of the Deployments,
or tuned to be applied to the individual Deployment groups.

Issue: [ch7938]


internal/operator/cluster/cluster.go
```diff
@@ -413,6 +413,46 @@ func ScaleDownBase(clientset kubeapi.Interface, replica *crv1.Pgreplica, namespa
...
+	if err != nil {
+		return err
+	}
+
+	// now update each deployment with the new annotations
+	for _, deployment := range deployments.Items {
+		log.Debugf("update annotations on [%s]", deployment.Name)
+		log.Debugf("new annotations: %v", annotations)
+
+		deployment.Spec.Template.ObjectMeta.SetAnnotations(annotations)
```
andrewlecuyer, line 434: Will this overwrite any existing annotations in the Pod?  At the moment specifically thinking about the `status` annotation written by Patroni, and not sure if any conflict we need to be concerned with:

```
$ kubectl get pods mycluster1-bc47bff86-k4rwg -o yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    status: '{"conn_url":"postgres://10.44.0.165:5432/postgres","api_url":"http://10.44.0.165:8009/patroni","state":"running","role":"master","version":"1.6.5","xlog_location":83886080,"timeline":1}'
```

pgo/cmd/cluster.go
```diff
@@ -504,6 +528,29 @@ func isTablespaceParam(param string) bool {
...
+		// 1: key=value (adding, editing)
+		// 2: key- (removing)
+		if strings.HasSuffix(annotation, "-") {
+			annotationGroup[strings.TrimSuffix(annotation, "-")] = ""
+			continue
+		}
+
+		parts := strings.Split(annotation, "=")
+
+		if len(parts) != 2 {
```
andrewlecuyer, line 545: We could leverage some of the validation logic included in `apimachinery` to further/fully validate the annotation key/values provided: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go

pgo/cmd/create.go
```diff
@@ -322,6 +337,24 @@ func init() {
...
 	// flags for "pgo create cluster"
+	createClusterCmd.Flags().StringSliceVar(&Annotations, "annotation", []string{},
+		"Add an Annotation to all of the managed deployments (PostgreSQL, pgBackRest, pgBouncer)"+
+			"The format to add an annotation is \"name=value\"\n"+
+			"The format to remove an annotation is \"name-\"\n\n"+
+			"For example, to add two annotations:\n"+
+			"--annotation=hippo=awesome --annotation=elephant=cool")
+	createClusterCmd.Flags().StringSliceVar(&AnnotationsBackrest, "annotation-pgbackrest", []string{},
+		"Add an Annotation to specifically to pgBackRest deployments"+
+			"The format to add an annotation is \"name=value\"\n"+
```
andrewlecuyer, line 348: Looks there's an extra `to` in this description before `specifically` (and same in those update commands that follow),  e.g. should be:

```
Add an Annotation specifically to pgBackRest deployments
```

internal/operator/backrest/repo.go
```diff
@@ -235,6 +237,32 @@ func getRepoDeploymentFields(clientset kubernetes.Interface, cluster *crv1.Pgclu
...
+
+	// now update the pgBackRest deployment
+	log.Debugf("update annotations on [%s]", deployment.Name)
+	log.Debugf("new annotations: %v", annotations)
+
+	deployment.Spec.Template.ObjectMeta.SetAnnotations(annotations)
+
+	// finally, update the Deployment. If something errors, we'll log that there
+	// was an error, but continue with processing the other deployments
+	if _, err := clientset.AppsV1().Deployments(deployment.Namespace).Update(deployment); err != nil {
```
andrewlecuyer, line 259: In general been thinking we prob want to start accounting for potential conflicts when using `Update()` (i.e. `PUT` requests), e.g. in case something else updates the Deployment underneath us.  Or the alternative is to use a patch instead.

Def not a blocker or anything (we have similar Updates all over the code base), but something I wanted to at least mention since it's been in the back of my mind with some recent changes (I've personally be leaning on patch a bit more).

internal/controller/pgcluster/pgclustercontroller.go
```diff
@@ -273,6 +273,15 @@ func (c *Controller) onUpdate(oldObj, newObj interface{}) {
 			return
 		}
 	}
+
+	// check to see if any of the annotations have been modified, in particular,
+	// the non-system annotations
+	if !reflect.DeepEqual(oldcluster.Spec.Annotations, newcluster.Spec.Annotations) {
```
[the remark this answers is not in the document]
jkatz: Sounds good. Let's cross that bridge when the time comes.

internal/operator/cluster/cluster.go
jkatz: Does Patroni reference that status, or is that merely for convenience?

pgo/cmd/cluster.go
jkatz: I may be missing something, but I don't see anything that validates the Annotation format in there.

jkatz: Add most I believe it would be `IsValidLabelValue`?

internal/operator/cluster/cluster.go
andrewlecuyer: I know that Patroni uses that annotation to obtain information about other instances in the cluster, e.g. when `patronictl list` is run.  To test this locally, I overwrote the status annotation in a pod and quickly ran `patronictl list`, which resulted in an error:

```
$ kubectl annotate pod mycluster1-uaql-7dccc96fc6-xcfxg status="something" --overwrite    
pod/mycluster1-uaql-7dccc96fc6-xcfxg annotated
[vagrant@client-centos crunchy-containers]$ kubectl exec -it mycluster1-bc47bff86-k4rwg -- patronictl list                      
Traceback (most recent call last):
  File "/usr/local/bin/patronictl", line 11, in <module>
    sys.exit(ctl())
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/patroni/ctl.py", line 829, in members
    output_members(cluster, cluster_name, extended, fmt)
  File "/usr/local/lib/python3.6/site-packages/patroni/ctl.py", line 763, in output_members
    len(set(m['host'] for m in cluster['members'])) < len(members)
  File "/usr/local/lib/python3.6/site-packages/patroni/ctl.py", line 763, in <genexpr>
    len(set(m['host'] for m in cluster['members'])) < len(members)
KeyError: 'host'
command terminated with exit code 1
```

However, right after that Patroni fixes the annotation and all starts working again.

```
[vagrant@client-centos crunchy-containers]$ kubectl exec -it mycluster1-bc47bff86-k4rwg -- patronictl list                      
+ Cluster: mycluster1 (6863113107911381134) -----+--------+---------+----+-----------+
|              Member              |     Host    |  Role  |  State  | TL | Lag in MB |
+----------------------------------+-------------+--------+---------+----+-----------+
|    mycluster1-bc47bff86-k4rwg    | 10.44.0.165 | Leader | running |  1 |           |
| mycluster1-uaql-7dccc96fc6-xcfxg | 10.44.0.168 |        | running |  1 |         0 |
+----------------------------------+-------------+--------+---------+----+-----------+
```

So my guess based on the above is that Patroni will just re-add it if we overwrite it when adding any custom annotations (and hopefully without Patroni clobbering any of our annotations when it is re-added).

jkatz: @andrewlecuyer So...the annotations are being added to the Deployment, which then get propagated to the Pod. I'm going to observe a bit of what happens to ensure everything is playing nicely.

internal/operator/backrest/repo.go
jkatz: Sounds like a good topic as we continue to evolve the Operator codebase.

internal/operator/cluster/cluster.go
jkatz: @andrewlecuyer OK, here is why it's OK. It is because, in fact, we are editing the annotations at the Deployment level, which get propagated to the Pod.

When the modifications take place, the Pod must be destroyed anyway, so regardless, Patroni will have to add a status label to it.

So this is fine.

internal/operator/backrest/repo.go
cbandy: I have similar thoughts about patching. As is, this reads well and has boundaries that look good, but I also know we should try to write entire objects less often.

As an example, here's one way to replace the annotations using a JSON Patch:

```go
patch, err := json.Marshal([]util.JSONPatchOperation{{
  Op: "add",
  Path: "/spec/template/metadata/annotations",
  Value: annotations,
}})

if err == nil {
  _, err = clientset.AppsV1().Deployments(cluster.Namespace).Patch(deploymentName, types.JSONPatchType, patch)
}

return err
```

docs/content/pgo-client/reference/pgo_update_cluster.md
```diff
@@ -22,6 +22,17 @@ pgo update cluster [flags]
 
 ```
       --all                              all resources.
+      --annotation strings               Add an Annotation to all of the managed deployments (PostgreSQL, pgBackRest, pgBouncer)The format to add an annotation is "name=value"
```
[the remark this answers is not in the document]
jkatz: This is the autoformatting from the command generation script, so not going to fix.

PR #1797, opened 2020-08-19
Remove `pgo load`

jkatz:
**Checklist:**

 
**Type of Changes:**

 
The functionality had not worked in awhile, the interface
could present challenges to use, and there are other was
that data can presently be loaded.

This gives us an opportunity to explore future implementations
on how we can permit data loading into PostgreSQL, but likely
this is application specific.

**Other information**:

Issue: [ch9061]


internal/apiserver/perms.go
```diff
@@ -117,7 +116,6 @@ func initializePerms() {
 		CLONE_PERM:        "yes",
 		DF_CLUSTER_PERM:   "yes",
 		LABEL_PERM:        "yes",
-		LOAD_PERM:         "yes",
```
cbandy, line 120: If a role has been granted this permission during an earlier version, will the role continue to work (for other granted functionality) after an upgrade?

jkatz: Well it wouldn't be considered at all.

andrewlecuyer: Maybe something our upgrade code should take care of, i.e. removing references to this now-defunct permission?

PR #1801, opened 2020-08-20
Update the Restore Workflow to Use the '--restore-from' Approach

andrewlecuyer:
When a PostgreSQL cluster restore is requested (e.g. via `pgo restore mycluster` when using the `pgo` client, when calling the `restore` REST endpoint directly, or when submitting a pgBackRest restore pgtask), the workflow utilized when creating a new cluster from an existing backup (i.e. the workflow used when the `--restore-from` flag is specified when creating a new cluster via the `pgo` client) is now leveraged.  This means the custom restore workflow, along with the `pgo-restore` container, are no longer utilized when performing a restore.

This change consolidates all restore logic for the Operator (with the exception of `clone`, which is now deprecated) into a single workflow.  Therefore, whether performing a restore for a existing cluster (e.g. using `pgo restore`) or performing a restore to create a brand new cluster (e.g. using `pgo create cluster --restore-from`), the same bootstrap logic, Job type, etc. is utilized to perform the restore and initialize the cluster.  And once a restore is successful, now the same logic used to initialize the newly restored as is used when creating a new cluster from scratch.  This means that there is no longer restore-specific logic for creating PostgreSQL Deployments, resources, etc. during a restore.

In support of this update to the restore process, the following changes have been made:

- All existing PG PersistentVolumesClaims (PVCs) for the cluster (the primary and all replicas) will be deleted.  Therefore, data from prior to the restore is only retained according to the retention policy for the PersistentVolumes associated with the original PVCs.  This allows for consistent naming and cluster creation across various workflows, while the removal of the replica PVCs ensures the proper recreation of replicas post-restore.

- The restore pgtask has been renamed from `backrest-restore-<fromCluster>-to-<toPVC>` to `backrest-restore-<cluster>` since the name of the cluster is now always utilized to create the restored primary.

- The `toPVC` parameter has been removed from the `RestoreRequest` struct utilized in conjunction with the `restore` endpoint in the `apiserver`.  This is because a restore is no longer performed to a a PVC with a randomly generated ID.  Instead a PVC is created (or recreated if at already exists) matching the name of the cluster being restored (the `FromCluster` parameter has been maintained for backwards compatibility, and should still reference the name of the cluster being restored).

- When submitting a pgtask for a restore, the following parameters no longer need to be specified, since they can be derived from the cluster being restored:
    - `pgbackrest-stanza`
    - `pgbackrest-db-path`
    - `pgbackrest-repo-path`
    - `pgbackrest-repo-host`
    - `backrest-s3-verify-tls`

- Upon completion of the restore Job, the restore workflow pgtask is updated to show `restored Primary created` and not `restored PVC created`.  This is because the bootstrap job now initializes an actual primary via Patroni, instead of simply restoring into a PVC.

- The pgBackRest repo is re-deployed and a stanza-create Job is run post-restore, per the standard `--restore-from` workflow (generally the post-restore stanza-create Job should be a no-op since the stanza should already exist from prior to the restore).

- The `<clusterName>-config` and `<clusterName>-leader` ConfigMaps for the cluster are now deleted and re-created (instead of creating new 'config' and 'leader' ConfigMaps with a different name).

- The `post-existing-init.sql` script is running following the restore, and any modules included in the PostgreSQL container are reloaded.

Ultimately this change paves the way for full removal of the `pgo-restore` container (which can be removed once `clone` is removed from the Operator), while further consolidating and streamlining bootstrap/restore and cluster creation processes and logic.

**Checklist:**

 
**Type of Changes:**

 
Restores use a custom workflow.

[ch8904]

Restores leverage the `--restore-from` workflow.

**Other information**:

N/A


internal/controller/pgtask/backresthandler.go
```diff
@@ -0,0 +1,65 @@
...
+	cluster, err := c.Client.CrunchydataV1().Pgclusters(namespace).Get(clusterName,
+		metav1.GetOptions{})
+	if err != nil {
+		log.Error("pgtask Controller: " + err.Error())
+		return
+	}
+
+	cluster, err = backrestoperator.PrepareClusterForRestore(c.Client, cluster, task)
+	if err != nil {
+		log.Errorf("pgtask Controller: " + err.Error())
```
cbandy, line 43: This and a few other lines use `Errorf` but no arguments. I think `go vet` (in recent versions?) issues a warning for this.

internal/apiserver/backrestservice/backrestimpl.go
```diff
@@ -539,14 +539,23 @@ func Restore(request *msgs.RestoreRequest, ns, pgouser string) msgs.RestoreRespo
...
+
 	//create a pgtask for the restore workflow
 	if _, err := apiserver.Clientset.CrunchydataV1().Pgtasks(ns).Create(pgtask); err != nil {
 		resp.Status.Code = msgs.Error
 		resp.Status.Msg = err.Error()
 		return resp
 	}
 
-	resp.Results = append(resp.Results, "restore performed on "+request.FromCluster+" to "+request.ToPVC+" opts="+request.RestoreOpts+" pitr-target="+request.PITRTarget)
+	resp.Results = append(resp.Results, fmt.Sprintf("restore request for %s with opts '%s' and pitr-target='%s'",
```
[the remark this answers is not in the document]
andrewlecuyer: Updated to use `%q`.

internal/operator/backrest/restore.go
```diff
@@ -61,235 +56,171 @@ type BackrestRestoreJobTemplateFields struct {
...
 	}
-
-	publishRestore(cluster.ObjectMeta.Labels[config.LABEL_PG_CLUSTER_IDENTIFIER], clusterName, task.ObjectMeta.Labels[config.LABEL_PGOUSER], namespace)
-
-	err = updateWorkflow(clientset, workflowID, namespace, crv1.PgtaskWorkflowBackrestRestoreJobCreatedStatus)
-	if err != nil {
-		log.Error(err)
-		log.Error("restore workflow: error in updating workflow status")
+	if err := clientset.CoreV1().ConfigMaps(namespace).Delete(fmt.Sprintf("%s-config",
+		clusterName), &metav1.DeleteOptions{}); err != nil && !kerrors.IsNotFound(err) {
```
cbandy, line 204: It looks like this delete already happened on L198 as part of the slice of names. Does this second call behave differently?

andrewlecuyer: Good catch, went ahead and removed this deletion.

PR #1802, opened 2020-08-20
Allow Resource Configuration for Crunchy Postgres Exporter Container

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
The PostgreSQL Operator does not have a way to set the resource
parameters for the Crunchy Postgres Exporter container.

This commit allows for resource configuration for the Crunchy Postgres
Exporter container, either during initial cluster creation, or after
creation as an update. It allows for initial CPU request and the
corresponding limit, as well as the initial memory request, again with
a corresponding limit. The flags include:

pgo create cluster --exporter-cpu
pgo update cluster --exporter-cpu
pgo create cluster --exporter-cpu-limit
pgo update cluster --exporter-cpu-limit
pgo create cluster --exporter-memory
pgo update cluster --exporter-memory
pgo create cluster --exporter-memory-limit
pgo update cluster --exporter-memory-limit

**Other information**:

[ch7035]


internal/apiserver/clusterservice/clusterimpl.go
```diff
@@ -1909,6 +1972,34 @@ func UpdateCluster(request *msgs.UpdateClusterRequest) msgs.UpdateClusterRespons
 			cluster.Spec.BackrestResources[v1.ResourceMemory] = quantity
 		}
 
+		// ensure there is a value for ExporterResources
+		if cluster.Spec.ExporterResources == nil {
+			cluster.Spec.ExporterLimits = v1.ResourceList{}
+			cluster.Spec.ExporterResources = v1.ResourceList{}
+		}
```
cbandy, line 1979: On plain Kubernetes containers, limit-only implies limit == request. Does our code allow one to specify only limits? If so how does it pass through the layers?

It looks like this block might incorrectly overwrite / zero-out `ExporterLimits` if they are populated at this point when `ExporterResources` are not.

tjmoore4: This is matching the existing pattern for the other resource requests.

PR #1839, opened 2020-09-01
Restrict check for pgBackRest primary Pods to running pods

### LINKED ISSUE #1835: Error: backrestrepo pod not found for cluster example-db
We just ran into an issue where we had more than one backrestrepo pod - i.e., 1 terminated and 1 active backrestrepo pod, so matching `!=1` (see commit link below). We thusly got:

    pgo show backup common-db -n example
    Error: backrestrepo pod not found for cluster example-db

When we threw away the terminated pod, the command succeeded again. This feels like a bug to me.

I found the following associated commit:
https://github.com/CrunchyData/postgres-operator/commit/22c792658d510bbf055286569ec4186aae6e4f40

Search for `backrestrepo pod not found` line, see:
https://github.com/CrunchyData/postgres-operator/commit/22c792658d510bbf055286569ec4186aae6e4f40#r41942528

jkatz:
There could be cases where a pgBackRest Pod is running while
another is terminating, and we want to ensure that the full
pgBackRest functionality is available when there is one (and
presently, only one) pgBackRest Pod available for the request.

Issue: [ch9155]
fixes #1835

### DIFF
--- a/internal/apiserver/backrestservice/backrestimpl.go
+++ b/internal/apiserver/backrestservice/backrestimpl.go
@@ -36,6 +36,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 )
 
 const containername = "database"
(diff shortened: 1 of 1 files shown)


internal/apiserver/backrestservice/backrestimpl.go
```diff
@@ -279,7 +280,13 @@ func getPrimaryPodName(cluster *crv1.Pgcluster, ns string) (string, error) {
 
 	//look up the backrest-repo pod name
 	selector := "pg-cluster=" + cluster.Spec.Name + ",pgo-backrest-repo=true"
-	repopods, err := apiserver.Clientset.CoreV1().Pods(ns).List(metav1.ListOptions{LabelSelector: selector})
+
+	options := metav1.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector("status.phase", string(v1.PodRunning)).String(),
+		LabelSelector: selector,
```
jmckulk, line 286: Why not add this to the options struct too? You would have to redefine `selector` somewhere else but could be cleaner
```suggestion
		LabelSelector: "pg-cluster=" + cluster.Spec.Name + ",pgo-backrest-repo=true",
```

jkatz: The main goal was to add the field selector and not make extensive changes to the code, so the primary issue could be fixed.

If I wer eto clean up the selector, I would follow a similar pattern to how we're constructing `FieldSelector`.

PR #1843, opened 2020-09-02
Fix --target Option Detection for Restores

andrewlecuyer:
A regex is now used to determine whether or not a `--target` flag already exists in the options provided for a pgBackRest restore. This avoids conflicts with other options containing the string "target" (e.g. `--target-action`), therefore ensuring the PITR target is properly injected into the pgBackRest options as needed.

**Checklist:**

 
**Type of Changes:**

 
PITR's fail if the `--target-action` flag is specified.

[ch9162]

PITR's no longer fail if the `--target-action` flag is specified.

**Other information**:

N/A

### DIFF
--- a/internal/operator/backrest/restore.go
+++ b/internal/operator/backrest/restore.go
@@ -18,6 +18,7 @@ package backrest
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -34,6 +35,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
+
+// restoreTargetRegex defines a regex to detect if a restore target has been specified
+// for pgBackRest using the '--target' option
+var restoreTargetRegex = regexp.MustCompile("--target(=| +)")
 
 type BackrestRestoreJobTemplateFields struct {
 	JobName                string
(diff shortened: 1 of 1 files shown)


internal/operator/backrest/restore.go
```diff
@@ -35,6 +36,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
+// restoreTargetRegex defines a regex to detect if a restore target has been specified
+// for pgBackRest using the '--target' option
+var restoreTargetRegex = regexp.MustCompile("--target(=| +)")
```
jkatz, line 41: any reason why it's not `\s+`? I did a few double takes on that.

jkatz (approved):

internal/operator/backrest/restore.go
andrewlecuyer: No specific reason, just different syntax.

cbandy: In this context, `\s` seems more appropriate. We could go the other direction and match `--target[^-]`, but it doesn't feel optimal either.

andrewlecuyer: Also, was really just looking to match on a spaces, and not necessarily also on tabs and newlines (not that those should really ever be in there).

cbandy (approved):

PR #1845, opened 2020-09-02
Shorter pgrestore Job Names

andrewlecuyer:
The name of the PVC being utilized for a pgrestore Job is now no longer included in the Job name.  This ensures the 63 character limit for Kubernetes Job naming is not easily exceeded when creating pgrestore Jobs, i.e. for clusters with names 
longer than simply a few characters.

**Checklist:**

 
**Type of Changes:**

 
pgrestore Jobs fail to create with a `must be no more than 63 characters` error if the cluster name is longer than a few characters.

[ch9163]

pgrestore Jobs no longer fail to create if the cluster name is longer than a few characters.

**Other information**:

N/A

### DIFF
--- a/internal/operator/pgdump/restore.go
+++ b/internal/operator/pgdump/restore.go
@@ -18,6 +18,7 @@ package pgdump
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"os"
 
 	"github.com/crunchydata/postgres-operator/internal/config"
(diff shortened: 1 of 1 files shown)


jkatz (approved):

internal/operator/pgdump/restore.go
```diff
@@ -75,7 +76,8 @@ func Restore(namespace string, clientset kubeapi.Interface, task *crv1.Pgtask) {
 	taskName := task.Name
 
 	jobFields := restorejobTemplateFields{
-		JobName:             "pgrestore-" + task.Spec.Parameters[config.LABEL_PGRESTORE_FROM_CLUSTER] + "-from-" + fromPvcName + "-" + util.RandStringBytesRmndr(4),
+		JobName: fmt.Sprintf("pgrestore-%s-%s", task.Spec.Parameters[config.LABEL_PGRESTORE_FROM_CLUSTER],
+			util.RandStringBytesRmndr(4)),
```
cbandy, line 80: fwiw, [`k8s.io/apimachinery/pkg/util/rand.String()`](https://pkg.go.dev/k8s.io/apimachinery@v0.17.4/pkg/util/rand?tab=doc#String) generates similar random strings. I believe it's how Pods get their names: https://github.com/kubernetes/apiserver/blob/v0.19.0/pkg/storage/names/generate.go#L53

cbandy (approved):

PR #1894, opened 2020-09-16
Prevent Panic if All PG Deployments Scaled Down Outside of the PostgreSQL Operator

andrewlecuyer:
If the primary and all replica Deployments are scaled to zero outside of the context of the PostgreSQL Operator (e.g. by using
`kubectl scale deployment --replicas=0` instead of `pgo update cluster --shutdown`) and the associated `pgcluster` for the Deployment(s) being scaled is not manually set to the proper `shutdown` status, then the PostgreSQL Operator could panic as it continues to attempt to synchronize Patroni/PostgreSQL configuration for the PostgreSQL cluster (specifically because no pods are found for the cluster).  This commit protects against a panic in this scenario by simply logging an error when no Pods are found during an attempt to sync configuration.

**Checklist:**

 
**Type of Changes:**

 
A panic can occur if the primary and all replica Deployments are scaled to zero outside of the context of the PostgreSQL Operator and the associated `pgcluster` for the Deployment(s) being scaled is not manually set to the proper `shutdown` status.

[ch9278]

A panic will no longer occur if the primary and all replica Deployments are scaled to zero outside of the context of the PostgreSQL Operator, and instead an error will be logged when the Operator attempts to synchronize configuration for the PG cluster.

**Other information**:

N/A

### DIFF
--- a/internal/operator/config/localdb.go
+++ b/internal/operator/config/localdb.go
@@ -217,10 +217,10 @@ func (l *LocalDB) Update(configName string, localDBConfig LocalDBConfig) error {
 	return nil
 }
 
-// apply applies the configuration stored in the CusterConig's configMap for a
-// specific database server to that server.  This is done by updating the contents of that
-// database server's local configuration with the configuration for that cluster stored in
-// the LocalDB's configMap, and the issuing a Patroni "reload" for that specific server.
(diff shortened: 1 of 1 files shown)


jkatz (approved):
Should be backpatched

cbandy (approved):

PR #1914, opened 2020-09-23
Allow for bad replicas to be scaled down

jkatz:
**Checklist:**

 
**Type of Changes:**

 
Previously, the scaledown command would only list out healthy
replicas, not replicas that were in a bad state. However, these
may be the exact replicas that one would want to scale down,
given they are unhealthy.

Issue: [ch9253]

### DIFF
--- a/internal/util/failover.go
+++ b/internal/util/failover.go
@@ -79,6 +79,14 @@ const (
 	// instanceReplicationInfoTypePrimaryStandby is the label used by Patroni to indicate that an
 	// instance is indeed a primary PostgreSQL instance, specifically within a standby cluster
 	instanceReplicationInfoTypePrimaryStandby = "Standby Leader"
+	// instanceRolePrimary indicates that an instance is a primary
+	instanceRolePrimary = "primary"
+	// instanceRoleReplica indicates that an instance is a replica
+	instanceRoleReplica = "replica"
+	// instanceRoleUnknown indicates taht an instance is of an unknown typ
+	instanceRoleUnknown = "unknown"
+	// instanceStatusUnavailable indicates an instance is unavailable
+	instanceStatusUnavailable = "unavailable"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


internal/util/failover.go
```diff
@@ -137,20 +137,34 @@ func GetPod(clientset kubernetes.Interface, deploymentName, namespace string) (*
...
+
+	// if we are not including the primary, determine if we are including busted
+	// replicas or not
 	if !includePrimary {
-		roleSelector = fmt.Sprintf("%s=%s", config.LABEL_PGHA_ROLE, config.LABEL_PGHA_ROLE_REPLICA)
-	} else {
-		roleSelector = config.LABEL_PG_DATABASE
+		if includeBusted {
+			// include all Pods that identify as a database, but **not** a primary
+			// pg-cluster=clusterName,pgo-pg-database,role!=config.LABEL_PGHA_ROLE_PRIMARY
```
jmckulk, line 160: I think having something like `selector + role!=config.LABEL_PGHA_ROLE_PRIMARY` would make it more clear what this line is doing

internal/util/failover.go
```diff
@@ -137,20 +137,34 @@ func GetPod(clientset kubernetes.Interface, deploymentName, namespace string) (*
...
+			selector += fmt.Sprintf(",%s!=%s", config.LABEL_PGHA_ROLE, config.LABEL_PGHA_ROLE_PRIMARY)
+		} else {
+			// include all Pods that identify as a database and have a replica label
+			// pg-cluster=clusterName,pgo-pg-database,role=replica
```
jmckulk, line 164: same here `selector + role=replica`

jmckulk (approved):
Have a couple comments about the comments but the changes make sense to me

internal/util/failover.go
jkatz: This shows the full selector string though.

PR #1944, opened 2020-10-07
Use Patch objects for small updates

cbandy:
We use a variety of techniques to apply small patches to Kubernetes objects: `fmt.Sprintf`, `util.JSONPatchOperation`, `jsonpatch.CreateMergePatch`. This PR introduces two new types `kubeapi.JSON6902` and `kubeapi.Merge7386` to replace those when making small updates.

The version of `github.com/evanphx/json-patch` that we depend on is actually a Go module, so it should be imported as `github.com/evanphx/json-patch/v4`. The new types added here happen to eliminate our direct use of `github.com/evanphx/json-patch`, making it easier to switch to Go modules ourselves.

**Checklist:**

 
**Other information**:

Issue: [ch9453]


andrewlecuyer:
This is a really nice patch.  Great to get rid of `evanphx/json-patch` and get on a single/consistent approach for patching.

The one question I have is in terms of simplicity/ease-of-use when it comes to patching (something I think we'll be doing a lot more of moving forward).  Specifically, when working with these various patch types recently (json and merge), I recently started reaching for the `Sprintf` approach more often, because of just how easy it made testing patches (especially those that are more complex).  For instance, with a `Sprintf` patch such as follows:

```bash
replicaPatch := fmt.Sprintf(`{"metadata":{"annotations":{"%s":null}},"spec":{"status":""},`+
		`"status":{"message":"%s","state":"%s"}}`, config.ANNOTATION_PGHA_BOOTSTRAP_REPLICA,
		"Cluster is being restored", crv1.PgclusterStateRestore)
```

I can quickly and easily copy/paste to turn that into a patch string to test with `kubectl`:

```bash
kubectl patch pgreplica mycluster1-rniv --type=merge -p='{"metadata":{"annotations":{"current-primary":null}},"spec":{"status":""},
    "status":{"message":"Cluster is being restored","state":"pgcluster Restoring"}}'
```

This can be especially helpful with the JSON patch type, which I've found can return some very generic error messages. For instance, and invalid `path` such as:

```bash
[{"op":"remove","path":"/invalid/path","value":"test"}]
```

...returns the following non-descriptive error:

```bash
"status":"Failure","message":"the server rejected our request due to an error in our request","reason":"Invalid"
```

When a JSON patch has many different `add`/`remove`/`replace's`, tracking down exactly what is causing this very generic error can be messy.  And since compiling, re-running, etc. the code to troubleshoot/track-down some minor issue can be also often be time consuming, the ability to quickly/easily reach for `kubectl` to first get the patch right via the CLI can be really nice.  With the `Sprintf` approach you pretty much know exactly what the final patch will look like (i.e. with a few quick string replacements you're seeing exactly what will be sent to the Kube API server), making it easy to quickly/easily adjust a patch as needed to until the desired result is achieved.

I will say that the solution you provided will likely prevent a lot of invalid things from slipping in in the first place, e.g. thanks to things like proper pointer escaping and the ability to use existing types.  Therefore, at a minimum I think we should consider adding some additional debugging to show what the actual patches look like prior to being applied.

As we expand our use of patching I can see these things getting more complex, and just want to be sure we have an effective way to easily test patches, and troubleshoot/diagnose any issues.  And knowing exactly what was sent to the Kube API server when troubleshooting one of those generic patch errors out in the wild will be very useful from a support perspective.

Anyway, just curious your thoughts on the above in general.  And regardless, definitely like the idea of standardizing on a single method for patching moving forward.

cbandy:
> This is a really nice patch. Great to get rid of `evanphx/json-patch` and get on a single/consistent approach for patching.

Thanks! I see [`evanphx/json-patch` still used quite often](https://pkg.go.dev/github.com/evanphx/json-patch?tab=importedby) and I didn't set out to eliminate it, but it turns out we were only using it for some very small patches. 

> The one question I have is in terms of simplicity/ease-of-use when it comes to patching (something I think we'll be doing a lot more of moving forward). Specifically, when working with these various patch types recently (json and merge), I recently started reaching for the `Sprintf` approach more often, because of just how easy it made testing patches (especially those that are more complex). For instance, with a `Sprintf` patch such as follows:
> 
> ```shell
> replicaPatch := fmt.Sprintf(`{"metadata":{"annotations":{"%s":null}},"spec":{"status":""},`+
> 		`"status":{"message":"%s","state":"%s"}}`, config.ANNOTATION_PGHA_BOOTSTRAP_REPLICA,
> 		"Cluster is being restored", crv1.PgclusterStateRestore)
> ```
> 
> I can quickly and easily copy/paste to turn that into a patch string to test with `kubectl`:
> 
> ```shell
> kubectl patch pgreplica mycluster1-rniv --type=merge -p='{"metadata":{"annotations":{"current-primary":null}},"spec":{"status":""},
>     "status":{"message":"Cluster is being restored","state":"pgcluster Restoring"}}'
> ```

That is easy, but I worry about quoting and escaping. The way we use it now with constants is safe, but we shouldn't ever send user-input into something like this.

I also found it hard to read, myself. When all the arguments are at the end, I find myself counting to try to match values to fmt verbs. A map of key/value pairs is much easier for me to follow.

That said, I don't exactly like the patch interfaces as they are now. The variadic path at the end is convenient to implement but my key/value-brain wants the path first followed by the value. [1]

> This can be especially helpful with the JSON patch type, which I've found can return some very generic error messages. For instance, and invalid `path` such as:
> 
> ```shell
> [{"op":"remove","path":"/invalid/path","value":"test"}]
> ```
> 
> ...returns the following non-descriptive error:
> 
> ```shell
> "status":"Failure","message":"the server rejected our request due to an error in our request","reason":"Invalid"
> ```
> 
> When a JSON patch has many different `add`/`remove`/`replace's`, tracking down exactly what is causing this very generic error can be messy.

I completely agree; JSON patch is not great to work with. I do not want to be creating a JSON patch with lots of operations. My impression is that if we want to "change a lot" of something we should be doing an Update()--_maybe_ an automated patch like `evanphx/json-patch`.

> I will say that the solution you provided will likely prevent a lot of invalid things from slipping in in the first place, e.g. thanks to things like proper pointer escaping and the ability to use existing types. Therefore, at a minimum I think we should consider adding some additional debugging to show what the actual patches look like prior to being applied.
> 
> As we expand our use of patching I can see these things getting more complex, and just want to be sure we have an effective way to easily test patches, and troubleshoot/diagnose any issues. And knowing exactly what was sent to the Kube API server when troubleshooting one of those generic patch errors out in the wild will be very useful from a support perspective.

Debug statements are a great idea. I probably wouldn't want them for _every_ patch, e.g. one-liner "set single field to X"; do you see any in this PR you'd want to be logged?

> Anyway, just curious your thoughts on the above in general. And regardless, definitely like the idea of standardizing on a single method for patching moving forward.

In general, my hope is that we do less patching and more declarative replacing (at/from the operator). 🤔 Metadata and Status feel right for patching and ConfigMaps/Secrets I suppose, but digging around in a CR, Pod, or Deployment less so...

If we do use patches to focus on particular fields, I think the automatic marshaling of k8s types is going to be vital.

---

[1]: 🤔 Perhaps there's room for more building, though I don't want to own code that's too complicated. I might try out tje last style below if I have some time. Thoughts?
```go
patch.At("some", "path").Add(value).
    At("another").Replace(value)

// Remove doesn't take a value. Perhaps the original spelling still makes sense.
patch.Remove("another")
patch.At("another").Remove()

// Closures are another way to fiddle with variadics.
patch.Add("some", "path")(value).
    Replace("another")(value)
```

cbandy:
> Perhaps there's room for more building…

I've gone ahead and made this change in another commit (to be squashed), and I think it turned out well.
- The value comes after the path.
- The extra parentheses make a stronger distinction between the path and the value.
- The compiler tells you when you've made a mistake.
  e.g. `kubeapi.NewMergePatch().Add("metadata", "labels").Bytes()` →
  `Bytes undefined (type func(interface {}) *kubeapi.Merge7386 has no field or method Bytes)`

The only downside is that in a couple places, we're bitten by https://golang.org/issue/41889:
- The last two hunks in `internal/kubeapi/patch_test.go`, and
- Two hunks in `internal/operator/backrest/restore.go`.

andrewlecuyer:
Nice, looks good.  My mind definitely also thinks path followed by value, so I can definitely appreciate this change.  I honestly also liked the `At()` idea as well, but at this point its really just a stylistic decision -  I'm really good with any of these various approaches at the end of the day, including the initial design.

As for debugging, I would honestly be a fan of logging all patch requests (specifically when the debug flag is `true`).  Since even a single add/remove/replace can result in a non-descriptive/generic error, I would rather just have as much information as possible when we run into patch issues in the wild (which has occurred in the past).  There is probably a relation between this and adding true logging levels to the Operator as well (as recently discussed), e.g. this might be a good candidate for a "trace" log level.  However, in the meantime definitely want to ensure our existing logging infrastructure can provide these details, and I would therefore rather lean on the side of providing as much information as possible.

cbandy:
Added a commit (to be squashed) that logs the patch body before every API patch call.

andrewlecuyer (approved):
Thanks for adding the debug statements.  Looks great otherwise.

PR #1946, opened 2020-10-07
Update OLM packaging to use python-yq

cbandy:
**Checklist:**

 
The scripts that make and test OLM CSVs rely heavily on an old version of `yq`. New contributors don't (and won't) have this version installed.

Those scripts now rely on a recent version of a _different_ `yq` that can be installed with `pip`. It behaves more consistently than the other utility and leverages `jq` to provide the same syntax and features we're already using.


jkatz (approved):

cbandy:
I see some things in the output that I don't like. Hold on this until I can make some changes.

cbandy:
Alright, I've pivoted completely and switched YAML utilities. This one is [_also_ called `yq`](https://kislyuk.github.io/yq/) but is more of a veneer over `jq`. It can be installed on any system using `pip install yq` and `brew install python-yq` on macOS.

jkatz (approved):

PR #1954, opened 2020-10-10
Update logic for pgbouncer password modification on creation

jkatz:
If the AddPgBouncer call is invoked as the Operator attempts to
add a new pgBouncer and it is determined that a pgBouncer Secret
already exists, do not attempt to update the password within
PostgreSQL.

Issue: [ch9457]


jkatz:
Per suggestion from @andrewlecuyer I have also set this up so that the value in the pgBouncer Secret, should it already exist, is synchronized with that of the representative `pgbouncer` user in the PostgreSQL database.

PR #1648, opened 2020-06-21
Goodbye dep, hello go modules

jkatz:
**Checklist:**

 
 - [x] Have you added an explanation of what your changes do and why you'd like them to be included?
 - [x] Have you updated or added documentation for the change, as applicable?
 - [x] Have you tested your changes on all related environments with successful results, as applicable?

**Type of Changes:**

 
 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] New feature (non-breaking change which adds functionality)
 - [x] Breaking change (fix or feature that would cause existing functionality to change)

**What is the current behavior? (link to any open issues here)**

Dep is the dependency manager.

**What is the new behavior (if this is a feature change)?**

An unintended side effect of #1576 and others is that the
directory structure is now conducive to using go modules. After
my latest wrangling with dep to try to do some routine upgrades,
I thought it prudent to attempt to make the switch. It turns
out that it was not too bad this go around, so now seems as likely
of a time to try to make the switch.

**Other information**:

Issue: #1647

PR #1979, opened 2020-10-17
Switch to Go modules to manage dependencies

### LINKED ISSUE #1647: Errors will happen when upgrading the libraries
(The purpose of this report is to alert `CrunchyData/postgres-operator` to the possible problems when `CrunchyData/postgres-operator` try to upgrade the following dependencies)

### github.com/spf13/cobra  
-Version: V1.0.0   (Latest commit a684a6d on 11 Apr)
-Where did you use it: 
https://github.com/CrunchyData/postgres-operator/search?q=spf13%2Fcobra%2Fdoc&unscoped_q=spf13%2Fcobra%2Fdoc
-Detail: 
>>[github.com/spf13/cobra/go.mod](https://github.com/spf13/cobra/blob/v1.0.0/go.mod#L6)
```
module github.com/spf13/cobra
go 1.12
require (
	github.com/cpuguy83/go-md2man/v2 v2.0.0
	github.com/inconshreveable/mousetrap v1.0.0
	github.com/mitchellh/go-homedir v1.1.0
	github.com/spf13/pflag v1.0.3
	github.com/spf13/viper v1.4.0
	gopkg.in/yaml.v2 v2.2.2
)
```

>>[github.com/spf13/cobra/doc/man_docs.go](https://github.com/spf13/cobra/blob/v1.0.0/doc/man_docs.go#L27)
```
package doc
import (
	"github.com/cpuguy83/go-md2man/v2/md2man"
	"github.com/spf13/cobra"
	"github.com/spf13/pflag"
)
```

This problem was introduced since **_spf13/cobra v0.0.6_** . Now you used version v0.0.5 . If you try to upgrade  **spf13/cobra** to version **v0.0.6** and **above**, you will get an error--- **_no package exists at "github.com/cpuguy83/go-md2man/v2/md2man"_**

## I investigated the libraries'  (**spf13/cobra >= v0.0.6**)  release information and found the root cause of this issue is that---- 
1.	These dependencies all added [Go modules](https://github.com/golang/go/wiki/Modules) in the recent versions.

2.	They all comply with the specification of "Releasing Modules for v2 or higher" available in the Modules documentation. Quoting the specification:
> A package that has migrated to Go Modules _must_ include the major version in the import path to reference any v2+ modules. For example,  Repo github.com/my/module migrated to Modules on version v3.x.y.  Then this repo should declare its module path with MAJOR version suffix "/v3"  (e.g., module `github.com/my/module/v3`), and its downstream project should use `"github.com/my/module/v3/mypkg"` to import this repo’s package.

3.	This "github.com/my/module/v3/mypkg" is not the `physical path`. So earlier versions of Go (including those that don't have minimal module awareness) plus all tooling (like dep, glide, govendor, etc) don't have `minimal module awareness` as of now and therefore don't handle import paths correctly See [golang/dep#1962](https://github.com/golang/dep/issues/1962), [golang/dep#2139](https://github.com/golang/dep/issues/2139).
> Note: creating a new branch is not required. If instead you have been previously releasing on master and would prefer to tag v3.0.0 on master, that is a viable option. (However, be aware that introducing an incompatible API change in master can cause issues for non-modules users who issue a go get -u given the go tool is not aware of semver prior to Go 1.11 or when module mode is not enabled in Go 1.11+).
> Pre-existing dependency management solutions such as dep currently can have problems consuming a v2+ module created in this way. See for example dep#1962.
https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher

### 1. Migrate to Go Modules.
Go Modules is the general trend of ecosystem, if you want a better upgrade package experience, migrating to Go Modules is a good choice.

Migrate to modules will be accompanied by the introduction of virtual paths(It was discussed above).
> This "github.com/my/module/v3/mypkg" is not the `physical path`. So Go versions older than 1.9.7 and 1.10.3 plus all third-party dependency management tools (like dep, glide, govendor, etc) don't have `minimal module awareness` as of now and therefore don't handle import paths correctly.

Then the downstream projects might be negatively affected in their building if they are module-unaware (Go versions older than 1.9.7 and 1.10.3; Or use third-party dependency management tools, such as: Dep, glide, govendor…).

### 2. Maintaining v2+ libraries that use Go Modules in Vendor directories. 
If `CrunchyData/postgres-operator` want to keep using the dependency manage tools (like dep, glide, govendor, etc), and still want to upgrade the dependencies, can choose this fix strategy.
Manually download the dependencies into the vendor directory and do compatibility dispose(materialize the virtual path or delete the virtual part of the path). Avoid fetching the dependencies by virtual import paths. This may add some maintenance overhead compared to using modules.

There are 3 module users downstream, such as jufi-git/postgres-operator, CrunchyData/pgo-osb, CrunchyData/postgres-operator…)
https://github.com/search?q=CrunchyData%2Fpostgres-operator+filename%3Ago.mod

As the import paths have different meanings between the projects adopting module repos and the non-module repos, materialize the virtual path is a better way to solve the issue, while ensuring compatibility with downstream module users. A textbook example provided by repo `github.com/moby/moby` is here:
https://github.com/moby/moby/blob/master/VENDORING.md
https://github.com/moby/moby/blob/master/vendor.conf
In the vendor directory, `github.com/moby/moby` adds the **/vN** subdirectory in the corresponding dependencies.
**This will help more downstream module users to work well with your package.**

### 3. Request upstream to do compatibility processing.
The `spf13/cobra` have 1212 module-unaware users in github, such as: maliceio/cli, Azure/acr-docker-credential-helper, alyyousuf7/spotifyshare…
https://github.com/search?q=spf13%2Fcobra+filename%3Avendor.conf+filename%3Avendor.json+filename%3Aglide.toml+filename%3AGodep.toml+filename%3AGodep.json&type=Code

## Summary
You can make a choice when you meet this DM issues by balancing your own development schedules/mode against the affects on the downstream projects.

For this issue, Solution 1 can maximize your benefits and with minimal impacts to your downstream projects the ecosystem. 

## References
* https://github.com/golang/go/wiki/Modules#semantic-import-versioning
* https://golang.org/cmd/go/#hdr-Module_compatibility_and_semantic_versioning
* https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher

Do you plan to upgrade the libraries in near future?
Hope this issue report can help you ^_^
Thank you very much for your attention.

Best regards,
Kate

cbandy:
**Checklist:**

 
**Type of Changes:**

 
The `go` command is sufficient to install dependencies, run tests, and build binaries. We no longer depend on the `GOPATH` or `GOBIN` environment variables, and the project can reside outside your GOPATH. 

**Other Information**:

Once you have these changes locally, you'll want to run `make clean-deprecated` or `make clean` to remove build artifacts leftover from the old ways.

Replaces: #1648
Issue: [ch9453]
fixes #1647


docs/content/contributing/developer-setup.md
```diff
@@ -21,14 +21,10 @@ The following environment variables are expected by the steps in this guide:
...
 Variable | Example | Description
 -------- | ------- | -----------
-`GOPATH` | $HOME/odev | Golang project directory
-`PGOROOT` | $GOPATH/src/github.com/crunchydata/postgres-operator | Operator repository location
+`PGOROOT` | $HOME/odev/postgres-operator | Operator repository location
 `PGO_CONF_DIR` | $PGOROOT/installers/ansible/roles/pgo-operator/files | Operator Config Template Directory
-`PGO_BASEOS` | {{< param centosBase >}} | Base OS for container images
 `PGO_CMD` | kubectl | Cluster management tool executable
-`PGO_IMAGE_PREFIX` | crunchydata | Container image prefix
 `PGO_OPERATOR_NAMESPACE` | pgo | Kubernetes namespace for the operator
```
andrewlecuyer, line 27: Was just curious why we're removing some of these (`PGO_BASEOS`, `PGO_OPERATOR_NAMESPACE` & `PGO_VERSION`) and not the others.  All of these appear to be defined in the default Makefile, so just wondering which we're keeping in this list.

Also, would `PGO_OPERATOR_NAMESPACE` be more Bash install thing vs. dev setup?  I do see we have a `Deployment` section in the instructions that references what is essentially the Bash install, which would require this env var.  Maybe we should just point to the bash install instructions instead (especially considering the various other setup, e.g. for things like RBAC, CRD's, etc, are really needed to properly deploy)?

Makefile
```diff
@@ -73,10 +71,12 @@ ifeq ("$(PGO_BASEOS)", "centos8")
 endif
 
 DEBUG_BUILD ?= false
-GCFLAGS=
+GO_BUILD = $(GO_CMD) build
```
andrewlecuyer, line 74: Any chance we might still be able to allow for `install` vs `build`?

I currently leverage this in conjunction with my `GO_BIN` to quickly/easily launch our various Operator applications as part of my typical development process, since `GO_BIN` is already part of my default`PATH` (with this change I suppose I could also make `$PGOROOT/bin` part of my standard `PATH`).

Makefile
```diff
@@ -94,76 +94,65 @@ images = pgo-apiserver \
...
 cleannamespaces:
-	cd deploy && ./cleannamespaces.sh
-
-bounce:
-	$(PGO_CMD) \
-		--namespace=$(PGO_OPERATOR_NAMESPACE) \
-		get pod \
-		--selector=name=postgres-operator \
-		-o=jsonpath="{.items[0].metadata.name}" \
-	| xargs $(PGO_CMD) --namespace=$(PGO_OPERATOR_NAMESPACE) delete pod
```
andrewlecuyer, line 123: With this section gone, can `PGO_OPERATOR_NAMESPACE` be removed from the Makefile?

bin/get-deps.sh
```diff
@@ -57,36 +37,21 @@ if ! [ -f $EVTDIR/nsqd -a -f $EVTDIR/nsqadmin ]; then
...
-	echo "=== Installing docker ==="
-	if [ -f /etc/centos-release ]; then
-		sudo yum -y install docker
-	else
-		sudo yum -y install docker --enablerepo=rhel-7-server-extras-rpms
-	fi
+if ! command -v docker &> /dev/null && ! command -v podman &> /dev/null; then
+	echo 'Cannot find `docker` nor `podman`. Perhaps:'
+	echo '  sudo yum install docker'
+	exit 1
```
andrewlecuyer, line 43: Since technically neither `docker` nor `podman` are required, should we remove this check?

For instance, if using this script from a CI server, I wouldn't want it to simply exit without the presence of either or these two tools, since I might be using something like `buildah` and `crio` only (and would still want to be able to grab all of the other dependencies, despite this check failing).

andrewlecuyer (commented):
Looks great, just a few questions/comments.

bin/get-deps.sh
cbandy: Yep, will do.

docs/content/contributing/developer-setup.md
cbandy: > Was just curious why we're removing some of these (`PGO_BASEOS`, `PGO_OPERATOR_NAMESPACE` & `PGO_VERSION`) and not the others. All of these appear to be defined in the default Makefile, so just wondering which we're keeping in this list.

I intended to remove only the variables related to building images because those are only used in the Makefile. However, I see that `PGO_IMAGE_PREFIX` is also used in `bin/push-to-gcr.sh`.

> Also, would `PGO_OPERATOR_NAMESPACE` be more Bash install thing vs. dev setup? I do see we have a `Deployment` section in the instructions that references what is essentially the Bash install, which would require this env var. Maybe we should just point to the bash install instructions instead (especially considering the various other setup, e.g. for things like RBAC, CRD's, etc, are really needed to properly deploy)?

I'm fine with linking to the Bash install docs for the Deployment section. I really just punted on figuring out when/why `PGO_CMD` and `PGO_OPERATOR_NAMESPACE` should be set because they are used throughout the `deploy` directory.

Makefile
cbandy: Yep. Looks like `PGO_CMD` can go, too.

Makefile
cbandy: I don't like `install` because it creates things outside the project directory.

I use `./bin/pgo` to run the client, but (most of) the other binaries go into subdirectories of `./bin` which I can see being a hassle. I'd like to move those up into `./bin` so that `$PGOROOT/bin` can be added to `PATH` as you suggest. Do you know of anything besides the Dockerfiles that would be affected by that?

cbandy: So, most executables are named the same as a non-empty subdirectory in `bin`. I can move `postgres-operator` up alongside `pgo` and `apiserver` at least. Would that alleviate most of the pain?

cbandy:
> 2. I'd be completely fine with eliminating `odev` and its reference entirely.

Done.

I've addressed most comments by amending the last two commits.

Makefile
andrewlecuyer: I am definitely good with that approach.  With `pgo`, `postgres-operator` and `apiserver` being the most common binaries run via my `PATH`, this should definitely alleviate most pain (I am also fine with this approach considering that some of the various other binaries such as `pgo-scheduler`, `pgo-rmdata` & `pgo-backrest` will likely change in some way or another moving forward, e.g. `pgo-backrest` will soon be moved to the Container Suite project as part of compaction).

cbandy: Done.

andrewlecuyer (approved):

PR #1989, opened 2020-10-21
Add kustomize installation method

### LINKED ISSUE #1984: Installing without ansible/pgo-deployer
**What is the motivation or use case for the feature? **

Currently the only documented installation methods are all via ansible. Even the helm chart just creates a job (pgo-deployer) that runs ansible. This is very unusual, and does not play well with generally accepted means for managing kubernetes clusters.

A collection of kubernetes manifests to apply to a cluster that include all components that will be running long term in the cluster. These could be templated via helm; or (preferably) ready to use with kustomize.

james-callahan:
This is a Work-In-Progress attempt at creating a kustomize base for postgres-operator.

I created this by manually running through the ansible scripts and attempting to create minimal sensible defaults. I'm still trying to figure out how it's all put together.

Closes #1984


ValClarkson:
Hi @james-callahan this is a great start, thanks for contributing.  

Looks like you are missing some secrets and SA's needed for the pgocli 
```
kubectl get secret -n pgo
NAME                            TYPE                                  DATA   AGE
default-token-w6xvg             kubernetes.io/service-account-token   3      5d5h
pgo-backrest-repo-config        Opaque                                5      5d5h
pgo-backrest-token-h6zbw        kubernetes.io/service-account-token   3      5d5h
pgo-default-token-w6vk8         kubernetes.io/service-account-token   3      5d5h
pgo-deployer-sa-token-75bq4     kubernetes.io/service-account-token   3      5d5h
pgo-pg-token-sdzss              kubernetes.io/service-account-token   3      5d5h
pgo-target-token-mxvwt          kubernetes.io/service-account-token   3      5d5h
pgo.tls                         kubernetes.io/tls                     2      5d5h
pgorole-pgoadmin                Opaque                                2      5d5h
pgouser-admin                   Opaque                                3      5d5h
postgres-operator-token-fvbw9   kubernetes.io/service-account-token   3      5d5h
```

And i see you have it set up for read-only currently will need to add dynamic and disabled.

thanks again for contributing and looking forward to further updates.

james-callahan:
> Looks like you are missing some secrets and SA's needed for the pgocli

I'm not sure which you mean?

  - `pgo.tls` is created if it doesn't exist; if the user would like to provide it, then can+should do so in an overlay


But also, why do you say "needed for the pgocli", in a base kustomization you'd only want the minimum required for the operator (daemon): what does the cli need that the daemon doesn't?

ValClarkson:
Hi @james-callahan  Sorry for any confusion.  currently all of our deployments deploy the apiserver along with the operator. The apiserver will need the pgorole-pgoadmin secret as well as the pgouser-admin secret.

rpahli:
Any update on this PR. Would be very helpful to have a kustomize based installation method.

james-callahan:
> Any update on this PR. Would be very helpful to have a kustomize based installation method.

I'll close it now; the operator just isn't ready for production deployments IMO.

PR #2004, opened 2020-10-27
Ease creating PostgreSQL clusters via Custom Resource workflow

jkatz:
This changes the methodology for how a pgBackRest Secret is
generated such that it is no longer required at the time a
cluster is created vis-a-vis the pgcluster custom resource.

Instead, the Operator follows this heuristic:

- If a pgBackRest Secret is provided, this Secret is used
- If the pgBackRest Secret is partially filled out, the
missing pieces are filled in
- If no Secret is provided, a Secret is generated.

Note that if you want to use S3 or a S3-like storage system,
you will need to still create the Secret with the appropriate
S3 credentials.

This also updates various documentation to show the easier
workflow.

This patch also introduces the creation of the `pgo-config` ConfigMap should one not be created at installation time. All of the default template files are already packaged within the Operator, so, should the Operator detect the lack of a `pgo-config` in its namespace, it will create this file.

Issue: [ch9451]


tjmoore4 (approved):
Both standard and S3 enabled clusters worked as expected. Tests with adding resource configurations and pgBouncer pods worked.

PR #2046, opened 2020-11-12
Initial Controller for the postgrescluster Custom Resource

andrewlecuyer:
Adds the initial controller for the `postgrescluster`  custom resource.  This includes the creation of a controller runtime "manager", which is responsible running and managing the `postgrescluster` controller (and will ultimately manage any additional controllers created in support of managing PostgreSQL clusters via the `postgrescluster` custom resource).

The controller itself is created using the controller runtime `builder` package, and includes a reconcile function that will be responsible for reconciling PostgreSQL clusters via events generated for `postgrescluster` custom resources.  For this initial controller implementation, the reconcile function simply logs a debug message when a `postgrescluster` event is received, providing a structure and layout for adding the business logic to reconcile an actual PostgreSQL cluster in subsequent commits. 
 Also, the controller demonstrates how the controller runtime client can be utilized to obtain a `postgrescluster` resource from the Kubernetes API, while also showing how native Kubernetes events can be generated for `postgrescluster` custom resources.

Also included are kustomize artifacts and associated specifications for Deploying the PostgreSQL Operator to run the new 
 controller runtime "manager" discussed above, and therefore the new `postgrescluster` controller.  This includes the various, namespaces, RBAC, etc. needed to support the Deployment of the operator, along with the Deployment artifact for the PostgreSQL Operator itself.  Make targets have also been created to facilitate deploying the PostgreSQL Operator via kustomize (e.g. `make deploy`, `make install`, etc.).  Additionally, a simple example has been created to create a `postgrescluster` custom resource using kustomize.

It should be noted that with this initial implementation the `postgres-operator` ServiceAccount is assigned a ClusterRole to to provide it with the ability manage 'postgrescluster' resources in any namespace.  However, by using the `PGO_TARGET_NAMESPACE` environment variable, the PostgreSQL Operator can be configured to only listen for `postgrescluster` resources in a specific namespace.

And finally, environment variables have been created to allow either the new `postgrescluster` controller, or the current `pgcluster` controller (along with supporting controllers, e.g. for `pgtask`, `pgreplica`, etc.) to be disabled.  This allows the PostgreSQL Operator to run the controllers for either PostgreSQL cluster custom resource as needed (including the ability to run both at the same time if desired).  The kustomize installers for the new `postgrescluster` custom resource are designed to only enable the `postgrescluster` controller, while the installers for the `pgcluster` custom resource have been updated to only enable to `pgcluster` controllers.

**Checklist:**

 
**Type of Changes:**

 
A controller does not exist for the `postgrescluster` custom resource.

[ch9475]

A controller exists for the `postgrescluster` custom resource.

**Other information**:

N/A


internal/controller/postgrescluster/controller.go
```diff
@@ -0,0 +1,87 @@
...
+	// call business logic to reconcile the postgrescluster
+
+	return reconcile.Result{}, nil
+}
+
+// SetupWithManager adds the PostgresCluster controller to the provided runtime manager
+func (r *Reconciler) SetupWithManager(mgr manager.Manager) error {
+
+	// create a controller for the PostgresCluster custom resource
+	if err := builder.ControllerManagedBy(mgr).
```
cbandy, line 76: I'm not opposed to using the component packages directly, but I see that the `sigs.k8s.io/controller-runtime` package exports some of these, too. I can't tell if the top-level ones are intended to be "easy/helpful" or the canonical way to use the module. 🤷 

- https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.7.0-alpha.6#pkg-variables
- https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.7.0-alpha.6#pkg-types

internal/controller/postgrescluster/controller.go
```diff
@@ -0,0 +1,87 @@
...
+		For(&v1alpha1.PostgresCluster{}).
+		WithEventFilter(predicate.ResourceVersionChangedPredicate{}).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: workerCount,
+		}).
+		Complete(r); err != nil {
+		return err
+	}
+
+	return nil
```
cbandy, line 86: This smells like it should not bother branching. Recommend:

```go
	return builder.ControllerManagedBy(mgr)...
		Complete(r)
}
```

internal/controller/runtime/runtime.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/crunchydata/postgres-operator/internal/controller/postgrescluster"
+	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1alpha1"
+)
+
+const (
+	// the name of the event recorder created for the postgrescluster controller
+	recorderName = "postgrescluster-controller"
+
+	// default refresh interval in minutes
+	refreshInterval = 60
```
cbandy, line 36: If we shouldn't set it smaller than one minute, we can say so in the comment.

```suggestion
	// default refresh interval
	refreshInterval = 60 * time.Minute
```

internal/controller/runtime/runtime.go
```diff
@@ -0,0 +1,99 @@
...
+
+	pgoScheme, err := createPGOScheme()
+	if err != nil {
+		return nil, err
+	}
+
+	// create controller runtime manager
+	syncPeriod := refreshInterval * time.Minute
+	mgr, err := manager.New(config.GetConfigOrDie(), manager.Options{
+		Namespace:  os.Getenv("PGO_TARGET_NAMESPACE"), // if empty then watching all namespaces
```
cbandy, line 53: `GetConfigOrDie()` and `os.Getenv()` seem like responsibilities that belong in `main()`. This file is good to go, but I'm going to think on it some more.

examples/postgrescluster/postgrescluster.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: postgres-operator.crunchydata.com/v1alpha1
+kind: PostgresCluster
+metadata:
+  name: mycluster
+spec:
+  instances:
+    - name: instance1
```
cbandy, line 7: I don't see these fields in the CRD, yet.

Makefile
```diff
@@ -9,6 +9,7 @@ PGO_VERSION ?= 4.5.0
 PGO_PG_VERSION ?= 12
 PGO_PG_FULLVERSION ?= 12.4
 PGO_BACKREST_VERSION ?= 2.29
+PGO_CMD=kubectl
```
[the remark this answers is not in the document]
andrewlecuyer: Thanks, changed to `?=`.

cmd/postgres-operator/main.go
```diff
@@ -50,6 +55,93 @@ func main() {
...
+	// exit with an error if neither the pgcluster nor postgrescluster controllers are enabled
+	if disablePGCluster && disablePostgresCluster {
+		log.Fatal("either the pgcluster or postgrescluster controller must be enabled")
+	}
+
+	var controllerManager *manager.ControllerManager
+	if !disablePGCluster {
+		controllerManager = enablePGClusterControllers(stopCxt.Done())
+	}
+	defer controllerManager.RemoveAll()
```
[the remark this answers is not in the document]
andrewlecuyer: Thanks, don't think I ever saw the panic since I was shutting down anyway :thinking:.

Anyway, I went ahead and moved the `defer` as suggested.

internal/controller/postgrescluster/controller.go
andrewlecuyer: Changed as suggested.

internal/controller/runtime/runtime.go
andrewlecuyer: This was just a starting point, so I don't have any specific constraints around this interval at the moment (maybe we make it configurable, as similar settings currently are?).

examples/postgrescluster/postgrescluster.yaml
andrewlecuyer: I just had to add _something_ to the spec to get the Kube API to accept the `postgrescluster`, so I just went ahead and added these per the currently proposed CRD design (figure these will be some of the first fields we add).

cmd/postgres-operator/main.go
```diff
@@ -105,44 +194,22 @@ func main() {
...
-	log.Infof("Signal received, now exiting")
+	return controllerManager
 }
 
-// createAndStartNamespaceController creates a namespace controller and then starts it
-func createAndStartNamespaceController(kubeClientset kubernetes.Interface,
-	controllerManager controller.Manager, stopCh <-chan struct{}) error {
+// enablePGClusterController enables all controllers needed to manage PostgreSQL clusters using
+// the 'postgrescluster' custom resource
+func enablePostgresClusterControllers(context context.Context) {
```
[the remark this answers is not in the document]
andrewlecuyer: Thanks, fixed.

cmd/postgres-operator/main.go
```diff
@@ -50,6 +55,93 @@ func main() {
 		log.Info("debug flag set to false")
 	}
 
+	// create a context that will be used to stop all controllers on a SIGTERM or SIGINT
+	stopCxt := signals.SetupSignalHandler()
```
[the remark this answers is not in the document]
andrewlecuyer: Went ahead and just changed to `ctx` per the generally accepted naming convention.

cmd/postgres-operator/main.go
```diff
@@ -105,44 +194,22 @@ func main() {
...
-		go nsController.RunWorker(stopCh)
+	log.Debug("starting controller runtime manager and will wait for signal to exit")
+	if err := mgr.Start(context); err != nil {
+		log.Fatal(err)
 	}
-
-	log.Debug("namespace controller is now running")
-
-	return nil
+	log.Debug("signal recieved, exiting")
```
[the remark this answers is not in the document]
andrewlecuyer: The similar message in `main()` is only called when the `pgcluster` controllers only are enabled, and this one is called when either just the `postgrescluster` controllers or both types of controllers are enabled.

internal/controller/postgrescluster/controller.go
```diff
@@ -0,0 +1,87 @@
...
+const workerCount = 2
+
+// Reconciler holds resources for the PostgresCluster reconciler
+type Reconciler struct {
+	Client   client.Client
+	Recorder record.EventRecorder
+}
+
+// Reconcile reconciles a ConfigMap in a namespace managed by the PostgreSQL Operator
+func (r *Reconciler) Reconcile(context context.Context,
```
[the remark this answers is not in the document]
andrewlecuyer: Changed to `ctx`.

internal/controller/postgrescluster/controller.go
andrewlecuyer: Changed as suggested.

Makefile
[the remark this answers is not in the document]
andrewlecuyer: Changed to `PGO_KUBE_CLIENT`.

cmd/postgres-operator/main.go
```diff
@@ -105,44 +194,22 @@ func main() {
...
-		}))
-	nsController, err := nscontroller.NewNamespaceController(controllerManager,
-		nsKubeInformerFactory.Core().V1().Namespaces(),
-		*operator.Pgo.Pgo.NamespaceWorkerCount)
+	mgr, err := runtime.CreateRuntimeManager()
 	if err != nil {
-		return err
+		log.Fatal(err)
 	}
+	log.Debug("controller runtime manager created")
```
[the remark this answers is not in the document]
andrewlecuyer: Debug log removed.

internal/controller/postgrescluster/controller.go
andrewlecuyer: Per discussion, we will use the component packages directly for the time being.

internal/controller/runtime/runtime.go
andrewlecuyer: `os.Getenv()` has been moved to `main()`, and `GetConfig()` is now used instead to capture the error and propagate to `main()`.

andrewlecuyer:
Thanks for the review @cbandy, all feedback has been addressed.

config/manager/manager.yaml
```diff
@@ -0,0 +1,18 @@
...
+  template:
+    spec:
+      containers:
+      - name: operator
+        image: postgres-operator
+        env:
+        - name: CRUNCHY_DEBUG
+          value: "true"
+        - name: PGO_DISABLE_PGCLUSTER
+          value: "true"
```
jkatz, line 17: Any chance these can be booleans or no?

installers/ansible/roles/pgo-operator/templates/deployment.json.j2
```diff
@@ -77,6 +77,10 @@
                                     }
                                 }
                             },
+                            {
+                                "name": "PGO_DISABLE_POSTGRESCLUSTER",
+                                "value": "true"
+                            },
```
jkatz, line 83: Do we need this?

internal/controller/postgrescluster/controller.go
```diff
@@ -0,0 +1,83 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1alpha1"
+)
+
+const workerCount = 2
```
jkatz, line 34: Should this be configurable?

internal/controller/runtime/runtime.go
jkatz: +1 on configurable.

internal/controller/runtime/runtime.go
```diff
@@ -0,0 +1,104 @@
...
+		Recorder: mgr.GetEventRecorderFor(recorderName)}
+	if err := r.SetupWithManager(mgr); err != nil {
+		return err
+	}
+	return nil
+}
+
+// createPGOScheme creates a scheme containing the resource types required by the PostgreSQL
+// Operator
+func createPGOScheme() (*runtime.Scheme, error) {
```
jkatz, line 88: As a bikeshed, my preference would be we don't abbreviate names, unless the actual scheme is called `pgo` (vs postgresoperator)

config/manager/manager.yaml
andrewlecuyer: These must be strings, otherwise the Deployment will be rejected by the API server, e.g.:

```bash
Error from server (BadRequest): error when creating "./config/default": Deployment in version "v1" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Env: []v1.EnvVar: v1.EnvVar.Value: ReadString: expects " or n, but found t, error found in #10 byte of ...|,"value":true},{"nam|..., bigger context ...|ainers":[{"env":[{"name":"CRUNCHY_DEBUG","value":true},{"name":"PGO_DISABLE_PGCLUSTER","value":true}|...
```

internal/controller/postgrescluster/controller.go
andrewlecuyer: We have not yet established our configuration approach/solution for `postgrescluster` (e.g. whether we use a `pgo.yaml`, where similar settings are stored for `pgcluster`), so I was not considering that in-scope for this initial controller implementation.

internal/controller/runtime/runtime.go
andrewlecuyer: I can go ahead and use `PostgresOperator` instead, e.g. `createPostgresOperatorScheme()`.

internal/controller/runtime/runtime.go
andrewlecuyer: Similar to my comment for `workerCount`, we have not yet established our configuration approach/solution for `postgrescluster`, so I was not considering that in-scope for this initial controller implementation.

installers/ansible/roles/pgo-operator/templates/deployment.json.j2
andrewlecuyer: This will simply prevent a `pgcluster`-based Deployment from also trying to run the `postgrescluster` controllers, which require their own specific resources (e.g. RBAC) that is managed and handled separately from this installer (specifically in the Kustomize installer for the `postgrescluster`-based Deployment).

internal/controller/runtime/runtime.go
cbandy: My gripe is with the lack of type. A constant amount of type should be a `time.Duration`.

internal/controller/runtime/runtime.go
cbandy: The scheme contains more than just our definitions. Perhaps just `createScheme()`.

internal/controller/runtime/runtime.go
andrewlecuyer: I'm cool with that.  Being that similar configurable settings for `pgcluster` are defined as `int's` within various operator configuration structs (e.g. for `pgo.yaml`), I essentially just had the same mindset here (considering this will likely eventually be configurable).  However, I'm OK with changing this to `time.Duration` until we figure out our configuration approach, and then adjust these settings accordingly.

internal/controller/runtime/runtime.go
andrewlecuyer: Just to clarify, the idea behind the name is that it'd create all resources required by the "PostgresOperator" as a whole (and not just for `postgrescluster`), which would would be a combination of our resources and the standard Kube resources.

cbandy (approved):

internal/controller/runtime/runtime.go
jkatz: WFM.

cbandy (approved):

jkatz:
Woohoo!

PR #2112, opened 2020-12-10
Wait for PG Deployment Deletion on Restore

andrewlecuyer:
Since the primary PVC for the cluster is now retained during an in-place PostgreSQL cluster restore in support of pgBackRest delta restores, when preparing a cluster for a restore we can no longer rely on the deletion of all PVC's as an indicator that the `config` and `leader` ConfigMaps created by Patroni can be removed.  Therefore, the Operator now specifically waits for all Deployments to be successfully removed prior to deleting these resources.

**Checklist:**

 
**Type of Changes:**

 
The deletion of all primary and replica PVCs is used to indicate that it is safe to remove the `config` and `leader` ConfigMaps when performing an in-place cluster restore.

Issue: [ch9926]

The Operator now specifically waits for all Deployments to be successfully removed prior to deleting the `config` and `leader` ConfigMaps when performing an in-place cluster restore.

**Other information**:

N/A

### DIFF
--- a/internal/operator/backrest/restore.go
+++ b/internal/operator/backrest/restore.go
@@ -184,7 +184,24 @@ func PrepareClusterForRestore(clientset kubeapi.Interface, cluster *crv1.Pgclust
 		}); err != nil {
 		return nil, err
 	}
-	log.Debugf("restore workflow: deleted primary and replicas %v", pgInstances)
+	log.Debugf("restore workflow: deleted primary and replica deployments for cluster %s",
+		clusterName)
+
+	// Wait for all primary and replica deployments to be removed.  If unable to verify that all
+	// deployments have been removed, then the restore cannot proceed and the function returns.
+	if err := wait.Poll(time.Second/2, time.Minute*3, func() (bool, error) {
+		for _, deployment := range pgInstances.Items {
+			if _, err := clientset.AppsV1().Deployments(namespace).
+				Get(ctx, deployment.GetName(), metav1.GetOptions{}); err == nil || !kerrors.IsNotFound(err) {
+				return false, nil
+			}
+		}
+		return true, nil
+	}); err != nil {
+		return nil, err
+	}
+	log.Debugf("restore workflow: finished waiting for primary and replica deployments for "+
+		"cluster %s to be removed", clusterName)
 
 	// delete all existing jobs
 	deletePropagation := metav1.DeletePropagationBackground


internal/operator/backrest/restore.go
```diff
@@ -184,7 +184,25 @@ func PrepareClusterForRestore(clientset kubeapi.Interface, cluster *crv1.Pgclust
...
+		clusterName)
+
+	// Wait for all primary and replica deployments to be removed.  If unable to verify that all
+	// deployments have been removed, then the restore cannot proceed and the function returns.
+	if err := wait.Poll(time.Second/2, time.Minute*3, func() (bool, error) {
+		notFound := true
+		for _, deployment := range pgInstances.Items {
+			if _, err := clientset.AppsV1().Deployments(namespace).
+				Get(ctx, deployment.GetName(), metav1.GetOptions{}); err == nil {
+				notFound = false
```
jkatz, line 197: We could just `return false, nil` here?

internal/operator/backrest/restore.go
```diff
@@ -184,7 +184,25 @@ func PrepareClusterForRestore(clientset kubeapi.Interface, cluster *crv1.Pgclust
...
+			}
+		}
+		return notFound, nil
```
jkatz, line 200:
```suggestion
		for _, deployment := range pgInstances.Items {
			if _, err := clientset.AppsV1().Deployments(namespace).
				Get(ctx, deployment.GetName(), metav1.GetOptions{}); err == nil {
				return false, nil
			}
		}
		return true, nil
```

internal/operator/backrest/restore.go
```diff
@@ -184,7 +184,25 @@ func PrepareClusterForRestore(clientset kubeapi.Interface, cluster *crv1.Pgclust
...
+	log.Debugf("restore workflow: deleted primary and replica deployments for cluster %s",
+		clusterName)
+
+	// Wait for all primary and replica deployments to be removed.  If unable to verify that all
+	// deployments have been removed, then the restore cannot proceed and the function returns.
+	if err := wait.Poll(time.Second/2, time.Minute*3, func() (bool, error) {
+		notFound := true
+		for _, deployment := range pgInstances.Items {
+			if _, err := clientset.AppsV1().Deployments(namespace).
+				Get(ctx, deployment.GetName(), metav1.GetOptions{}); err == nil {
```
cbandy, line 196: Might want to check for NotFound here. Other errors can be retried.

andrewlecuyer: Now specifically checking for `IsNotFound`.

internal/operator/backrest/restore.go
andrewlecuyer: Recommendation included in latest push.

internal/operator/backrest/restore.go
andrewlecuyer: Recommendation included in latest push.

jkatz (approved):

PR #2114, opened 2020-12-10
Use poll utility provided by Kubernetes for waiting

jkatz:
This moves several home-constructed methods to using a similarly
constructed one that is maintained upstream. Provides more
consistency across the code that can serve future implementation.

### DIFF
--- a/internal/operator/cluster/clusterlogic.go
+++ b/internal/operator/cluster/clusterlogic.go
@@ -43,6 +43,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 )
 
@@ -259,7 +260,6 @@ func getBootstrapJobFields(clientset kubeapi.Interface,
 func getClusterDeploymentFields(clientset kubernetes.Interface,
 	cl *crv1.Pgcluster, dataVolume, walVolume operator.StorageResult,
 	tablespaceVolumes map[string]operator.StorageResult) operator.DeploymentTemplateFields {
-
 	namespace := cl.GetNamespace()
 
 	log.Infof("creating Pgcluster %s in namespace %s", cl.Name, namespace)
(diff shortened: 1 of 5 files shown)


internal/operator/cluster/clusterlogic.go
```diff
@@ -804,22 +799,17 @@ func waitForDeploymentReady(clientset kubernetes.Interface, namespace, deploymen
...
-				return nil
-			}
+	if err := wait.Poll(periodSecs, timeoutSecs, func() (bool, error) {
+		// check to see if the deployment is ready
+		d, err := clientset.AppsV1().Deployments(namespace).Get(ctx, deploymentName, metav1.GetOptions{})
+		if err != nil {
+			log.Warn(err)
 		}
+
+		return d.Status.Replicas == d.Status.ReadyReplicas, nil
```
cbandy, line 809: This used to loop when `err != nil`. When there's an error, `d` is probably empty which would return true:
`0 == d.Status.Replicas == d.Status.ReadyReplicas`.

```suggestion
		return err == nil && d.Status.Replicas == d.Status.ReadyReplicas, nil
```

internal/operator/cluster/upgrade.go
```diff
@@ -318,21 +319,15 @@ func deleteBeforeUpgrade(clientset kubeapi.Interface, clusterName, currentPrimar
...
-		case <-timeout:
-			return fmt.Sprintf("Timed out waiting for deployment to be deleted: [%s]", deploymentName)
-		case <-tick.C:
-			_, err := clientset.AppsV1().Deployments(namespace).Get(ctx, deploymentName, metav1.GetOptions{})
-			if err != nil {
-				return fmt.Sprintf("Deployment %s has been deleted.", deploymentName)
-			}
-		}
+
+	if err := wait.Poll(periodSecs*time.Second, timeoutSecs*time.Second, func() (bool, error) {
```
cbandy, line 323: Hmm. `time.Duration * time.Duration`. It looks like the only two calls are a bit higher up in this file. Please fix this while you're here.

cbandy (changes requested):

internal/operator/cluster/clusterlogic.go
jkatz: Ah good catch. That would panic otherwise.

cbandy (approved):

PR #2154, opened 2020-12-27
Simplification of user secret creation

jkatz:
### Operator Generates bootstrap user Secrets if not present

Previously, the PostgreSQL user Secrets required for bootstrap
(superuser, replication user, standard user) needed to be present
before adding a custom resource. Now, the PostgreSQL Operator can
create the missing Secrets before bootstrapping a cluster.

The API server behavior does not change, as the API server can
both create the user credentials and return them to the end user.

This also modifies the examples to not include a requirement to
generate the replication user credential. As this is really a
managed service credential, there is no reason to put this onus
on the user. The superuser credential is still included in the
exmaples, as there are certain situations where one should log
in as the superuser.

### Eliminate user Secret name attributes from pgcluster CRD

These Secrets have been fully managed by the Postgres Operator
for a long time and there is no reason why they need to be specified
by the user as specific custom resource attributes. This further
simplifies the number of steps required to create a PostgreSQL
cluster with the operator through a "custom resource" workflow.


internal/operator/cluster/cluster.go
```diff
@@ -649,6 +657,62 @@ func annotateBackrestSecret(clientset kubernetes.Interface, cluster *crv1.Pgclus
...
+	// derive the secret name
+	secretName := crv1.UserSecretName(cluster, username)
+
+	// if the secret already exists, skip it
+	if _, err := clientset.CoreV1().Secrets(cluster.Spec.Namespace).Get(
+		ctx, secretName, metav1.GetOptions{}); err == nil {
+		log.Infof("user secret %q exists for user %q for cluster %q",
+			secretName, username, cluster.Spec.Name)
+		return nil
+	}
```
jkatz, line 677: Hm, this needs to catch errors that are not a "does not exist" error.

internal/operator/cluster/cluster.go
```diff
@@ -649,6 +657,65 @@ func annotateBackrestSecret(clientset kubernetes.Interface, cluster *crv1.Pgclus
...
+		return err
+	}
+
+	// great, now we can create the secret! if we can't, return an error
+	if err := util.CreateSecret(clientset, cluster.Spec.Name, secretName,
+		username, password, cluster.Spec.Namespace); err != nil {
+		return err
+	}
+
+	return nil
```
jkatz, line 696:
```suggestion
	returnutil.CreateSecret(clientset, cluster.Spec.Name, secretName,
		username, password, cluster.Spec.Namespace)
```

ValClarkson (approved):
lgtm

docs/content/custom-resources/_index.md
```diff
@@ -188,73 +131,57 @@ spec:
...
+# namespace that the cluster is running in
+export cluster_namespace=pgo
+# name of the cluster
+export pgo_cluster_name=hippo
+# name of the user whose password we want to get
+export pgo_cluster_username=hippo
 
-#### Step 2: Creating the PostgreSQL User Secrets
+kubectl -n "${cluster_namespace}" get secrets \
+  "${pgo_cluster_name}-${pgo_cluster_username}-secret" -o "jsonpath={.data['password']}" | base64 -d
```
cbandy, line 175: 📝 `kubectl` can do the decoding from the golang template format: 

```
-o 'go-template={{ index .data "password" | base64decode }}'
```

jkatz: I think I got that reference from official docs on `kubectl`. Noted for the future.

PR #2156, opened 2020-12-28
Move TLS key generation to ECDSA

jkatz:
This moves the default TLS key generation for the API server
to use ECDSA keys with a P-256 curve and a SHA384 signature.
This only affects newly created Operator deployments, or
Operator deployments that have deleted their TLS secret.


ValClarkson (approved):
lgtm

PR #2163, opened 2021-01-01
Move `pgo failover` to use updated switchover plumbing

jkatz:
The manual failover/switchover command, `pgo failover`, now
uses the updated "switchover" plumbing that was introduced as
part of the rolling update changes. This ensures a unified
experience when performing an action that involves a failover.

Additionally, `pgo failover` now occurs inline: it does not
create a pgtask custom resource. This is due to both being a much
simpler process and the transactional nature of an immediate failover.

`pgo failover` can now also be executed with a --target flag. This
was actually always supported, but unavailable based upon a restriction
in the `pgo` client. When the `--target` flag is not used, the
PostgreSQL Operator will choose the best candidate for failing over.

This also adds the `--force` flag to `pgo failover` so one can force
a failover when there is no healthy candidate to promote.


internal/operator/switchover.go
```diff
@@ -0,0 +1,156 @@
...
+	log "github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+/*
+ Copyright 2021 Crunchy Data Solutions, Inc.
```
cbandy, line 20: This should be higher in the file. Before `package` or `import` wfm.

internal/operator/switchover_test.go
```diff
@@ -0,0 +1,45 @@
+package operator
+
+import (
+	"reflect"
+	"testing"
+)
+
+/*
+ Copyright 2021 Crunchy Data Solutions, Inc.
```
cbandy, line 9: This, too.

andrewlecuyer:
There is one important consequence of switching to `switchover` that I think we'll want to explicitly point out in the release notes and/or docs.  Specifically, while `failover` can be used to failover to a specific node even if it is unhealthy (for instance, because it is lagging), this is not the case with `switchover`, which will only succeed if the cluster is healthy.  Therefore, users will need to be aware that `pgo failover` can no longer be used in scenarios where the cluster is unhealthy.

cbandy (approved):

jkatz:
@andrewlecuyer Feedback resolved.

PR #2173, opened 2021-01-05
Allow for empty --service-type on `pgo create pgbouncer`

jkatz:
This was an oversight in the logic for allowing a default pgBouncer
Service Type to be set.

Issue: [ch10065]


jmckulk (approved):
without running any tests it looks like it should work

Issue #1203, opened 2020-01-12
postgres-operator ignores cluster CR deletion


lkhomenk:
**Describe the bug**
If PgCluster CR of the cluster is deleted - operator does nothing. all object remain

**To Reproduce**
1. Provision any pg_cluster with `pgo` or by creating CR from yaml
2. Wait for cluster creation to complete
3. Delete pg_cluster CR
4. Wait for cluster deletion

**Expected behavior**
Before 4.2 operator was deleting cluster objects after CR was deleted

**Please tell us about your environment:**

* Operating System: centos7
* Where is this running ( Local, Cloud Provider): AWS
* Storage being used (NFS, Hostpath, Gluster, etc):
* Container Image Tag: centos7-10.11-4.2.0
* PostgreSQL Version: 10-11
* Platform (Docker, Kubernetes, OpenShift): Openshift 
* Platform Version: 3.11

stoetti:
Are there any updates on the timeline for this feature?

jkatz:
This will be available in 4.6.0

PR #2190, opened 2021-01-10
Enable cluster deletion from custom resource workflow

jkatz:
This behavior, which had been removed circa 4.1, is now reintroduced
to use the present machienry when a PostgreSQL cluster is being
deleted. Deleting a pgcluster resoruce directly will now delete a
PostgreSQL cluster and all of its associated objects.

To retain the PVCs for the PostgreSQL data directory or the backups
repository, one can set the following Annotations on a pgclusters
custom resource:

- `keep-backups`: indicates to keep the pgBackRest PVC when deleting
the cluster.
- `keep-data`: indicates to keep the PostgreSQL data PVC when deleting
the cluster.

Issue: [ch9435]
closes: #1203

### DIFF
--- a/internal/controller/pgcluster/pgclustercontroller.go
+++ b/internal/controller/pgcluster/pgclustercontroller.go
@@ -18,7 +18,6 @@ limitations under the License.
 import (
 	"context"
 	"encoding/json"
-	"io/ioutil"
 	"reflect"
 	"strings"
 
@@ -117,8 +116,6 @@ func (c *Controller) processNextItem() bool {
 			"'initialized' or 'bootstrapping' state", cluster.GetName())
 		return true
 	}
-
-	addIdentifier(cluster)
 
 	// If bootstrapping from an existing data source then attempt to create the pgBackRest repository.
 	// If a repo already exists (e.g. because it is associated with a currently running cluster) then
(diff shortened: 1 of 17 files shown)


internal/controller/pgcluster/pgclustercontroller.go
```diff
@@ -348,11 +345,58 @@ func (c *Controller) onUpdate(oldObj, newObj interface{}) {
...
+	// iterate through the list of Jobs and see if any are currently active
+	// a succeeded Job could be a remnaint of an old Job for the cluser of a
+	// same name, in which case, we can continue with deleting the cluster
+	for _, job := range jobs.Items {
+		// if the Job is currently active or if the Job is not active but never has
+		// completed, exit
+		if job.Status.Active > 0 {
+			return
+		} else if job.Status.Succeeded < 1 {
+			return
```
ValClarkson, line 377: this part here seems confusing, took me a few mins to understand it
if the job is active exit, if the job is not active but never succeeded exit...do we log that somewhere that we can't delete b/c of an active job or a job never succeeded?

otherwise lgtm

ValClarkson (approved):

internal/controller/pgcluster/pgclustercontroller.go
jkatz: @ValClarkson I clarified the comment. However, looking through it, I realized that the "failure" cause was not great, so I added this additional logic. Please let me know what you think.

ValClarkson: I agree this is much better with the backofflimit added

ValClarkson (approved):

PR #2191, opened 2021-01-10
Tighten up ephemeral storage limits

### LINKED ISSUE #2188: ephemeral storage limit
Hello,
i am testing pgo operator.

pod is always evicted (see attached desc)
[yk-5c9b5bbbbc-vncbp.log](https://github.com/CrunchyData/postgres-operator/files/5790593/yk-5c9b5bbbbc-vncbp.log)

i did not find any plaecs where i can change this limit value....
thx

jkatz:
This removes an ephemeral storage volume that is no longer used
and places a cap of 64Mi on the pgBadger ephemeral storage
is not mounted unless there is a pgBadger instance directory.

This does not add a size limit to shared memory (dshm), as that
is a much more complicated topic.

closes #2188

### DIFF
--- a/installers/ansible/roles/pgo-operator/files/pgo-configs/cluster-deployment.json
+++ b/installers/ansible/roles/pgo-operator/files/pgo-configs/cluster-deployment.json
@@ -176,9 +176,6 @@
                         }, {
                             "mountPath": "/pgconf",
                             "name": "pgconf-volume"
-                        }, {
-                            "mountPath": "/recover",
-                            "name": "recover-volume"
                         },
                         {
                             "mountPath": "/dev/shm",
@@ -282,11 +279,11 @@
                       {{ end }}
                     {{ end }}
                     {
-                        "name": "recover-volume",
-                        "emptyDir": { "medium": "Memory" }
-                    }, {
                         "name": "report",
-                        "emptyDir": { "medium": "Memory" }
+                        "emptyDir": {
+                          "medium": "Memory",
+                          "sizeLimit": "64Mi"
+                        }
                     },
                     {
                       "name": "dshm",


installers/ansible/roles/pgo-operator/files/pgo-configs/cluster-deployment.json
```diff
@@ -282,11 +279,11 @@
...
                     {{ end }}
                     {
-                        "name": "recover-volume",
-                        "emptyDir": { "medium": "Memory" }
-                    }, {
                         "name": "report",
-                        "emptyDir": { "medium": "Memory" }
+                        "emptyDir": {
+                          "medium": "Memory",
+                          "sizeLimit": "64Mi"
```
cbandy, line 285: As I understand it, this will cause the pod to be evicted if it uses more than this amount: https://issue.k8s.io/63126

It looks like Kubernetes 1.20 has an alpha feature to finally consider this amount when creating the volume: https://pr.k8s.io/94444

- https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1967-size-memory-backed-volumes

jkatz: From my understanding, it would evcit the Pod. But it sounds like there are restrictions that if we don't set a size limit, it will not schedule the Pod.

We could set the sizeLimit to be a higher value. That said, I had mainly tested generating pgBadger files to see how large they were. From an initial run, it was ~1MB. From my experience with pgBadger, the assets around it ar enot that large -- the input (PostgreSQL logs) tends to be the large bit.

cbandy: > From my experience with pgBadger, the assets around it ar enot that large -- the input (PostgreSQL logs) tends to be the large bit.

Good point.

cbandy (approved):

PR #2194, opened 2021-01-11
Update Helm example

jkatz:
This reduces the number of requirements that are needed to deploy
a PostgreSQL cluster with this Helm chart. It does add additional
attributes to help further customize the PostgreSQL cluster, as
well as updated instructions for how to run the example.


ValClarkson (approved):
awesome update to the helm example!

PR #2195, opened 2021-01-11
Upgrade fixes for 4.6

jkatz:
- Do not restrict max upgrade version when running `pgo upgrade`. This is primarily an internal change.
- Do not delete primary/backup PVCs during upgrade.
- Ensure a PostgreSQL cluster comes back up as a leader once the upgrade is complete.


tjmoore4 (approved):
Looks good to me

PR #2204, opened 2021-01-13
Make "Effect" optional when setting a toleration via `pgo` client

jkatz:
Per Kubernetes documentation, one does not need to set an Effect
when setting a Toleration, so four CLI should allow for
this to be optional.

[ch10147]


cbandy:
nit: Effect is required on Taint. It's optional on Toleration where it means "any effect".

jkatz:
@cbandy OK, I will modify the commit message.

jmckulk (approved):
looks like this should work

PR #2205, opened 2021-01-13
pgBackRest and associated jobs should respect cluster tolerations

jkatz:
pgBackRest is a close corollary to the PostgreSQL cluster itself
and should respect similar tolerations to the PostgreSQL cluster.

Future work could break out tolerations specific for pgBackRest,
but given the two do need to work in concert, it is prudent to
start with this approach.

Issue: [ch10146]


jmckulk (approved):
lgtm will test again in beta.3

PR #2212, opened 2021-01-15
Adjust replica creation logic as part of a standby cluster

jkatz:
This ensures an explicit call of standby creation after receiving
a notification that a stanza for the cluster was successfully
created.

Co-authored-by: @andrewlecuyer


jmckulk (approved):
tested out fantastically (with both the main and standby cluster on the same local kube cluster)

PR #2219, opened 2021-01-18
Labeling consistency an ease-of-use

jkatz:
- Ensure all labeling functions are using the same API validation (babf44e and f3f19da backpatchable)
- Move labeling CLI interface to be similar to that of annotations.
    - `pgo create cluster`, `pgo label`, and `pgo delete label` all
now have a single `--label` flag. `--label` can be specified multiple
times.
    - The API call itself takes a mapping of key/value pairs.
    - The API endpoint parameter for `pgo label` and `pgo delete label`
is now called `Labels`

Issue: [ch10201]
Issue: [ch10202]


tjmoore4 (approved):
Looks good to me

internal/util/cluster_test.go
```diff
@@ -114,3 +117,60 @@ func TestGenerateNodeAffinity(t *testing.T) {
...
+					t.Fatalf("expected an ErrLabelInvalid error.")
+				}
+			})
+		}
+	})
+}
+
+// formatLabelStr is a helper function that formats the label string so you
+// can see it in the test
+func formatLabelStr(input map[string]string) string {
```
cbandy, line 168: https://pkg.go.dev/k8s.io/apimachinery/pkg/labels#FormatLabels is similar if you want.

internal/util/cluster_test.go
```diff
@@ -114,3 +117,60 @@ func TestGenerateNodeAffinity(t *testing.T) {
...
+			t.Run(formatLabelStr(input), func(t *testing.T) {
+				err := ValidateLabels(input)
+
+				if err == nil {
+					t.Fatalf("expected an invalid input error.")
+				}
+
+				if !errors.Is(err, ErrLabelInvalid) {
+					t.Fatalf("expected an ErrLabelInvalid error.")
+				}
```
cbandy, line 160: These two can be combined. It's good to mention the failing value in this case.

```suggestion
				if !errors.Is(err, ErrLabelInvalid) {
					t.Fatalf("expected an ErrLabelInvalid error, got %T: %v", err, err)
				}
```

internal/apiserver/common.go
```diff
@@ -44,6 +44,8 @@ var (
 	// ErrDBContainerNotFound is an error that indicates that a "database" container
 	// could not be found in a specific pod
 	ErrDBContainerNotFound = errors.New("\"database\" container not found in pod")
+	// ErrLabelInvalid indicates that a label is invalid
+	ErrLabelInvalid = errors.New("invalid label")
```
cbandy, line 48: This looks like a copy of `internal/util.ErrLabelInvalid`. Is it used?

cbandy (approved):

internal/apiserver/common.go
jkatz: Oh I forgot to remove that during the refactor. I can handle that.

Issue #1956, opened 2020-10-11
Would be great to have a list of supported containers for each version of pgo


alrooney:
When I try to bring up a new version of pgo the first issue I hit is what is the container tag I should be using?  I stumbled across this page: https://www.crunchydata.com/developers/download-postgres/containers/postgis (we are using postgis containers) but would be helpful if this info was linked somewhere from the pgo docs.  I thought there might be a way to get a listing of what is in:  registry.developers.crunchydata.com but lots of googling hasn't suggested a good way to do that.  Without this info, invariably I spend a good deal of time trying to pull various images to figure out if I can.

jkatz:
This is now added: https://access.crunchydata.com/documentation/postgres-operator/v5/references/components/

VladMasarik:
@jkatz Many thanks for the fix! But is it actually a fix? I have been away for far too long (PGO just works too well I guess :D ), so I dont know all the details, but I believe this issue was created because we wanted to know the specific Docker Image tags that we can use, and their source was not mentioned in the docs. The page you linked does provide more information, but it does not seem to be enough. 

E.g. if I wanted to use this image, how do I know the tag should have `centos8` and `4.7.0` parts as well?
```
registry.developers.crunchydata.com/crunchydata/crunchy-postgres-gis:centos8-12.7-2.5-4.7.0
```
If the info is there, but I don't see it, just reply "Info is there", and I won't bother you with this issue again. Many thanks though!

PR #2225, opened 2021-01-20
Added information about where to find image tags

VladMasarik:
**Checklist:**

 
**Type of Changes:**
Documentation

Currently, as said in #1956 there is no clear source of information where to find the image tags for PostgreSQL nor PostGIS.

I have included two references to the website which issue #1956 referenced, and I was also able to find only that website, but let me know if there is a better source.

**Other information**:
I am not sure if these are the best parts to include this information, but when I was going though the documentation myself, I was looking into these parts especially.


jkatz:
This is now added: https://access.crunchydata.com/documentation/postgres-operator/v5/references/components/

PR #2243, opened 2021-01-30
Clean Stanza Create pgtask & Job After Cluster Initialization

### LINKED ISSUE #2106: 4.5.0 BackrestBackups loop until partitions full
In Crunchy 4.4.0 and below I noticed the default behavior in regards to backrest backups being made was one backup was created on cluster creation prior to replica, then any following backup would be made from a failover or some pgo command. In Crunchy 4.5.0 I see backrestBackups being created every 10-15 minutes, and without settings limits in a custom pgBackRest Custom Configuration, it will repeatedly make backups until the local partitions are full and postgres will enter a non function state until all of the wal files are deleted in the local storage folder on each node. I'm not sure if there's some operator installation config that I missed when upgrading from 4.4 or if the automatic backups are intended behavior but I was wondering if there was a way to ensure backups only occur like in 4.4 which is on cluster creation, failover and from pgo commands rather then periodic.

I confirmed that this issue occurs when no custom configs used and and no data was ever added to the database by just running:

```
pgo create cluster test-cluster -n crunchy-test --cpu 15m --memory 690Mi --memory-limit 2Gi --storage-config testlocalstorage --replica-storage-config testlocalstorage --pgbackrest-cpu 30m --pgbackrest-memory 20Mi --pgbackrest-memory-limit 60Mi
```

I'm guessing this shouldn't be happening, as it doesnt happen with the same creation method on 4.4, and I haven't seen any issues similar to this.

* Operating System:
* Where is this running ( Local , Cloud Provider): 3 node local cluster
* Storage being used (NFS, Hostpath, Gluster, etc): primary and replica use localstorage, backrest-shared-repo uses openebs-sc
* Container Image Tag: centos7-13.0-4.5.0
* PostgreSQL Version: 13.0
* Platform (Docker, Kubernetes, OpenShift): OpenShift
* Platform Version: 3.11

If possible please run the following on the kubernetes or OpenShift (oc) commands and provide the result:

andrewlecuyer:
When a PostgreSQL cluster is initialized for the first time, the stanza create pgtask and the Kubernetes Job associated with it are now automatically deleted.  This ensures that subsequent update events in the Job controller for a stanza-create Job do not trigger additional "initial backup" Jobs for the cluster.

**Checklist:**

 
**Type of Changes:**

 
In OpenShift 3.11, periodic update events are sent to the Job controller for all Jobs, which can trigger the automated full backup that is typically only run once during PostgreSQL cluster initialization.

Issue: [ch10277]
fixes #2106

The stanza create Job is removed when successfully completed, preventing any subsequent Job update events from triggering a full backup for the cluster.

**Other information**:

N/A

### DIFF
--- a/internal/controller/job/backresthandler.go
+++ b/internal/controller/job/backresthandler.go
@@ -167,8 +167,11 @@ func (c *Controller) handleBackrestStanzaCreateUpdate(job *apiv1.Job) error {
 			return err
 		}
 
-		_, _ = backrest.CreateInitialBackup(c.Client, job.ObjectMeta.Namespace,
-			clusterName, backrestRepoPodName)
+		if _, err := backrest.CreateInitialBackup(c.Client, job.ObjectMeta.Namespace,
+			clusterName, backrestRepoPodName); err != nil {
+			log.Error(err)
+			return err
+		}
 
 		// now that the initial backup has been initiated, proceed with deleting the stanza-create
 		// pgtask and associated Job.  This will ensure any subsequent updates to the stanza-create
--- a/internal/controller/job/backresthandler.go
+++ b/internal/controller/job/backresthandler.go
@@ -170,6 +170,14 @@ func (c *Controller) handleBackrestStanzaCreateUpdate(job *apiv1.Job) error {
 		_, _ = backrest.CreateInitialBackup(c.Client, job.ObjectMeta.Namespace,
 			clusterName, backrestRepoPodName)
 
+		// now that the initial backup has been initiated, proceed with deleting the stanza-create
+		// pgtask and associated Job.  This will ensure any subsequent updates to the stanza-create
+		// Job do not trigger more initial backup Jobs.
+		if err := backrest.CleanStanzaCreateResources(namespace, clusterName, c.Client); err != nil {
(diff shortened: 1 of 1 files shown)


internal/controller/job/backresthandler.go
```diff
@@ -167,8 +167,20 @@ func (c *Controller) handleBackrestStanzaCreateUpdate(job *apiv1.Job) error {
 			return err
 		}
 
-		_, _ = backrest.CreateInitialBackup(c.Client, job.ObjectMeta.Namespace,
+		_, err = backrest.CreateInitialBackup(c.Client, job.ObjectMeta.Namespace,
 			clusterName, backrestRepoPodName)
+		if err != nil {
```
jkatz, line 172:
```suggestion
		if _, err := backrest.CreateInitialBackup(c.Client, job.ObjectMeta.Namespace,
			clusterName, backrestRepoPodName); err != nil {
```

jkatz (approved):

internal/controller/job/backresthandler.go
andrewlecuyer: Suggestion applied!

PR #2264, opened 2021-02-08
CCP image tag update and CM upgrade fix

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
During both pgcluster create and upgrade, the `--ccp-image-tag`
flag can be used to specify an image tag that overrides the
value set in the PostgreSQL Operator default configuration.
In certain cases, this value is not authoritative because the
underlying `ccpimagetag` CRD value is not honored as expected.

Also, during the upgrade of an existing pgcluster, the existing
`<clustername>-pgha-config` configmap is used to populate
the struct object so that key values are upgraded, as needed.
In certain cases where the configmap exists but was is not
initialized as expected, the exist function can hit a runtime
panic when referencing uninitialized structs or maps.

First, this update standardizes the behavior throughout the Operator so
that the CLI flag and subsequent CRD value is authoritative both
during initial cluster creation/upgrade and also used during
subsequent opertions, such as backups and restores.

This update is designed to conform to the standard ccpimagetag
pattern. For example, the tag
`centos8-12.5-4.6.0`
will correspond to the image OS version, PostgreSQL version and
Crunchy PostgreSQL Operator version. Similarly, a PostGIS enabled
image would be expected to have a tag similar to
`centos8-12.5-3.0-4.6.0`
where the additional value, 3.0, corresponds to the PostGIS version.
This is particularly important to note when using PostGIS enabled
clusters, as the Operator will automatically handle the removal of
the '3.0' value when creating sidecars or companion pods, such as
pgBadger or pgBouncer.

Finally, the tag validation used during the Operator's automated
upgrade process has been adjusted to properly perform tag validation
based on the provide image tag when the `--ccp-image-tag` flag is used.

Second, this update checks that these structs and maps are not nil
before attempting to set any values.

**Other information**:
[ch10368]


internal/operator/cluster/upgrade.go
```diff
@@ -879,16 +879,30 @@ func updateClusterConfig(clientset kubeapi.Interface, pgcluster *crv1.Pgcluster,
...
+		if _, ok := dcsConf.PostgreSQL.Parameters["unix_socket_directories"]; ok {
+			dcsConf.PostgreSQL.Parameters["unix_socket_directories"] = "/tmp"
+		}
+		if _, ok := dcsConf.PostgreSQL.Parameters["archive_command"]; ok {
+			dcsConf.PostgreSQL.Parameters["archive_command"] = `source /opt/crunchy/bin/postgres-ha/pgbackrest/pgbackrest-set-env.sh && pgbackrest archive-push "%p"`
+		}
+		if _, ok := dcsConf.PostgreSQL.RecoveryConf["restore_command"]; ok {
+			dcsConf.PostgreSQL.RecoveryConf["restore_command"] = `source /opt/crunchy/bin/postgres-ha/pgbackrest/pgbackrest-set-env.sh && pgbackrest archive-get %f "%p"`
+		}
+	}
```
jkatz, line 895:
```suggestion
	if dcsConf.PostgreSQL != nil && dcsConf.PostgreSQL.Parameters != nil {
		dcsConf.PostgreSQL.Parameters["unix_socket_directories"] = "/tmp"
		dcsConf.PostgreSQL.Parameters["archive_command"] = `source /opt/crunchy/bin/postgres-ha/pgbackrest/pgbackrest-set-env.sh && pgbackrest archive-push "%p"`
		dcsConf.PostgreSQL.RecoveryConf["restore_command"] = `source /opt/crunchy/bin/postgres-ha/pgbackrest/pgbackrest-set-env.sh && pgbackrest archive-get %f "%p"`
	}
```

jkatz (changes requested):
Looks good. One change.

internal/operator/cluster/upgrade.go
tjmoore4: updated, thanks.

jkatz (approved):

PR #2268, opened 2021-02-09
Create rc file as part of client_setup.sh

jmckulk:
This change updates the client_setup script to create a client_vars.rc file and
store it in the output directory. This will allow users to source the variables
directly instead of manually running them or copy/pasting them into their bashrc
files. You can now source the rc file or add the following line to your bashrc:

source $HOME/.pgo/$NAMESPACE/client_vars.rc

This change will be necessary for automated environments where manual steps would
otherwise need to be hardcoded.

**Checklist:**

 
**Type of Changes:**

 
**Other information**:


installers/kubectl/client-setup.sh
```diff
@@ -78,9 +78,14 @@ kubectl_get_private "${OUTPUT_DIR}/pgouser" secret -n "${PGO_OPERATOR_NAMESPACE}
...
+EOF
+)
+
 echo "pgo client files have been generated, please add the following to your bashrc"
-echo "export PATH=${OUTPUT_DIR}:\$PATH"
-echo "export PGOUSER=${OUTPUT_DIR}/pgouser"
-echo "export PGO_CA_CERT=${OUTPUT_DIR}/client.crt"
-echo "export PGO_CLIENT_CERT=${OUTPUT_DIR}/client.crt"
-echo "export PGO_CLIENT_KEY=${OUTPUT_DIR}/client.key"
+echo "$client_vars" | tee ${OUTPUT_DIR}/client_vars.rc
```
andrewlecuyer, line 91: No need for `tee` here, e.g. can use 

```bash
echo "$client_vars" > ${OUTPUT_DIR}/client_vars.rc`
```

andrewlecuyer (changes requested):
One quick note, otherwise looks good!

installers/kubectl/client-setup.sh
jmckulk: I used `tee` so that the lines would still be printed to `stdout` **and** sent to the file. That way I'm not changing the current usage of the file but still adding the new functionality.

If I change how the script works then I will need to go in and update docs related to this script. I can do that but I was trying to limit this change.

jmckulk: Another option (without `tee`)
```
echo "$client_vars"
echo "$client_vars" > ${OUTPUT_DIR}/client_vars.rc
```

andrewlecuyer: Cool, that definitely makes sense.  Appreciate the explanation!

jkatz (approved):

PR #2269, opened 2021-02-09
Existence check for primary Deployment before scaling

jkatz:
During the cluster initialization process, there is a step where
the Deployment object representing the primary is explicitly
scaled from 0 to 1 Pods. However, there is a case that may
trigger the Deployment scaling before the Deployment is created.
As such, we can add a guard to check that the Deployment is
actually created before proceeding with the scaling operation.

This patch also modifies a debug line around the cluster
Deployment creation, as the debug output came after the
point at which the code may have exited due to error.

### DIFF
--- a/internal/controller/pod/inithandler.go
+++ b/internal/controller/pod/inithandler.go
@@ -18,6 +18,7 @@ limitations under the License.
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/crunchydata/postgres-operator/internal/config"
 	"github.com/crunchydata/postgres-operator/internal/controller"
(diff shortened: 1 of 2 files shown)


internal/controller/pod/inithandler.go
```diff
@@ -291,3 +304,40 @@ func (c *Controller) labelPostgresPodAndDeployment(newpod *apiv1.Pod) {
...
+			fields.OneTermEqualSelector(config.LABEL_PG_DATABASE, config.LABEL_TRUE),
+			fields.OneTermEqualSelector(config.LABEL_DEPLOYMENT_NAME, primaryDeploymentName),
+		).String(),
+	}
+
+	// as the poll will initially wait 5 seconds, do a quick check to see if the
+	// deployment already exists, as most cases will fall into this
+	if d, err := c.Client.AppsV1().Deployments(cluster.Namespace).List(ctx, options); err == nil && len(d.Items) > 0 {
+		return nil
+	}
```
cbandy, line 326: :memo: [`wait.PollImmediate()`](https://pkg.go.dev/k8s.io/apimachinery@v0.20.1/pkg/util/wait#PollImmediate) will perform the check then wait.

jkatz: @cbandy thanks, i'll update to use that

andrewlecuyer (approved):

PR #2305, opened 2021-02-27
Automatically detect proper setting for "disable FSGroup"

jkatz:
In particular, we want to disable explicitly setting the FSGroup
when running in OpenShift, i.e. when using the "restricted" SCC
(which is the default).

This adds detection to see if we are within an OpenShift cluster
and, if so, will automatically disable the FSGroup.

This also removes the default explicit setting of the disable
FSGroup feature from the installer. However, one can still set
this option should they so desire.

Issue: [ch10571]


cbandy (approved):

docs/content/quickstart/_index.md
```diff
@@ -326,7 +326,7 @@ primary_storage: "nfsstorage"
 replica_storage: "nfsstorage"
 ```
 
-If you are using either Openshift or CodeReady Containers and you have a `restricted` Security Context Constraint, you will need to set `disable_fsgroup` to `true` in order to deploy the PostgreSQL Operator.
+In OpenShift and CodeReady Containers, the PostgreSQL Operator will automatically set `disable_fsgroup` to `true` so that it will deploy PostgreSQL clusters correctly under the `restricted` Security Context Constraint (SCC). Though we recommend using `restricted`, If you are using the `anyuid` SCC, you will need to set `disable_fsgroup` to `false` in order to deploy the PostgreSQL Operator.
```
andrewlecuyer, line 329: Lowercase "if", i.e. `..if you are using`.

docs/content/installation/configuration.md
```diff
@@ -46,7 +46,7 @@ Operator.
 | `delete_operator_namespace` | false |  | Set to configure whether or not the PGO operator namespace (defined using variable `pgo_operator_namespace`) is deleted when uninstalling the PGO. |
 | `delete_watched_namespaces` | false |  | Set to configure whether or not the PGO watched namespaces (defined using variable `namespace`) are deleted when uninstalling the PGO. |
 | `disable_auto_failover` | false |  | If set, will disable autofail capabilities by default in any newly created cluster |
-| `disable_fsgroup` | false |  | Set to `true` for deployments where you do not want to have the default PostgreSQL fsGroup (26) set. The typical usage is in OpenShift environments that have a `restricted` Security Context Constraints. |
+| `disable_fsgroup` |  |  | Set to `true` for deployments where you do not want to have the default PostgreSQL fsGroup (26) set. The typical usage is in OpenShift environments that have a `restricted` Security Context Constraints. If you use the `anyuid` SCC, you would want to set this to `false`. The Postgres Operator will set this value by default, except for the `anyuid` SCC. |
```
andrewlecuyer, line 49: Should the last sentence be:

>  The Postgres Operator will set this value properly by default, except for when using the `anyuid` SCC.

andrewlecuyer (approved):
Just a couple small notes, otherwise looks good.

PR #2306, opened 2021-02-28
Security context modifications

jkatz:
This adds some explicit additions to the security contexts to the containers within the PostgreSQL Operator ecosystem, including:

- Ensuring the containers are run as non-root users.
- Disallow privilege escalation, since it's not needed.
- Disallow writing to the root container filesystem.
- Ensure the containers run in unprivileged mode, though this is the environmental default.

Issue: [ch7571]
Issue: [ch10570]

### DIFF
--- a/deploy/deployment.json
+++ b/deploy/deployment.json
@@ -33,7 +33,8 @@
                         "image": "$PGO_IMAGE_PREFIX/pgo-apiserver:$PGO_IMAGE_TAG",
                         "imagePullPolicy": "IfNotPresent",
                         "securityContext": {
-                          "allowPrivilegeEscalation": false
+                          "allowPrivilegeEscalation": false,
+                          "readOnlyRootFilesystem": true
                         },
                         "ports": [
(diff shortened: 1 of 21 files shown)


cbandy (approved):
⛏️  I'm not excited about some of these being set on the Pod and some on the Container. While `runAsNonRoot` can be on either, the setting on Container takes precedence. I blame the Kubernetes API for making two places for one security setting.

:camping: `privileged` defaults to `false` on Container; should we make that explicit, too?

andrewlecuyer (approved):

jkatz:
> ⛏️ I'm not excited about some of these being set on the Pod and some on the Container. While `runAsNonRoot` can be on either, the setting on Container takes precedence. I blame the Kubernetes API for making two places for one security setting.

If someone is going in and modifying them specifically at the container level, that is their prerogative.

> 🏕️ `privileged` defaults to `false` on Container; should we make that explicit, too?

Yeah, that seems reasonable.

PR #2340, opened 2021-03-18
Maintain YAML & JSON Ordering for DCS Config

andrewlecuyer:
Testing has shown that with versions of PostgreSQL earlier than v13, changes to the ordering of the various Patroni and PostgreSQL settings stored within a cluster's DCS (as stored in the `config` annotation of the `<cluster-name>-config` ConfigMap) can cause a replica to automatically restart.  And being that the go standard library does not guarantee ordering when encoding/decoding JSON, the DCS configuration can be re-ordered when the `<cluser-name>-pgha-config` ConfigMap is updated, and the operator subsequently syncs its contents with the `<cluster-name>-config` ConfigMap.  This, in turn, can lead to the automatic restart of a replica following a configuration change in the `<cluser-name>-pgha-config`, since the settings might be patched in a different order than initially stored.

While in general order for JSON typically should not matter, this commit ensures order is maintained when marshalling and unmarshalling the DCS configuration found within both the `<cluster-name>-config` and `<cluster-name>-pgha-config` ConfigMaps into both YAML and JSON.  With this change,a replica will no longer automatically following a configuration change, but will rather show a pending restart as expected.

**Checklist:**

 
**Type of Changes:**

 
When using PG v12 or less, after cluster creation replicas are sometimes automatically restarted following configuration changes that require a PG restart.

[ch10876]

When using any supported PG version, any change to a PG setting requiring a PG restart results in a pending restart for all instances (including replicas).

**Other information**:

N/A

### DIFF
--- a/internal/operator/config/dcs.go
+++ b/internal/operator/config/dcs.go
@@ -27,12 +27,16 @@ import (
 	"github.com/crunchydata/postgres-operator/internal/util"
 	"github.com/iancoleman/orderedmap"
 	log "github.com/sirupsen/logrus"
-	goyaml "gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/yaml"
+
+	// This yaml package is needed for a specific requirement within this package for maintaining
+	// the order of YAML keys when encoding/decoding.  Outside of this specific use-case, the
+	// 'sigs.k8s.io/yaml' should be leveraged for all other required YAML functionality.
+	goyaml "gopkg.in/yaml.v2"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


internal/operator/config/dcs.go
```diff
@@ -25,7 +25,9 @@ import (
 	"github.com/crunchydata/postgres-operator/internal/config"
 	"github.com/crunchydata/postgres-operator/internal/kubeapi"
 	"github.com/crunchydata/postgres-operator/internal/util"
+	"github.com/iancoleman/orderedmap"
 	log "github.com/sirupsen/logrus"
+	goyaml "gopkg.in/yaml.v2"
```
jkatz, line 30: Can we use the YAML library provided by the Kube sig? This is:

```
"sigs.k8s.io/yaml"
```

You can see references in the code.

jkatz (changes requested):

jkatz (approved):

andrewlecuyer:
@jkatz added a comment explaining why the `gopkg.in/yaml.v2` package is needed.

PR #2355, opened 2021-03-28
Allow for a S3 bucket name to be edited after cluster creation

jkatz:
This allows for the name of a S3 bucket to be edited after a
cluster is created, i.e. through the custom resource. Should this
action be taken, the following occurs:

- The pgBackRest report Pod is redeployed
- The PostgreSQL cluster is redeployed using a rolling update
- Once the above two actions complete, a new stanza is created and
an initial backup in that stanza is created.

A create stanza must be attempted as the new bucket may be
entirely empty.

Issue: [ch11061]

### DIFF
--- a/docs/content/architecture/high-availability/_index.md
+++ b/docs/content/architecture/high-availability/_index.md
@@ -443,5 +443,6 @@ modification to the custom resource:
 - Custom annotation changes
 - Enabling/disabling the monitoring sidecar on a PostgreSQL cluster (`--metrics`)
 - Enabling/disabling the pgBadger sidecar on a PostgreSQL cluster (`--pgbadger`)
+- S3 bucket name updates
 - Tablespace additions
 - Toleration modifications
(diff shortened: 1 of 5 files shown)


andrewlecuyer (changes requested):
Looks great.  However, it does look like there is one more piece needed in support of "restore from" functionality.

Specifically, annotations on the pgBackRest Secret are currently utilized to store pertinent pgBackRest configuration information as needed to ensure subsequent cluster can use it to bootstrap once  (this Secret is retained along with the pgBackRest repo PVC when deleting a cluster with `--keep-backups`):

https://github.com/CrunchyData/postgres-operator/blob/04cd81d903fe6dd82bb8fc65c26a3b5d4898b1ea/internal/operator/cluster/cluster.go#L666

We'll therefore simply want to patch/update that same annotation in the pgBackRest config secret when the S3 bucket is updated in the spec.

```go
patch, err := kubeapi.NewMergePatch().Add("metadata", "annotations")(map[string]string{
		config.ANNOTATION_S3_BUCKET:           newBucketName,
}).Bytes()
```

jkatz:
@andrewlecuyer Feedback addressed.

andrewlecuyer (approved):

PR #2356, opened 2021-03-28
Adds nss_wrapper Support for readOnlyRootFileSystem in OpenShift

### LINKED ISSUE #2357: Stanza cannot connect to primary
**Describe the bug**
After upgrade from 4.6.1 to 4.6.2 the cluster upgrade can recreate psql cluster but stanza cannot run.
It said it cannot connect to (primary pod IP), and unable to find primary cluster.
It is the same if I run pgo backup on this cluster.

**To Reproduce**
1. Upgrade operator (oc apply -f postgres-operator.yml)
2. Upgrade any postgres cluster
3. Stanza error

Openshift 4.7, Crunchy 4.6.2

**Additional context**
After I downgrade to 4.6.1 it is working again.

andrewlecuyer:
Adds support for the `nss_wrapper` as implemented within the Crunchy Container Suite, specifically as needed for compatibility with the `restricted` SCC in OpenShift (and therefore the random UID's that are assigned to containers by OpenShift). This is done in accordance with RedHat guidance (see https://access.redhat.com/articles/4859371), and ensures a `readOnlyRootFileSystem` can be enabled when the PostgreSQL Operator is deployed in an OpenShift environment.

To support the `nss_wrapper` solution, applicable PostgreSQL and pgBackRest Deployment templates have been updated with the proper `nss_wrapper` environment variables. Additionally, the default SSHD configuration file has also been updated to ensure the proper `nss_wrapper` environment variables are set when pgBackRest executes various commands over SSH.

**Checklist:**

 
**Type of Changes:**

 
`readOnlyRootFileSystem` **_cannot be_** utilized with the `restricted` SecurityContextConstraint in OpenShift.

[ch9887]

`readOnlyRootFileSystem` **_can_** be utilized with the `restricted` SecurityContextConstraint in OpenShift.

**Other information**:

One side-effect of including the `NSS_WRAPPER` environment variables in the Deployment templates is that the `nss_wrapper` will be configured before the UID scripts (which configured the proper `nss_wrapper` `passwd` and `group` files) have been run, which can lead to harmless (as verified by testing) errors early in the initialization of containers with these environment variables set in their associated Pod spec (which currently includes the `crunchy-postgres-ha` and `crunchy-pgbackrest-repo` containers):

```bash
$ kubectl logs pgo-test-4ndsm-664df48dc6-vdnkd -n pgouser1
NWRAP_ERROR(1) - nwrap_files_cache_reload: Unable to open '/tmp/nss_wrapper/postgres-ha/passwd' readonly -1:No such file or directory
NWRAP_ERROR(1) - nwrap_files_getpwuid: Error loading passwd file
nss_wrapper: user exists
nss_wrapper: group exists
nss_wrapper: environment configured
...
```

However, setting these environment variables in the Pod spec ensures that `pgbackrest` commands function normally, without requiring wrapper scripts to call `pgbackrest` commands (e.g. users can still call `pgbackrest info` instead of having to use `source /opt/crunchy/bin/nss_wrapper.end && pgbackrest info` to get repository information).  [_One potential way to address this in the future would be to utilize an init container to generate the `nss_wrapper` `passwd` and `group` files, which will ensure the `nss_wrapper` configuration is ready by the time a container utilizing the `NSS_WRAPPER` environment variables is run._]

Additionally, the use of `ForceCommand` in the `sshd_config` prevents the ability to create interactive login shells bvetween the pgBackRest deidicated repo host and various instance Pods, but non-interactive commands (such as those used by `pgbackrest`) can still be utilized to test SSH connectivity.  Interactive-logins could be enabled if need be (e.g. as describe [here](https://unix.stackexchange.com/a/526855)), but this would be solely for troubleshooting purposes, and not for any functional requirement within pgBackRest.

And finally, please note that the `SetEnv` setting in OpenSSH is not available in all versions of OpenSSH utilized by the Operator across both CentOS 7 & 8 and UBI 7 & 8, with `ForceCommand` providing an approach that is compatible among versions, which also ensures the proper `nss_wrapper` environment is _always_ set server-side (i.e. without giving the client the ability to control pertinent `nss_wrapper` environment variables).

closes #2357


bin/common/nss_wrapper.sh
```diff
@@ -0,0 +1,67 @@
...
+
+CRUNCHY_DIR=${CRUNCHY_DIR:-'/opt/crunchy'}
+
+# The following sets up an nss_wrapper environment in accordance with OpenShift
+# guidance for supporting arbitrary user ID's
+
+# define nss_wrapper directory and passwd & group files that will be utilized by nss_wrapper
+NSS_WRAPPER_DIR="/tmp/nss_wrapper/${NSS_WRAPPER_SUBDIR}"
+NSS_WRAPPER_PASSWD="${NSS_WRAPPER_DIR}/passwd"
+NSS_WRAPPER_GROUP="${NSS_WRAPPER_DIR}/group"
```
jkatz, line 24: Would appreciate a comment about why we need this here vs. the env var sourcing that occurs later.

build/crunchy-postgres-exporter/Dockerfile
```diff
@@ -19,13 +20,22 @@ RUN if [ "$DFSET" = "centos" ] ; then \
...
         postgresql${PGVERSION} \
+		gettext \
+		nss_wrapper \
     && ${PACKAGER} -y clean all ; \
 fi
 
+# needed for nss_wrapper
+RUN if [ "$BASEOS" = "rhel7" ] ; then \
+	${PACKAGER} -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm ; \
+fi
```
jkatz, line 31: I beileve we need this for `ubi7`/`ubi8`. In fact, I believe we can remove all of the RHEL bits.

build/crunchy-postgres-exporter/Dockerfile
```diff
@@ -19,13 +20,22 @@ RUN if [ "$DFSET" = "centos" ] ; then \
...
+
 RUN if [ "$DFSET" = "rhel" ] ; then \
 	${PACKAGER} install -y \
         --setopt=skip_missing_names_on_install=False \
         postgresql${PGVERSION} \
+		gettext \
+		nss_wrapper \
     && ${PACKAGER} -y clean all ; \
 fi
```
jkatz, line 40: Need to set this up for ubi7/ubi8.

installers/ansible/roles/pgo-operator/files/pgo-backrest-repo/sshd_config
```diff
@@ -125,3 +125,6 @@ Subsystem	sftp	/usr/libexec/openssh/sftp-server
 #	AllowTcpForwarding no
 #	PermitTTY no
 #	ForceCommand cvs server
+
+# ensure nss_wrapper env vars are set when executing commands as needed for OpenShift compatibility
+ForceCommand NSS_WRAPPER_SUBDIR=ssh . /opt/crunchy/bin/nss_wrapper_env.sh && $SSH_ORIGINAL_COMMAND
```
jkatz, line 130: Feels like we need to note that existing clusters need to reload their sshd_config? I believe this is recreated as part of the upgrade process?

installers/ansible/roles/pgo-operator/files/pgo-configs/cluster-deployment.json
```diff
@@ -145,6 +145,18 @@
...
                         "name": "PGHOST",
                         "value": "/tmp"
+                    }, 
+                    {
+                      "name": "LD_PRELOAD",
+                      "value": "/usr/lib64/libnss_wrapper.so"
+                    },
+                    {
+                      "name": "NSS_WRAPPER_PASSWD",
+                      "value": "/tmp/nss_wrapper/postgres-ha/passwd"
```
jkatz, line 155: Noting that this needs to be reviewed in the context of https://github.com/CrunchyData/crunchy-containers/pull/1337 based on where some of those settings are being stored based on that name (e.g. `postgres-ha`).

jkatz (changes requested):

build/crunchy-postgres-exporter/Dockerfile
andrewlecuyer: I did not have a problem installing `nss_wrapper` in `ubi7/ubi8` without installing `epel-release`.

build/crunchy-postgres-exporter/Dockerfile
andrewlecuyer: `DFSET` `rhel`  includes both `ubi7` and `ubi8`.

jkatz: OK.

build/crunchy-postgres-exporter/Dockerfile
jkatz: OK, we can remove this line then.

jkatz, line 31:
```suggestion
```

installers/ansible/roles/pgo-operator/files/pgo-configs/cluster-deployment.json
andrewlecuyer: Thanks, great catch - think the answer here will be to keep the directory the same across all PG containers, whether HA, GIS, etc.  I'll adjust accordingly.

build/crunchy-postgres-exporter/Dockerfile
andrewlecuyer: These lines have been removed.

andrewlecuyer: Removed as suggested.

installers/ansible/roles/pgo-operator/files/pgo-configs/cluster-deployment.json
andrewlecuyer: Updated the path to `/tmp/nss_wrapper/postgres`, as will now be utilized for all variants of the PG container (HA, PostGIS, etc.).

bin/common/nss_wrapper.sh
andrewlecuyer: Comment added.

jkatz (approved):
LGTM.

@andrewlecuyer Are there any special considerations when upgrading?

installers/ansible/roles/pgo-operator/files/pgo-backrest-repo/sshd_config
andrewlecuyer: Another [commit](https://github.com/CrunchyData/postgres-operator/pull/2356/commits/1bfab07ac063c506510342f24755b5e42c0e5ba7) has been pushed to address upgrade scenarios (it turns out that the `recreateBackrestRepoSecret()` function does not actually recreate the `sshd_config` in the pgBackRest repo secret).

In testing this I also found that the recent change to `UsePAM` was not being brought over during the upgrade, so the commit addresses that issue as well (i.e. it ensures `UsePAM` is set to `no`).

internal/operator/cluster/upgrade.go
```diff
@@ -51,6 +52,24 @@ const (
...
+var (
+	// nssWrapperRegex is the regular expression that is utilized to determine if the nss_wrapper
+	// ForceCommand setting is missing from the sshd_config (as it would be for versions prior to
+	// v4.7)
+	nssWrapperRegex = regexp.MustCompile(nssWrapperForceCommand)
+
+	// nssWrapperRegex is the regular expression that is utilized to determine if the UsePAM
+	// setting is set to 'yes' in the sshd_config (as it might be for versions up to v4.6.1,
+	// v4.5.2 and v4.4.2)
+	usePAMRegex = regexp.MustCompile(`(?m)^UsePAM\s*yes`)
```
jkatz, line 70: @andrewlecuyer I would make the `yes` case insensitive

internal/operator/cluster/upgrade.go
```diff
@@ -51,6 +52,24 @@ const (
...
+// the following regex expressions are used when upgrading the sshd_config file for a PG cluster
+var (
+	// nssWrapperRegex is the regular expression that is utilized to determine if the nss_wrapper
+	// ForceCommand setting is missing from the sshd_config (as it would be for versions prior to
+	// v4.7)
+	nssWrapperRegex = regexp.MustCompile(nssWrapperForceCommand)
+
+	// nssWrapperRegex is the regular expression that is utilized to determine if the UsePAM
+	// setting is set to 'yes' in the sshd_config (as it might be for versions up to v4.6.1,
+	// v4.5.2 and v4.4.2)
```
jkatz, line 69:
```suggestion
	// v4.5.2 and v4.4.3)
```

internal/operator/cluster/upgrade.go
andrewlecuyer: OK, though that would mean we're essentially fixing a reconfigured OpenSSH server, since OpenSSH requires these to be lowercase (i.e. the only reason we'd see anything other than `yes` would be a mis-configuration).

jkatz: I believe `Yes` worked in EL7. So I'm OK with this.

internal/operator/cluster/upgrade.go
andrewlecuyer: Version updated, thanks!

internal/operator/cluster/upgrade.go
andrewlecuyer: Made the entire thing case-insensitive after reading this in the `sshd` manual:

> note that keywords are case-insensitive and arguments are case-sensitive

jkatz (approved):

PR #2368, opened 2021-04-12
Add support for using GCS with pgBackRest

### LINKED ISSUE #1791: Option to backup and restore from gcp storage bucket (GCS)
**What is the motivation or use case for the change? **
I do not see an option using the crunchydata operator to backup and restore from GCP storage bucket.

I would to see an option while using the postgres-operator to backup and restore from GCP storage bucket.

* Operating System: Linux based 
* Where is this running ( Local, Cloud Provider) - GCP GKE cluster
* Storage being used (NFS, Hostpath, Gluster, etc):GCP PVC, 
* Container Image Tag: centos7-4.3.0 / centos7-12.2-4.3.0
* PostgreSQL Version:12.2
* Platform (Docker, Kubernetes, OpenShift): Kubernetes
* Platform Version: 1.16.11-gke.5

**Additional context**
Add any other context or screenshots about the enhancement request here.

jkatz:
pgBackRest 2.33 introduced support for GCS storage as a means
for taking backups. PGO now adds orchestration for using GCS
with its pgBackRest integration.

This adds a few new attributes to the pgclusters.crunchydata.com
custom resource to enable GCS support with PGO, including:

- BackrestGCSBucket (required)
- BackrestGCSEndpoint
- BackrestGCSKeyType

The pgBackRest repository Secret now supports a key called
"gcs-key", which references the GCS credential.

Similarly, additional flags are now available in the
`pgo create cluster` command to enable GCS support, including:

- `--pgbackrest-gcs-bucket`
- `--pgbackrest-gcs-endpoint`
- `--pgbackrest-gcs-key`
- `--pgbackrest-gcs-key-type`

Note that `--pgbackrest-gcs-key` references a file path in your
local environment. The GCS credential is a JSON file; for
convenience, the PGO client will accept the file and handle the
upload.

There are also several installation configuration parameters
available if, for example, you are using the same bucket.

The two parameters that are required are the GCS bucket name and
the GCS key; pgBackRest can figure out the rest.

This supports all of the same things that PGO supports around S3;
you can even create standbys using GCS.

However, note that in a "hybrid" setup, you can only use "posix,gcs";
"s3,gcs" is not supported at this time. In other words, the following
storage types are supported:

- posix
- s3
- gcs
- posix,s3
- posix,gcs

Issue: [ch11177]
closes #1791


andrewlecuyer (approved):
Looks and works great!  Just a few comments.

internal/config/annotations.go
```diff
@@ -45,6 +45,12 @@ const (
 	ANNOTATION_REPO_PATH = "repo-path"
 	// ANNOTATION_PG_PORT is for storing the PostgreSQL port for a cluster
 	ANNOTATION_PG_PORT = "pg-port"
+	// ANNOTATION_GCS_BUCKET is for storing the name of the GCS bucket used by
```
andrewlecuyer, line 48: :question: Do we also want to track `gcs-key-type`?  I'm not sure if there is any practical reason this would ever be anything but the default of `service` out in the wild (the pgBackRest docs mention that `token` is for local testing, so thinking not...), but if so, we would also want to track which type of key is in the secret to know how to use it in a subsequent restore.

jkatz: Yup, that's why I didn't add it, but it seems like it's fairly easy to add, so...may as well.

PR #2405, opened 2021-04-26
Update package requirements for exporter image

jkatz:
There is an additional requirement for the exporter image to
include findutils.


crunchyheath (approved):
This is reasonable to me

PR #2408, opened 2021-04-27
Give names to the ports

### LINKED ISSUE #2407: Add name of port for Prometheus Operator
**What is the motivation or use case for the feature? **
Easier monitoring of databases with external Prometheus using Prometheus Operator 

Add port name to database port. Now when database is deployed with --metrics flag we have:
    image: registry.connect.redhat.com/crunchydata/crunchy-postgres-exporter:ubi7-4.5.1
    imagePullPolicy: IfNotPresent
    name: postgres-exporter
    ports:
    - containerPort: 9187
      protocol: TCP
      
and Prometheus Operator need name of that port so it shoud look like:
    - containerPort: 9187
      name: postexport (in my version it has to be less than 15 characters) 
      protocol: TCP
    
    https://github.com/prometheus-operator/prometheus-operator
    https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/user-guides/getting-started.md
    
That will be useful if someone want collect metrics but using own stack without deploying solutions from Crunchy.

jkatz:
This makes it easier for PGO Pods to work with other systems.

closes #2407

### DIFF
--- a/deploy/deployment.json
+++ b/deploy/deployment.json
@@ -38,7 +38,7 @@
                           "readOnlyRootFilesystem": true
                         },
                         "ports": [
-                            { "containerPort": $PGO_APISERVER_PORT }
+                            { "containerPort": $PGO_APISERVER_PORT, "name": "api" }
                         ],
                         "readinessProbe": {
                             "httpGet": {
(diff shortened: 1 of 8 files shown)


cbandy (commented):
Names lgtm.

PR #2422, opened 2021-05-03
Remove TLS cert generation for API server from installer

jkatz:
The apiserver will reconcile its own TLS certificate -- it has been
doing so for awhile. This unifies the method to ensure that only
the apiserver will generate its certificate, unless the user explicitly
provides one.

Issue: [ch11380]


jmckulk (approved):
This looks like it should work to me :+1: 

One thing i would like to see/consider is moving the `client-setup.sh` script up a level in the installers directory. Every installer that we have uses this script (or at least the main ones) so it makes sense to me to have it in a more communal place. Definitely non-blocking :thinking:

PR #2423, opened 2021-05-03
Include additional dependencies into minimal image

jkatz:
This includes some of the process debugging utils, vi, less,
and ensuring tzdata is present[1]

[1] https://access.redhat.com/solutions/5616681

Issue: [ch11367]


crunchyheath (approved):
+1 from me

PR #2558, opened 2021-07-21
Implement Related Images for OLM Installer

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
This commit adds operator support for the "related images" concept
as defined by the Operator Lifecycle Manager (OLM). This specifically
allows PGO to run in a restricted environment (e.g. the images can be
installed "offline"). Essentially, when the image parameters are not
defined on the postgrescluster spec, the Operator attempts to retrieve
the image value from a "RELATED_IMAGE_*" environment variable for the
relevant image. For the pgBouncer, pgBackRest and Postgres Exporter
images there will be one variable each. The PostgreSQL image will have
one unique variable for each supported Postgres version, and the PostGIS
enabled PostgreSQL image will have 2 variables for each Postgres version,
one for each supported PostGIS version per PostgreSQL version.

**Other information**:
[ch11988]

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -13,6 +13,16 @@ spec:
         env:
         - name: CRUNCHY_DEBUG
           value: "true"
+        - name: RELATED_IMAGE_POSTGRES_13
+          value: "registry.developers.crunchydata.com/crunchydata/crunchy-postgres-ha:centos8-13.3-0"
+        - name: RELATED_IMAGE_POSTGRES_13_GIS_3.1
+          value: "registry.developers.crunchydata.com/crunchydata/crunchy-postgres-gis-ha:centos8-13.3-3.1-0"
+        - name: RELATED_IMAGE_PGBACKREST
+          value: "registry.developers.crunchydata.com/crunchydata/crunchy-pgbackrest:centos8-2.33-0"
+        - name: RELATED_IMAGE_PGBOUNCER
+          value: "registry.developers.crunchydata.com/crunchydata/crunchy-pgbouncer:centos8-1.15-0"
+        - name: RELATED_IMAGE_PGEXPORTER
+          value: "registry.developers.crunchydata.com/crunchydata/crunchy-postgres-exporter:ubi8-5.0.0-0"
         securityContext:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
--- a/docs/content/references/crd.md
+++ b/docs/content/references/crd.md
@@ -100,6 +100,11 @@ PostgresClusterSpec defines the desired state of PostgresCluster
         <td>Specifies a data source for bootstrapping the PostgreSQL cluster.</td>
         <td>false</td>
       </tr><tr>
+        <td><b>image</b></td>
+        <td>string</td>
+        <td>The image name to use for PostgreSQL containers. When omitted, the value comes from an operator environment variable. For standard PostgreSQL images, the format is RELATED_IMAGE_POSTGRES_{postgresVersion}, e.g. RELATED_IMAGE_POSTGRES_13. For PostGIS enabled PostgreSQL images, the format is RELATED_IMAGE_POSTGRES_{postgresVersion}_GIS_{postGISVersion}, e.g. RELATED_IMAGE_POSTGRES_13_GIS_3.1.</td>
+        <td>false</td>
+      </tr><tr>
         <td><b><a href="#postgresclusterspecimagepullsecretsindex">imagePullSecrets</a></b></td>
         <td>[]object</td>
         <td>The image pull secrets used to pull from a private registry Changing this value causes all running pods to restart. https://k8s.io/docs/tasks/configure-pod-container/pull-image-private-registry/</td>
@@ -130,6 +135,11 @@ PostgresClusterSpec defines the desired state of PostgresCluster
         <td>The port on which PostgreSQL should listen.</td>
         <td>false</td>
       </tr><tr>
+        <td><b>postGISVersion</b></td>
+        <td>string</td>
+        <td>The PostGIS extension version installed in the PostgreSQL image. When image is not set, indicates a PostGIS enabled image will be used.</td>
+        <td>false</td>
+      </tr><tr>
         <td><b><a href="#postgresclusterspecproxy">proxy</a></b></td>
         <td>object</td>
         <td>The specification of a proxy that connects to PostgreSQL.</td>
@@ -150,11 +160,6 @@ PostgresClusterSpec defines the desired state of PostgresCluster
         <td>PostgreSQL backup configuration</td>
         <td>true</td>
       </tr><tr>
-        <td><b>image</b></td>
-        <td>string</td>
-        <td>The image name to use for PostgreSQL containers</td>
-        <td>true</td>
-      </tr><tr>
         <td><b><a href="#postgresclusterspecinstancesindex">instances</a></b></td>
         <td>[]object</td>
         <td></td>
@@ -162,7 +167,7 @@ PostgresClusterSpec defines the desired state of PostgresCluster
       </tr><tr>
         <td><b>postgresVersion</b></td>
         <td>integer</td>
-        <td>The major version of PostgreSQL installed in the PostgreSQL container</td>
+        <td>The major version of PostgreSQL installed in the PostgreSQL image</td>
         <td>true</td>
       </tr></tbody>
 </table>
@@ -559,15 +564,15 @@ PGMonitorSpec defines the desired state of the pgMonitor tool suite
         <td>Projected volumes containing custom PostgreSQL Exporter configuration.  Currently supports the customization of PostgreSQL Exporter queries. If a "queries.yaml" file is detected in any volume projected using this field, it will be loaded using the "extend.query-path" flag: https://github.com/prometheus-community/postgres_exporter#flags Changing the values of field causes PostgreSQL and the exporter to restart.</td>
         <td>false</td>
       </tr><tr>
+        <td><b>image</b></td>
+        <td>string</td>
+        <td>The image name to use for crunchy-postgres-exporter containers. The image may also be set using the RELATED_IMAGE_PGEXPORTER environment variable.</td>
+        <td>false</td>
+      </tr><tr>
         <td><b><a href="#postgresclusterspecmonitoringpgmonitorexporterresources">resources</a></b></td>
         <td>object</td>
         <td>Changing this value causes PostgreSQL and the exporter to restart. More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers</td>
         <td>false</td>
-      </tr><tr>
-        <td><b>image</b></td>
-        <td>string</td>
-        <td>The image name to use for crunchy-postgres-exporter containers</td>
-        <td>true</td>
       </tr></tbody>
 </table>
 
@@ -1072,6 +1077,11 @@ Defines a PgBouncer proxy and connection pooler.
         <td>A secret projection containing a certificate and key with which to encrypt connections to PgBouncer. The "tls.crt", "tls.key", and "ca.crt" paths must be PEM-encoded certificates and keys. Changing this value causes PgBouncer to restart. More info: https://kubernetes.io/docs/concepts/configuration/secret/#projection-of-secret-keys-to-specific-paths</td>
         <td>false</td>
       </tr><tr>
+        <td><b>image</b></td>
+        <td>string</td>
+        <td>Name of a container image that can run PgBouncer 1.15 or newer. Changing this value causes PgBouncer to restart. The image may also be set using the RELATED_IMAGE_PGBOUNCER environment variable. More info: https://kubernetes.io/docs/concepts/containers/images</td>
+        <td>false</td>
+      </tr><tr>
         <td><b><a href="#postgresclusterspecproxypgbouncermetadata">metadata</a></b></td>
         <td>object</td>
         <td>Metadata contains metadata for PostgresCluster resources</td>
@@ -1096,11 +1106,6 @@ Defines a PgBouncer proxy and connection pooler.
         <td>[]object</td>
         <td>Tolerations of a PgBouncer pod. Changing this value causes PgBouncer to restart. More info: https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration</td>
         <td>false</td>
-      </tr><tr>
-        <td><b>image</b></td>
-        <td>string</td>
-        <td>Name of a container image that can run PgBouncer 1.15 or newer. Changing this value causes PgBouncer to restart. More info: https://kubernetes.io/docs/concepts/containers/images</td>
-        <td>true</td>
       </tr></tbody>
 </table>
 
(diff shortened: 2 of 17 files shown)


internal/config/config_test.go
```diff
@@ -0,0 +1,175 @@
...
+func TestSetContainerImage(t *testing.T) {
+
+	t.Run("env var not set and spec empty", func(t *testing.T) {
+
+		assert.Equal(t, SetContainerImage("", CrunchyPostgresHA_PG10), "")
+	})
+
+	t.Run("image name from spec", func(t *testing.T) {
+
+		os.Setenv(CrunchyPostgresHA_PG10, "envVarImage")
```
cbandy, line 46: 🔧 Go 1.17 has a `testing.T.Setenv()` that makes sure to undo the effect when the test ends. https://tip.golang.org/doc/go1.17#testing

We should do something similar with a note to replace it with the standard one once we are on Go 1.17.
https://github.com/golang/go/blob/3e48c0381fd1/src/testing/testing.go#L986

```go
func setEnv(t testing.TB, key, value string) {
	t.Helper()
	… os.LookupEnv(key)
	… os.Setenv(key, value)
	t.Cleanup(…)
}

File: internal/config/config_test.go [hunk 2]
```diff
@@ -0,0 +1,175 @@
...
+		},
+		Spec: v1beta1.PostgresClusterSpec{
+			Metadata:        &v1beta1.Metadata{},
+			PostgresVersion: 12,
+			InstanceSets:    []v1beta1.PostgresInstanceSetSpec{},
+		},
+	}
+
+	cluster1.SetGroupVersionKind(v1.SchemeGroupVersion.WithKind("postgrescluster"))
+	if err := tClient.Create(ctx, cluster1); err != nil {
```
cbandy, line 95: ♻️ I suspect we don't need an entire Kubernetes for these tests because the code does not call the API. If so, drop the `+build envtest`, too.

internal/config/config_test.go
```diff
@@ -0,0 +1,175 @@
...
+			naming.PostGISVersion: "3.0",
+		}
+		os.Setenv(CrunchyPostgresHA_PG12_GIS_3_0, "envVarGISImage")
+		assert.Equal(t, SetPGContainerImage(cluster1), "envVarGISImage")
+	})
+
+	t.Run("return GIS image name from spec", func(t *testing.T) {
+
+		cluster1.Spec.Image = "specGISImageName"
+		os.Setenv(CrunchyPostgresHA_PG12_GIS_3_0, "envVarGISImage")
```
cbandy, line 136: ⛏️ This test would be more clear if it set the PostGIS annotation as well. It's relying on the prior test as-is.

cbandy (commented):
A bunch of nitpicks about naming and two comments on the tests.

internal/config/config.go
```diff
@@ -0,0 +1,137 @@
...
+// must be made available. Because a user could potentially use any of the images
+// below, each is defined separately. (See
+//	https://docs.openshift.com/container-platform/4.7/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-restricted-network_osdk-generating-csvs)
+// This approach works with the concept of image streams and can allow for automatic
+// updates when the container image is changed (See
+//  https://docs.openshift.com/container-platform/4.7/openshift_images/images-understand.html#images-imagestream-use_images-understand)
+
+const (
+	// One variable for each version of Postgres
+	CrunchyPostgresHA_PG10 = "RELATED_IMAGE_CRUNCHY_POSTGRES_HA_PG_10"
```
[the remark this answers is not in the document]
jkatz: :+1:

internal/postgres/reconcile.go
```diff
@@ -168,7 +169,7 @@ func InstancePod(ctx context.Context,
 		// Patroni will set the command and probes.
 
 		Env:       Environment(inCluster),
-		Image:     inCluster.Spec.Image,
+		Image:     config.SetPGContainerImage(inCluster),
```
cbandy, line 172: ❓ Did we decide the CRD should now say `image` fields are optional? "Required and blank" to trigger this functionality doesn't sound right, so I think we said "yes?"

andrewlecuyer: Agree that we had decided to make `image` fields options in support of this change :+1:

internal/naming/annotations.go
```diff
@@ -45,4 +45,8 @@ const (
 	// timestamp), which will be stored in the PostgresCluster status to properly track completion
 	// of the Job.
 	PGBackRestRestore = annotationPrefix + "pgbackrest-restore"
+
+	// PostGISVersion is the annotation that is used to determine the proper PostGIS container version
+	// to use with the
+	PostGISVersion = annotationPrefix + "postgis-version"
```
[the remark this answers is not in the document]
tjmoore4: Sure, wfm. Updated.

internal/config/config.go
```diff
@@ -0,0 +1,137 @@
...
+
+	if postgresCluster.Spec.Image != "" {
+		return postgresCluster.Spec.Image
+	}
+
+	var version string
+	var ok bool
+	if annotations := postgresCluster.Spec.Metadata.GetAnnotationsOrNil(); annotations != nil {
+		version, ok = annotations[naming.PostGISVersion]
+	}
```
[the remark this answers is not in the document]
tjmoore4: :+1: updated.

internal/config/config.go
```diff
@@ -0,0 +1,137 @@
...
+	CrunchyPGBackRest = "RELATED_IMAGE_CRUNCHY_PGBACKREST"
+	CrunchyPGExporter = "RELATED_IMAGE_CRUNCHY_PGEXPORTER"
+)
+
+// SetContainerImage returns the container image to use for non-PostgreSQL
+// based images. It takes the current image defined in the postgrescluster spec
+// and the associated related image override environment variable. If the image
+// defined on the spec is empty, the value stored in the environment variable
+// is returned.
+func SetContainerImage(specImageName, imageOverrideEnvVar string) string {
```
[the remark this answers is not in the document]
tjmoore4: I do like that approach, particularly because it's closer to the PG image approach already in place. I'll work on this update.

internal/config/config_test.go
tjmoore4: Yes, that seems to be the case with these tests, I'll remove the envtest portions.

internal/config/config_test.go
tjmoore4: :+1: note added. At this point, I'll have to add a call to `os.Unsetenv` on certain tests to make sure I'm accounting for locally set variables. :thinking: I'm not sure if there might be a better option.

internal/config/config.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// SetPGContainerImage returns the container image to use for PostgreSQL
+// based images. It takes in the postgrescluster CRD and first checks for the
+// image value on the spec. If empty, it attempts to retrievew the relevant
+// environment variable value for the required image. This is determined by
+// gathering the defined Postgres version from the spec and checking for the
+// PostGIS annotation. Depending on these configured items, the relevant value
+// is pulled from the environment variable.
+func SetPGContainerImage(postgresCluster *v1beta1.PostgresCluster) string {
```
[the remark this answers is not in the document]
tjmoore4: :+1: wfm, updated.

internal/config/config.go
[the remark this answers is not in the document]
tjmoore4: Sure, I can make that change. I'll go with the `RELATED_IMAGE_POSTGRES_10` format.

internal/postgres/reconcile.go
tjmoore4: Agreed, updated.

internal/config/config_test.go
tjmoore4: Switched to using a spec value, but included the GIS version explicitly for clarity.

internal/config/config.go
```diff
@@ -0,0 +1,137 @@
...
+	var ok bool
+	if annotations := postgresCluster.Spec.Metadata.GetAnnotationsOrNil(); annotations != nil {
+		version, ok = annotations[naming.PostGISVersion]
+	}
+
+	switch postgresCluster.Spec.PostgresVersion {
+	case 10:
+		if ok {
+			if version == "2.3" {
+				return os.Getenv(CrunchyPostgresHA_PG10_GIS_2_3)
```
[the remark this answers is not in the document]
tjmoore4: Updated. Seems to be working as expected, I'm not seeing any issues with setting the variable name during operator installation.

config/manager/manager.yaml
```diff
@@ -13,6 +13,14 @@ spec:
...
         - name: CRUNCHY_DEBUG
           value: "true"
+        - name: RELATED_IMAGE_POSTGRES_13
+          value: "registry.developers.crunchydata.com/crunchydata/crunchy-postgres-ha:centos8-13.3-0"
+        - name: RELATED_IMAGE_POSTGRES_13_GIS_3.1
+          value: "registry.developers.crunchydata.com/crunchydata/crunchy-postgres-gis-ha:centos8-13.3-3.1-0"
+        - name: RELATED_IMAGE_PGBACKREST
+          value: "registry.developers.crunchydata.com/crunchydata/crunchy-pgbackrest:centos8-2.33-0"
+        - name: RELATED_IMAGE_PGBOUNCER
+          value: "registry.developers.crunchydata.com/crunchydata/crunchy-pgbouncer:centos8-1.15-0"
```
cbandy, line 23: 📝 Recent Kustomize (v3, I think?) (in `kubectl` v1.21) can be configured to replace these using the [images](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/images/) kustomization.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -75,8 +75,17 @@ type PostgresClusterSpec struct {
...
+	// The image name to use for PostgreSQL containers. The image may also be set
+	// using an environment variable. For standard PostgreSQL images, the format
+	// is RELATED_IMAGE_POSTGRES_<Version> where Version is the PostgreSQL
+	// version set in PostgresVersion.
+	// Example: RELATED_IMAGE_POSTGRES_13
+	// For PostGIS enabled PostgreSQL images, the format is
+	// RELATED_IMAGE_POSTGRES_<Version>_GIS_<GIS Version> where "Version" is the
+	// PostgreSQL version set in PostgresVersion and "GIS Version" is the PostGIS
+	// version set in PostGISVersion.
+	// Example: RELATED_IMAGE_POSTGRES_13_GIS_3.1
```
cbandy, line 87: 📝 Line breaks in these comments are removed when added to the CRD, and it looks like `kubectl explain` might ... unwrap (?) even more. `Example: RELATED_IMAGE_POSTGRES_13` bleeds into the following sentence making it a little confusing.

🌱 Seeing it now, I'm not sure I like talking about operator environment variables in the CRD spec. The person deploying PostgresClusters may not know anything about how the operator was deployed. A section about configuring/installing the operator in our [Markdown] documentation seems like the right way to go, but it's not necessary for this PR.

🔧 We should refer to other fields by their JSON/YAML names, like `postgresVersion` rather than `PostgresVersion`. Is it plausible to use those names when describing the env var format?

```suggestion
	// The image name to use for PostgreSQL containers. When omitted, the value
	// comes from an operator environment variable. For standard PostgreSQL images,
	// the format is RELATED_IMAGE_POSTGRES_<postgresVersion>,
	// e.g. RELATED_IMAGE_POSTGRES_13. For PostGIS enabled PostgreSQL images,
	// the format is RELATED_IMAGE_POSTGRES_<postgresVersion>_GIS_<postGISVersion>,
	// e.g. RELATED_IMAGE_POSTGRES_13_GIS_3.1.
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -108,6 +117,12 @@ type PostgresClusterSpec struct {
 	// +kubebuilder:validation:Maximum=13
 	PostgresVersion int `json:"postgresVersion"`
 
+	// The PostGIS extension version installed in the PostgreSQL container.
+	// When set, this value indicates that a PostGIS enabled image will be
+	// used when created the cluster.
```
cbandy, line 122: ⛏️ I think this and `postgresVersion` should say "the PostgreSQL image" rather than "the PostgreSQL container".

❓ The field can be used even after the cluster is created, but it only has meaning when `image` is omitted... Can we get away with leaving that explanation in `image`?

```suggestion
	// The PostGIS extension version installed in the PostgreSQL image.
```

cbandy (changes requested):

docs/content/references/crd.md
```diff
@@ -117,7 +122,7 @@ PostgresClusterSpec defines the desired state of PostgresCluster
       </tr><tr>
         <td><b>openshift</b></td>
         <td>boolean</td>
-        <td>Whether or not the PostgreSQL cluster is being deployed to an OpenShift environment</td>
+        <td>Whether or not the PostgreSQL cluster is being deployed to an OpenShift envioronment</td>
```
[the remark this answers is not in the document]
tjmoore4: :+1: updated.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -75,8 +75,17 @@ type PostgresClusterSpec struct {
...
+	// is RELATED_IMAGE_POSTGRES_<Version> where Version is the PostgreSQL
+	// version set in PostgresVersion.
+	// Example: RELATED_IMAGE_POSTGRES_13
+	// For PostGIS enabled PostgreSQL images, the format is
+	// RELATED_IMAGE_POSTGRES_<Version>_GIS_<GIS Version> where "Version" is the
+	// PostgreSQL version set in PostgresVersion and "GIS Version" is the PostGIS
+	// version set in PostGISVersion.
+	// Example: RELATED_IMAGE_POSTGRES_13_GIS_3.1
+	// +optional
 	Image string `json:"image"`
```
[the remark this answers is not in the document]
tjmoore4: :+1: updated.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -108,6 +117,12 @@ type PostgresClusterSpec struct {
 	// +kubebuilder:validation:Maximum=13
 	PostgresVersion int `json:"postgresVersion"`
 
+	// The PostGIS extension version installed in the PostgreSQL container.
+	// When set, this value indicates that a PostGIS enabled image will be
+	// used when created the cluster.
+	// +optional
+	PostGISVersion string `json:"postGISVersion"`
```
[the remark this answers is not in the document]
tjmoore4: :+1: updated.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
tjmoore4: The suggestion looks good to me, so I've updated to that. 

As for the question regarding the documentation/use of the JSON/YAML names, for the time being I think the suggested update is sufficient. It's clear which variables are used to update the environment variable names, and we have corresponding examples in `config/manager/manager.yaml` that should make the usage fairly straightforward.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
tjmoore4: I'll make the updates from 'container' to 'image'.

For field question, I do agree rewording may be helpful, but perhaps adding a note about it only being relevant when using a related image?

tjmoore4: I was thinking this would be appropriate:
```
	// The PostGIS extension version installed in the PostgreSQL image.
	// When image is not set, indicates a PostGIS enabled image will be used.
	// +optional
```

cbandy (approved):

PR #2565, opened 2021-07-23
Allow v4.x & v5.x Operator to Run in a Single Namespace

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
It is not currently possible to run both the v4.x and v5.x PostgreSQL
Operator in a single namespace due to a naming conflict with the
operators' respectively deployments.

This commit updates the PostgreSQL Operator v5.x installers to name
the v5.x Operator deployment 'pgo' instead of 'postgres-operator' to
allow both versions to run in the same namespace.

Also, to facilitate independent installation and removal of the 4.x
and 5.x Operators, the default service account name has been updated
in 5.x to 'pgo' in place of the current 'postgres-cluster'.

**Other information**:
[ch11969]

For existing v5.0.0 deployments, the follow steps can be taken to update the Operator
deployment and serviceaccount:

1. `git checkout` the updated changes
2. run
```
make deploy
kubectl delete deployment postgres-operator -n <operator-namespace>
kubectl delete sa postgres-operator -n <operator-namespace>
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -2,7 +2,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: postgres-operator
+  name: pgo
 spec:
   replicas: 1
   template:
@@ -17,4 +17,4 @@ spec:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
           runAsNonRoot: true
-      serviceAccountName: postgres-operator
+      serviceAccountName: pgo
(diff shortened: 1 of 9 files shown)


andrewlecuyer (commented):
Looks good.  However, one other thing that does need to be considered is the Helm installer under `/examples`.

However, this is probably also a good opportunity to discuss where we want the Helm installer to live moving forward, considering it is also in the examples repo.

We can therefore discuss further on Monday's standup.

tjmoore4:
Helm install/uninstall updates added.

andrewlecuyer (approved):
A `devel` branch has been created in the examples repo:

https://github.com/CrunchyData/postgres-operator-examples/tree/devel

PR #2566, opened 2021-07-26
Consistently name startup instance when restoring from a different cluster

jmckulk:
In some environments, caching issues when GETing a cluster from the client, lead to errors when applying changes to the restore job. Using a hash of `cluster.Name` and `instanceSet.Name` to generate the startup instance name ensures that the `StartupInstance` status is set consistently and the `prepareForRestore` function is idempotent. This logic is used when using a `dataSource` to restore from another cluster.

**Checklist:**

 
 - [ ] Have you updated or added documentation for the change, as applicable? NA

**Type of Changes:**

 
**Other information**:
[ch12088]

### DIFF
--- a/internal/naming/names.go
+++ b/internal/naming/names.go
@@ -17,6 +17,7 @@ package naming
 
 import (
 	"fmt"
+	"hash/fnv"
 
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


internal/naming/names.go
```diff
@@ -209,6 +209,17 @@ func GenerateInstance(
 	}
 }
 
+// GenerateStartupInstance returns a hashed name that matches the random
+// instance returned by GenerateInstance
```
andrewlecuyer, line 213: :pick: Description is a bit confusing here, e.g. this returns an _instance name_ that matches the naming pattern utilized by GenerateInstance, only with a pre-defined suffix (i.e. a hash) rather than one that is randomly generated.

:thinking: Also was wondering if this could be one function, e.g.:

```go
// GenerateInstance returns a name for a member of cluster and set.  If no
// suffix is provided, one is randomly generated.
func GenerateInstance(
	cluster *v1beta1.PostgresCluster, set *v1beta1.PostgresInstanceSetSpec,
	suffix string) metav1.ObjectMeta {
	if suffix == "" {
		suffix = rand.String(4)
	}
	return metav1.ObjectMeta{
		Namespace: cluster.Namespace,
		Name:      cluster.Name + "-" + set.Name + "-" + suffix,
	}
}
```

andrewlecuyer (approved):

internal/naming/names.go
cbandy: > 🤔 Also was wondering if this could be one function

I don't like how `suffix` (or `hash` in the original) could be the wrong size.

I like the idea that might have been in an earlier iteration, `GenerateStartupInstance`. We're being passed an entire cluster and instance spec, so let's do the hashing inside/here. A useful property of [`hash.Hash`](https://pkg.go.dev/hash#Hash) is that its `Write()` never returns an error.

```go
import "hash/fnv"

func GenerateStartupInstance(
	cluster *v1beta1.PostgresCluster, set *v1beta1.PostgresInstanceSetSpec,
) metav1.ObjectMeta {
	// Calculate a stable name that's shaped like GenerateInstance above.
	// hash.Hash.Write never returns an error: https://pkg.go.dev/hash#Hash.
	hash := fnv.New32()
	_, _ = hash.Write([]byte(cluster.Name + set.Name))
	suffix := rand.SafeEncodeString(fmt.Sprint(hash.Sum32()))[:4]

	return metav1.ObjectMeta{
		Namespace: cluster.Namespace,
		Name:      cluster.Name + "-" + set.Name + "-" + suffix,
	}
}
```

cbandy (approved):

PR #2570, opened 2021-07-27
OpenShift Auto-detection

jmckulk:
Update the operator to automatically detect if it is running in an
OpenShift environment. When the postgres-operator binary starts, it will
check if its being run in an OpenShift environment and pass a value into
the reconciler. When a postgrescluster is reconciled, this value will be
used to set `cluster.Spec.Openshift` if the field is not set by the user
in the spec.

**Checklist:**

 
**Type of Changes:**

 
**Other information**:
[ch11680]

### DIFF
--- a/cmd/postgres-operator/main.go
+++ b/cmd/postgres-operator/main.go
@@ -16,10 +16,12 @@ limitations under the License.
 */
 
 import (
+	"context"
 	"os"
 	"strings"
 
 	"go.opentelemetry.io/otel"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
 	cruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -75,7 +77,7 @@ func main() {
 	assertNoError(err)
 
 	// add all PostgreSQL Operator controllers to the runtime manager
-	err = addControllersToManager(mgr)
+	err = addControllersToManager(ctx, mgr)
 	assertNoError(err)
 
 	log.Info("starting controller runtime manager and will wait for signal to exit")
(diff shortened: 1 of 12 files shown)


cmd/postgres-operator/main.go
```diff
@@ -29,6 +30,8 @@ import (
 	"github.com/crunchydata/postgres-operator/internal/logging"
 )
 
+const openShiftAPIGroupSuffix = ".openshift.io"
```
cbandy, line 33: ⛏️  I'd prefer this inside the one function that uses it.

cmd/postgres-operator/main.go
```diff
@@ -87,10 +90,29 @@ func main() {
...
 	r := &postgrescluster.Reconciler{
-		Client:   mgr.GetClient(),
-		Owner:    postgrescluster.ControllerName,
-		Recorder: mgr.GetEventRecorderFor(postgrescluster.ControllerName),
-		Tracer:   otel.Tracer(postgrescluster.ControllerName),
+		Client:      mgr.GetClient(),
+		Owner:       postgrescluster.ControllerName,
+		Recorder:    mgr.GetEventRecorderFor(postgrescluster.ControllerName),
+		Tracer:      otel.Tracer(postgrescluster.ControllerName),
+		IsOpenshift: isOpenshift(mgr.GetConfig()),
```
cbandy, line 97: ⛏️ OpenShift has a capital `S`, so we should do the same.

```suggestion
		IsOpenShift: isOpenShift(mgr.GetConfig()),
```

cbandy (approved):

andrewlecuyer (approved):

internal/controller/postgrescluster/controller.go
```diff
@@ -144,6 +145,10 @@ func (r *Reconciler) Reconcile(
 		return result, nil
 	}
 
+	if cluster.Spec.OpenShift == nil {
+		cluster.Spec.OpenShift = &r.IsOpenshift
```
cbandy, line 149: 🔧 This should move to ~L109 so that it doesn't get sent in patches that update the PostgresCluster status.

Failing test shows `\"f:spec\":{\"f:openshift\":{}}` in the managed fields.

cbandy (changes requested):

internal/controller/postgrescluster/controller.go
jmckulk: Wow that makes sense, I will move it. I fought with this for a while (just now seeing your comment this morning) before I figured out the following
```
	if cluster.Spec.OpenShift == nil && r.IsOpenshift {
		cluster.Spec.OpenShift = &r.IsOpenshift
	}
```
I'm thinking with this the openshift will only automatically be set if its true.

cbandy: That would avoid the failure in Kubernetes and trigger the failure in OpenShift? I'd prefer that we treat both environments the same.

cmd/postgres-operator/main.go
```diff
@@ -85,12 +87,36 @@ func main() {
...
+	assertNoError(err)
+
+	groups, _, err := client.ServerGroupsAndResources()
+	assertNoError(err)
+
+	// If we detect that any API group name ends with "openshift.io", we'll
+	// return that this is an OpenShift environment
+	for _, g := range groups {
+		if strings.HasSuffix(g.Name, openShiftAPIGroupSuffix) {
+			log.V(1).Info("detected OpenShift environment")
```
cbandy, line 116: 🔧 I'd like to see this even when debug is off.

```suggestion
			log.Info("detected OpenShift environment")
```

cbandy (approved):

PR #2573, opened 2021-07-28
Reconcile PVCs Using Labels

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
Currently, when reconciling the various PVCs for a PostgresCluster,
 a specific naming convention is used. Specifically, each PVC is
reconciled as follows:

PostgreSQL Data: <clusterName>-<instanceName>-pgdata
  e.g., hippo-instance1-abcd-pgdata
pgBackRest Repository: <clusterName>-<repoName>
  e.g., hippo-repo1

While this approach simplifies reconciliation logic, it does mean that
existing PVC's that do not follow this convention cannot be used.

To allow for more versatility, this commit updates all PVC reconciliation
logic to reconcile PVC's based on their labels. This is done by first
identifying the PVC by their labels, then using the names of the PVCs
during reconciliation. This means any existing PVC in the environment
should be detected and reconciled, while a new PVC will be generated if
no existing PVC is found.

**Other information**:
[ch11966]

### DIFF
--- a/internal/controller/postgrescluster/instance.go
+++ b/internal/controller/postgrescluster/instance.go
@@ -518,7 +518,7 @@ func (r *Reconciler) reconcileInstanceSets(
 			rootCA, clusterPodService, instanceServiceAccount,
 			patroniLeaderService, primaryCertificate,
 			findAvailableInstanceNames(set, instances, clusterVolumes),
-			numInstancePods)
+			numInstancePods, clusterVolumes)
 		if err != nil {
 			return err
 		}
@@ -910,6 +910,7 @@ func (r *Reconciler) scaleUpInstances(
 	primaryCertificate *v1.SecretProjection,
 	availableInstanceNames []string,
 	numInstancePods int,
+	clusterVolumes []corev1.PersistentVolumeClaim,
 ) ([]*appsv1.StatefulSet, error) {
 	log := logging.FromContext(ctx)
 
@@ -953,7 +954,7 @@ func (r *Reconciler) scaleUpInstances(
 			clusterConfigMap, clusterReplicationSecret,
 			rootCA, clusterPodService, instanceServiceAccount,
 			patroniLeaderService, primaryCertificate, instances[i],
-			numInstancePods,
+			numInstancePods, clusterVolumes,
 		)
 	}
 	if err == nil {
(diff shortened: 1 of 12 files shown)


andrewlecuyer (changes requested):
Looks good overall, just have some notes around error handling.

Also, I did pull down these changes and tested a few scenarios where I pre-created the PVC's, and all worked as expected :+1:

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -511,8 +511,9 @@ func (r *Reconciler) setScheduledJobStatus(ctx context.Context,
 // generateRepoHostIntent creates and populates StatefulSet with the PostgresCluster's full intent
 // as needed to create and reconcile a pgBackRest dedicated repository host within the kubernetes
 // cluster.
-func (r *Reconciler) generateRepoHostIntent(postgresCluster *v1beta1.PostgresCluster,
-	repoHostName string) (*appsv1.StatefulSet, error) {
+func (r *Reconciler) generateRepoHostIntent(ctx context.Context,
```
cbandy, line 514: ⛏️ This calls the API when it did not before. Is it possible to leave API calls to the caller?

internal/controller/postgrescluster/instance.go
```diff
@@ -1197,7 +1197,7 @@ func addPGBackRestToInstancePodSpec(cluster *v1beta1.PostgresCluster,
 		}
 	}
 	if !dedicatedRepoEnabled {
-		if err := pgbackrest.AddRepoVolumesToPod(cluster, template,
+		if err := pgbackrest.AddRepoVolumesToPod(cluster, template, repoPVCNames,
```
cbandy, line 1200: 📝 If I understand correctly, this (volumes but no dedicated) is the pgBackRest "mode" we intend to remove. If so, we should not worry about optimizing it in this PR.

andrewlecuyer:
@tjmoore4 @cbandy @jmckulk I pushed a few commits to address the feedback discussed today.

Lets touch base first thing in the morning to discuss next steps.

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -1463,12 +1464,16 @@ func (r *Reconciler) reconcilePostgresClusterDataSource(ctx context.Context,
 		Name:      instanceName,
 		Namespace: cluster.GetNamespace(),
 	}}
+	clusterVolumes, err := r.observePersistentVolumeClaims(ctx, cluster)
```
cbandy, line 1467: ❓ I see `observePersistentVolumeClaims` is called before `reconcileDataSource`. Is it possible to pass those PVCs down here rather than make this API call?

andrewlecuyer (approved):

Issue #2542, opened 2021-07-16
 secret <cluster>-postgres-token  missing


tamagottschi:
When we create a new cluster the secret <cluster>-postgres-token for the postgres superuser will not be generated.

We are running RKE (Rancher) with Kubernetes 1.19.9 and using the postgres operator version 5.0.

We created just the cluster "hippo" exactly as described in the tutorial with postgresql 13. We have the same problem with our own postgres cluster, too.

Is the missing token a bug or is there a reason for this?

Thx, Jan

jkatz:
The superuser is not made available by default. This is for a variety of reasons.

However, an upcoming release will allow you to specify additional users/roles (e.g. superuser) you would like to create should you need to connect with a superuser.

flotothemoon:
Just to clarify @jkatz, this means that in the current 5.0 version it's not possible at all to create additional users? While not critical, it would be very convenient to add superusers e.g. in our test environment to create & destroy test databases (is there another way to achieve this?).

jkatz:
You can still create new users in 5.0 by logging in directly to the database and doing so.

However, #2574 should make this process simpler and will be available in the next release.

Issue #1925, opened 2020-09-27
Add CRD support to define multiple users/databases per cluster


GrandChaman:
**What is the motivation or use case for the feature? **
So that one can deploy the Operator without user interaction.

**Describe the solution you'd like**
Either modify the current `pgcluster` CRD to allow to define multiple users and multiples databases

or

A new CRD, that would allow to define databases and users for clusters.

**Describe any alternatives you've considered**
Currently, the alternative would be to use the `pgo` client.

jkatz:
- What is your workflow/use case?
- What does `pgo [create|show|update|delete] user` lack that you are looking for?
- What is the preference for the Operator managing databases vs. using the `postgres` superuser account to `CREATE DATABASE`?
- How would you want sensitive information (e.g. user credentials) to be managed by a CRD?

GrandChaman:
Hi,
 - Workflow : We use `hashicorp/terraform` to deploy our kubernetes infrastructure, to limit user interaction. Having CRD would allow us to reduce user interactions even more.
- Is `pgo` lacking ? Not at all
- We could use `terraform` to directly create databases and roles, even though, I thought it would make a nice addition to the operator.
- Sensitive information could be read from kubernetes secrets and/or `hashicorp/vault`

jkatz:
Follow-up question: is there a reason why you cannot log in directly as the `postgres` superuser (created as part of the standard Operator workflow) and create additional users or databases?

Are you looking for the PostgreSQL users to have Secrets managed by the Operator?

GrandChaman:
Hi,

Indeed, one could use the superuser to create the necessary roles & databases, but I was more looking for a simple solution where the operator would manage roles and their associated secrets.

The reason we would prefer using CRD to define roles & databases would be to further the interoperability of our Infrastructure-as-code codebase.

jkatz:
Follow-up question :smile:

Would being able to interact directly with the API server that the Operator exposes be sufficient? `pgo * user` interfaces with that.

GrandChaman:
It would do of course, it would be less practical IMO though

jkatz:
Thanks for bearing with me while I've asked a lot of questions as I'm trying to better understand this case.

There are several issues here I am trying to balance:

- Duplicating information: PostgreSQL itself keeps a catalog of databases (pg_database) and users (pg_authid), and there is the very real potential that this could get out of sync if we're storing it in two places.
[...]

PR #2574, opened 2021-07-29
Reconcile multiple PostgreSQL users and databases

cbandy:
**Checklist:**

 
**Type of Changes:**

 
We create one PostgreSQL user and database that match the name of the PostgresCluster. Credentials and connection details are stored in the `{cluster}-pguser` Secret, but the password cannot be rotated. This cannot be disabled.

This change allows one to specify any number of PostgreSQL users that can connect to any number of databases. We reconcile the users' role attributes and passwords, allowing one to grant superuser access or manually trigger credential rotations.

By default, we still create a user and database matching the cluster name, but only when the cluster name is also a valid user name. One can disable this entirely by specifying an empty user list.

We store the password of each user and their connection details in a separate Secret (per user). In addition to the `postgres-operator.crunchydata.com/pguser={username}` label from before, each Secret now has a `postgres-operator.crunchydata.com/role=pguser` label and the Secret name includes the username.

The default Secret has moved to conform to this new scheme, but an old copy remains in place so one can upgrade. Where one referenced `{cluster}-pguser` in the past, they should now reference `{cluster}-pguser-{cluster}`. Once all references are updated, the old copy can be deleted manually.

This also lifts an artificial restriction we had that disallowed clusters named `postgres`. One may now include `postgres` in the user list to generate a Secret and password and allow it to connect over the network.

**Other information**:

Issue: [ch11512]
Issue: [ch11806]
Fixes: #1925
Fixes: #2542
Closes: #2579


jmckulk (commented):
Didn't finish reviewing but its 5:50 on Friday. Here is what I have so far. There were some names that I had a hard time following but I'm not sure they are worth changing. Maybe that would be something to chat about?

jmckulk:
To finish the thought on the naming comment. I feel like I read the word user a lot and in a few different contexts. I'm wondering if there is a way to simplify/clarify. There may not be

internal/naming/selectors.go
```diff
@@ -88,17 +88,23 @@ func ClusterPatronis(cluster *v1beta1.PostgresCluster) metav1.LabelSelector {
...
 
+// ClusterPostgresUsers selects things labeled for PostgreSQL users in cluster.
+func ClusterPostgresUsers(cluster string) metav1.LabelSelector {
+	return metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			LabelCluster: cluster,
+		},
+		MatchExpressions: []metav1.LabelSelectorRequirement{
+			// The now-deprecated default PostgreSQL user secret lacks a LabelRole.
+			// The existence of a LabelPostgresUser matches it and current secrets.
```
jmckulk, line 99: > The existence of a LabelPostgresUser matches it and current secrets.

:question: I'm not sure what this comment is trying to say. Should this selector be able to grab a deprecated user secret? iiuc, LabelPostgresUser will only work for new secrets

internal/postgres/exec.go
```diff
@@ -66,3 +66,27 @@ done <<< "${databases}"
 		append([]string{"bash", "-ceu", "--", script, "-"}, args...)...)
 	return stdout.String(), stderr.String(), err
 }
+
+// Execute uses "psql" to execute sql. The sql statement(s) are passed via stdin
+// and may contain psql variables that are assigned from the variables map.
+// - https://www.postgresql.org/docs/current/app-psql.html#APP-PSQL-VARIABLES
+func (exec Executor) Execute(
```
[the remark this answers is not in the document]
cbandy: Matching sounds good. I'll go with `Exec`.

internal/controller/postgrescluster/postgres.go
```diff
@@ -18,18 +18,441 @@ package postgrescluster
...
+func (r *Reconciler) reconcilePostgresDatabases(
+	ctx context.Context, cluster *v1beta1.PostgresCluster, instances *observedInstances,
+) error {
+	const container = naming.ContainerDatabase
+	var podExecutor postgres.Executor
+
+	// Find the PostgreSQL instance that can execute SQL that writes system
+	// catalogs. When there is none, return early.
+
+	for _, instance := range instances.forCluster {
```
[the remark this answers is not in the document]
cbandy: > 🌱 Seems like a check for "can i run sql and where" could be pretty common. Maybe a place for common or util function?

Agreed. The current uses seem to be for writing. 🤔 Reading could involve replicas, but it's safer to interact with the primary.

> 💭 maybe move the check into the exec (assuming thats the only place we need to know the primary/writable instance) and have that return an error or something if it didn't work or couldn't find the right instance.

This method returns `nil` when a good candidate cannot be found. It's not fatal that a cluster lacks a primary, so we shouldn't treat it as an error. The extreme case is a standby cluster; it is never writable.

internal/naming/selectors.go
cbandy: I renamed the constant from `LabelUserSecret` to `LabelPostgresUser`. Both old and new have it; the value is the Postgres user name.

jmckulk: Ah, I see. Sorry didn't catch that

docs/content/architecture/user-management.md
```diff
@@ -0,0 +1,31 @@
+---
+title: "User Management"
+date:
+draft: false
+weight: 125
+---
+
+PGO manages PostgreSQL users that you define in [`PostgresCluster.spec.users`]({{< relref . "/references/crd#postgresclusterspecusers" >}}).
```
cbandy, line 8: 🔧 This link is wrong.

```suggestion
PGO manages PostgreSQL users that you define in [`PostgresCluster.spec.users`]({{< relref "/references/crd#postgresclusterspecusersindex" >}}).
```

docs/content/architecture/user-management.md
```diff
@@ -0,0 +1,31 @@
...
+[`DROP OWNED`](https://www.postgresql.org/docs/current/sql-drop-owned.html) and
+`DROP USER` or [`DROP ROLE`](https://www.postgresql.org/docs/current/sql-droprole.html)
+in PostgreSQL.
+
+Similarly, removing a database from `PostgresCluster.spec.users` does not drop
+that database nor revoke any access to it. To completely remove it, you must run
+[`DROP DATABASE`](https://www.postgresql.org/docs/current/sql-dropdatabase.html)
+as a superuser in PostgreSQL.
+
+The built-in `postgres` superuser can be managed the same way, which one restriction:
```
jmckulk, line 28:
```suggestion
The built-in `postgres` superuser can be managed the same way, with one restriction:
```

possibly?

jmckulk (approved):

PR #2583, opened 2021-08-02
Fix Type Assertion for pg_hba Config

### LINKED ISSUE #2534: Add Optional TLS Usage in 5.0
## Overview

From the version 5.0 documentation:

> All connections in PGO use TLS to encrypt communication between components. 

We would like to have an option to disable TLS since it can degrade performance significantly.
Our workloads are very performance-aware and any degradation can outweigh the benefits of added security.

## Use Case

An option to disable TLS.

## Desired Behavior

Maybe a PostgresCluster (CR) field to enable/disable TLS?

- Platform Version: `5.0`

andrewlecuyer:
In accordance with the default types used when unmarshalling JSON into an interface value (https://pkg.go.dev/encoding/json#Unmarshal), the logic within the `patroni.DynamicConfiguration` function that detects
user-provided `pg_hba` configuration (as provided using the `spec.patroni.dynamicConfiguration` section of the PostgresCluster spec) now checks for an array of interface types.

This ensures user-provided `pg-hba.conf` settings are properly detected, and then added to the pg-hba section of the PostgresCluster's DCS.

closes #2534
[ch12109]

### DIFF
--- a/internal/patroni/config.go
+++ b/internal/patroni/config.go
@@ -280,8 +280,13 @@ func DynamicConfiguration(
 	for i := range pgHBAs.Mandatory {
 		hba[i] = pgHBAs.Mandatory[i].String()
 	}
-	if section, ok := postgresql["pg_hba"].([]string); ok {
-		hba = append(hba, section...)
+	if section, ok := postgresql["pg_hba"].([]interface{}); ok {
+		for i := range section {
+			// any pg_hba values that are not strings will be skipped
+			if value, ok := section[i].(string); ok {
+				hba = append(hba, value)
+			}
+		}
 	}
 	// When the section is missing or empty, include the recommended defaults.
 	if len(hba) == len(pgHBAs.Mandatory) {
--- a/internal/patroni/config_test.go
+++ b/internal/patroni/config_test.go
@@ -419,7 +419,7 @@ func TestDynamicConfiguration(t *testing.T) {
 			name: "postgresql.pg_hba: no default when input",
 			input: map[string]interface{}{
 				"postgresql": map[string]interface{}{
-					"pg_hba": []string{"custom"},
+					"pg_hba": []interface{}{"custom"},
 				},
 			},
 			hbas: postgres.HBAs{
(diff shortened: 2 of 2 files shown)


jkatz (approved):

internal/patroni/config.go
```diff
@@ -280,8 +280,10 @@ func DynamicConfiguration(
 	for i := range pgHBAs.Mandatory {
 		hba[i] = pgHBAs.Mandatory[i].String()
 	}
-	if section, ok := postgresql["pg_hba"].([]string); ok {
-		hba = append(hba, section...)
+	if _, ok := postgresql["pg_hba"].([]interface{}); ok {
+		for _, section := range postgresql["pg_hba"].([]interface{}) {
+			hba = append(hba, section.(string))
```
cbandy, line 285: 🔧 This could panic when the input contains not-string.

I'm not sure what we should do if the user gives `pg_hba: ["str1", 5, "str2"]`. The following ignores the values that are not-string:

```go
	if section, ok := postgresql["pg_hba"].([]interface{}); ok {
		for i := range section {
			if value, ok := section[i].(string); ok {
				hba = append(hba, value)
			}
		}
	}
```

cbandy: 🤔 Alternatively, we could stop caring about the type of the values:

```go
	hba := make([]interface{}, len(pgHBAs.Mandatory))
	for i := range pgHBAs.Mandatory {
		hba[i] = pgHBAs.Mandatory[i].String()
	}
	if section, ok := postgresql["pg_hba"].([]interface{}); ok {
		hba = append(hba, section...)
	}
```

andrewlecuyer: Its starting to feel like we should define a proper structure here for capturing this configuration (e.g. so that we can add proper validation for fields like this), rather than capturing the raw input :thinking:

cbandy (changes requested):
We should include a test that has mixed types in `pg_hba` input.

internal/patroni/config.go
cbandy: Agreed. I want to avoid baking Patroni structures into the CRD, when possible.

Postgres parameters and HBA aren't Patroni-specific features. We should elevate them out of this area as a separate concern.

andrewlecuyer: I updated the logic to simply ignore anything that isn't a string.  I honestly don't know what sort for rule someone would add that would not be a string outside of a mis-configuration (e.g. the format essentially forces a string), and considering a bad `pg_hba.cong` configuration can prevent PG from even starting, this doesn't seem like such a bad thing.  This also aligns with what this would likely look like if defined elsewhere in the spec (array of strings) - which I agree should be elevated outside of this section.

andrewlecuyer:
@cbandy feedback addressed.

cbandy (approved):

PR #2593, opened 2021-08-04
Allow supplemental groups to be specified per cluster

cbandy:
**Checklist:**

 
**Type of Changes:**

 
Supplemental groups are hard-coded to what some systems call `nfsnobody`.

Supplemental groups are an optional field in `PostgresCluster.spec`. When omitted, Pods run without any supplemental groups. The root GID (0) is disallowed.

**Other information**:

Issue: [ch11761]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -242,6 +242,7 @@ generate-crd:
 		paths='./pkg/apis/postgres-operator.crunchydata.com/...' \
 		output:dir='config/crd/bases' # config/crd/bases/{group}_{plural}.yaml
 
+# TODO(cbandy): Run config/crd through Kustomize to pickup any patches there.
 generate-crd-docs:
 	GOBIN='$(CURDIR)/hack/tools' go install fybrik.io/crdoc@v0.4.0
 	$(CURDIR)/hack/tools/crdoc \
--- a/config/crd/validation.yaml
+++ b/config/crd/validation.yaml
@@ -1,0 +1,14 @@
+# PostgresCluster "v1beta1" is in "/spec/versions/0"
+
+# Containers should not run with a root GID.
+# - https://kubernetes.io/docs/concepts/security/pod-security-standards/
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/supplementalGroups/items/minimum
+  value: 1
+
+# Supplementary GIDs must fit within int32.
+# - https://releases.k8s.io/v1.18.0/pkg/apis/core/validation/validation.go#L3659-L3663
+# - https://releases.k8s.io/v1.22.0/pkg/apis/core/validation/validation.go#L3923-L3927
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/supplementalGroups/items/maximum
+  value: 2147483647 # math.MaxInt32
--- a/internal/controller/postgrescluster/apply.go
+++ b/internal/controller/postgrescluster/apply.go
@@ -19,7 +19,9 @@ import (
 	"context"
 	"reflect"
 
-	v1 "k8s.io/api/core/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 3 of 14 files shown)


Makefile
```diff
@@ -243,11 +244,13 @@ generate-crd:
 		output:dir='config/crd/bases' # config/crd/bases/{group}_{plural}.yaml
 
 generate-crd-docs:
+	install -d ./config/crd/generated
+	$(PGO_KUBE_CLIENT) kustomize ./config/crd > ./config/crd/generated/all.yaml
```
cbandy, line 248: ❓ I think I'm going to remove these changes. The template we use now doesn't contain anything about validation, and that's all we get by running `kustomize` on this directory.

I dislike the `generated` directory that contains one file. I'd prefer to create a `generated.yaml` file that we commit and review during PRs (except the CRD file is already 400KB...) When https://github.com/fybrik/crdoc/commit/a3ed20a543e80 is released, we can reconsider what file(s) to commit, whether or not we want validation in the CRD docs, etc.

internal/controller/postgrescluster/apply.go
```diff
@@ -64,3 +76,30 @@ func (r *Reconciler) apply(ctx context.Context, object client.Object) error {
 	}
 	return err
 }
+
+// applyPodSecurityContext is called by apply to work around issues with server-side apply.
```
andrewlecuyer, line 80: :question: Do we have the specific SSA issue captured anywhere?

cbandy: Not that I know of. I looked for a specific issue at k8s that might explain the difference between v1.18 and v1.19, ~but I couldn't find one~. Found it: https://pr.k8s.io/92661

In v1.18.15, sending a `null` in a slice field causes the field/key to be dropped from `managedFields` without changing the value on the object.

I also saw this on an older version of v1.18 when we added then [removed](https://github.com/CrunchyData/postgres-operator/commit/b7aba1e4d99e739087a4feee4fb399f1190b3e6c) the wrapper script from command/args on some containers. As I recall, we deferred at the time because we were adding more than removing and not yet concerned with breaking changes.

jmckulk (approved):

config/crd/validation.yaml
```diff
@@ -0,0 +1,7 @@
+# PostgresCluster "v1beta1" is in "/spec/versions/0"
+
+# Containers should not run with a root GID.
+# - https://kubernetes.io/docs/concepts/security/pod-security-standards/
+- op: add
+  path: /spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/supplementalGroups/items/minimum
+  value: 1
```
tjmoore4, line 7: :question: Is there any value to adding validation for the upper bound (i.e. 65535)?
The example PSP [here](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#example-policies) seems to set a limit.

tjmoore4 (approved):
:+1: Looks good, one question.

config/crd/validation.yaml
cbandy: I read a bit about this&sup1; while trying to find why we had 65534 hard-coded. Apparently, `uid_t` and `gid_t` used to be 16 bits and are now 32 bits. It looks like Kubernetes [chose `int64`](https://github.com/kubernetes/kubernetes/blob/v1.6.0/staging/src/k8s.io/apimachinery/pkg/types/unix_user_id.go) "as a safe bet against wrap-around" and [validates](https://github.com/kubernetes/kubernetes/blob/v1.22.0/pkg/apis/core/validation/validation.go#L3923-L3927) that values are `<= math.MaxInt32`.

🔧 I think we should add a maximum. No more than int32 (to match Kubernetes). 🤔 I see recommendations to stay within 16 bits, but I'm not certain we should enforce it. I'm inclined to let/expect users of this field Know What They Are Doing.

&sup1; I learned a lot from this page: https://systemd.io/UIDS-GIDS/

cbandy: I tried a value that was too large, and I think hit https://issue.k8s.io/97309. The StatefulSet was accepted/valid, but it could not create the Pod:

```
Pod "hippo-00-z9fc-0" is invalid: spec.securityContext.supplementalGroups[0]: Invalid value: 4000000000: must be between 0 and 2147483647, inclusive
```

Makefile
cbandy: Replaced with a TODO mentioning the CRD patches. bc27ad6524d55c4a8f3c993c1871064be9056a96

internal/controller/postgrescluster/apply.go
cbandy: Added a link to the upstream issue. [`45138beb`](https://github.com/CrunchyData/postgres-operator/pull/2593/commits/45138beb5dbe1257257de55c109e5e693fdf752a)

config/crd/validation.yaml
cbandy: Added a maximum with links. [`bc27ad65`](https://github.com/CrunchyData/postgres-operator/pull/2593/commits/bc27ad6524d55c4a8f3c993c1871064be9056a96)

cbandy:
Added one more link to an article at Red Hat.

PR #2596, opened 2021-08-04
Create a service to expose PostgreSQL replicas

cbandy:
**Checklist:**

 
**Type of Changes:**

 
There is now a `{cluster}-replicas` service that points to any/all available PostgreSQL replicas.

**Other information**:

Issue: [ch12144]

### DIFF
--- a/internal/controller/postgrescluster/cluster.go
+++ b/internal/controller/postgrescluster/cluster.go
@@ -20,6 +20,7 @@ import (
 	"io"
 
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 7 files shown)


andrewlecuyer (approved):
Looks good.

And just to clarify, the replica service will now be created regardless if any actual replicas in the cluster?

Otherwise, the only other thing I'm wondering is if we should have a "Connect to Replica" section/example/etc. of sorts in the docs to specifically point out this service, especially if its something users have been looking for :thinking: Any thoughts?

internal/controller/postgrescluster/cluster.go
```diff
@@ -189,6 +190,59 @@ func (r *Reconciler) reconcileClusterPrimaryService(
...
+// +kubebuilder:rbac:groups="",resources="services",verbs={create,patch}
+
+// reconcileClusterReplicaService writes the Service that exposes PostgreSQL
+// replica instances.
+func (r *Reconciler) reconcileClusterReplicaService(
+	ctx context.Context, cluster *v1beta1.PostgresCluster,
+) error {
+	service := generateClusterReplicaServiceIntent(cluster)
+
+	err := errors.WithStack(r.setControllerReference(cluster, service))
```
jmckulk, line 239: We have been setting the controller reference inside of generate intent functions. Is there an argument to do it this way, i.e. generate the intent using one function, then set the controller reference outside of that function :question:

cbandy, line 239: > And just to clarify, the replica service will now be created regardless if any actual replicas in the cluster?

Yes. This is where I was thinking we'd handle it if we want to add and remove it depending on the number of instances specified.

internal/controller/postgrescluster/cluster_test.go
```diff
@@ -701,3 +702,59 @@ func TestContainerSecurityContext(t *testing.T) {
...
+  port: 9876
+  protocol: TCP
+  targetPort: postgres
+selector:
+  postgres-operator.crunchydata.com/cluster: pg2
+  postgres-operator.crunchydata.com/role: replica
+type: ClusterIP
+	`))
+
+	t.Run("Labels", func(t *testing.T) {
```
jmckulk, line 738: :pick: add in a test for cluster annotations. Looks like everything else is covered by the spec asserts

internal/controller/postgrescluster/cluster_test.go
```diff
@@ -701,3 +702,59 @@ func TestContainerSecurityContext(t *testing.T) {
...
+	`))
+	assert.Assert(t, marshalMatches(service.ObjectMeta, `
+creationTimestamp: null
+labels:
+  postgres-operator.crunchydata.com/cluster: pg2
+  postgres-operator.crunchydata.com/role: replica
+name: pg2-replicas
+namespace: ns1
+	`))
+	assert.Assert(t, marshalMatches(service.Spec, `
```
jmckulk, line 726: I like the `marshalMatches` here. Definitely want to make use of this for `generate*Intent` functions :100:

jmckulk (approved):
Couple comments but looks good!

internal/controller/postgrescluster/cluster.go
cbandy: With it outside of the generate function, the generate function returns no errors and does not depend on Reconciler.

andrewlecuyer: Good thought/question, I too would consider the controller reference as part of the intent.

cbandy: The `SetControllerReference` helper provided by `controller-runtime` is geared toward the create-or-update pattern there. It returns errors when there's already a controller attached to the object (the Go struct) or you try to set ownership across namespaces.

We're doing SSA, so the Go struct always lacks a controller. That validation will never fail, and the upsert it performs is also always an append. 😐 

https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.3/pkg/controller/controllerutil/controllerutil.go#L59

🌱 At some point, we'll want (need?) to be using [applyconfigurations](https://pkg.go.dev/k8s.io/client-go/applyconfigurations) to build our SSA intents.

andrewlecuyer: To me it simply feels like a "generate intent" function should generate the full intent, not just part of it - otherwise the function feels confusing.  I know we've also done similar with other/similar functions (i.e. moved setting the controller ref into the "generate intent" function).

If that helper function is the reason we can't or don't want to, maybe we should ditch it?

andrewlecuyer: And yes ... I want applyconfigurations!

Any updates out there as to when we'll see something similar in controller runtime?

cbandy: > Any updates out there as to when we'll see something similar in controller runtime?

Not that I can see.

cbandy: > We have been setting the controller reference inside of generate intent functions.

> To me it simply feels like a "generate intent" function should generate the full intent, not just part of it - otherwise the function feels confusing.

Moved in [`ee784a1c`](https://github.com/CrunchyData/postgres-operator/pull/2596/commits/ee784a1c3f8940891722123ac169df78a777e61d).

PR #2611, opened 2021-08-14
Upgrade Guide

andrewlecuyer:
Adds an upgrade guide to document any additional steps required when upgrading PGO.

Issue: [ch12285]

### DIFF
--- a/docs/content/installation/upgrade.md
+++ b/docs/content/installation/upgrade.md
@@ -1,0 +1,42 @@
+---
+title: "Upgrade"
+date:
+draft: false
+weight: 50
+---
+
+# Overview
+
+Upgrading to a new version of PGO is typically as simple as following the various installation
+guides defined within the PGO documentation:
+
+- [PGO Kustomize Install]({{< relref "./kustomize.md" >}})
+- [PGO Helm Install]({{< relref "./helm.md" >}})
+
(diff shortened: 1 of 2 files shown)


docs/content/installation/upgrade.md
```diff
@@ -0,0 +1,42 @@
...
+
+```bash
+kubectl -n postgres-operator delete deployment,serviceaccount postgres-operator
+```
+
+Then, once both the Deployment and ServiceAccount have been deleted, proceed with upgrading PGO
+by applying the new version of the Kustomize installer:
+
+```bash
+kubectl -n postgres-operator apply -k kustomize/install/bases
```
cbandy, line 41: ❓ I forget, does `-n` work here or does `apply -k` always use what's in the Kustomization file?

Our [install docs](https://github.com/CrunchyData/postgres-operator/blob/74562db/docs/content/installation/kustomize.md#configuration) say to change the contents of the file. I suspect we should omit the `-n` flag here to be consistent.

andrewlecuyer: Good call - while `-n` can be specified with `apply -k`, it make sense to remove it here as we have on the Kustomize install page.  I therefore went ahead and removed `-n` for this command/example.

cbandy (approved):

PR #2613, opened 2021-08-16
Allow PostgreSQL and PgBouncer to be exposed outside of Kubernetes

cbandy:
**Checklist:**

 
**Type of Changes:**

 
Services that route to PostgreSQL and PgBouncer are accessible only from within the Kubernetes cluster (via ClusterIP).

The PostgreSQL primary service and PgBouncer service can be configured (separately) to have a NodePort or LoadBalancer.

**Other Information**:

Changing `Service.Spec.Type` to `ClusterIP` when it was already `NodePort` or `LoadBalancer` is tricky prior to Kubernetes 1.20. The API allocates and fills in a node port for the latter service types, but forgets/fails to remove it for other types. In these cases, the service definition (patch body) must contain `nodePort: null` for the API to accept it.

There's no way to convince the patch generators to send a `null` for that field, so we have to manipulate the patch body ourselves. This uses `github.com/evanphx/json-patch` (a new direct dependency) to do that. Like other SSA workarounds, we're able to detect the broken API _after_ sending a plain/normal patch. When we're talking to a newer/fixed API, we do no extra work; there's no slowdown. 

<br/>

Much of this PR is splitting the "reconcile service" methods into "generate" followed by "reconcile" to make it easier to test changes like the following. This adds tests for the PgBouncer service (we had none), and replaces the basic Ginkgo tests we had for the primary and leader services in `controller_test.go`.

```diff
diff --git internal/controller/postgrescluster/pgbouncer.go internal/controller/postgrescluster/pgbouncer.go
index b1169b7d4..4328c6ef3 100644
--- internal/controller/postgrescluster/pgbouncer.go
+++ internal/controller/postgrescluster/pgbouncer.go
@@ -276,14 +265,18 @@ func (r *Reconciler) reconcilePGBouncerService(
 			naming.LabelRole:    naming.RolePGBouncer,
 		})
 
-	// Allocate an IP address and let Kubernetes manage the Endpoints by selecting
-	// Pods with the PgBouncer role.
+	// Allocate an IP address and/or node port and let Kubernetes manage the
+	// Endpoints by selecting Pods with the PgBouncer role.
 	// - https://docs.k8s.io/concepts/services-networking/service/#defining-a-service
-	service.Spec.Type = corev1.ServiceTypeClusterIP
 	service.Spec.Selector = map[string]string{
 		naming.LabelCluster: cluster.Name,
 		naming.LabelRole:    naming.RolePGBouncer,
 	}
+	if spec := cluster.Spec.Proxy.PGBouncer.Service; spec != nil {
+		service.Spec.Type = corev1.ServiceType(spec.Type)
+	} else {
+		service.Spec.Type = corev1.ServiceTypeClusterIP
+	}
 
 	// The TargetPort must be the name (not the number) of the PgBouncer
 	// ContainerPort. This name allows the port number to differ between Pods,
 ```

<br/>

Issue: [ch12074]
Issue: [ch12074]
Fixes: #2555

### CODE REVIEW

**Review by tjmoore4 — APPROVED** [review]
:+1: Looks good

File: pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go [hunk 1]
```diff
@@ -0,0 +1,24 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package v1beta1
+
+type ServiceSpec struct {
```
jmckulk, line 18: :+1: Pulling shared types into its own file

cbandy: I was on the fence about this one. Two single fields that seem similar might not be an appropriate threshold to share. 🤔 

These types are exported, but we don't have an importable module... 🤷🏻

internal/controller/postgrescluster/patroni.go
```diff
@@ -148,53 +148,64 @@ func (r *Reconciler) reconcilePatroniDynamicConfiguration(
...
+		service.Spec.Type = v1.ServiceType(cluster.Spec.Service.Type)
+	} else {
+		service.Spec.Type = v1.ServiceTypeClusterIP
+	}
 
 	// The TargetPort must be the name (not the number) of the PostgreSQL
 	// ContainerPort. This name allows the port number to differ between
 	// instances, which can happen during a rolling update.
-	leaderService.Spec.Ports = []v1.ServicePort{{
+	service.Spec.Ports = []v1.ServicePort{{
```
jmckulk, line 181: :pick: corev1 ?

internal/controller/postgrescluster/patroni_test.go
```diff
@@ -39,8 +37,192 @@ import (
...
+		})
+
+		// Labels not in the selector.
+		assert.Assert(t, service.Spec.Selector == nil,
+			"got %v", service.Spec.Selector)
+	})
+
+	types := []struct {
+		Type   string
+		Expect func(testing.TB, *corev1.Service)
```
jmckulk, line 128: Why `testing.TB` here :question:

cbandy: I have a habit of using this interface for helpers. Assertion libraries use a much smaller one so they can test themselves and stay compatible with different versions of Go: 1. [gotest.tools](https://pkg.go.dev/gotest.tools/v3@v3.0.3/assert#TestingT), 2. [testify](https://pkg.go.dev/github.com/stretchr/testify@v1.7.0/assert#TestingT), 3. [gomega](https://pkg.go.dev/github.com/onsi/gomega@v1.15.0/types#GomegaTestingT)

jmckulk: hm good to know :+1:

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
jmckulk: I like it. Makes me think of other places like `spec.Metadata` and maybe even `PostgresClusterDataSource` (used in postgrescluster and pgbackrest)

internal/controller/postgrescluster/pgbouncer.go
```diff
@@ -295,10 +288,34 @@ func (r *Reconciler) reconcilePGBouncerService(
...
+}
+
+// +kubebuilder:rbac:groups="",resources="services",verbs={get}
+// +kubebuilder:rbac:groups="",resources="services",verbs={create,delete,patch}
+
+// reconcilePGBouncerService writes the Service that resolves to PgBouncer.
+func (r *Reconciler) reconcilePGBouncerService(
+	ctx context.Context, cluster *v1beta1.PostgresCluster,
+) (*corev1.Service, error) {
+	service, specified, err := r.generatePGBouncerService(cluster)
```
jmckulk, line 303: ❓ What made you want to have the generate function determine if we need the service? I would think reconciles purpose would be to check if a pgbouncer service is needed then use `generatePGBouncerService` get that service. 

Should generate care if pgbouncer service is asked for in the spec?

jmckulk (approved):
LGTM

internal/controller/postgrescluster/pgbouncer.go
cbandy: With this split, generate does all the interpreting of the cluster spec and is the one place that decides the service name. Reconcile makes the API calls.

The `cluster.Spec.Proxy.PGBouncer == nil` guard is the crux of it for me. If we move that to reconcile, then it seems like generate should take a `PGBouncerPodSpec` (i.e. only call this when there's a spec). But generate also/still needs the cluster to set the metadata and ownership. Sending both seems confusing (does the bouncer spec have to match what's in the cluster spec?) or prone to mistakes (one line reads from the bouncer spec, another line reads from the cluster).

PR #2605, opened 2021-08-09
documented how to manually restart PostgreSQL 

ValClarkson:
documented how to manually restart PostgreSQL and updated connection pooling to point to manual restart directions instead of the delete cluster directions

**Checklist:**

 
 - [x ] Have you added an explanation of what your changes do and why you'd like them to be included?
 - [ ] Have you updated or added documentation for the change, as applicable?
 - [ ] Have you tested your changes on all related environments with successful results, as applicable?

**Type of Changes:**

 
 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [x ] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change)

**What is the current behavior? (link to any open issues here)**

**What is the new behavior (if this is a feature change)?**
Added documentation to manually restart PostgreSQL

**Other information**:

issue: [ch10812]

PR #2622, opened 2021-08-19
Document how to rotate custom TLS certificates

cbandy:
**Checklist:**

 
**Other Information**:

- [x] This should merge after #2605.

Issue: [ch11971]

### DIFF
--- a/docs/content/tutorial/administrative-tasks.md
+++ b/docs/content/tutorial/administrative-tasks.md
@@ -11,11 +11,82 @@ There are times when you might need to manually restart PostgreSQL. This can be 
 
 For example, if you have a cluster named `hippo` in the namespace `postgres-operator`, all you need to do is patch the hippo postgrescluster with the following:
 
-```
-kubectl patch postgrescluster/hippo --type=merge --patch='{"spec":{"metadata":{"annotations":{"restart":"'"$(date)"'"}}}}' -n postgres-operator
+```shell
+kubectl patch postgrescluster/hippo -n postgres-operator --type merge \
+  --patch '{"spec":{"metadata":{"annotations":{"restarted":"'"$(date)"'"}}}}'
 ```
 
 Watch your hippo cluster: you will see the rolling update has been triggered and the restart has begun.
+
+
+## Rotating TLS Certificates
+
+Credentials should be invalidated and replaced (rotated) as often as possible
+to minimize the risk of their misuse. Unlike passwords, every TLS certificate
+has an expiration, so replacing them is inevitable. When you use your own TLS
+certificates with PGO, you are responsible for replacing them appropriately.
+Here's how.
+
+
+PostgreSQL needs to be restarted after its server certificates change.
+There are a few ways to do it:
+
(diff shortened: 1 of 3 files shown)


docs/content/tutorial/administrative-tasks.md
```diff
@@ -0,0 +1,69 @@
...
+   ```yaml
+   spec:
+     customTLSSecret:
+       name: hippo.new.tls
+   ```
+
+2. Replace the old certificates in the current Secret. PGO doesn't notice when
+   the contents of your Secret change, so you need to
+   [trigger a rolling restart]({{< relref "/tutorial/administrative-tasks.md" >}}#manually-restarting-postgresql).
+
```
ValClarkson, line 31: with this being enumerated it feels as though both are to be done vs these are both options 🤔 
same thing for the pgbouncer section.  Otherwise LGTM

ValClarkson (approved):

docs/content/tutorial/administrative-tasks.md
cbandy: Agreed. I meant to put an "_or_" between 1. and 2. there. Would that help? 🤔 In the PgBouncer section, that would be (1 or 2 (i or ii)) which is ... a lot of "or". 😬

cbandy: Added some "_or_" and ditched `kubectl rollout` to simplify things. [`39464bf7`](https://github.com/CrunchyData/postgres-operator/pull/2622/commits/39464bf7e06c511e3806ae7f44f40ca6ca4db414)

ValClarkson: this is less complicated for sure 👍

docs/content/tutorial/administrative-tasks.md
```diff
@@ -0,0 +1,69 @@
...
+## Rotating TLS Certificates
+
+Credentials should be invalidated and replaced (rotated) as often as possible
+to minimize the risk of their misuse. Unlike passwords, every TLS certificate
+has an expiration, so replacing them is inevitable. When you use your own TLS
+certificates with PGO, you are responsible for replacing them appropriately.
+Here's how.
+
+
+PostgreSQL needs to be restarted after its server certificates change.
```
tjmoore4, line 16: :question: Is there anything we need to call out regarding the `customReplicationTLSSecret` in this section? I believe the process for replacing them is more or less the same as described for `customTLSSecret`, but it may be worth saying so explicitly.

tjmoore4 (commented):
:+1: Looks good overall, one question.

docs/content/tutorial/administrative-tasks.md
cbandy: Nothing needs to reload or restart when the replication certificate changes. The CA needs to continue to match, so maybe something about that?

tjmoore4: :thinking: Yeah, that I think that would be helpful, just to be safe. And with that, I think this will be good to go. :+1:

tjmoore4 (approved):

Issue #2615, opened 2021-08-18
Restore fails if max_connections is higher than 100


tpatrascu:
## Overview

Restore fails if max_connections on the source cluster is higher than 100, even though we have set in the restore instance config the same max_connections parameter as on the source cluster.

Restore instance settings definition:

```
spec:
  dataSource:
    postgresCluster:
      clusterName: clustername
      repoName: repo1
  patroni:
    leaderLeaseDurationSeconds: 360 
    syncPeriodSeconds: 20
    dynamicConfiguration:
      synchronous_mode: true 
      synchronous_mode_strict: true
      maximum_lag_on_failover: 0
      master_start_timeout: 300
      master_stop_timeout: 300
      postgresql:
        parameters:
          max_connections: 1000
          max_parallel_workers: 0
          max_worker_processes: 4
          shared_buffers: 2GB
          work_mem: 2MB
```

## Environment

Please provide the following details:

- Platform: Kubernetes
- Platform Version: 1.20
- PGO Image Tag: ubi8-5.0.1-0
- Postgres Version 13
- Storage: hostpath PVC

### REPRO

Create a cluster with the postgres parameter max_connections set higher than 100 and try to restore it.

### EXPECTED

The restore job configures the restore postgres instance with the required settings from the source cluster or that we have a way of specifying the postgres restore instance postgres parameters. Or maybe just set a higher default for the restore job.

### ACTUAL

Restore fails.

## Logs

```
2021-08-18 06:29:37.291 P00 DETAIL: sync path '/pgdata/pg13/pg_wal'
34
2021-08-18 06:29:37.291 P00 DETAIL: sync path '/pgdata/pg13/pg_wal/archive_status'
33
2021-08-18 06:29:37.291 P00 DETAIL: sync path '/pgdata/pg13/pg_xact'
32
2021-08-18 06:29:37.291 P00   INFO: restore global/pg_control (performed last to ensure aborted restores cannot be started)
31
2021-08-18 06:29:37.291 P00 DETAIL: sync path '/pgdata/pg13/global'
30
2021-08-18 06:29:37.393 P00   INFO: restore command end: completed successfully (1243ms)
29
waiting for server to start....2021-08-18 06:29:37.422 GMT [15] LOG:  starting PostgreSQL 13.3 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 8.4.1 20200928 (Red Hat 8.4.1-1), 64-bit
28
[...]

PR #2657, opened 2021-08-31
Fix restore with custom PostgreSQL parameters

cbandy:
**Checklist:**

 
**Type of Changes:**

 
When `max_connections` or `max_worker_processes` is configured higher than the default during backup, the restore fails because PostgreSQL refuses to start with the default values (which are now too small).

During restore, those settings (and a few others) are now configured to match exactly what PostgreSQL is expecting from the backup and WAL archive.

**Other Information**:

Issue: [ch12371]
Fixes: #2615

### DIFF
--- a/internal/pgbackrest/config.go
+++ b/internal/pgbackrest/config.go
@@ -191,16 +191,49 @@ func JobConfigVolumeAndMount(pgBackRestConfigMap *v1.ConfigMap, pod *v1.PodSpec,
 //   Patroni config when bootstrapping a cluster using an existing data directory.
 func RestoreCommand(pgdata string, args ...string) []string {
 
+	// After pgBackRest restores files, PostgreSQL starts in recovery to finish
+	// replaying WAL files. "hot_standby" is "on" (by default) so we can detect
+	// when recovery has finished. In that mode, some parameters cannot be
+	// smaller than they were when PostgreSQL was backed up. Configure them to
+	// match the values reported by "pg_controldata". Those parameters are also
+	// written to WAL files and may change during recovery. When they increase,
+	// PostgreSQL exits and we reconfigure and restart it.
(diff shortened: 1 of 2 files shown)


internal/pgbackrest/config.go
```diff
@@ -191,16 +191,51 @@ func JobConfigVolumeAndMount(pgBackRestConfigMap *v1.ConfigMap, pod *v1.PodSpec,
...
+max_prepared_transactions = '${max_ptxn}'
+max_worker_processes = '${max_work}'
+unix_socket_directories = '/tmp'
+EOF
+if [ "$(< "${pgdata}/PG_VERSION")" -ge 12 ]; then
+read -r max_wals <<< "${control##*max_wal_senders setting:}"
+echo >> /tmp/postgres.restore.conf "max_wal_senders = '${max_wals}'"
+fi
+
+pg_ctl start --silent --wait --options='--config-file=/tmp/postgres.restore.conf'
```
andrewlecuyer, line 233: :question: Adding the `--wait` here makes sense, any specific reason for the `--silent`?

I tend to like informational information in these scenarios, simply to have a clear idea of what was going on at various points in the script.

andrewlecuyer (approved):
Looks great, hoping we don't see the complexity grow to much further around starting the DB for these Jobs :crossed_fingers:

internal/pgbackrest/config.go
cbandy: It and PostgreSQL both log to stdout/stderr here, so the "starting postgres" message from `pg_ctl` was overwriting messages from PostgreSQL. It emits errors when PostgreSQL can't start, so it's not _totally_ "silent".

`pg_ctl stop` is similar. PostgreSQL logs that it got the signal to shut down and the progress of shutdown, so the "stopping postgres" message was just noise.

Issue #2531, opened 2021-07-12
Turning backups off fills up disk with WAL logs until crash


james-callahan:
## Overview

If I do not specify a backup repo (e.g. because I don't want backups), then the WAL will grow until storage is exhausted.

Eventually the database crashes with "No space left on device" and doesn't restart, logging "FATAL:  could not write lock file "postmaster.pid": No space left on device"

## Environment

Please provide the following details:

- Platform: EKS
- Platform Version: 5.0.0
- PGO Image Tag: ubi8-5.0.0-0
- Postgres Version 13
- Storage: ??

### REPRO

Provide steps to get to the error condition:

1. create cluster:
```
apiVersion: postgres-operator.crunchydata.com/v1beta1
kind: PostgresCluster
metadata:
  name: hippo
spec:
  image: registry.developers.crunchydata.com/crunchydata/crunchy-postgres-ha:centos8-13.3-4.7.0
  postgresVersion: 13
  monitoring:
    pgmonitor:
      exporter:
        image: registry.developers.crunchydata.com/crunchydata/crunchy-postgres-exporter:ubi8-5.0.0-0
  instances:
    - name: db
      replicas: 1
      dataVolumeClaimSpec:
        accessModes: [ReadWriteOnce]
        resources:
          requests:
            storage: 1Gi
  backups:
    pgbackrest:
      image: registry.developers.crunchydata.com/crunchydata/crunchy-pgbackrest:centos8-2.33-0
```

2. perform 1GB or so of writes 
3. observe DB is down due to out-of-space

### EXPECTED

WAL would not be retained if there is no backup repo.

### ACTUAL

WAL logs fill up disk

## Additional Information

I think this could be fixed via by setting `archive_command` to `/bin/true` when there are no backup repos?

Alternatively, you could just make it possible for me to override `archive_mode` or `archive_command` myself. Currently they are forced to be on:
https://github.com/CrunchyData/postgres-operator/blob/21905f0c2962e22f72aec34a9cad733e31afcdf0/internal/pgbackrest/postgres.go#L39-L40

james-callahan:
Other flag that might be useful to fix: "Maximum Archive Push Queue Size Option" (`--archive-push-queue-max`)

As docs say:
>  The purpose of this feature is to prevent the log volume from filling up at which point Postgres will stop completely. Better to lose the backup than have PostgreSQL go down.

james-callahan:
@jkatz is this something you might be able to look into?

[...]

PR #2662, opened 2021-09-01
Require At Least One Repo to be Configured

andrewlecuyer:
At least one pgBackRest repository must be defined within a PostgresCluster spec for the proper operation of any/all PostgreSQL clusters created by PGO.

This commit therefore updates the validation for the `repos` section of the spec to require that at least one repo always
be defined when creating or modifying a PostgresCluster.

[ch12418]

### DIFF
--- a/config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
+++ b/config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
@@ -1422,6 +1422,7 @@ spec:
                           required:
                           - name
                           type: object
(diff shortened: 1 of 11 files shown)


tjmoore4 (approved):

cbandy (approved):

james-callahan:
I'm not sure this was the correct fix: why should I *have* to have a repo configured if I don't want backups?
As mentioned in https://github.com/CrunchyData/postgres-operator/issues/2531:
> `archive_command` to `/bin/true` when there are no backup repos

PR #2667, opened 2021-09-02
Custom Database Initialization SQL

### LINKED ISSUE #2652: How to automatically enable extensions
I have recently deployed a Postgres cluster including timescale extension using PGO. However, I had to manually `CREATE EXTENSION IF NOT EXISTS timescaledb`. Is there a recommended way how to automate such a task? I initially thought of an initContainer, but this seems a little bulky and I expect such a routine requirement to be handled by the operator.

My current config:
```
apiVersion: postgres-operator.crunchydata.com/v1beta1
kind: PostgresCluster
metadata:
  name: test
spec:
  image: registry.developers.crunchydata.com/crunchydata/crunchy-postgres-ha:centos8-13.4-4.7.2
  postgresVersion: 13
  instances:
    - name: instance1
      replicas: 3
      resources:
        limits:
          cpu: "1"
          memory: 1Gi
        requests:
          cpu: 250m
          memory: 512Mi
      dataVolumeClaimSpec:
        accessModes:
        - "ReadWriteOnce"
        resources:
          requests:
            storage: 8Gi
      affinity:
        podAntiAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
          - weight: 1
            podAffinityTerm:
              topologyKey: kubernetes.io/hostname
              labelSelector:
                matchLabels:
                  postgres-operator.crunchydata.com/cluster: test
                  postgres-operator.crunchydata.com/instance-set: instance1
  backups:
    pgbackrest:
      image: registry.developers.crunchydata.com/crunchydata/crunchy-pgbackrest:centos8-2.33-1
      repoHost:
        dedicated: {}
      repos:
      - name: repo1
        volume:
          volumeClaimSpec:
            accessModes:
            - "ReadWriteOnce"
            resources:
              requests:
                storage: 20Gi
  monitoring:
    pgmonitor:
      exporter:
       image: registry.developers.crunchydata.com/crunchydata/crunchy-postgres-exporter:ubi8-5.0.2-0
  patroni:
    dynamicConfiguration:
      postgresql:
        parameters:
          shared_preload_libraries: timescaledb
```

jmckulk:
As part of the reconcile loop, the DatabaseInitSQL field is checked.
When found, the name and fieldPath from this field will be used to get
data from the named Kubernetes ConfigMap. This data will be treated as a
SQL file and executed in the database. The SQL statement(s) are passed
in via stdin and may contain app-psql variables defined in the link
below.

If the psql call returns with exit code 0, the DatabaseInitSQL status
will be set on the cluster. If the call returns with a failure exit
code, the reconcile loop will continue to get the SQL from the ConfigMap
and execute until either the call succeeds or DatabaseInitSQL is removed
from the spec. Unless the app-psql variable `ON_ERROR_STOP` is defined
(using `\set ON_ERROR_STOP` in the SQL file), the psql call will succeed
whenever the last SQL statement succeeds. Failed statements made before
the final statement will be ignored when setting the status.

If the DatabaseInitSQL status is lost or removed and DatabaseInitSQL
still exists in the spec, the SQL will be executed in the database
again. Considering this, SQL statements should be idempotent. When
DatabaseInitSQL is removed from the spec, the DatabaseInitSQL status
will also be removed.

https://www.postgresql.org/docs/current/app-psql.html#APP-PSQL-VARIABLES

**Checklist:**

 
**Type of Changes:**

 
**Other information**:
[ch10815]
closes #2652


internal/controller/postgrescluster/postgres.go
```diff
@@ -602,3 +591,104 @@ func (r *Reconciler) reconcilePostgresWALVolume(
...
+
+	// If write succeeds, keep track of the configmap using a status. This
+	// helps to ensure SQL doesn't get run again. SQL can be run again if the
+	// status is lost and the custom init script trigger exists on the cluster.
+	if err == nil {
+		status := cluster.Spec.CustomDatabaseInitScript.ResourceRef.Name
+		cluster.Status.CustomDatabaseInit = &status
+	}
+
+	return nil
```
jmckulk, line 693:
```suggestion
	return err
```

internal/controller/postgrescluster/instance.go
```diff
@@ -266,6 +266,27 @@ func newObservedInstances(
 	return &observed
 }
 
+func (observed *observedInstances) writablePod(container string) (*corev1.Pod, *Instance) {
```
tjmoore4, line 269: :wrench: This function should have a comment.

internal/controller/postgrescluster/postgres.go
```diff
@@ -602,3 +591,103 @@ func (r *Reconciler) reconcilePostgresWALVolume(
...
+
+		return nil, cm.Data[path]
+	}
+
+	if err, data = getDataFromFieldRef(); err != nil {
+		return err
+	} else if data == "" {
+		return nil
+	}
+	_, _ = sql.WriteString(data)
```
tjmoore4, line 650: :question: Are we sure we don't want to handle any returned errors here?

internal/controller/postgrescluster/postgres.go
```diff
@@ -602,3 +591,103 @@ func (r *Reconciler) reconcilePostgresWALVolume(
...
+	var podExecutor postgres.Executor
+	if pod, _ := instances.writablePod(naming.ContainerDatabase); pod != nil {
+		podExecutor = func(
+			_ context.Context, stdin io.Reader, stdout, stderr io.Writer, command ...string,
+		) error {
+			return r.PodExec(pod.Namespace, pod.Name, naming.ContainerDatabase, stdin, stdout, stderr, command...)
+		}
+	}
+	if podExecutor == nil {
+		log.Info("podExecutor is nil. Is there a pod with a writable database container?")
```
tjmoore4, line 663: :thinking: I'm wondering when "podExecutor" might be nil. Didn't we just check that the pod returned was writable?

internal/controller/postgrescluster/postgres_test.go
```diff
@@ -365,3 +368,254 @@ volumeMode: Filesystem
...
+	// Define a fully configured cluster that would lead to SQL being run in
+	// the database. This test cluster will be modified as needed for testing
+	testCluster := testCluster()
+	testCluster.Namespace = ns.Name
+	testCluster.Spec.DatabaseInitSQL = &v1beta1.DatabaseInitSQL{
+		Name:      cm.Name,
+		FieldPath: path,
+	}
+
+	// Start Tests
```
tjmoore4, line 457: :+1: Nice test coverage.

tjmoore4 (commented):
Overall looks good. A couple of questions and one update requested.

internal/controller/postgrescluster/postgres_test.go
jmckulk: tyty :rocket:

internal/controller/postgrescluster/instance.go
jmckulk: agreed, i'll add one in. Not sure about the name tbh, `writablePod` but it spits out a pod and instance. Couldn't think of a better one if anyone has ideas (`writablePodAndInstance` is one but not particularly good)

internal/controller/postgrescluster/postgres.go
jmckulk: yup, `func (b *Buffer) WriteString(s string) (n int, err error)`, we don't care about n and err is guaranteed to be nil.

https://pkg.go.dev/bytes?utm_source=gopls#Buffer.WriteString

internal/controller/postgrescluster/postgres.go
jmckulk: hmm, yeah we are defining `podExecutor` so it should never be nil - assuming pod isn't nil and we check that. Think the same applies earlier in the file too

internal/controller/postgrescluster/instance_test.go
```diff
@@ -261,6 +262,176 @@ func TestNewObservedInstances(t *testing.T) {
 	})
 }
 
+func TestWritablePod(t *testing.T) {
+	env, cc, config := setupTestEnv(t, ControllerName)
```
cbandy, line 266: 🔧 This test doesn't use the API, reconciler, nor manager. Setting up envtest just slows things down.

internal/controller/postgrescluster/instance.go
tjmoore4: :+1: 
As for the name, we have several functions in the form of `Is`_something_ (`IsRunning`, `IsPrimary`, etc). Perhaps `isWritable`? `IsWritablePod`?

jmckulk: the `Is*` checks return bool values so that feels off. What about something like `GetWritablePod(container)`?

internal/controller/postgrescluster/postgres.go
```diff
@@ -602,3 +591,103 @@ func (r *Reconciler) reconcilePostgresWALVolume(
...
+				Namespace: cluster.Namespace,
+			},
+		}
+		err := r.Client.Get(ctx, client.ObjectKeyFromObject(cm), cm)
+		if err != nil {
+			return err, ""
+		}
+
+		path := cluster.Spec.DatabaseInitSQL.FieldPath
+		if _, ok := cm.Data[path]; !ok {
```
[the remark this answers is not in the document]
jmckulk: We talked on a call and decided that moving away from the `FieldPath` variable to `Key` is the ideal solution for now. The only  resources we would use to accept sql files are configmaps or secrets and `key` will work for both.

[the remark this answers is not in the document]
andrewlecuyer: :+1: 

> We talked on a call and decided that moving away from the FieldPath variable to Key is the ideal solution for now.

Would like to ensure we document and fully understand (as a team) the reasoning behind this decision, especially if we're going against certain conventions.

internal/controller/postgrescluster/instance.go
tjmoore4: Yeah, I could see that. Not a blocker either way.

tjmoore4 (approved):

jkatz:
@jmckulk Are there still a few commits that need to be squashed? 🤔

jmckulk:
> @jmckulk Are there still a few commits that need to be squashed? thinking

@jkatz05 yes. I was planning on doing squashing as part of the merge after everything gets approved

cbandy (commented):
Some questions and nitpicks about the docs. Looking good.

docs/content/tutorial/customize-cluster.md
```diff
@@ -185,6 +185,105 @@ spec:
...
+
+After you add the ConfigMap reference to your spec, apply the change with `kubectl apply -k kustomize/postgres` and PGO will create your `hippo` cluster. Once the cluster has started and PGO can find a suitable database container, your initialization SQL will be executed. You can verify that your SQL has been run by checking the `databaseInitSQL` status on your Postgres cluster. While the is set, your init SQL will not be run again. You can check cluster status with the `kubectl describe` command:
+
+```
+kubectl -n postgres-operator describe postgresclusters.postgres-operator.crunchydata.com hippo
+```
+
+{{% notice warning %}}
+
+Status is generated based on the current state of the cluster as defined by the spec. Kubernetes does not guarantee that status will not be lost. If the cluster status is lost, PGO will execute any init SQL defined in the spec as part of the reconciliation process. This means your SQL could be run in the database more than once. Please ensure that the statements defined in your init SQL are idempotent.
```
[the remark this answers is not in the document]
jmckulk: I did get that phrasing from somewhere. I'll see if i can find a link

docs/content/tutorial/customize-cluster.md
```diff
@@ -185,6 +185,105 @@ spec:
...
+
+{{% notice tip %}}
+If you edit your ConfigMap and your changes aren't showing up, you may be waiting for PGO to reconcile your cluster. After some time, PGO will automatically reconcile the cluster or you can trigger reconciliation by applying any change to your cluster (e.g. with `kubectl apply -k kustomize/postgres`).
+{{% /notice %}}
+
+To ensure that `psql` returns a failure exit code when any of your statements fail, set the `ON_ERROR_STOP` variable, found [here](https://www.postgresql.org/docs/current/app-psql.html), as part of your SQL file:
+
+```
+\set ON_ERROR_STOP
+\echo `Any error will lead to exit code 3`
```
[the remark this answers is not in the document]
jmckulk: nope that i not intentional

internal/controller/postgrescluster/postgres.go
```diff
@@ -602,3 +589,101 @@ func (r *Reconciler) reconcilePostgresWALVolume(
...
+		}
+
+		return nil, cm.Data[path]
+	}
+
+	if err, data = getDataFromConfigMap(); err != nil {
+		return err
+	} else if data == "" {
+		return nil
+	}
```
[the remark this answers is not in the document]
jmckulk: I don't think so but I would have to test to be sure. If the SQL file is empty i would think that `data == ""` would lead to a return nil. 

Are you suggesting turning the log message into an error?
> log.V(1).Info("ConfigMap did not contain expected data at fieldPath", "ConfigMap", cm.Name, "Key", path)

docs/content/tutorial/customize-cluster.md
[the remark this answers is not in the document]
jmckulk: Tested that out and am seeing the same behavior. I can remove the backticks...or i can remove the `echo` all together. I added it to provide a second example for meta-commands but it could be extra

internal/controller/postgrescluster/postgres.go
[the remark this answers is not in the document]
jmckulk: I will add in a new log message for this if case. Should I create an error `if data == ""`?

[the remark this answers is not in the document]
cbandy: I think it should be valid to specify an empty SQL file, and I suspect we should run it (i.e. not consider its contents). Why does this return early when it is empty?

> Are you suggesting turning the log message into an error?

That's the one. Is this `if data == ""` about getting a zero value from a map or is it about the contents of the existing/present SQL file?

docs/content/tutorial/customize-cluster.md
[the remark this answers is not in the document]
cbandy: Did you find a link?

internal/controller/postgrescluster/postgres.go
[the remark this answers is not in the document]
jmckulk: `if data == ""` is making sure the SQL file actually contains something. Huh, I'm tracking now. Your comment had two different unrelated pieces
> Is the status set when the SQL file is empty?

Currently no the status is not set when the SQL file is empty. `if data == ""` i.e. the file is empty we return early without setting the status. If they specify an empty file then I can see setting the status. They said to run an empty file and that task has been completed so set the status.

It does seem weird to run a pod exec on something that we know is empty and wont do anything. Whats the point? That does mean we are considering the contents :thinking: I'm really on the fence

> Perhaps we should error when the ConfigMap lacks the key?

I can replace the log message with a new error :+1:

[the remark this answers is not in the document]
andrewlecuyer: I also like the idea of not checking the contents at all, and just running whatever is provided.  This would essentially mean running `psql` with an empty SQL file if `data` happens to be empty (though we would not specifically check or care).  So essentially the equivalent of doing:

```
bash-4.4$ echo "" | psql --file=-
bash-4.4$ echo $?
0
```

Which means when running with an empty sql file the `exec` call would exit cleanly and we would then set `status`, etc. just like any other valid SQL file that ran successfully.

This appears to keep the behavior of the SQL file provided consistent with that of `psql`, and does not require us to do anything but run whatever is provided (i.e. no need to check contents).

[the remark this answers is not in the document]
jmckulk: alright cool, i'll pull that check out

internal/controller/postgrescluster/postgres.go
```diff
@@ -615,3 +602,102 @@ func (r *Reconciler) reconcilePostgresWALVolume(
...
+		}
+
+		return nil, cm.Data[key]
+	}
+
+	if err, data = getDataFromConfigMap(); err != nil {
+		log.Error(err, "Could not get data from ConfigMap",
+			"ConfigMap", cluster.Spec.DatabaseInitSQL.Name,
+			"Key", cluster.Spec.DatabaseInitSQL.Key)
+		return err
```
cbandy, line 659: 🌱 We should log an error or return it, not both. In the future, we want errors that are returned to become warning events on the cluster.

I picked up that mantra from thoughts like these:
- https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully
- https://dave.cheney.net/2015/11/05/lets-talk-about-logging

PR #2687, opened 2021-09-11
PVC validation

cbandy:
**Checklist:**

 
**Type of Changes:**

 
Our CRD does not require PVC fields that are required in the Kubernetes API. For example, `dataVolumeClaimSpec: {}` was valid in our spec but failed to create a PVC.

These are now invalid:

```yaml
dataVolumeClaimSpec: {}
dataVolumeClaimSpec: { accessModes: [] }
dataVolumeClaimSpec: { resources: {} }
```

**Other information**:

Patches to the CRD are now applied during `make generate-crd`.

Issue: [ch11146]


andrewlecuyer:
I definitely see your point about there being no diff to approve in the PR for the added CRD, which does make wonder if the build process should simply produce the CRD's with all added validation included (rather than patching validation in via kustomize during the actual install).  This way we have one canonical version of our CRD with all validation, defaults, etc. already included.

One thing that comes to mind here are the various other places we have to use these CRD's, e.g. for our various other installers (Helm, OLM, etc.).  Specifically, it seems like having the CRD's pre-populated with all expected validation, etc. would facilitate copying those definitions to these various other installers where they also need to live.

cbandy:
> I definitely see your point about there being no diff to approve in the PR for the added CRD, which does make wonder if the build process should simply produce the CRD's with all added validation included (rather than patching validation in via kustomize during the actual install). This way we have one canonical version of our CRD with all validation, defaults, etc. already included.

Agreed. I'll see if there's an easy way to move this patch into the `generate` target before merging.

andrewlecuyer (approved):

cbandy:
Rebased to resolve conflicts in `docs/content/references/crd.md`.

PR #2697, opened 2021-09-15
Init and sidecar resources

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
This commit allows a user to configure proper resources for all init and sidecar
containers in a PostgreSQL cluster. This ensures all pods can be assigned a
particular [quality of service (QoS) class](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/)

All resource configuration sections follow the same pattern as the standard
Kubernetes structure for setting [container resources](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/).

**Other information**:
[sc-11688]


pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbackrest_types.go
```diff
@@ -139,6 +139,17 @@ type PGBackRestArchive struct {
 	// Defines details for performing an in-place restore using pgBackRest
 	// +optional
 	Restore *PGBackRestRestore `json:"restore,omitempty"`
+
+	// Defines pgBackRest sidecars
```
andrewlecuyer, line 143: :pick: Can we change the wording here to "Configuration for pgBackRest sidecars" (or similart), vs "Defines pgBackRest sidecars"?

The latter sounds like we're allowing users to define what sidecars are actually deployed (and we do actually have a separate request to be able to add custom sidecars), rather than just allowing for the configuration of certain details within the sidecars we deploy.

(similar thought would apply to PgBouncer and instances)

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -361,6 +365,17 @@ type PostgresInstanceSetSpec struct {
 	WALVolumeClaimSpec *corev1.PersistentVolumeClaimSpec `json:"walVolumeClaimSpec,omitempty"`
 }
 
+// InstanceSidecars defines PostgreSQL instance sidecar containers
+type InstanceSidecars struct {
+	// Defines the configuration for the replica cert copy sidecar container
+	// +optional
+	ReplicaCertCopy *Sidecar `json:"replicacertcopy,omitempty"`
```
andrewlecuyer, line 372: :wrench: Thinking we should camelCase here (e.g. `replicaCertCopy`) for consistency with other areas of the spec.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbouncer_types.go
```diff
@@ -123,6 +127,13 @@ type PGBouncerPodSpec struct {
 	TopologySpreadConstraints []corev1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
 }
 
+// PGBouncerSidecars defines pgBouncer sidecar containers
+type PGBouncerSidecars struct {
+	// Defines the configuration for the pgBouncer config sidecar container
+	// +optional
+	PGBouncerConfig *Sidecar `json:"pgbouncerconfig,omitempty"`
```
andrewlecuyer, line 134: :wrench: Thinking we should camelCase here (e.g. `pgbouncerConfig`) for consistency with other areas of the spec.

andrewlecuyer (commented):
Looks great.  Just a few thoughts/notes in the spec.

docs/content/tutorial/resize-cluster.md
```diff
@@ -15,9 +15,19 @@ Let's dive in.
...
 
-For example, let's say we want to update our `hippo` Postgres cluster so that each instance has a limit of `2.0` CPUs and `4Gi` of memory. We can make the following changes to the manifest:
+- `spec.instances.resources` section, which sets the resource values for the PostgreSQL container, as well as any init containers in the associated pod.
+- `spec.instances.sidecars.replicacertcopy.resources` section, which sets the resources for the `replica-cert-copy` sidecar container.
+- `spec.instances.sidecars.exporter.resources` section, which sets the resources for the `exporter` sidecar container.
+- `spec.backups.pgbackrest.repoHost.resources` section, which sets the resources for the pgBackRest repo host container, as well as any init containers in the associated pod.
+- `spec.backups.pgbackrest.sidecars.pgbackrest.resources` section, which sets the resources for the `pgbackrest` sidecar container.
+- `spec.proxy.pgBouncer.resources` section, which sets the resources for the `pgbouncer` container.
+- `spec.proxy.pgBouncer.sidecars.pgbouncerconfig.resources` section, which sets the resources for the `pgbouncer-config` sidecar container.
```
cbandy, line 26: 🤔 I go back and forth on what field names make the most sense to me here. For this one, I think I'd want to eliminate the stutter and just call it `config`: `spec.proxy.pgBouncer.sidecars.config.resources`.

The stutter in the pgBackRest sidecar name makes me think we should call it ~~`postgres` or~~ `instance` because that is where the sidecar runs.

`replicacertcopy` seems very specific to our current implementation, but it's accurate. I think we can manage deprecating it or renaming it in the controller in the future.

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -326,6 +328,13 @@ func addPGMonitorExporterToInstancePodSpec(
 			MountPath: "/conf",
 		}},
 	}
+
+	// set sidecar container resources
+	if sidecars != nil && sidecars.Exporter != nil &&
+		sidecars.Exporter.Resources != nil {
+		exporterContainer.Resources = *sidecars.Exporter.Resources
```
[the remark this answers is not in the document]
jmckulk: The exporter spec already has a field for resources. I like the idea of having all sidecar resources (possibly other fields we want to set on a container) defined in one place. In that case, I'm thinking this section in the exporter spec would need to be considered for removal/upgrade.
https://github.com/CrunchyData/postgres-operator/blob/c654a820bb6d9f4014a393481a89b2aedcca19b0/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go#L534

docs/content/tutorial/resize-cluster.md
```diff
@@ -15,9 +15,19 @@ Let's dive in.
 
 Memory and CPU resources are an important component for vertically scaling your Postgres cluster. Couple with [tweaks to your Postgres configuration file]({{< relref "./customize-cluster.md" >}}), allowing your cluster to have more memory and CPU allotted to it can help it to perform better under load.
 
-It's important for instances in the same high availability set to have the same resources. PGO  lets you adjust CPU and memory within the `spec.instances.resources` section of the `postgresclusters.postgres-operator.crunchydata.com` custom resource. The layout of `spec.instances.resources` should be familiar: it follows the same pattern as the standard Kubernetes structure for setting [container resources](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/).
+It's important for instances in the same high availability set to have the same resources. PGO lets you adjust CPU and memory within the `resources` sections of the `postgresclusters.postgres-operator.crunchydata.com` custom resource. These include:
```
jmckulk, line 18: Weren't we going to put these new docs in the `Customize-cluster.md` docs :question:

cbandy:
Please include the `Issue: ` [trailer](https://git-scm.com/docs/git-interpret-trailers) in commit messages.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbackrest_types.go
tjmoore4: :+1: That works for me, updated in all three places.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
tjmoore4: :+1: updated.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbouncer_types.go
tjmoore4: :+1: updated.

internal/postgres/reconcile.go
```diff
@@ -44,13 +44,21 @@ func InitCopyReplicationTLS(postgresCluster *v1beta1.PostgresCluster,
...
-		})
+
+	container := v1.Container{
+		Command:         []string{"bash", "-c", cmd},
+		Image:           config.PostgresContainerImage(postgresCluster),
+		Name:            naming.ContainerClientCertInit,
+		SecurityContext: initialize.RestrictedSecurityContext(),
+	}
+
+	// set the NSS wrapper container resources to the 'database' container settings
```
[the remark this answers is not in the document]
tjmoore4: I've reworded the comment since it is confusing. The replication cert container will use the same resource configuration settings as the `database` container rather than having a its own `resources` section.

internal/postgres/reconcile.go
```diff
@@ -44,13 +44,21 @@ func InitCopyReplicationTLS(postgresCluster *v1beta1.PostgresCluster,
...
+	for i, c := range template.Spec.Containers {
+		if c.Name == "database" {
```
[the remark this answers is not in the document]
tjmoore4: :+1: updated.

internal/controller/postgrescluster/util.go
```diff
@@ -128,13 +128,20 @@ func addNSSWrapper(image string, template *v1.PodTemplateSpec) {
...
+	container := v1.Container{
+		Command:         []string{"bash", "-c", nssWrapperCmd},
+		Image:           image,
+		Name:            naming.ContainerNSSWrapperInit,
+		SecurityContext: initialize.RestrictedSecurityContext(),
+	}
+
+	// set the NSS wrapper container resources to the 'database' container settings
+	for i, c := range template.Spec.Containers {
+		if c.Name == "database" || c.Name == "pgbackrest" {
```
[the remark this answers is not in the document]
tjmoore4: :+1: updated.

docs/content/tutorial/resize-cluster.md
tjmoore4: That was the original idea, but on review we already use had a resource configuration example in `resize-cluster.md`, so it didn't make sense to me to add it again in another section.

internal/controller/postgrescluster/pgmonitor.go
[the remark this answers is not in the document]
tjmoore4: `exporter` is unique among the sidecars as it can be toggled on/off. With that in mind, I think it makes the most sense to leave the existing method of configuring `resources` in place.

docs/content/tutorial/resize-cluster.md
```diff
@@ -15,9 +15,19 @@ Let's dive in.
...
 
-For example, let's say we want to update our `hippo` Postgres cluster so that each instance has a limit of `2.0` CPUs and `4Gi` of memory. We can make the following changes to the manifest:
+- `spec.instances.resources` section, which sets the resource values for the PostgreSQL container, as well as any init containers in the associated pod.
+- `spec.instances.sidecars.replicacertcopy.resources` section, which sets the resources for the `replica-cert-copy` sidecar container.
+- `spec.instances.sidecars.exporter.resources` section, which sets the resources for the `exporter` sidecar container.
```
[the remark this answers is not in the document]
tjmoore4: :+1: updated.

internal/pgbouncer/reconcile_test.go
```diff
@@ -327,6 +327,111 @@ volumes:
...
+      fi
+    done
+    }; export directory="$1"; export -f monitor; exec -a "$0" bash -ceu monitor
+  - pgbouncer-config
+  - /etc/pgbouncer
+  image: image-town
+  name: pgbouncer-config
+  resources:
+    requests:
+      cpu: 200m
```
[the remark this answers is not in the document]
tjmoore4: Agreed, I looked into other options, but, given the existing default resource values, I felt it was needed. I assume leaving the same image name and TLS secret won't be an issue given the order of the tests, but I could rework that if necessary.

docs/content/tutorial/resize-cluster.md
tjmoore4: I'm hesitant to change `pgbouncerConfig` to `config` if only because I'd like to keep the names as close as possible to the actual container names. In the event we add more containers, it should always be abundantly clear with sidecar we're referring to, independent of documentation. The same idea applies to the instance `pgbackrest` sidecar.

:thinking: Unless you're saying we should rename the actual containers?

andrewlecuyer (approved):

PR #2702, opened 2021-09-17
Pod priority classes

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
Add Support for Pod Priority Classes

This update adds the ability to configure a Priority Class Name field in the
PostgresCluster spec so that users can set the priority on pods.  Priority
indicates the importance of a Pod relative to other Pods.

The priority class name can be set on all current Pods, Jobs included:

Instances
Dedicated Repo Host
PgBouncer
Backup (manual and scheduled)
Restore (data source or in-place)
Data Migration Jobs

More information [here](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/)

**Other information**:

Issue: [sc-12515]


internal/controller/postgrescluster/pgbackrest_test.go
```diff
@@ -1419,6 +1432,16 @@ func TestReconcileManualBackup(t *testing.T) {
...
 
+					// verify the priority class
+					var foundPriorityClass bool
+					for _, job := range jobs.Items {
+						if job.Spec.Template.Spec.PriorityClassName == "some-priority-class" {
+							foundPriorityClass = true
+							break
+						}
+					}
+					assert.Assert(t, foundPriorityClass)
```
cbandy, line 1443: ⛏️ This loop and the one above it don't seem necessary. This assertion passes when at least one job has the priority class, and we've already asserted that `len(jobs.Items) == 1` on L1411.

If I understand correctly, this loop can be replaced with a single assertion:

```go
assert.Equal(t, jobs.Items[0].Spec.Template.Spec.PriorityClassName, "some-priority-class")
```

tjmoore4: :+1: updated.

jmckulk (approved):

cbandy (approved):

PR #2703, opened 2021-09-17
Set Resources on all Jobs

jmckulk:
This change allows users to define resource requests and limits for all
jobs created by PGO. This includes, manual backup, replica
create backup, restore, and move jobs for pgdata, pgwal and repoHost
volumes.

A new field has been added to the spec to define resource options for
backup jobs. `spec.backups.pgbackrest.jobs.resources` is a Kubernetes
ResourceRequirements field that is passed to all backup jobs.

Resources for move jobs will be set based on the type of volume the job
is working on. pgdata and pgwal volume move jobs will use the resource
field defined for the first instance defined for the cluster (i.e.
`spec.instances[0])`.  repoHost volume move jobs will use resources
defined for the repo host
(`spec.backups.pgbackrest.repoHost.resources`).

Resources for restore jobs are configurable using
`spec.backups.pgbackrest.restore.resources`.

**Checklist:**

 
**Type of Changes:**

 
**Other information**:
[sc-12554]


internal/controller/postgrescluster/pgbackrest_test.go
```diff
@@ -2194,6 +2194,46 @@ func TestReconcilePostgresClusterDataSource(t *testing.T) {
...
+	t.Run("empty", func(t *testing.T) {
+		_, err := generateBackupJobSpecIntent(
+			&v1beta1.PostgresCluster{},
+			"", "", "", "", "",
+			nil, nil,
+		)
+		assert.NilError(t, err)
+	})
+
+	cluster := &v1beta1.PostgresCluster{
```
jmckulk, line 2207: I want to add a test around here to generate the spec when jobs is defined but resources is not. I'll push it up with the docs changes. :recycle:

internal/controller/postgrescluster/volumes.go
```diff
@@ -449,6 +449,7 @@ func (r *Reconciler) reconcileMovePGDataDir(ctx context.Context,
 					Image:        config.PostgresContainerImage(cluster),
 					Name:         naming.ContainerJobMovePGDataDir,
 					VolumeMounts: []corev1.VolumeMount{postgres.DataVolumeMount()},
+					Resources:    cluster.Spec.InstanceSets[0].Resources,
```
tjmoore4, line 452: :thinking: Is there any chance `InstanceSets` might ever be empty? Perhaps a `len(cluster.Spec.InstanceSets) > 0` check above might be useful?

internal/controller/postgrescluster/volumes.go
```diff
@@ -548,6 +549,7 @@ func (r *Reconciler) reconcileMoveWALDir(ctx context.Context,
 					Image:        config.PostgresContainerImage(cluster),
 					Name:         naming.ContainerJobMovePGWALDir,
 					VolumeMounts: []corev1.VolumeMount{postgres.WALVolumeMount()},
+					Resources:    cluster.Spec.InstanceSets[0].Resources,
```
tjmoore4, line 552: :thinking: Same as above

internal/controller/postgrescluster/volumes.go
```diff
@@ -652,6 +654,7 @@ func (r *Reconciler) reconcileMoveRepoDir(ctx context.Context,
 					Image:        config.PGBackRestContainerImage(cluster),
 					Name:         naming.ContainerJobMovePGBackRestRepoDir,
 					VolumeMounts: []corev1.VolumeMount{pgbackrest.RepoVolumeMount()},
+					Resources:    cluster.Spec.Backups.PGBackRest.RepoHost.Resources,
```
tjmoore4, line 657: :wrench:  Similar to the above, but I believe it's likely there are cases where `RepoHost` will be `nil`.

tjmoore4 (commented):
Looks good overall. A few notes, and I'll be on the lookout for the mentioned updates.

internal/controller/postgrescluster/volumes.go
jmckulk: :thinking: I don't think that is possible based on spec validation.. Maybe in a shutdown scenario? Either way would be good to have a check

tjmoore4 (approved):

internal/controller/postgrescluster/pgbackrest_test.go
```diff
@@ -2194,6 +2194,56 @@ func TestReconcilePostgresClusterDataSource(t *testing.T) {
...
+		job, err := generateBackupJobSpecIntent(
+			cluster,
+			"", "", "", "", "",
+			nil, nil,
+		)
+		assert.NilError(t, err)
+		assert.DeepEqual(t, job.Template.Spec.Containers[0].Resources,
+			corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{
+					corev1.ResourceStorage: resource.MustParse("1Gi"),
```
cbandy, line 2241: ⛏️ A container doesn't normally allow `storage` here.

```
Pod "mycluster-00-xcxl-0" is invalid: [spec.containers[0].resources.requests[storage]: Invalid value: "storage": must be a standard resource for containers
```

cbandy (approved):

PR #2705, opened 2021-09-20
Move service account from rbac overlay into manager base

james-callahan:
The ServiceAccount is already specified by name in the manager Deployment.
There is no flexibility gained by putting it in the rbac overlay; while it makes it awkward to do a custom rbac setup


andrewlecuyer:
In reviewing this PR, we have determined that the `ServiceAccount` is best aligned with the RBAC per Kubebuilder and Kustomize conventions.  For instance, you can see this in the `rbac` folder of various Kubebuilder installer examples:

- https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3-config/config/rbac/service_account.yaml
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4-config/config/rbac/service_account.yaml

And while the `ServiceAccount` is specified in the `Deployment` as you stated, it is also specified in the `RoleBinding`.  We therefore feel as though it aligning the `ServiceAccount` with the `rbac` represents the most sensible structure for the PGO Kustomize installer.

However, if you have any additional questions, or if you would like to discuss your use case further, please let us know.  For instance, is there a specific reason you need to customize the RBAC (especially considering PGO's default RBAC should provide the exact permissions required for PGO to operate)?

PR #2717, opened 2021-09-22
Update golangci-lint 

jmckulk:
The following changes fix linter errors and warnings that appear for golangci-lint version v1.42.1.

**Other information**:

### DIFF
--- a/internal/controller/postgrescluster/apply_test.go
+++ b/internal/controller/postgrescluster/apply_test.go
@@ -29,7 +29,6 @@ import (
 	"gotest.tools/v3/assert"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -54,7 +53,7 @@ func TestServerSideApply(t *testing.T) {
 	cc, err := client.New(config, client.Options{})
 	assert.NilError(t, err)
 
-	ns := &v1.Namespace{}
+	ns := &corev1.Namespace{}
 	ns.GenerateName = "postgres-operator-test-"
 	assert.NilError(t, cc.Create(ctx, ns))
 	t.Cleanup(func() { assert.Check(t, cc.Delete(ctx, ns)) })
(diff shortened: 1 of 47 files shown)


internal/controller/postgrescluster/patroni.go
```diff
@@ -261,23 +260,23 @@ func (r *Reconciler) reconcilePatroniStatus(
...
 			Name:      cluster.Spec.CustomReplicationClientTLSSecret.Name,
 			Namespace: cluster.Namespace,
 		}}
 		err := errors.WithStack(r.Client.Get(ctx,
 			client.ObjectKeyFromObject(custom), custom))
 		if err == nil {
-			return custom, err
+			return custom, nil
 		}
 		return nil, err
```
cbandy, line 276: ⛏️ Same here:

```go
		custom := …
		err := …

		return custom, err
```

internal/patroni/config.go
```diff
@@ -246,9 +246,9 @@ func DynamicConfiguration(
 	postgresql["parameters"] = parameters
 
 	// Copy the "postgresql.pg_hba" section after any mandatory values.
-	hba := make([]string, len(pgHBAs.Mandatory))
+	hba := make([]string, 0)
```
cbandy, line 249: ❓ We should allocate when possible. Does the following quiet the linter as well?
```suggestion
	hba := make([]string, 0, len(pgHBAs.Mandatory))
```

internal/controller/postgrescluster/apply_test.go
```diff
@@ -29,7 +29,6 @@ import (
 	"gotest.tools/v3/assert"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
```
cbandy, line 32: 💯 `corev1` What a doozie! 👏🏻 

❓ We can enforce it with `importas`. What do you think of adding these rules?

```patch
diff --git .golangci.yaml .golangci.yaml
index fca0785dc..09a82f418 100644
--- .golangci.yaml
+++ .golangci.yaml
@@ -7,6 +7,7 @@ linters:
   enable:
     - gomodguard
     - gosimple
+    - importas
     - misspell
   presets:
     - bugs
@@ -28,6 +29,14 @@ linters-settings:
             reason: >
               k8s.io/kubernetes is for managing dependencies of the Kubernetes
               project, i.e. building kubelet and kubeadm.
+  importas:
+    alias:
+      - pkg: k8s.io/api/(\w+)/(v[\w\d]+)
+        alias: $1$2
+      - pkg: k8s.io/apimachinery/pkg/apis/(\w+)/(v[\w\d]+)
+        alias: $1$2
+      - pkg: k8s.io/apimachinery/pkg/api/errors
+        alias: apierrors
 
 run:
   build-tags:
```

I think we already comply with the first two rules. The last one requires a couple of changes in `internal/controller/postgrescluster/controller_ref_manager.go` and `internal/controller/postgrescluster/pgbackrest_test.go`.

jmckulk: Yeah figured it would be good to go ahead and get that out of the way!

I like importas :+1: i'll add that in. Are there any other rules we can add?

cbandy: None that I know of right off. Very open to adding more later.

andrewlecuyer (approved):

cbandy (approved):

PR #2719, opened 2021-09-22
Reload PostgreSQL server certificates when they change

cbandy:
**Checklist:**

 
**Type of Changes:**

 
One must restart PostgreSQL pods when updating its server certificate.

A sidecar detects changes to the certificate and reloads it automatically, without downtime.

**Other Information**:

Issue: [sc-12369]


internal/postgres/reconcile.go
```diff
@@ -158,9 +41,31 @@ func WALVolumeMount() corev1.VolumeMount {
...
+		ReadOnly:  true,
+	}
+	certVolume := corev1.Volume{
+		Name: certVolumeMount.Name,
+		VolumeSource: corev1.VolumeSource{
+			Projected: &corev1.ProjectedVolumeSource{
+				// PostgreSQL expects client certificate keys to not be readable
+				// by any other user.
+				// - https://www.postgresql.org/docs/current/libpq-ssl.html
+				DefaultMode: initialize.Int32(0o600),
```
jmckulk, line 60: this `o` looks out of place. Should it be this?
```suggestion
				DefaultMode: initialize.Int32(0600),
```

internal/postgres/config.go
```diff
@@ -99,6 +100,54 @@ func Environment(cluster *v1beta1.PostgresCluster) []corev1.EnvVar {
 	}
 }
 
+// reloadCommand returns an entrypoint that convinces PostgreSQL to reload
+// certificate files when they change. The process will appear as name in `ps`
```
jmckulk, line 104: >The process will appear as name in `ps`

Should this be something other than name :question:

jmckulk (approved):
just had the one question but overall lgtm

internal/postgres/reconcile.go
cbandy: `0o` indicates that this is an octal number, like `0x` for hexadecimal. https://golang.org/ref/spec#Integer_literals

I prefer it over just `0` because the latter is too easy to typo. For example, `chmod` accepts a leading `1` to set the sticky bit, but doing that in Go would change to a decimal number.

 ```console
 $ touch cbandy
$ chmod 0600 cbandy
$ ls -ld cbandy
-rw-------. 1 vagrant vagrant 0 Sep 22 21:05 cbandy
$ chmod 1600 cbandy
$ ls -ld cbandy
-rw------T. 1 vagrant vagrant 0 Sep 22 21:05 cbandy
```

https://play.golang.org/p/bEOWEpT0jZ0
```go
fmt.Println(0600)   // 384
fmt.Println(1600)   // 1600
fmt.Println(0o1600) // 896
```

jmckulk: Ok cool thanks for the explanation :+1:

internal/postgres/config.go
cbandy: This is what it looks like with the wrapper. Without it, the entire multi-line script appears: `bash -c declare …`

```console
$ kubectl exec pod/mycluster-00-bvw9-0 -- ps ux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
postgres      39  0.2  0.8 606752 33128 ?        Ssl  21:11   0:01 /usr/bin/python3.6 /usr/local/bin/patroni /etc/patroni
postgres      46  0.0  0.0  14844  3196 ?        Ss   21:11   0:00 replication-cert-copy -ceu monitor
…
```

andrewlecuyer (approved):

cbandy:
[Added a comment](https://github.com/CrunchyData/postgres-operator/compare/5bf2a5e1ac8fe9b21d0680029419bad7b4c09586..45e9574d9a6b847e0c490f2b4897f2ce2aa75937) about replication certificates.

PR #2720, opened 2021-09-22
Default scheduling constraints for HA

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
This commit implements default constraints for the various Pods comprising a
PostgreSQL cluster, specifically as needed to ensure the Operator always
attempts to deploy High-Availability cluster architecture by default.

Using Pod Topology Spread Constraints, the general scheduling guidelines are
as follows:

- Pods are only considered from the same cluster.
- PgBouncer pods are only considered amongst other PgBouncer pods.
- Postgres pods are considered amongst all Postgres pods and pgBackRest repo
    host Pods.
- pgBackRest repo host Pods are considered amongst all Postgres pods and
    pgBackRest repo hosts Pods.
- Pods are scheduled across the different kubernetes.io/hostname and
    topology.kubernetes.io/zone failure domains.
- Pods are scheduled when there are fewer nodes than pods, e.g. single node.

We apply default pod topology spread constraints per the rules above when the
new spec.disableDefaultPodScheduling field is not specified at all, or is set
to false. By default it will be set to false.

**Other information**:
Issue: [sc-11183]

### DIFF
--- a/internal/controller/postgrescluster/instance.go
+++ b/internal/controller/postgrescluster/instance.go
@@ -1143,6 +1143,7 @@ func generateInstanceStatefulSetIntent(_ context.Context,
 			naming.LabelCluster:     cluster.Name,
 			naming.LabelInstanceSet: spec.Name,
 			naming.LabelInstance:    sts.Name,
+			naming.LabelData:        naming.DataPostgres,
 		})
 
 	// Don't clutter the namespace with extra ControllerRevisions.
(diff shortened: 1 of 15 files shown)


internal/controller/postgrescluster/instance.go
```diff
@@ -1168,6 +1169,36 @@ func generateInstanceStatefulSetIntent(_ context.Context,
...
+			!*cluster.Spec.DisableDefaultPodScheduling) {
+		sts.Spec.Template.Spec.TopologySpreadConstraints = append(sts.Spec.Template.Spec.TopologySpreadConstraints,
+			corev1.TopologySpreadConstraint{
+				MaxSkew:           int32(1),
+				TopologyKey:       "kubernetes.io/hostname",
+				WhenUnsatisfiable: corev1.ScheduleAnyway,
+				LabelSelector: &metav1.LabelSelector{
+					MatchExpressions: []metav1.LabelSelectorRequirement{
+						{Key: naming.LabelCluster, Operator: "In", Values: []string{cluster.Name}},
+						{Key: naming.LabelData, Operator: "Exists"},
```
cbandy, line 1185: 🔧 This Exists should be `in (postgres, pgbackrest)`

internal/naming/names.go
```diff
@@ -158,6 +158,16 @@ const (
 	sshSecretNameSuffix = "%s-ssh"
 )
 
+const (
+	// key value used with the "Data" label to indicate PostgreSQL data is attached
+	// to the relevant Pod
+	DataPostgres = "postgres"
```
cbandy, line 164: ⛏️ I'd prefer these label values be in label.go and mention "LabelData".

```go
// DataPostgres is a LabelData value that indicates the object has PostgreSQL data.
```

internal/naming/labels.go
```diff
@@ -36,6 +36,9 @@ const (
 	// LabelClusterCertificate is used to identify a secret containing a cluster certificate
 	LabelClusterCertificate = labelPrefix + "cluster-certificate"
 
+	// LabelData is used to identify
```
cbandy, line 39: ⛏️ Missing the rest of this comment.

tjmoore4: :+1: updated, thanks.

internal/controller/postgrescluster/instance.go
tjmoore4: :+1: updated.

internal/naming/labels.go
```diff
@@ -244,6 +247,7 @@ func PGBackRestDedicatedLabels(clusterName string) labels.Set {
 	commonLabels := PGBackRestLabels(clusterName)
 	operatorConfigLabels := map[string]string{
 		LabelPGBackRestDedicated: "",
+		LabelData:                DataPGBackRest,
```
cbandy, line 250: 🔧 These labels also go into the repo host STS selector which is not allowed to change. When the operator deploys with this change, an existing cluster can no longer reconcile:

```
error="StatefulSet.apps \"mycluster-repo-host\" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden"
```

tjmoore4: I updated the way this label is applied so that it is now set directly in `generateRepoHostIntent`. This should correct the selector issue as I understand it.

jmckulk (approved):

internal/naming/names.go
tjmoore4: :+1: updated

cbandy (approved):

Issue #2729, opened 2021-09-24
FQDN added to default certificate generated by CRD


ackerr01:
Would it be possible to have the FQDN added as a dns alt name to the default certificate that gets created?

{db}-primary.{namespace}.svc.cluster.local ?

PR #2732, opened 2021-09-27
Include more SANs in the PostgreSQL TLS certificate

cbandy:
**Checklist:**

 
**Type of Changes:**

 
The certificate generated for the PostgreSQL primary contains only a single DNS name.

closes CrunchyData/postgres-operator#2729

The FQDN and every possible local (search domain) name is included as SANs in the certificate.

**Other information**:

Issue: [sc-12692]

### DIFF
--- a/internal/controller/postgrescluster/pki_test.go
+++ b/internal/controller/postgrescluster/pki_test.go
@@ -19,6 +19,7 @@ package postgrescluster
 import (
 	"bytes"
 	"context"
+	"crypto/x509"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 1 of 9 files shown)


internal/controller/postgrescluster/pki_test.go
```diff
@@ -365,14 +369,14 @@ func TestReconcileCerts(t *testing.T) {
 		assert.NilError(t, err)
 
 		t.Run("check standard secret projection", func(t *testing.T) {
-			secretCertProj, err := r.reconcileClusterCertificate(ctx, initialRoot, cluster1)
+			secretCertProj, err := r.reconcileClusterCertificate(ctx, initialRoot, cluster1, primaryService)
```
andrewlecuyer, line 372: :thinking: Looks like our tests currently validate the projection - do we have anything to validate the contents of the certs themselves?  Was looking for something along those lines to verify the DNS names, but looks like our tests are with the projection only.

Should we be doing further validation of the certs themselves here? I do see some other tests in here that appear to grab the cert contents for some additional validation.

andrewlecuyer (approved):

internal/controller/postgrescluster/pki_test.go
cbandy: Added some checks around L440 in this file.

cbandy:
Added a commit to fix a test that occasionally fails in K3s 1.22.

andrewlecuyer (approved):

PR #2735, opened 2021-09-28
Update OpenShift Auto-Detect Logic

andrewlecuyer:
Updates the OpenShift auto-detection logic to remove the potential for false positives.  Specifically, PGO now specifically checks for the presence of the `SecurityContextConstraints` Kind within the `security.openshift.io` Group to determine if the Kubernetes cluster it is running within is an OpenShift cluster.

This prevents false positives by looking for a specific Group/Kind that will only ever be present in an OpenShift environment (note that the current logic is looking at a Group suffix that could exist in a non-OpenShift environment, e.g. if Rook is installed, resulting in a `replication.storage.openshift.io` Group).

[sc-12715]

### DIFF
--- a/cmd/postgres-operator/main.go
+++ b/cmd/postgres-operator/main.go
@@ -101,20 +101,25 @@ func addControllersToManager(ctx context.Context, mgr manager.Manager) error {
 func isOpenshift(ctx context.Context, cfg *rest.Config) bool {
 	log := logging.FromContext(ctx)
 
-	const openShiftAPIGroupSuffix = ".openshift.io"
+	const sccGroup, sccKind = "security.openshift.io", "SecurityContextConstraints"
 
 	client, err := discovery.NewDiscoveryClientForConfig(cfg)
 	assertNoError(err)
 
-	groups, _, err := client.ServerGroupsAndResources()
+	_, resourceLists, err := client.ServerGroupsAndResources()
 	assertNoError(err)
 
-	// If we detect that any API group name ends with "openshift.io", we'll
-	// return that this is an OpenShift environment
-	for _, g := range groups {
-		if strings.HasSuffix(g.Name, openShiftAPIGroupSuffix) {
-			log.Info("detected OpenShift environment")
-			return true
+	// If we detect that the "SecurityContextConstraints" Kind is present in the
+	// "security.openshift.io" Group, we'll return that this is an OpenShift
+	// environment
+	for _, rl := range resourceLists {
+		if strings.HasPrefix(rl.GroupVersion, sccGroup+"/") {
+			for _, r := range rl.APIResources {
+				if r.Kind == sccKind {
+					log.Info("detected OpenShift environment")
+					return true
+				}
+			}
 		}
 	}
 


cmd/postgres-operator/main.go
```diff
@@ -101,20 +102,27 @@ func addControllersToManager(ctx context.Context, mgr manager.Manager) error {
...
-	for _, g := range groups {
-		if strings.HasSuffix(g.Name, openShiftAPIGroupSuffix) {
-			log.Info("detected OpenShift environment")
-			return true
+	// If we detect that the "SecurityContextConstraints" Kind is present in the
+	// "security.openshift.io" Group, we'll return that this is an OpenShift
+	// environment
+	for _, rl := range resourceLists {
+		gv, err := schema.ParseGroupVersion(rl.GroupVersion)
+		assertNoError(err)
```
cbandy, line 118: 🤔 I wonder if we should avoid parsing and errors here. Would `strings.HasPrefix(rl.GroupVersion, sccGroup+"/")` be a reasonable substitute?

cmd/postgres-operator/main.go
```diff
@@ -101,20 +102,27 @@ func addControllersToManager(ctx context.Context, mgr manager.Manager) error {
...
 	log := logging.FromContext(ctx)
 
-	const openShiftAPIGroupSuffix = ".openshift.io"
+	const sccGroup, sccKind = "security.openshift.io", "SecurityContextConstraints"
 
 	client, err := discovery.NewDiscoveryClientForConfig(cfg)
 	assertNoError(err)
 
-	groups, _, err := client.ServerGroupsAndResources()
+	resourceLists, err := client.ServerResources()
```
cbandy, line 110: 🔧 `ServerResources()` is deprecated in favor of `ServerGroupsAndResources()`. It looks like the former calls the latter, so the behavior should be the same with either.

https://pkg.go.dev/k8s.io/client-go@v0.20.8/discovery#DiscoveryClient.ServerResources

andrewlecuyer: Darn, didn't realize that.  Will switch it back.  Thanks.

cmd/postgres-operator/main.go
andrewlecuyer: I actually went back and forth on this, and actually had `strings.HasPrefix()` initially! ... However, ultimately this felt like the safest way to ensure the proper Group was extracted from a GV, so I stuck with that approach.

Are you anticipating any issues with the current approach? (I do feel like an error would appropriate if we discover a GV that we can't parse :thinking:)

cmd/postgres-operator/main.go
andrewlecuyer: Switched back to `ServerGroupsAndResources()`.

cmd/postgres-operator/main.go
cbandy: No issues. I expect it is "impossible" for the API to return a GV that can't be parsed.

"Impossible" from the API and "impossible" in our code are different, though. I would prefer that auto-detection never halt the server because users do have a way forward when it is broken. It's not a blocker, though.

cbandy (approved):

cmd/postgres-operator/main.go
andrewlecuyer: Heh, I went back and forth again, and I def see what you mean!

While I too have no idea why an invalid GV would ever be in there in the first place, I think I agree - why let that stop PGO from running even if there were one.  We have a very high-degree of confidence those we do rely on (e.g. from the native Kube  and OpenShift API's) will be right (including the specific one we're looking for here), so why worry about errors and let others get in the way.

Updated to use `strings.HasPrefix()`.

cbandy (approved):

PR #2737, opened 2021-09-29
Update Backup Method in Upgrade Docs

andrewlecuyer:
The "Backup" method in the "PGO v4 to PGO v5" guide no longer has specific steps for adjusting the permissions of the PGO v4 pgBackRest repo for those using NFS or HostPath storage.  This is because the the specific steps provided do not work consistently in all Kubernetes envs (e.g. OpenShift will not allow `chown` to be run when using the `restricted` SCC).  Additionally, the specific permissions applied could also vary depending on the specific type of storage being utilized (e.g. NFS or HostPath), as well as how the storage is configured, provisioned, etc.

Instead, a warning is now displayed for NFS and HostPath users indicating that they should properly adjust their repo volume permissions prior to creating a PGO v5 cluster using that repo. Please note that the specific wording utilized was pulled from the "Migrate Data Volumes to New Clusters" guide to keep all guidance consistent.

### DIFF
--- a/docs/content/faq/_index.md
+++ b/docs/content/faq/_index.md
@@ -33,6 +33,23 @@ For additional information regarding commercial support and Crunchy PostgreSQL f
 
 The PGO source code is available under the [Apache License 2.0](https://github.com/CrunchyData/postgres-operator/blob/master/LICENSE.md).
 
+### Where are the release tags for PGO v5?
+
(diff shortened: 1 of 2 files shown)


jkatz (approved):

tjmoore4 (commented):
:+1: Looks good to me.

PR #2744, opened 2021-09-30
Bump pgMonitor to v4.5

cbandy:
**Checklist:**

 
**Type of Changes:**

 
The exporter image is built with [pgMonitor v4.5](https://github.com/CrunchyData/pgmonitor/releases/tag/v4.5). Fewer messages are logged when applying the SQL so that errors are more obvious.

**Other Information**:

Issue: [sc-12625]

### DIFF
--- a/internal/pgmonitor/postgres.go
+++ b/internal/pgmonitor/postgres.go
@@ -93,20 +93,31 @@ func EnableExporterInPostgreSQL(ctx context.Context, exec postgres.Executor,
 
 	stdout, stderr, err := exec.ExecInAllDatabases(ctx,
 		strings.Join([]string{
+			// Quiet NOTICE messages from IF EXISTS statements.
+			// - https://www.postgresql.org/docs/current/runtime-config-client.html
+			`SET client_min_messages = WARNING;`,
+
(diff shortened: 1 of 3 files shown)


andrewlecuyer (approved):

jmckulk (approved):
seems good to me with a quick look

internal/pgmonitor/postgres.go
```diff
@@ -93,20 +93,31 @@ func EnableExporterInPostgreSQL(ctx context.Context, exec postgres.Executor,
 
 	stdout, stderr, err := exec.ExecInAllDatabases(ctx,
 		strings.Join([]string{
+			// Quiet NOTICE messages from IF EXISTS statements.
+			// - https://www.postgresql.org/docs/current/runtime-config-client.html
+			`SET client_min_messages = WARNING;`,
```
benjaminjb, line 98: From the name, I assume this is only valid for the client, which is why you have to set it both here and on L119 and also why it doesn't have to be undone after, yeah?

cbandy: Yes. It affects only one/current connection and has no effect after the connection closes.

PR #2761, opened 2021-10-07
PGMonitor Service

edenkoveshi:
**Type of Changes:**

 
Enabling PGMonitor sets up the relevant container, however it can only be scraped using Prometheus kubrenetes pod discovery.

Enabling PGMonitor sets up a new service that exposes the relevant port, enabling scraping in any way you'd like.

**Other information**:
The new service name is <cluster-name>-pgmonitor.
The service reconciliation was written very similarly to the existing PGBouncer service.
It is documented and has been tested by me, a few local tests in my minikube environment. Seems OK.

### DIFF
--- a/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
+++ b/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
@@ -591,4 +591,9 @@ type ExporterSpec struct {
 	// More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers
 	// +optional
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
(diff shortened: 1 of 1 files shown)


jkatz:
I'm OK with the concept overall. The code does require unit tests and a bit more documentation around the function names (following Go standards).

It seems that it was copy and pasted from the pgBouncer bits, as the term "pgBouncer" is in the code comments themselves.

I do think that this should be an opt-in feature, i.e. you should choose if you want to expose the exporter port. That would likely require a CRD change.

edenkoveshi:
Added the service feature as optional using a new property in the CRD.
setting `spec.monitoring.pgmonitor.exporter.expose` to `true` will create the service, and it won't be created otherwise.
As I have said, this is very similar to pgBouncer service creation and so comments are quite the same.. changed pgBouncer to pgMonitor though.

PR #2792, opened 2021-10-19
Restart PostgreSQL when changing postmaster parameters

cbandy:
**Checklist:**

 
**Type of Changes:**

 
One must manually restart PostgreSQL when changing a parameter that can only be set at server start.

PGO rolls out such changes while considering whether the primary or replica must apply it first.

**Other information**:

Issue: [sc-11430]

### DIFF
--- a/config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
+++ b/config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
@@ -4855,18 +4855,24 @@ spec:
               patroni:
                 properties:
                   dynamicConfiguration:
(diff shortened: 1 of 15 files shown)


cbandy:
> @cbandy it feels like this should now work itself out?

I suspect so. `shared_preload_libraries` should be detected, but I didn't test it. If it is detected and we automatically restart, should the language in the warning change?

andrewlecuyer:
> If it is detected and we automatically restart, should the language in the warning change?

Yea, I think we'll want to change that warning if we're now handling the restart for the user (e.g. to prevent them from attempting an unneeded restart).

PR #2808, opened 2021-10-27
Initial E2E test

jmckulk:
This change adds a quick end-to-end test that use the Kuttl test
framework. The layout for Kuttl tests is `testing/kuttl/e2e` with
individual tests in sub-directories of the `e2e` directory.

Run test(s) by using the `make check-kuttl` target.

The initial test creates a cluster then checks the `postgrescluster` status and
that the initial backup completes.

[sc-10754]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,7 @@ GO ?= go
 GO_BUILD = $(GO_CMD) build -trimpath
 GO_CMD = $(GO_ENV) $(GO)
 GO_TEST ?= $(GO) test
+KUTTL_TEST ?= kuttl test
 
 # Disable optimizations if creating a debug build
 ifeq ("$(DEBUG_BUILD)", "true")
@@ -202,6 +203,11 @@ check-envtest-existing:
 	USE_EXISTING_CLUSTER=true $(GO_TEST) -count=1 -cover -p=1 -tags=envtest ./...
 	${PGO_KUBE_CLIENT} delete -k ./config/dev
 
+# Expects operator to be running
+.PHONY: check-kuttl
+check-kuttl:
+	${PGO_KUBE_CLIENT} ${KUTTL_TEST} \
+		--config testing/kuttl/kuttl-test.yaml
 
 .PHONY: check-generate
 check-generate: generate-crd generate-deepcopy generate-rbac
--- a/testing/kuttl/e2e/cluster-start/00-assert.yaml
+++ b/testing/kuttl/e2e/cluster-start/00-assert.yaml
@@ -1,0 +1,28 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: test
+status:
+  instances:
+    - name: instance1
+      readyReplicas: 1
+      replicas: 1
+      updatedReplicas: 1
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  labels:
+    postgres-operator.crunchydata.com/cluster: test
+    postgres-operator.crunchydata.com/pgbackrest-backup: replica-create
+spec:
+  template:
+    spec:
+      containers:
+      - name: pgbackrest
+        command: ["/opt/crunchy/bin/pgbackrest"]
+        resources:
+          requests:
+            cpu: 1m
+status:
+  succeeded: 1
--- a/testing/kuttl/e2e/cluster-start/00-assert.yaml
+++ b/testing/kuttl/e2e/cluster-start/00-assert.yaml
@@ -1,7 +1,7 @@
 apiVersion: postgres-operator.crunchydata.com/v1beta1
 kind: PostgresCluster
 metadata:
-  name: test
+  name: cluster-start
 status:
   instances:
     - name: instance1
@@ -13,16 +13,7 @@ apiVersion: batch/v1
 kind: Job
 metadata:
   labels:
-    postgres-operator.crunchydata.com/cluster: test
+    postgres-operator.crunchydata.com/cluster: cluster-start
     postgres-operator.crunchydata.com/pgbackrest-backup: replica-create
-spec:
-  template:
-    spec:
-      containers:
-      - name: pgbackrest
-        command: ["/opt/crunchy/bin/pgbackrest"]
-        resources:
-          requests:
-            cpu: 1m
 status:
   succeeded: 1
--- a/testing/kuttl/e2e/cluster-start/00-cluster.yaml
+++ b/testing/kuttl/e2e/cluster-start/00-cluster.yaml
@@ -1,7 +1,7 @@
 apiVersion: postgres-operator.crunchydata.com/v1beta1
 kind: PostgresCluster
 metadata:
-  name: test
+  name: cluster-start
 spec:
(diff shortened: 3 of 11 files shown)


Makefile
```diff
@@ -202,6 +203,11 @@ check-envtest-existing:
 	USE_EXISTING_CLUSTER=true $(GO_TEST) -count=1 -cover -p=1 -tags=envtest ./...
 	${PGO_KUBE_CLIENT} delete -k ./config/dev
 
+# Expects operator to be running
+.PHONY: check-kuttl
+check-kuttl:
```
jkatz, line 208: Perhaps

```suggestion
check-e2e:
```

Unless we're planning on adding other e2e testing libs?

jmckulk: I was leaving it open to add other e2e libs but I don't have any in mind at the moment. I was also following what we did for `envtest` with `make check-envtest{-existing}`

testing/kuttl/kuttl-test.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestSuite
+testDirs:
+- testing/kuttl/e2e/
+timeout: 120
+parallel: 2
+# comment out namespace if you want kuttl tests to run in its
+# own generated namespace
+namespace: postgres-operator
```
jmckulk, line 9: Should the default here be to run in multiple namespaces :question: Then have a comment saying to add the line if you want them in a single ns.

testing/kuttl/e2e/cluster-start/01-psqlconnect.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- script: /home/vagrant/postgres-operator/testing/kuttl/script/portforward.sh
+  background: true
+# this sleep is needed when running in cloud environments, may not be needed
+# when running locally
```
jmckulk, line 7: Can you update the comment to say why the sleep is needed? aiui, in this case the `connectpsql.sh` will fail if the port-forward hasn't started properly and it tries to connect.

Can you add a `TODO` saying to find a better way to make sure that the port-forward has started :pick: ?

testing/kuttl/kuttl-test.yaml
ValClarkson: from dealing with previous e2e testing, I dislike running in generated namespace because it has made debugging issues a little more difficult but I can see in a CI/CD env it being useful to be in generated namespaces. I'll make an update to default to multiple namespaces and add the comment.

testing/kuttl/e2e/cluster-start/01-psqlconnect.yaml
ValClarkson: will do

Makefile
ValClarkson: I like leaving it at check-kuttl this will allow us to add more e2e libs in the future and we could potentially add a check-e2e to then run all of them.

benjaminjb (commented):
Just to make sure I'm understanding kuttl, the 01-psqlconnect TestStep will fail if any of the commands fail, right? That's why we don't have to do any 01-assert.yaml?

Makefile
jkatz: OK. Suggestion withdrawn.

jmckulk:
> Just to make sure I'm understanding kuttl, the 01-psqlconnect TestStep will fail if any of the commands fail, right? That's why we don't have to do any 01-assert.yaml?

Yup that is the idea. @ValClarkson is that what you saw when you were testing?

testing/kuttl/kuttl-test.yaml
```diff
@@ -0,0 +1,20 @@
...
+suppress: 
+- events
```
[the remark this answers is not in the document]
ValClarkson: Yes i think leaving it will be better for the moment. It spits out quite a bit of information when it is not suppressed.

testing/kuttl/e2e/cluster-start/00-assert.yaml
```diff
@@ -0,0 +1,28 @@
...
+    postgres-operator.crunchydata.com/pgbackrest-backup: replica-create
+spec:
+  template:
+    spec:
+      containers:
+      - name: pgbackrest
+        command: ["/opt/crunchy/bin/pgbackrest"]
+        resources:
+          requests:
+            cpu: 1m
```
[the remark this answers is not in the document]
ValClarkson: updated

testing/kuttl/e2e/cluster-start/00-cluster.yaml
```diff
@@ -0,0 +1,31 @@
...
+        resources:
+          requests:
+            storage: 1Gi
+  backups:
+    pgbackrest:
+      image: registry.developers.crunchydata.com/crunchydata/crunchy-pgbackrest:centos8-2.35-0
+      jobs:
+        resources:
+          requests:
+            cpu: 1m
```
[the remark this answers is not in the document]
ValClarkson: update

testing/kuttl/e2e/cluster-start/00-cluster.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: test
```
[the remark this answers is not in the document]
ValClarkson: kept cluster-start

> > Just to make sure I'm understanding kuttl, the 01-psqlconnect TestStep will fail if any of the commands fail, right? That's why we don't have to do any 01-assert.yaml?
> 
> Yup that is the idea. @ValClarkson is that what you saw when you were testing?

yes if anything fails the whole test fails

testing/kuttl/script/portforward.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
ValClarkson: updated

testing/kuttl/script/connectpsql.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+PG_CLUSTER_USER_SECRET_NAME=test-pguser-test
+
+PGPASSWORD=$(kubectl get secrets -n postgres-operator "${PG_CLUSTER_USER_SECRET_NAME}" -o go-template='{{.data.password | base64decode}}') \
+PGUSER=$(kubectl get secrets -n postgres-operator "${PG_CLUSTER_USER_SECRET_NAME}" -o go-template='{{.data.user | base64decode}}') \
+PGDATABASE=$(kubectl get secrets -n postgres-operator "${PG_CLUSTER_USER_SECRET_NAME}" -o go-template='{{.data.dbname | base64decode}}') \
+psql -h localhost -c "create table test (x int);"
```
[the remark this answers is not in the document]
ValClarkson: updated

testing/kuttl/kuttl-test.yaml
ValClarkson: updated

benjaminjb (approved):
LGTM

PR #2814, opened 2021-10-28
Reconcile pgAdmin4 users through its Python models

cbandy:
**Checklist:**

 
**Type of Changes:**

 
Only the default administrator exists in pgAdmin.

The default administrator is disabled and there is a pgAdmin user for everything in `PostgresCluster.spec.users`.

**Other Information**:

Issue: [sc-12535]

~I left it split into a few commits in case that's easier to review, but they should be squashed before merging.~


internal/pgadmin/users.go
```diff
@@ -0,0 +1,128 @@
...
+
+  db.session.add(admin)
+  db.session.commit()
+
+  for line in sys.stdin:
+    if not line.strip(): continue
+
+    data = json.loads(line)
+    user = (
+      db.session.query(User).filter_by(email=data['username']).first() or
```
jkatz, line 88: `data['username']` -- if for whatever reason this is blank, this will throw a `KeyError`.

If you want to fail softly, I'd suggest:

```suggestion
      db.session.query(User).filter_by(email=data.get('username')).first() or
```

jkatz: et al. for the other ones.

internal/controller/postgrescluster/pgadmin_test.go
```diff
@@ -604,6 +606,136 @@ volumeMode: Filesystem
...
+
+		// Container not running
+		pod.Status.ContainerStatuses = nil
+		r.Client = fake.NewClientBuilder().
+			WithLists(&corev1.PodList{Items: []corev1.Pod{*pod}}).
+			Build()
+
+		assert.NilError(t, r.reconcilePGAdminUsers(ctx, cluster, nil, nil))
+
+		// Pod terminating
```
benjaminjb, line 651: Would the "pod terminating" section test a different path in the code, since the `running` bool is still false and the condition we're looking for in pgadmin.go#330 is `running && !terminating`?

internal/pgadmin/users.go
cbandy: By "blank" you mean missing from the dictionary, right?

```pycon
>>> data = dict(empty='', full='asdf')
>>> data['empty']
''
>>> data['full']
'asdf'
>>> data['other']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'other'
```

🤔 I think an exception here is okay. It describes the problem and the line that caused it, and the `python` command exits non-zero. What else might we do if the input JSON is incorrect?

internal/controller/postgrescluster/postgres.go
```diff
@@ -259,6 +259,13 @@ func (r *Reconciler) reconcilePostgresUsers(
 	if err == nil {
 		err = r.reconcilePostgresUsersInPostgreSQL(ctx, cluster, instances, users, secrets)
 	}
+	if err == nil {
+		// TODO(cbandy): If I were looking for pgAdmin stuff, I wouldn't expect
```
benjaminjb, line 263: Is there a better place to put this code? I think you lay out the case for why it's here, so I'm not sure there's a TODO here.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -502,6 +492,27 @@ type PostgresStandbySpec struct {
...
+}
+
+// Default sets the defaults for any user interfaces that are set.
+func (s *UserInterfaceSpec) Default() {
+	if s.PGAdmin != nil {
+		s.PGAdmin.Default()
+	}
+}
+
+// PostgresUserInterfaceStatus is a union of the supported PostgreSQL user interfaces.
```
benjaminjb, line 509:
```suggestion
// PostgresUserInterfaceStatus is a union of the supported PostgreSQL user interfaces statuses.
```
?

internal/controller/postgrescluster/pgadmin_test.go
cbandy: The idea was to quickly call out both criteria, but this fails if I remove `running` from the implementation. I'll rework it.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgadmin_types.go
```diff
@@ -96,3 +96,10 @@ func (s *PGAdminPodSpec) Default() {
 		*s.Replicas = 1
 	}
 }
+
+// PGAdminPodStatus represents the observed state of a pgAdmin deployment.
+type PGAdminPodStatus struct {
+
+	// Identifies the users that have been installed into pgAdmin.
```
andrewlecuyer, line 103: :wrench: This wording threw me off a bit at first, since its not really possible to directly identify any specific users using the value/hash stored in this field.

Can we clarify that its a hash used to track the users that have been installed into pgAdmin?

internal/pgadmin/users_test.go
```diff
@@ -0,0 +1,150 @@
...
+			_ context.Context, stdin io.Reader, stdout, stderr io.Writer, command ...string,
+		) error {
+			assert.Assert(t, stdin != nil, "should send stdin")
+			assert.Assert(t, stdout != nil, "should capture stdout")
+			assert.Assert(t, stderr != nil, "should capture stderr")
+
+			assert.Check(t, !strings.ContainsRune(strings.Join(command, ""), '\t'),
+				"Python should not be indented with tabs")
+
+			assert.DeepEqual(t, command, []string{"python3", "-c", `
```
andrewlecuyer, line 46: :thinking: Is there a linter we can run Python code through, e.g. like we do with our `bash` code using `shellcheck`?

internal/pgadmin/users.go
jkatz: Yes I meant missing.

We could gracefully handle the failure or if we keep it as a KeyError, provide a more informative message.

internal/pgadmin/users_test.go
cbandy: I know of `pylint` and `flake8`, and I see CentOS packages for both. I'll try them out.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgadmin_types.go
cbandy: Makes sense; identity/identify. [StatefulSetStatus](https://pkg.go.dev/k8s.io/api/apps/v1#StatefulSetStatus) uses "indicates the version" on similar fields...

How about the following?

```suggestion
	// Hash indicating which users have been installed into pgAdmin.
```

andrewlecuyer: Works for me!

internal/pgadmin/users_test.go
andrewlecuyer: Sounds good!  Definitely not needed for this specific PR, but figure a linter could help any non-Python coders keep these parts of the code clean (I'm also happy to add an item to the backlog for this, so just let me know).

cbandy: [Pylint expects to run modules](https://pylint.pycqa.org/en/latest/user_guide/run.html). It wants a comment describing the module. The assignments are top-level, so it wants them spelled like constants.

[Flake8](https://github.com/pycqa/flake8) has fewer complaints, so I used that and addressed issues it raised. Neither tool comes pre-installed in the [Linux provided by GitHub Actions](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md). We'll need to do more if we want that test to run there.

```console
$ pylint script.py
************* Module script
script.py:4:0: C0301: Line too long (101/100) (line-too-long)
script.py:1:0: C0114: Missing module docstring (missing-module-docstring)
script.py:1:0: C0410: Multiple imports on one line (importlib.util, os, sys) (multiple-imports)
script.py:3:0: C0103: Constant name "spec" doesn't conform to UPPER_CASE naming style (invalid-name)
script.py:4:0: C0103: Constant name "root" doesn't conform to UPPER_CASE naming style (invalid-name)
…
```

internal/controller/postgrescluster/postgres.go
cbandy: Agreed. I dropped the TODO.

internal/controller/postgrescluster/pgadmin_test.go
cbandy: Split into two subtests.

cbandy:
Rebased to bring in the Deployment → StatefulSet change. This PR now does a `Get` to find the pgAdmin pod rather than a `List`.

Addressed review feedback and squashed.

internal/pgadmin/users.go
andrewlecuyer: I do feel as though the current behavior/exception is exactly what we're looking for, i.e. we get the non-zero exit with an error that describes the problem well enough.  I do also like keeping these embedded scripts as terse as possible.

Therefore, not seeing the need for any additional error handling here at this time.

internal/pgadmin/users_test.go
andrewlecuyer: Excellent, that sounds great!

internal/controller/postgrescluster/pgadmin_test.go
```diff
@@ -31,6 +32,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
```
andrewlecuyer, line 35: :thinking: Curious about the use of `fake` here vs. `env` test, especially considering we've favored `envtest` per the recommendation for `envtest` in the [go docs for `fake`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/fake):

> When in doubt, it's almost always better not to use this package and instead use envtest.Environment with a real client and API server.

I suppose this might be a scenario where there is no doubt (and I do see one other existing test that uses `fake`) since the Kube API interaction is limited to a single `Get()`, and we're not actually creating anything and not exec'ing into an actual Pod?

I suppose I had been thinking _any_ Kube API interaction would be good reason to hit a real API server via `envtest`, but if there are scenarios where we feel like it might make sense to reach for `fake` instead (e.g. assuming here this was done to keep the test simple for what is a simple/standard API interaction), would definitely be worth discussing a bit among the team.

andrewlecuyer (approved):

internal/controller/postgrescluster/pgadmin_test.go
cbandy: > I suppose this might be a scenario where there is no doubt (and I do see one other existing test that uses `fake`) since the Kube API interaction is limited to a single `Get()`, and we're not actually creating anything and not exec'ing into an actual Pod?

Right. I want the Pod in particular states, but perhaps `fake` is an indication that this method is doing too much. Its responsibility right now is to find a Pod, pass an exec to the other package, and hash/cache the result. 

We could isolate the API call by making it from another method and passing the Pod into this one. 🤔 How would we test that method then? A basic exists + not-exists using `envtest` makes sense, but it'd be nice to cut out the other stuff about PodExec and the result. 🤔 Maybe it's enough to no-op PodExec.

> I suppose I had been thinking _any_ Kube API interaction would be good reason to hit a real API server via `envtest`, but if there are scenarios where we feel like it might make sense to reach for `fake` instead (e.g. assuming here this was done to keep the test simple for what is a simple/standard API interaction), would definitely be worth discussing a bit among the team.

This one is looking for a Pod and trying to PodExec. 🤔 I'm pretty sure to get either of those from the API requires a real cluster. Perhaps coverage of these calls is best reserved for an E2E?

💭 Correct me if I'm wrong, but I think we're comfortable with mocking out PodExec. Perhaps something about mocking out Pod API objects?

💭 If we did all our observations up front, this one `Get()` would not be happening here.

andrewlecuyer: We should be able to create/mock a Pod using just `envtest`, e.g. as was done here:

https://github.com/CrunchyData/postgres-operator/blob/a2c8ad911b7add6134dcbd1f412e35eb2d220de4/internal/controller/postgrescluster/pgbackrest_test.go#L2179-L2198

However, no denying that is more cumbersome than using `fake`.  And while I agree moving the API call could help in one sense (e.g. no need to worry about), like you said, that just moves the problem elsewhere.

So overall I have no issue with using `fake`; just seems like something worth discussing a bit with the team so we have an idea of when/where its use might make sense (and def no need to change anything for this PR).

And great point about this once again being a place where up-front observation could help us architecturally (e.g. with testing) - definitely agree there.   Maybe we start flagging these points in the code (also going to look at getting a story creating to track our thoughts)?

PR #2816, opened 2021-10-29
Add upgrade check request and basic docs

benjaminjb:
Add upgrade check request and basic docs

**Checklist:**

**Type of Changes:**

No automated checks for versions

Automated checks for versions (need to revisit the client when the server is set)

**Other information**:
Issue [sc-12817]

### DIFF
--- a/internal/util/http.go
+++ b/internal/util/http.go
@@ -1,0 +1,188 @@
+package util
+
+/*
+ Copyright 2017 - 2021 Crunchy Data Solutions, Inc.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/crunchydata/postgres-operator/internal/logging"
+)
+
+const (
+	clientHeader = "X-Client-Upgrade-State"
+)
+
+var (
+	client HTTPClient
+
+	// With these Backoff settings, wait.ExponentialBackoff will
+	// * use one second as the base time;
+	// * increase delays between calls by a power of 2 (1, 2, 4, etc.);
+	// * and retry four times.
+	// Note that there is no indeterminacy here since there is no Jitter set).
+	// With these parameters, the calls will occur at 0, 1, 3, and 7 seconds
+	// (i.e., at 1, 2, and 4 second delays for the retries).
+	backoff = wait.Backoff{
+		Duration: 1 * time.Second,
+		Factor:   float64(2),
+		Steps:    4,
+	}
+
+	// TODO(benjaminjb): Get real URL
+	upgradeCheckURL    = "http://localhost:8080"
+	upgradeCheckPeriod = 24 * time.Hour
+)
+
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+func init() {
+	// Since we create this client once during startup,
+	// we want each connection to be fresh, hence the non-default transport
+	// with DisableKeepAlives set to true
+	// See https://github.com/golang/go/issues/43905 and https://github.com/golang/go/issues/23427
+	// for discussion of problems with long-lived connections
+	client = &http.Client{
+		Timeout: 5 * time.Second,
+		Transport: &http.Transport{
+			DisableKeepAlives: true,
+		},
+	}
+}
+
+// Extensible struct for client upgrade data
+type clientUpgradeData struct {
+	Version string `json:"version"`
+}
+
+func addHeader(req *http.Request, versionString string) (*http.Request, error) {
+	upgradeInfo := &clientUpgradeData{
+		Version: versionString,
+	}
+	marshaled, err := json.Marshal(upgradeInfo)
+	if err == nil {
+		upgradeInfoString := string(marshaled)
+		req.Header.Add(clientHeader, upgradeInfoString)
+	}
+
+	return req, err
+}
+
+func checkForUpgrades(versionString string, backoff wait.Backoff) (message string, err error) {
+	var res *http.Response
+	var bodyBytes []byte
+
+	// Guard against panics within the checkForUpgrades function to allow the
+	// checkForUpgradesScheduler to reschedule a check
+	defer func() {
+		if panicErr := recover(); panicErr != nil {
+			err = fmt.Errorf("%s", panicErr)
+		}
+	}()
+
+	// Prep request
+	req, err := http.NewRequest("GET",
+		upgradeCheckURL,
+		nil)
+	if err == nil {
+		req, err = addHeader(req, versionString)
+	}
+
+	// wait.ExponentialBackoff will retry the func according to the backoff object until
+	// (a) func returns done as true or
+	// (b) the backoff settings are exhausted,
+	// i.e., the process hits the cap for time or the number of steps
+	// The anonymous function here sets certain preexisting variables (res, err)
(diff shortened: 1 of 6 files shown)


internal/util/http_test.go
```diff
@@ -0,0 +1,234 @@
...
+func (m *MockClient) Do(req *http.Request) (*http.Response, error) {
+	return funcFoo()
+}
+
+func TestCheckForUpdates(t *testing.T) {
+	log := logging.Discard()
+	// Setting retries to 2 to speed up testing
+	orig := os.Getenv(updateCheckMaxRetriesEnvVar)
+	os.Setenv(updateCheckMaxRetriesEnvVar, "2")
+	backoff := getBackoff(log)
```
cbandy, line 106: 🤔 This gets backoff parameters from the environment. Will the test behave differently in different environments?

Normally, the environment for a running process won't change. Perhaps we can avoid the above question by keeping settings in a variable/instance rather than loading them every call? Like a type that has the backoff settings, maybe a client, and methods.

```go
func (checker) CheckForUpdates()
type checker struct {
  backoff wait.Backoff
  client  *http.Client

  url, version string
}
```

The tests here can assign any backoff, client, etc, and not worry about restoring env vars. A "new-from-env" constructor could do all the parsing and keep behavior similar to now: `New(…).CheckForUpdates()`.

internal/util/http.go
```diff
@@ -0,0 +1,228 @@
...
+				// We can't close the body of this response in this block since we use it outside
+				// so ignore this linting error
+				res, err = client.Do(req) //nolint:bodyclose
+				// This is a very basic check, ignoring nuances around
+				// certain StatusCodes that should either prevent or impact retries
+				if err == nil && res.StatusCode == http.StatusOK {
+					return true, nil
+				}
+				if err == nil {
+					err = fmt.Errorf("received StatusCode %d", res.StatusCode)
```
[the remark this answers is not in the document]
benjaminjb: I'm doing the `client.Do` inside the `wait.ExponentialBackoff`, but since ExponentialBackoff takes a [conditionFunc](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#ConditionFunc) that only returns a bool + err, I decided to assign the results of the client.Do to preexisting vars, so I could check them outside of the wait.ExponentialBackoff.

(This feels like it's verging on breaking the Golang precept "share memory by communicating, don't communicate by sharing memory," but only _feels_ like that to me.)

internal/util/http_test.go
benjaminjb: This comes back to the issues I was wrestling with last week: on one hand, isn't it great to have things that are configurable but also provide defaults so you don't have to configure them; but on the other, under what circumstance would anyone really want to change these settings (besides the core configuration of turning this on/off)? 

We might want different settings for testing, but an even easier way to handle that is to just make them vars, set them one way in the code and another way in the testing, which again avoids any question of the environment accidentally setting these one way or another.

internal/util/http.go
```diff
@@ -0,0 +1,174 @@
...
+func init() {
+	// Since we create this client once during startup,
+	// we want each connection to be fresh, hence the non-default transport
+	// See https://github.com/golang/go/issues/43905 and https://github.com/golang/go/issues/23427
+	// for discussion of problems with long-lived connections
+	client = &http.Client{
+		Timeout: 5 * time.Second,
+		Transport: &http.Transport{
+			DisableKeepAlives:   true,
+			MaxIdleConnsPerHost: -1,
```
cbandy, line 62: ❓ I don't see negative `MaxIdleConnsPerHost` documented in `net/http` or mentioned in the linked issues. (Perhaps I missed it?) Do we have something that ensures it works the way we expect? Do we need another link that shows it?

internal/util/http.go
```diff
@@ -0,0 +1,174 @@
...
+	return "", err
+}
+
+// CheckForUpdatesScheduler invokes the check func when the operator starts
+// and then on the given period schedule
+func CheckForUpdatesScheduler(versionString string, channel chan bool) {
+	ctx := context.Background()
+	log := logging.FromContext(ctx)
+	defer func() {
+		if err := recover(); err != nil {
```
cbandy, line 144: ❓ I imagine we want to continue trying to check for updates when something panics. Should there be a similar defer+recover in `checkForUpdates`?

🤔 Recovering from panic in each HTTP call seems like too much. One recovery for all of `checkForUpdates` is probably enough. Perhaps there it can convert to an `error` and return.

internal/util/http_test.go
```diff
@@ -0,0 +1,197 @@
...
+		assert.Equal(t, res, `{"display_name":"PGO 5.0.3","major_version":5}`)
+	})
+}
+
+func TestCheckForUpdatesScheduler(t *testing.T) {
+	t.Run("panic recovers", func(t *testing.T) {
+		done := make(chan bool, 1)
+		// capture logs
+		var calls []string
+		logging.SetLogFunc(1, func(input genericr.Entry) {
```
cbandy, line 153: 🌱 These `SetLogFunc` seem alright for now. I worry about unintended side-effects in other tests; could there be any really? Perhaps this suggests that `CheckForUpdatesScheduler` should take a `context.Context` argument. The purpose of one would be ambiguous, I think: Does the scheduler respect it's `ctx.Done()`? Would canceling it stop the scheduler? etc. Things for some other day.

cbandy (commented):
Two questions and some other musings. Looking good.

internal/util/http.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/crunchydata/postgres-operator/internal/logging"
+)
+
+const (
+	clientHeader = "X-Client-Update-State"
+)
+
+var (
+	client  HTTPClient
+	backoff = wait.Backoff{
```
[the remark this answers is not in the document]
benjaminjb: Adding a comment (and ran some tests just to make sure I had the times right)

internal/util/http.go
```diff
@@ -0,0 +1,174 @@
...
+	Do(req *http.Request) (*http.Response, error)
+}
+
+func init() {
+	// Since we create this client once during startup,
+	// we want each connection to be fresh, hence the non-default transport
+	// See https://github.com/golang/go/issues/43905 and https://github.com/golang/go/issues/23427
+	// for discussion of problems with long-lived connections
+	client = &http.Client{
+		Timeout: 5 * time.Second,
```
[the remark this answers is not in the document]
benjaminjb: This timeout is not related to the Backoff settings. (In fact, a little experimentation with Backoff settings tells me that they're entirely independent from how long the function takes to complete -- the Backoff setting `Cap` isn't a maximum for how long a function takes or even how long the whole ExponentialBackoff takes, but just a cap on how long a delay for a single call is).

I could definitely see bumping this up, or putting in a comment to revisit this when we have the API to experiment with (as well as maybe running a few tests through proxies, maybe).

internal/util/http.go
benjaminjb: Originally I was going off a pattern I had seen elsewhere for clean/new http connections that set both, but after reading the Transport code more, it seems like DisableKeepAlives provides the behavior we want (that is, given the if statement `if t.DisableKeepAlives || t.MaxIdleConnsPerHost < 0 {` we don't need both (and there's no other place where a greater-than-zero MaxIdleConnsPerHost interferes with DisableKeepAlives.

internal/util/http.go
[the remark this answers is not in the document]
benjaminjb: Note: Cap is a cap on a single call, but it looks like it's checking if the next call is within the duration: https://github.com/kubernetes/apimachinery/blob/94020522c95cdeeeaeb7ad405e66f321f0ef9d57/pkg/util/wait/wait.go#L277-L283

internal/util/http.go
benjaminjb: OK, makes a lot of sense, I added a defer+recover to `checkForUpdates`, so now a single panicking call shouldn't interfere with future calls.

The downside of this is that since I am recovering from panics in `checkForUpdates`, I'm not sure how to test panic recovery from `CheckForUpdatesScheduler` since I was originally hijacking my mocked client call inside `checkForUpdates` to cause a panic.

internal/util/http.go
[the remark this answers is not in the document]
benjaminjb: Eh, you know what, we can make this simpler by just leaving out cap entirely.

internal/util/http.go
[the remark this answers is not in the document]
benjaminjb: (Also, I dropped the Cap field -- it felt unnecessary for this)

cbandy (approved):

PR #2823, opened 2021-11-02
Change pgAdmin Deployment to StatefulSet

andrewlecuyer:
Certain properties of a pgAdmin deployment make it better suited to be deployed via a StatefulSet rather than a Deployment (e.g. it includes its own embedded database, and we also want to ensure only a single replica is running at any given time).  This change therefore updates the intent generated by PGO for deploying pgAdmin to use a StatefulSet rather than a Deployment.  The StatefulSet uses the default `RollingUpdate` update strategy, and is also configured with a proper `ServiceName` as required for a stable network ID.

[sc-12915]

### DIFF
--- a/internal/controller/postgrescluster/pgadmin.go
+++ b/internal/controller/postgrescluster/pgadmin.go
@@ -46,7 +46,7 @@ func (r *Reconciler) reconcilePGAdmin(
 		dataVolume, err = r.reconcilePGAdminDataVolume(ctx, cluster)
 	}
 	if err == nil {
-		err = r.reconcilePGAdminDeployment(ctx, cluster, dataVolume)
+		err = r.reconcilePGAdminStatefulSet(ctx, cluster, dataVolume)
 	}
 	return err
 }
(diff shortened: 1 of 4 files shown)


ValClarkson (approved):
lgtm

internal/controller/postgrescluster/pgadmin.go
```diff
@@ -185,53 +185,58 @@ func (r *Reconciler) reconcilePGAdminDeployment(
...
 	// if the shutdown flag is set, set pgAdmin replicas to 0
 	if cluster.Spec.Shutdown != nil && *cluster.Spec.Shutdown {
-		deploy.Spec.Replicas = initialize.Int32(0)
+		sts.Spec.Replicas = initialize.Int32(0)
 	} else {
-		deploy.Spec.Replicas = cluster.Spec.UserInterface.PGAdmin.Replicas
+		sts.Spec.Replicas = cluster.Spec.UserInterface.PGAdmin.Replicas
 	}
 
 	// Don't clutter the namespace with extra ReplicaSets.
```
cbandy, line 193: ⛏️ StatefulSet doesn't have ReplicaSets.

```suggestion
	// Don't clutter the namespace with extra ControllerRevisions.
```

internal/controller/postgrescluster/pgadmin.go
```diff
@@ -185,53 +185,58 @@ func (r *Reconciler) reconcilePGAdminDeployment(
...
-		MaxUnavailable: intstr.ValueOrDefault(nil, intstr.FromInt(0)),
+	sts.Spec.RevisionHistoryLimit = initialize.Int32(0)
+
+	// Give the Pod a stable DNS record based on its name.
+	// - https://docs.k8s.io/concepts/workloads/controllers/statefulset/#stable-network-id
+	// - https://docs.k8s.io/concepts/services-networking/dns-pod-service/#pods
+	sts.Spec.ServiceName = naming.ClusterPodService(cluster).Name
+
+	// Set the StatefulSet update strategy to "RollingUpdate", and the Partition size for the
+	// update strategy to 0 (note that these are the defaults for a StatefulSet).
```
cbandy, line 202: 🤔 I had to think for a moment about the partition "size" of zero. That made me think of an empty set and no-op. I think it's better to describe the effect of the values used.

```suggestion
	// Delete and recreate every pod of the StatefulSet when the pod template changes.
```

Should we also call out that known issue mentioned in the docs? Perhaps the link to the docs is enough?

> If you update the Pod template to a configuration that never becomes Running and Ready (for example, due to a bad binary or application-level configuration error), StatefulSet will stop the rollout and wait.

cbandy (approved):
Some nitpicks in comments.

internal/controller/postgrescluster/pgadmin.go
andrewlecuyer: Comment updated per suggestion.

internal/controller/postgrescluster/pgadmin.go
andrewlecuyer: Appended to the description to describe the effect, and also added a direct link to the "Forced rollback" section in the STS docs.

PR #2828, opened 2021-11-03
PodDisruptionBudgets for Instance Sets & PGBouncer

jmckulk:
Reconcile Pod Disruption Budget's (PDB) for PG Instance sets and
PGBouncer deployments. This ensures a minimum number of PG Instance Pods
in a set and replicas for PGBouncer will remain running during voluntary
disruptions of the Kubernetes cluster.

PDBs are configured by setting the `minAvailable` field in the spec
(`spec.instances.minAvailable` for instances and
`spec.proxy.pgbouncer.minAvailable` for PGBouncer). If the field is set
to 0 a PDB will not be created and any existing PDBs for that resource
will be deleted.

If `minAvailable` is not defined a default value will be set. If your
resource has more than one replica a PDB with a minAvailable of 1 will
be created.

**Checklist:**

 
**Type of Changes:**

 
**Other information**:
[sc-12872]


internal/controller/postgrescluster/instance.go
```diff
@@ -538,14 +539,20 @@ func (r *Reconciler) reconcileInstanceSets(
...
+			ctx, cluster, instances, set,
 			clusterConfigMap, clusterReplicationSecret,
 			rootCA, clusterPodService, instanceServiceAccount,
 			patroniLeaderService, primaryCertificate,
-			findAvailableInstanceNames(set, instances, clusterVolumes),
+			findAvailableInstanceNames(*set, instances, clusterVolumes),
 			numInstancePods, clusterVolumes)
+
+		if err == nil {
+			// TODO (jmckulk): reconcilePodDisruptionBudget
```
andrewlecuyer, line 553: :wrench: Looks like this `TODO` can be removed?

internal/controller/postgrescluster/instance.go
```diff
@@ -559,6 +566,12 @@ func (r *Reconciler) reconcileInstanceSets(
 		return err
 	}
 
+	// Cleanup Instance Set resources that are no longer needed
+	err = r.deleteOrphanedPodDisruptionBudgets(ctx, cluster)
```
andrewlecuyer, line 570: :pick: "Orphaned" in a Kube context makes me think resources that don't have an owner, which doesn't really apply here IIUC ... maybe can just be something like `cleanupPodDisruptionBudgets()`?

internal/naming/names.go
```diff
@@ -279,6 +280,14 @@ func InstanceCertificates(instance metav1.Object) metav1.ObjectMeta {
 	}
 }
 
+func InstanceSetPodDisruptionBudget(cluster *v1beta1.PostgresCluster,
```
andrewlecuyer, line 283: :wrench: Comment needed here.

internal/naming/selectors.go
```diff
@@ -103,6 +115,15 @@ func ClusterPatronis(cluster *v1beta1.PostgresCluster) metav1.LabelSelector {
 	}
 }
 
+func ClusterPGBouncerSelector(cluster *v1beta1.PostgresCluster) metav1.LabelSelector {
```
andrewlecuyer, line 118: :wrench: Function comment needed here.

andrewlecuyer (commented):
:wrench: I'm not seeing a call to `Owns(&policyv1beta1.PodDisruptionBudget{})` in `SetupWithManager()`.  Is that intentional?  I assume we want to reconcile when the PDB's we create change?

internal/controller/postgrescluster/pod_disruption_budget.go
```diff
@@ -0,0 +1,99 @@
...
+	pdb := &policyv1beta1.PodDisruptionBudget{
+		ObjectMeta: meta,
+		Spec: policyv1beta1.PodDisruptionBudgetSpec{
+			// We can't pass the IntOrString value directly from our spec
+			// into the PDB spec because of restrictions placed on Arbitrary
+			// Controllers.
+			// - only .spec.minAvailable can be used, not .spec.maxUnavailable.
+			// - only an integer value can be used with .spec.minAvailable, not
+			// a percentage.
+			// https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors
```
andrewlecuyer, line 49: :thinking: The "Arbitrary Controllers and Selectors" section in the docs is throwing me off a bit.  Specifically, I'm not sure if this applies to us considering our controller does not actually _own_ the Pods, but rather PGO owns the StatefulSet or Deployment that owns the Pods (or the resource that then creates the Pods, i.e. a ReplicaSet in the case of a Deployment).

And looking at [other parts of the PDB docs](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets) I see that it actually discovers the owning workload for the Pod, which wouldn't be us (but rather a StatefulSet or ReplicaSet):

> The control plane discovers the owning workload resource by examining the .metadata.ownerReferences of the Pod.

Additionally, I did an experiment (this was with a local Kube 1.22 cluster) where I tried violating both of the restrictions described for the PDB.  Specifically, I tried setting `minAvailable` as a percentage and also tried setting `maxUnavialable` (i.e. instead of `minAvailable`), and the PDB appeared to work.

For instance, here is the PDB with `maxUnavailable`:

```bash
$ kubectl get pdb
NAME        MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
instance1   N/A             1                 0                     126m
```

And when I go to drain I only see one of my two PG replicas terminate, and the PDB is respected:

```bash
error when evicting pods/"hippo-instance1-pk26-0" -n "postgres-operator" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
```

Maybe the limitations don't apply to us since we do not directly own the Pods?  Though if anything I am missing here, please let me know.

cbandy: @andrewlecuyer I think [this is the code](https://github.com/kubernetes/kubernetes/blob/v1.22.0/pkg/controller/disruption/disruption.go#L665) that finds the specified/expected replicas/count, and you're right it is finding each controller for each pod and doing a sum.

This is the [line of the docs](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors) that had me concerned (emphasis mine):
> You can skip this section if you only use PDBs with the built-in application controllers (Deployment, ReplicationController, ReplicaSet, and StatefulSet), with **the PDB selector matching the controller's selector**.

andrewlecuyer: Yup, completely agree - that line gave me pause too!

And thanks for pulling that code, great to verify what they're doing there down at that level.

And after seeing that code, I get the feeling the doc isn't 100% accurate ... i.e. thinking it really should say something like:

> ...with the PDB selector selecting Pods that are also selected by one or more built-in application controllers

(or something like that ... feels like to much use of "select" in there :laughing:)

jmckulk: @andrewlecuyer Were you using the betav1 version of policy/PDBs? I wonder if the behavior would be different. They note that there is a difference in how selectors work (specifically the empty selector).

I'm not fully tracking what change is necessary here. Should I update the comment or are we talking about a code change?

internal/controller/postgrescluster/instance_test.go
```diff
@@ -1573,3 +1576,178 @@ func TestFindAvailableInstanceNames(t *testing.T) {
...
+		cluster.Namespace = ns.Name
+		spec := &cluster.Spec.InstanceSets[0]
+		spec.MinAvailable = &intstr.IntOrString{IntVal: int32(1)}
+		assert.NilError(t, r.reconcileInstanceSetPodDisruptionBudget(ctx, cluster, spec))
+		assert.Assert(t, foundPDB(t, cluster, spec, 1))
+
+		t.Run("delete pdb", func(t *testing.T) {
+			spec.MinAvailable = &intstr.IntOrString{IntVal: int32(0)}
+			err := r.reconcileInstanceSetPodDisruptionBudget(ctx, cluster, spec)
+			if apierrors.IsConflict(err) {
```
benjaminjb, line 1649: I wouldn't say no to a comment on this pattern as to why we need it or what errors you were seeing (presumably in existing envs?).

internal/controller/postgrescluster/pgbouncer_test.go
```diff
@@ -365,3 +371,87 @@ func TestReconcilePGBouncerDeployment(t *testing.T) {
 	})
 
 }
+
+func TestReconcilePGBouncerDisruptionBudget(t *testing.T) {
+	env, cc, config := setupTestEnv(t, ControllerName)
+	t.Cleanup(func() { teardownTestEnv(t, env) })
+
+	r := &Reconciler{}
+	ctx, cancel := setupManager(t, config, func(mgr manager.Manager) {
```
cbandy, line 380: 🔧 This test is calling reconcile methods directly, so there should be no Manager. (Manager watches for PostgresCluster objects and calls Reconcile.)

internal/controller/postgrescluster/pgbouncer_test.go
```diff
@@ -365,3 +371,87 @@ func TestReconcilePGBouncerDeployment(t *testing.T) {
...
+		cluster.Namespace = ns.Name
+		cluster.Spec.Proxy.PGBouncer.Replicas = initialize.Int32(1)
+		cluster.Spec.Proxy.PGBouncer.MinAvailable = &intstr.IntOrString{IntVal: int32(0)}
+		assert.NilError(t, r.reconcilePGBouncerPodDisruptionBudget(ctx, cluster))
+		assert.Assert(t, !foundPDB(t, cluster, 0))
+	})
+
+	t.Run("created", func(t *testing.T) {
+		cluster := testCluster()
+		cluster.UID = types.UID("hippouid")
```
cbandy, line 440: 🤔 This UID is for the controller reference, right? I think this test races with the garbage collector in a full Kubernetes cluster. Better to call `cc.Create(ctx, cluster)` without a Manager.

internal/controller/postgrescluster/pgbouncer_test.go
```diff
@@ -365,3 +371,87 @@ func TestReconcilePGBouncerDeployment(t *testing.T) {
...
+	) bool {
+		pdb, err := r.generatePodDisruptionBudget(
+			cluster,
+			naming.ClusterPGBouncer(cluster),
+			minAvailable,
+			naming.ClusterPGBouncerSelector(cluster),
+		)
+		assert.NilError(t, err)
+		got := &policyv1beta1.PodDisruptionBudget{}
+		err = r.Client.Get(ctx, client.ObjectKeyFromObject(pdb), got)
```
cbandy, line 403: ⛏️ `pdb` is used only for its `ObjectKey`, but we're also building/passing the name on L397. Can the call to `generatePodDisruptionBudget` be replaced with the following?

```suggestion
		err = r.Client.Get(ctx, naming.AsObjectKey(naming.ClusterPGBouncer(cluster)), got)
```

internal/controller/postgrescluster/instance_test.go
```diff
@@ -1573,3 +1576,178 @@ func TestFindAvailableInstanceNames(t *testing.T) {
...
+		})
+	})
+}
+
+func TestDeleteOrphanedPodDisruptionBudgets(t *testing.T) {
+	env, cc, config := setupTestEnv(t, ControllerName)
+	t.Cleanup(func() { teardownTestEnv(t, env) })
+
+	r := &Reconciler{}
+	ctx, cancel := setupManager(t, config, func(mgr manager.Manager) {
```
cbandy, line 1663: 🔧 This test is calling reconcile methods directly, so there should be no Manager.

internal/controller/postgrescluster/instance_test.go
```diff
@@ -1573,3 +1576,178 @@ func TestFindAvailableInstanceNames(t *testing.T) {
 		})
 	}
 }
+
+func TestReconcileInstanceSetPodDisruptionBudget(t *testing.T) {
+	env, cc, config := setupTestEnv(t, ControllerName)
+	t.Cleanup(func() { teardownTestEnv(t, env) })
+
+	r := &Reconciler{}
+	ctx, cancel := setupManager(t, config, func(mgr manager.Manager) {
```
cbandy, line 1585: 🔧 This test is calling reconcile methods directly, so there should be no Manager.

internal/controller/postgrescluster/instance_test.go
```diff
@@ -1573,3 +1576,178 @@ func TestFindAvailableInstanceNames(t *testing.T) {
...
+	) bool {
+		pdb, err := r.generatePodDisruptionBudget(
+			cluster,
+			naming.InstanceSetPodDisruptionBudget(cluster, spec),
+			minAvailable,
+			naming.ClusterInstanceSet(cluster.Name, spec.Name),
+		)
+		assert.NilError(t, err)
+		got := &policyv1beta1.PodDisruptionBudget{}
+		err = r.Client.Get(ctx, client.ObjectKeyFromObject(pdb), got)
```
cbandy, line 1609: ⛏️ Similar here; can we omit the call to `generatePodDisruptionBudget`?

```suggestion
		err = r.Client.Get(ctx, naming.AsObjectKey(naming.InstanceSetPodDisruptionBudget(cluster, spec)), got)
```

internal/controller/postgrescluster/instance_test.go
```diff
@@ -1573,3 +1576,178 @@ func TestFindAvailableInstanceNames(t *testing.T) {
...
+	}
+
+	t.Run("pdbs not found", func(t *testing.T) {
+		cluster := testCluster()
+		assert.NilError(t, r.cleanupPodDisruptionBudgets(ctx, cluster))
+	})
+
+	t.Run("pdbs found", func(t *testing.T) {
+		cluster := testCluster()
+		cluster.UID = types.UID("hippouid")
```
cbandy, line 1721: ⛏️ `cluster` is being created on L1726; no need for a custom UID here.

config/rbac/cluster/role.yaml
```diff
@@ -88,6 +88,16 @@ rules:
   - list
   - patch
   - watch
+- apiGroups:
```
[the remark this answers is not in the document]
jmckulk: Yeah I would think so, I can check and change if necessary.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbouncer_types.go
```diff
@@ -106,6 +107,10 @@ type PGBouncerPodSpec struct {
 	// +kubebuilder:validation:Minimum=0
 	Replicas *int32 `json:"replicas,omitempty"`
 
+	// Minimum number of replicas that should be available at a time.
+	// +optional
+	MinAvailable *intstr.IntOrString `json:"minAvailable,omitempty"`
```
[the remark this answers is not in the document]
jmckulk: Since this wold require a tool updated maybe we punt?

We have decided on using IntOrString now and passing through that value without validation.

internal/naming/names.go
```diff
@@ -279,6 +280,16 @@ func InstanceCertificates(instance metav1.Object) metav1.ObjectMeta {
 	}
 }
 
+// InstanceSetPodDisruptionBudget returns the ObjectMeta necessary to lookup the
+// Pod Disruption Budget associated with a cluster and instance set.
+func InstanceSetPodDisruptionBudget(cluster *v1beta1.PostgresCluster,
+	set *v1beta1.PostgresInstanceSetSpec) metav1.ObjectMeta {
+	return metav1.ObjectMeta{
+		Name:      set.Name,
```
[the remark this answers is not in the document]
jmckulk: Going with `<cluster-name>-set-<instance-name>`

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbouncer_types.go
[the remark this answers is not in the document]
cbandy: Yep, punt on validation.

cbandy (commented):
I noticed we don't have any RBAC for PDB `get`. Everything is working (I think because our controller-runtime client does all its GETs from its cache, populated by watches), but we should probably request that permission anyway.

internal/controller/postgrescluster/instance.go
```diff
@@ -1759,3 +1804,51 @@ func (r *Reconciler) prepareForUpgrade(ctx context.Context,
...
+		spec.Metadata.GetAnnotationsOrNil())
+
+	selector := naming.ClusterInstanceSet(cluster.Name, spec.Name)
+	pdb, err := r.generatePodDisruptionBudget(cluster, meta, minAvailable, selector)
+
+	// If 'minAvailable' is set to '0', we will not reconcile the PDB. If one
+	// already exists, we will remove it.
+	if err == nil &&
+		(minAvailable.Type == 0 && minAvailable.IntVal == 0) ||
+		(minAvailable.Type == 1 && (minAvailable.StrVal == "0%" || minAvailable.StrVal == "00%")) {
```
benjaminjb, line 1842: Does an "" register as anything with intOrString?

internal/controller/postgrescluster/instance_test.go
```diff
@@ -2271,3 +2274,201 @@ func fakeUpgradeCluster(clusterName, namespace, clusterUID string) *v1beta1.Post
...
+		}
+
+		assert.NilError(t, r.Client.Create(ctx, cluster))
+		t.Cleanup(func() { assert.Check(t, r.Client.Delete(ctx, cluster)) })
+
+		assert.NilError(t, r.reconcileInstanceSetPodDisruptionBudget(ctx, cluster, spec))
+		assert.Assert(t, foundPDB(cluster, spec))
+
+		t.Run("delete pdb", func(t *testing.T) {
+			spec.MinAvailable = &intstr.IntOrString{IntVal: int32(0)}
```
benjaminjb, line 2367: May be silly of me, but I'd like to see a string value of "0%" as part of the delete here -- I mean, it was created with a string, why not delete it with a string too?

internal/controller/postgrescluster/pgbouncer_test.go
```diff
@@ -365,3 +368,113 @@ func TestReconcilePGBouncerDeployment(t *testing.T) {
...
+		}
+
+		assert.NilError(t, r.Client.Create(ctx, cluster))
+		t.Cleanup(func() { assert.Check(t, r.Client.Delete(ctx, cluster)) })
+
+		assert.NilError(t, r.reconcilePGBouncerPodDisruptionBudget(ctx, cluster))
+		assert.Assert(t, foundPDB(cluster))
+
+		t.Run("deleted", func(t *testing.T) {
+			cluster.Spec.Proxy.PGBouncer.MinAvailable = &intstr.IntOrString{IntVal: int32(0)}
```
benjaminjb, line 467: Ditto re: str val for cleaning up PDB

internal/controller/postgrescluster/pod_disruption_budget.go
jmckulk: Updated to pass the `IntOrString` value through from the spec to the PDB

internal/controller/postgrescluster/pod_disruption_budget_test.go
```diff
@@ -0,0 +1,125 @@
...
+			cluster,
+			meta,
+			minAvailable,
+			selector,
+		)
+		assert.NilError(t, err)
+		assert.Equal(t, pdb.Name, cluster.Name)
+		assert.Equal(t, pdb.Namespace, cluster.Namespace)
+		assert.Assert(t, labels.Set(pdb.Labels).Has("label-key"))
+		assert.Assert(t, labels.Set(pdb.Annotations).Has("anno-key"))
```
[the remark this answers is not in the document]
jmckulk: I don't think we can. Object meta from the generated object includes the controller reference. I would have to update the test `meta` to also have that.

internal/controller/postgrescluster/instance.go
jmckulk: `""` is an invalid value when the spec is created. We aren't doing validation on our spec so it is possible to pass it through and end up with an error when you try to parse the  `IntOrString` or apply the PDB spec. So far we have been punting on validation but I do need to add in a check for the error. I can also make a note to valid in the future/with a webhook.

cbandy (approved):

benjaminjb (approved):

PR #2837, opened 2021-11-08
Validate instance set names

cbandy:
**Checklist:**

 
**Type of Changes:**

 
`spec.instances.name` is optional but only when there's a single instance set. Two or more instance sets without a name are rejected: `spec.instances[1]: Duplicate value: map[string]interface {}{"name":""}`.

Call out the above limitation in the field description. Values like `UPPER` and `lower.with.dots` used to be accepted but would not build a working cluster. Reject them instead.

**Other Information**:

Issue: [sc-13112]

### DIFF
--- a/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
+++ b/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
@@ -91,6 +91,8 @@ type PostgresClusterSpec struct {
 	// +optional
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
 
+	// Specifies one or more sets of PostgreSQL pods that replicate data for
+	// this cluster.
 	// +listType=map
 	// +listMapKey=name
 	// +kubebuilder:validation:MinItems=1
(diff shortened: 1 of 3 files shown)


andrewlecuyer (approved):

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -374,8 +376,15 @@ type PostgresInstanceSetSpec struct {
 	// +optional
 	Metadata *Metadata `json:"metadata,omitempty"`
 
+	// This value goes into the name of an appsv1.StatefulSet and the hostname
+	// of a corev1.Pod. The pattern below is IsDNS1123Label wrapped in "()?" to
+	// accommodate the empty default.
+
+	// Name that associates this set of PostgreSQL pods. This field is optional
+	// when only one instance set is defined.
```
benjaminjb, line 384: Maybe we should be explicit about it being unique?

cbandy: I looked for similar language upstream, but not many things use the `listMap` marker.

The official docs [add an italicized line](https://github.com/kubernetes-sigs/reference-docs/blob/69e29015936e49b5b5f445097f043d64ba70feda/gen-resourcesdocs/pkg/outputs/kwebsite/section.go#L93) above the description, "unique values on key(s) `name` will be kept during a merge." 😕 That doesn't explain very much at all, in my opinion. Ex:

- **conditions** of [PodDisruptionBudgetStatus](https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/pod-disruption-budget-v1/#PodDisruptionBudgetStatus)
- **ports** of [Container](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#ports)

Neither of these mention anything about uniqueness beyond that line.

The other prominent example I could think of was `Pod.spec.containers`. [That field](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#containers) has a line about the patch strategy but nothing in the description. The [`Pod.spec.containers.name` field](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#Container) says: "Each container in a pod must have a unique name." That's a little different though, now that I think about it; every container name (init, ephemeral, otherwise) is [checked against every other container name](https://github.com/kubernetes/kubernetes/blob/v1.22.0/pkg/apis/core/validation/validation.go#L3419-L3421).

🤷🏻 Anyway, I went with the last one in 23da44b2b967247853bcaa168a8494f1549e55b5: "Each instance set in a cluster must have a unique name."

benjaminjb (approved):

PR #2841, opened 2021-11-09
Remove requirement for external WAL volume in v4 => v5

jkatz:
pgBackRest 2.36 eliminates this requirement.


benjaminjb (approved):
Do we want to say something about "no need for external WAL if you use pgBackRest >=2.36"?

jkatz:
> Do we want to say something about "no need for external WAL if you use pgBackRest >=2.36"?

Nope. If it comes up, we just recommend upgrading to v5.0.4 or newer.

PR #2851, opened 2021-11-13
Send a simple customer data payload

benjaminjb:
**Checklist:**

**Type of Changes:**

The upgrade check just sends PGO version as a header.

This adds some material to the upgrade check header, including time, number of clusters managed, and an anonymous ID.

**Other information**:
Issue [sc-12903]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ deploy:
 	$(PGO_KUBE_CLIENT) apply -k ./config/default
 
 # Deploy the PostgreSQL Operator locally
-deploy-dev: build-postgres-operator
+deploy-dev: build-postgres-operator createnamespaces
 	$(PGO_KUBE_CLIENT) apply -k ./config/dev
 	hack/create-kubeconfig.sh postgres-operator pgo
 	env \
@@ -189,18 +189,18 @@ pgo-base-docker: pgo-base-build
 #======== Utility =======
 .PHONY: check
 check:
-	$(GO_TEST) -cover ./...
+	PGO_NAMESPACE="postgres-operator" $(GO_TEST) -cover ./...
 
 # - KUBEBUILDER_ATTACH_CONTROL_PLANE_OUTPUT=true
 .PHONY: check-envtest
 check-envtest: hack/tools/envtest
-	KUBEBUILDER_ASSETS="$(CURDIR)/$^/bin" $(GO_TEST) -count=1 -cover -tags=envtest ./...
+	KUBEBUILDER_ASSETS="$(CURDIR)/$^/bin" PGO_NAMESPACE="postgres-operator" $(GO_TEST) -count=1 -cover -tags=envtest ./...
 
 # - PGO_TEST_TIMEOUT_SCALE=1
 .PHONY: check-envtest-existing
-check-envtest-existing:
+check-envtest-existing: createnamespaces
 	${PGO_KUBE_CLIENT} apply -k ./config/dev
-	USE_EXISTING_CLUSTER=true $(GO_TEST) -count=1 -cover -p=1 -tags=envtest ./...
+	USE_EXISTING_CLUSTER=true PGO_NAMESPACE="postgres-operator" $(GO_TEST) -count=1 -cover -p=1 -tags=envtest ./...
 	${PGO_KUBE_CLIENT} delete -k ./config/dev
 
 # Expects operator to be running
--- a/cmd/postgres-operator/main.go
+++ b/cmd/postgres-operator/main.go
@@ -29,7 +29,7 @@ import (
 	"github.com/crunchydata/postgres-operator/internal/controller/postgrescluster"
 	"github.com/crunchydata/postgres-operator/internal/controller/runtime"
 	"github.com/crunchydata/postgres-operator/internal/logging"
-	"github.com/crunchydata/postgres-operator/internal/util"
+	"github.com/crunchydata/postgres-operator/internal/upgradecheck"
 )
 
 var versionString string
@@ -88,7 +88,10 @@ func main() {
 	done := make(chan bool, 1)
 	if upgradeCheckingEnabled {
 		log.Info("upgrade checking enabled")
-		go util.CheckForUpgradesScheduler(versionString, done)
+		go upgradecheck.CheckForUpgradesScheduler(done, versionString,
+			mgr.GetClient(), mgr.GetConfig(), isOpenshift(ctx, mgr.GetConfig()),
+			mgr.GetCache(),
+		)
 	} else {
 		log.Info("upgrade checking disabled")
 	}
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/evanphx/json-patch v4.9.0+incompatible
 	github.com/go-logr/logr v0.4.0
 	github.com/google/go-cmp v0.5.4
+	github.com/google/uuid v1.1.2
 	github.com/onsi/ginkgo v1.14.1
 	github.com/onsi/gomega v1.11.0
 	github.com/pkg/errors v0.9.1
@@ -39,7 +40,6 @@ require (
 	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e // indirect
 	github.com/golang/protobuf v1.4.3 // indirect
 	github.com/google/gofuzz v1.1.0 // indirect
-	github.com/google/uuid v1.1.2 // indirect
 	github.com/googleapis/gnostic v0.5.1 // indirect
 	github.com/hashicorp/golang-lru v0.5.4 // indirect
 	github.com/imdario/mergo v0.3.10 // indirect
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -91,3 +91,10 @@ func PostgresContainerImage(cluster *v1beta1.PostgresCluster) string {
 
 	return defaultFromEnv(image, key)
 }
+
+// PGONamespace returns the namespace where the PGO is running,
+// based on the env var from the DownwardAPI
+// If no env var is found, returns ""
+func PGONamespace() string {
+	return os.Getenv("PGO_NAMESPACE")
+}
--- a/internal/naming/names.go
+++ b/internal/naming/names.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/rand"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/crunchydata/postgres-operator/internal/config"
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
@@ -476,5 +477,13 @@ func MovePGBackRestRepoDirJob(cluster *v1beta1.PostgresCluster) metav1.ObjectMet
 	return metav1.ObjectMeta{
 		Namespace: cluster.GetNamespace(),
 		Name:      cluster.Name + "-move-pgbackrest-repo-dir",
+	}
+}
+
+// UpgradeCheckConfigMap returns the ObjectMeta for the PGO ConfigMap
+func UpgradeCheckConfigMap() metav1.ObjectMeta {
+	return metav1.ObjectMeta{
+		Namespace: config.PGONamespace(),
+		Name:      "pgo-upgrade-check",
 	}
 }
--- a/internal/upgradecheck/header.go
+++ b/internal/upgradecheck/header.go
@@ -1,0 +1,196 @@
+package upgradecheck
+
+/*
+ Copyright 2017 - 2021 Crunchy Data Solutions, Inc.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+
+	"github.com/go-logr/logr"
+	googleuuid "github.com/google/uuid"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/crunchydata/postgres-operator/internal/naming"
+	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
+)
+
+const (
+	clientHeader = "X-Crunchy-Client-Metadata"
+)
+
+var (
+	// Using apimachinery's UUID package, so our deployment UUID will be a string
+	deploymentID string
+)
+
+// Extensible struct for client upgrade data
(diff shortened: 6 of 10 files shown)


brandur:
@benjaminjb Hey! Just wanted to let you know that the server-side is more or less ready for consumption here.

Here are the endpoints you can hit:

* Production: https://operator-maestro.crunchydata.com/pgo-versions
* Staging: https://operator-maestro-staging.herokuapp.com/pgo-versions

I'd recommend checking the response structure of the list with `curl -i https://operator-maestro.crunchydata.com/pgo-versions` and adapting your code to it. For now, the response is extremely simple, containing only a version tag.

I also have a couple minor asks for some tweaks to the format here:

* Can you rename the header from `X-Client-Upgrade-State` to `X-Crunchy-Client-Metadata`? I think the naming is a little more apt given the payload and it's a good idea to include a vendor name in there.
* In the header payload, could we rename `id` to `deployment_id`? It's better to be a little more specific about what type of identifier this is in case we want to add other IDs in the future.

internal/upgradecheck/header.go
```diff
@@ -0,0 +1,174 @@
...
+			} else {
+				// If we encounter any error except NotFound, just go ahead with the generated ID
+				// and skip the ConfigMap
+				return deploymentID, err
+			}
+		}
+	}
+
+	if err == nil {
+		configMapUUID, err = uuid.Parse(cm.Data["deployment_id"])
```
andrewlecuyer, line 139: :wrench: One issue I'm seeing here is that if the `deployment_id` is deleted from the ConifgMap (but the ConfigMap itself still exists), we'll never update the ConfigMap with what we have in memory.  Can we use a Server-Side Apply in this function instead of using `Create()`?

I do realize our `apply()` function as used for SSA is in the `postgrescluster` package.  Maybe we move it elsewhere so that it can also be used here as well?

I also wouldn't be against just creating `apply()` and `patch()` funcitons in this package that serve the specific use case of the upgrade checker (e.g. this thing only deals with ConfigMap's, so no need to worry about the workarounds for StatefulSets, Jobs and Services in the current `apply()` function).

The benefit of a SSA is that we'll always being applying our intent, which means we'll always set the `deployment_id` if its missing.  And if we need to add additional fields to the ConfigMap in the future, we can simply update our intent, and those changes will be patched (and overall this aligns with our use of SSA for creating/patching resources throughout other parts of PGO v5).

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/evanphx/json-patch v4.9.0+incompatible
 	github.com/go-logr/logr v0.4.0
 	github.com/google/go-cmp v0.5.4
+	github.com/google/uuid v1.1.2
```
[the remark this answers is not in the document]
benjaminjb: No, it's a relatively easy swap out -- actually since apimachinery's uuid is mostly focused on types.UID (which is itself a string type), using that and just making sure everything is a string makes some things cleaner. (I didn't see us use any particular UUID package, though I feel like this has come up before...

internal/upgradecheck/header.go
```diff
@@ -0,0 +1,174 @@
...
+	PGOVersion       string        `json:"pgo_version"`
+	IsOpenShift      bool          `json:"is_open_shift"`
+}
+
+// generateHeader aggregates data and returns a struct of that data
+// If any errors are encountered, it logs those errors and uses the default values
+func generateHeader(log logr.Logger, version string, crclient crclient.Client,
+	ctx context.Context, cfg *rest.Config, isOpenShift bool) *clientUpgradeData {
+
+	var ensureDeploymentIDErr, getManagedClustersErr, getServerVersionErr error
```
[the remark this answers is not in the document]
benjaminjb: So this was an interesting case: I decided it made sense to NOT stop on error but just use a default/nil value; but I also wanted to get the errors from each individual func in order to log an apt error message; so I figured I could use individual errs for each func (as I did here) or to set the err back to nil after each time it was set to a non-nil value. 

Do you think that second option would be cleaner?

Or stepping back: what do you think about not stopping?

internal/upgradecheck/header.go
```diff
@@ -0,0 +1,174 @@
...
+		if err = crclient.Get(ctx, naming.AsObjectKey(
+			naming.UpgradeCheckConfigMap(),
+		), cm); err != nil {
+			if apierrors.IsNotFound(err) {
+				cm = &corev1.ConfigMap{ObjectMeta: naming.UpgradeCheckConfigMap()}
+				cm.SetGroupVersionKind(corev1.SchemeGroupVersion.WithKind("ConfigMap"))
+				cm.Data = map[string]string{
+					"deployment_id": deploymentID.String(),
+				}
+				err = errors.WithStack(crclient.Create(ctx, cm))
```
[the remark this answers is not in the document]
benjaminjb: Yes, PGO creates only one configmap, and never deletes OR updates it. (i.e., if it's there and has a value, use that value, don't overwrite that value ever).

internal/upgradecheck/header.go
```diff
@@ -0,0 +1,174 @@
...
+				if err != nil {
+					return deploymentID, err
+				}
+
+			} else {
+				// If we encounter any error except NotFound, just go ahead with the generated ID
+				// and skip the ConfigMap
+				return deploymentID, err
```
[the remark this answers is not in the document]
benjaminjb: My thinking on this was that `ensureDeploymentID` always returns _some_ ID -- the worst case scenario would be that it just creates an ID and saves that in memory -- so the error becomes a message for the calling function. 

I figured it made sense to leave it up to the caller to decide "I cannot trust this ID because of the error" or shrug and just use the ID. (Which is how `generateHeader` uses `ensureDeploymentID` -- log the error and use whatever ID it gives.)

(That said, I don't know who would ever use this `ensureDeploymentID` func, and it would need to be changed a bit since the configmap name is hard-coded in; so there's definitely an argument to be made that I should just log the error here and return only the ID.)

internal/upgradecheck/header.go
benjaminjb: That makes sense to me, with the caveat that we don't want to overwrite our deployment_id unless it's... not a valid UUID? Like:
CM doesn't exist -- create
CM exists, has deployment id UUID -- leave
CM exists, deployment id doesn't exist or is malformed -- update

benjaminjb: And for checking that some string is a valid UUID, I'd probably use that google/uuid package

Makefile
```diff
@@ -107,6 +107,7 @@ deploy:
 
 # Deploy the PostgreSQL Operator locally
 deploy-dev: build-postgres-operator
+	$(PGO_KUBE_CLIENT) create ns postgres-operator || true
```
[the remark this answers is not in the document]
benjaminjb: Oh, I'll just add `createnamespaces` to these targets (deploy-dev, check-envtest-existing -- the two targets where the absence of a postgres-operator namespace keep tripping me up)

internal/upgradecheck/header.go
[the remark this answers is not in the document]
andrewlecuyer: After thinking about this a bit, I am wondering if stopping is the best approach here.  We already have retry logic built in, correct?  Or does that only retry if we can't successfully contact the version endpoint?

It does also seem as though stopping is the best way to ensure we get clean data (e.g. consistent ID's, accurate cluster counts, etc.).

Lets take a few mins to discuss following the standup this morning.

internal/upgradecheck/header.go
andrewlecuyer: Ahh yes!  I do see we would need the Google package to check the UUID.

Also, I concur with the following :+1: 

> CM doesn't exist -- create
> CM exists, has deployment id UUID -- leave
> CM exists, deployment id doesn't exist or is malformed -- update

internal/upgradecheck/header.go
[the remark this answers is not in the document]
andrewlecuyer: Understood - lets hash out a bit following standup, considering the thread above regarding stopping on errors.

As it stands, I personally like the idea of logging a warning/info message and not returning an error if a valid deployment ID is returned. In other words, if the function is returning some valid UUID we can/will use, then I feel as though there shouldn't be an error (I'm also not sure how the consumer would be able to tell if the UUID should be used simply based on the error message ... e.g. if I get a UUID but also some transient Kube error, what does that mean?  I.e. can I use that to determine if its a valid UUID?).

internal/config/config.go
```diff
@@ -91,3 +100,8 @@ func PostgresContainerImage(cluster *v1beta1.PostgresCluster) string {
 
 	return defaultFromEnv(image, key)
 }
+
+// PGO_Namespace returns the namespace that PGO is running in.
+func PGO_Namespace() string {
+	return defaultIfNotEnv("postgres-operator", "PGO_NAMESPACE")
```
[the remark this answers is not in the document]
benjaminjb: Updating this according to our chat yesterday and updated all testing make targets

internal/upgradecheck/header.go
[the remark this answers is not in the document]
benjaminjb: Following discussion, we decided to stick with a standard Go pattern of logging the error rather than returning an error that is then just logged by the calling function.

cmd/postgres-operator/main.go
```diff
@@ -88,7 +88,8 @@ func main() {
 	done := make(chan bool, 1)
 	if upgradeCheckingEnabled {
 		log.Info("upgrade checking enabled")
-		go util.CheckForUpgradesScheduler(versionString, done)
+		go upgradecheck.CheckForUpgradesScheduler(done, versionString,
+			mgr.GetClient(), mgr.GetConfig(), isOpenshift(ctx, mgr.GetConfig()))
```
andrewlecuyer, line 92: :question: I have noticed that if there an attempt to use the client before the manager has started (i.e. before `mgr.Start()` is called a few lines down), and error will be thrown because the cache's for the manager/client have not yet synced.

Have you run into anything like this with your testing?  It seems like this would only _potentially_ be a concern for the first attempt to hit the endpoint when PGO starts.

benjaminjb: I have not seen that sort of error in testing this (either running the tests or running the program), but that might be just luck so far.

Actually, I just put in a sleep before the `mgr.Start` to try to force an error and that "successfully" errored; so I put in a WaitForCacheSync for-loop and that solved the error.

(I'm also beginning to think I should just pass the mgr in to retrieve the context, client, and cache, but that kind of obscures what we need, so I'll leave it for now, with some 🌱 for the future.)

internal/upgradecheck/header.go
```diff
@@ -0,0 +1,200 @@
...
+		return retrievedCM
+	// If the ConfigMap is missing the deploymentID, we'll update it
+	case retrievedCM.Data["deployment_id"] == "":
+		break
+	// If the ConfigMap exists and has a deploymentID, check it for correctness
+	default:
+		_, parseErr := googleuuid.Parse(retrievedCM.Data["deployment_id"])
+		// No error -- the ConfigMap has a valid deploymentID, so return the ConfigMap as is
+		if parseErr == nil {
+			return retrievedCM
```
andrewlecuyer, line 120: :wrench: Since this function is now using SSA, I would expect the intent to be applied each time this logic is run (instead of returning the current ConfigMap as is done here).

Therefore, in the `default` case here where the ConfigMap exists and has valid ID, I expect that we would simply proceed on to the logic below to generate the intent, specifically using the existing ID we parse here as `deployment_id` in the intent.  That means we will always be applying our full intent each time this is run (which most of the time will not change, but I could see us eventually adding other labels, annotation and/or data to the intent).

internal/naming/names.go
```diff
@@ -478,3 +479,11 @@ func MovePGBackRestRepoDirJob(cluster *v1beta1.PostgresCluster) metav1.ObjectMet
 		Name:      cluster.Name + "-move-pgbackrest-repo-dir",
 	}
 }
+
+// UpgradeCheckConfigMap returns the ObjectMeta for the PGO ConfigMap
+func UpgradeCheckConfigMap() metav1.ObjectMeta {
+	return metav1.ObjectMeta{
+		Namespace: config.PGONamespace(),
+		Name:      "pgo-upgrade-check-configmap",
```
andrewlecuyer, line 487: :wrench: Can we drop `configmap` from the name? (feels like a bit of stutter otherwise).

:thinking: Just wondering aloud if we should name more generically - e.g. just `pgo` - as the "PGO ConfigMap" (which maybe could eventually be used for other things - though nothing else comes to mind at the moment (which is why I def don't have a problem sticking with the current naming).

cmd/postgres-operator/main.go
andrewlecuyer: Sounds great, definitely on board with using `WaitForCacheSync()` here!

andrewlecuyer (changes requested):

internal/upgradecheck/http.go
```diff
@@ -163,7 +159,15 @@ func CheckForUpgradesScheduler(versionString string, channel chan bool) {
 		}
 	}()
 
-	info, err := checkForUpgrades(versionString, backoff)
+	// Make sure cache is synced before using the client
+	for {
```
andrewlecuyer, line 163: :question: Is the `for` loop needed here?  From what I can tell `WaitForCacheSync()` will simply block and not return until the cache finally syncs, but just let me know if there is anything I'm missing.

benjaminjb: As far as I can tell, yeah, `WaitForCacheSync` is well named -- I can remove the loop, it makes it cleaner without, I just put it in because it felt a little odd to run a function without checking the return.

Actually, the more I look at this without any for loop, I wonder about the off chance that syncing will encounter an error: if we're in a loop, then we just keep erroring until we don't. I almost feel like this would be another good place for a retry with an exponential backoff. (The other solution is to error and exit the process; an operator in that condition would never check for an upgrade.)

My feeling right now is... maybe remove the for-loop, and create a ticket for further enhancements?

andrewlecuyer: My thought is that if `WaitForCacheSync()` returns `false`, we would simply log an error and then _not_ proceed with running the "check for upgrade" logic.

My thinking here is that the `manager` will also fail to start in the event the caches don't sync successfully, so we're probably looking at a failure to start PGO as a whole anyway (and therefore likely a container restart).

internal/upgradecheck/http_test.go
```diff
@@ -144,22 +150,28 @@ func TestCheckForUpgrades(t *testing.T) {
...
+			fakeClient, ctx, cfg, false)
 		assert.Equal(t, counter, 2)
 		assert.NilError(t, err)
-		assert.Equal(t, res, `{"display_name":"PGO 5.0.3","major_version":5}`)
+		assert.Equal(t, res, `{"pgo_versions":[{"tag":"v5.0.3"},{"tag":"v5.0.2"},{"tag":"v5.0.1"},{"tag":"v5.0.0"}]}`)
 	})
 }
 
 func TestCheckForUpgradesScheduler(t *testing.T) {
+	fakeClient := setupFakeClientWithPGOScheme(t, false)
```
andrewlecuyer, line 170: :thinking: Can we use `envtest` here instead of `fake`?  It is specifically the `CacheWithWait` interface that I'm thinking about, i.e. if we use `envtest` and start the manager, can we just then grab/use the client as we do in our other tests?  With this approach the cache should always sync before we grab the client for use in our tests, so wondering if that would allow us to avoid the need to create the interface simply to facilitate mocking.

Maybe this is something we can consider for a future commit (switch to `envtest`)?  I definitely don't want to blow up the current tests at this point or anything (since there are a lot of great tests in here), so lets be sure to discuss before making any changes.  Also, if anything that I'm missing, just let me know.

andrewlecuyer (approved):
I have one comment about using `envtest` that I would like to hash out before merging, so just let me know when you have a chance to discuss

internal/upgradecheck/http_test.go
benjaminjb: I made a note to change this in the future; I did a quick version of changing it over and it was mostly clean but there were some failures, so as we discussed, I'll merge in now and we can iterate at a later point.

PR #2853, opened 2021-11-15
Major PG Upgrade Using pg_upgrade

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
This update adds support for major PostgreSQL upgrades using [pg_upgrade](https://www.postgresql.org/docs/current/pgupgrade.html). 
This is done by configuring a new `upgrade` section of the spec to enable the upgrade,
which identifies the current version being upgraded from and the crunchy-upgrade
container image to use. Note this image value is optional to allow for the use of the 
`RELATED_IMAGE_PGUPGRADE` environment variable.

Once configured, the section will be similar to:
```
  upgrade:
    enabled: true
    fromPostgresVersion: <current Postgres version>
    image: <image prefix/tag>
```
Before applying, the `spec.postgresVersion` and the image value will need to be
updated to the new version, i.e. if performing a 12 to 13 upgrade,
`spec.upgrade.fromPostgresVersion` should equal 12 and `spec.postgresVersion`
should equal 13. Once all updates are in place, applying the PostgresCluster
manifest update will start the upgrade process.

The upgrade is performed by scaling down any replicas, stopping the primary
database instance, running the upgrade Job, then, if the Job completed successfully,
starting the primary database instance Pod again.

**Follow on work to the current PR include:**
- Documentation
- Post-upgrade tasks, including:
  - stanza upgrade
  - initial backup for the new PG version
  - pgData cleanup
  - scaling up requested replicas

**Other information**:
[sc-12374]


internal/postgres/reconcile.go
```diff
@@ -258,3 +261,146 @@ func PodSecurityContext(cluster *v1beta1.PostgresCluster) *corev1.PodSecurityCon
...
+echo -e "Step 5: Running pg_upgrade check...\n"
+time /usr/pgsql-$NEWVERSION/bin/pg_upgrade --old-bindir /usr/pgsql-$OLDVERSION/bin --new-bindir \
+/usr/pgsql-$NEWVERSION/bin --old-datadir /pgdata/pg$OLDVERSION --new-datadir /pgdata/pg$NEWVERSION --link --check
+
+echo -e "\nStep 6: Running pg_upgrade...\n"
+time /usr/pgsql-$NEWVERSION/bin/pg_upgrade --old-bindir /usr/pgsql-$OLDVERSION/bin --new-bindir \
+/usr/pgsql-$NEWVERSION/bin --old-datadir /pgdata/pg$OLDVERSION --new-datadir /pgdata/pg$NEWVERSION --link
+
+echo -e "\nStep 7: Copying patroni.dynamic.json...\n"
+cp /pgdata/pg$OLDVERSION/patroni.dynamic.json /pgdata/pg$NEWVERSION
```
andrewlecuyer, line 347: :pick: Some links to the Patroni docs would be good around here to show where steps such as this one (deleting the `patroni.dynamic.json`) come from.

internal/postgres/reconcile.go
```diff
@@ -258,3 +261,146 @@ func PodSecurityContext(cluster *v1beta1.PostgresCluster) *corev1.PodSecurityCon
...
+			PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+				ClaimName: inDataVolume.Name,
+				ReadOnly:  false,
+			},
+		},
+	}
+	//while true; do sleep 30; done;
+	script := fmt.Sprintf(`OLDVERSION=%d
+NEWVERSION=%d
+echo -e "Performing PostgreSQL upgrade from version $OLDVERSION to $NEWVERSION for cluster %s.\n"
```
andrewlecuyer, line 323: :pick: Some links to the [`pg_upgrade` docs](https://www.postgresql.org/docs/14/pgupgrade.html) would be good here.

:thinking: Also, looking at the [`pg_upgrade` docs](https://www.postgresql.org/docs/14/pgupgrade.html) I see the following called out:

> If extension updates are available, pg_upgrade will report this and create a script that can be run later to update them.

Was therefore curious - did you ever see an extension upgrade script generated via the upgrade process?

internal/postgres/reconcile.go
```diff
@@ -258,3 +261,146 @@ func PodSecurityContext(cluster *v1beta1.PostgresCluster) *corev1.PodSecurityCon
...
+	dataVolume := corev1.Volume{
+		Name: dataVolumeMount.Name,
+		VolumeSource: corev1.VolumeSource{
+			PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+				ClaimName: inDataVolume.Name,
+				ReadOnly:  false,
+			},
+		},
+	}
+	//while true; do sleep 30; done;
```
andrewlecuyer, line 320: :wrench: Looks like a leftover comment here.

internal/postgres/reconcile.go
```diff
@@ -258,3 +261,146 @@ func PodSecurityContext(cluster *v1beta1.PostgresCluster) *corev1.PodSecurityCon
...
+/usr/pgsql-$NEWVERSION/bin --old-datadir /pgdata/pg$OLDVERSION --new-datadir /pgdata/pg$NEWVERSION --link --check
+
+echo -e "\nStep 6: Running pg_upgrade...\n"
+time /usr/pgsql-$NEWVERSION/bin/pg_upgrade --old-bindir /usr/pgsql-$OLDVERSION/bin --new-bindir \
+/usr/pgsql-$NEWVERSION/bin --old-datadir /pgdata/pg$OLDVERSION --new-datadir /pgdata/pg$NEWVERSION --link
+
+echo -e "\nStep 7: Copying patroni.dynamic.json...\n"
+cp /pgdata/pg$OLDVERSION/patroni.dynamic.json /pgdata/pg$NEWVERSION
+echo -e "\npg_upgrade Job Complete!"
+`, cluster.Spec.Upgrade.FromPostgresVersion, cluster.Spec.PostgresVersion, cluster.Name)
```
andrewlecuyer, line 349: :seedling: This looks like one of those places where some additional validation could be beneficial before we run the Job, i.e. to ensure the "from version" is less than the "to version".

Can we add a note here indicating additional validation via webhook could be beneficial?

:question: Also, I assume right now the `--check` will simply fail if versions don't make sense (e.g. if the "from version" is greater than the "to version")?

internal/postgres/reconcile.go
```diff
@@ -258,3 +261,146 @@ func PodSecurityContext(cluster *v1beta1.PostgresCluster) *corev1.PodSecurityCon
...
+		Name: dataVolumeMount.Name,
+		VolumeSource: corev1.VolumeSource{
+			PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+				ClaimName: inDataVolume.Name,
+				ReadOnly:  false,
+			},
+		},
+	}
+	//while true; do sleep 30; done;
+	script := fmt.Sprintf(`OLDVERSION=%d
```
andrewlecuyer, line 321: :pick:  It'd be good to move this script into its own function (e.g. such as `startupCommand()` or `reloadCommand()`) so that we can run the script through a shellcheck test (such as we do in `TestStartupCommand()`, etc.).

:wrench: In anticipation of an eventual ShellCheck test, we should align with their guidance.  Therefore, the bash vars should be defined using `"${NEWVERSION}"` rather than `$NEWVERSION`.

internal/naming/annotations.go
```diff
@@ -50,4 +50,10 @@ const (
 	// timestamp), which will be stored in the PostgresCluster status to properly track completion
 	// of the Job.
 	PGBackRestRestore = annotationPrefix + "pgbackrest-restore"
+
+	// PGUpgrade is the annotation that is added to a PostgresCluster to initiate a major
+	// PostgreSQL upgrade.The value of the annotation will be a unique identfier for an
+	// upgrade Job (e.g. a timestamp), which will be stored in the PostgresCluster status
+	// to properly track completion of the Job.
+	PGUpgrade = annotationPrefix + "pgupgrade"
```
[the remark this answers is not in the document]
tjmoore4: No, I missed this during an earlier refactor. I'll remove it.

internal/postgres/reconcile.go
tjmoore4: Yep, it is. Thanks for the catch, removing.

config/manager/manager.yaml
```diff
@@ -30,6 +30,8 @@ spec:
           value: "registry.developers.crunchydata.com/crunchydata/crunchy-pgbouncer:centos8-1.16-0"
         - name: RELATED_IMAGE_PGEXPORTER
           value: "registry.developers.crunchydata.com/crunchydata/crunchy-postgres-exporter:ubi8-5.0.4-0"
+        - name: RELATED_IMAGE_PGUPGRADE
+          value: "registry.developers.crunchydata.com/crunchydata/crunchy-upgrade:centos8-14.1-0"
```
jmckulk, line 34: ~Should this match the postgres/gis versions of 13.5 :question:~ ...Actually why do we have both 13 and 14 versions here :confused:

internal/postgres/reconcile.go
tjmoore4: :+1: I added a comment and a link for this step

internal/postgres/reconcile.go
tjmoore4: I've added a link to the documentation and more detailed comments on exactly what is happening.

I have seen a generated extension upgrade script, but not necessarily every time. I will add some detail to the upcoming documentation to discuss this a bit more, but this iteration of the upgrade won't be utilizing any 'post-upgrade' scripts.

internal/controller/postgrescluster/instance.go
```diff
@@ -1341,3 +1384,361 @@ func (r *Reconciler) reconcileInstanceCertificates(
...
+		} else if len(runners) > 0 {
+			cluster.Status.StartupInstance = runners[0].GetName()
+			cluster.Status.StartupInstanceSet =
+				runners[0].GetLabels()[naming.LabelInstanceSet]
+		} else if len(cluster.Spec.InstanceSets) > 0 {
+			// Generate a hash that will be used make sure that the startup
+			// instance is named consistently
+			cluster.Status.StartupInstance = naming.GenerateStartupInstance(cluster,
+				&cluster.Spec.InstanceSets[0]).Name
+			cluster.Status.StartupInstanceSet = cluster.Spec.InstanceSets[0].Name
```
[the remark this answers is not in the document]
tjmoore4: Yes, I think that approach does make sense.  I'm thinking I could use the "PGUpgradeFailed" condition with a separate message, if that sounds like a reasonable way to handle it.

internal/controller/postgrescluster/instance.go
```diff
@@ -1341,3 +1384,361 @@ func (r *Reconciler) reconcileInstanceCertificates(
...
+		setPreparingClusterCondition("removing runners")
+		for _, runner := range runners {
+			err := r.Client.Delete(ctx, runner,
+				client.PropagationPolicy(metav1.DeletePropagationForeground))
+			if client.IgnoreNotFound(err) != nil {
+				return errors.WithStack(err)
+			}
+		}
+		return nil
+	}
```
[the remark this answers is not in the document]
tjmoore4: :+1: I added a TODO for now.

internal/postgres/reconcile.go
tjmoore4: I dug into this a bit more and found that the current code, while preventing the upgrade from running, would allow the cluster to effectively be re-deployed with the wrong version. To handle this, I added some better use of conditions and events in `internal/controller/postgrescluster/instance.go`, which is where I put a TODO regarding webhook validation.

As for the `--check`, that's correct, the upgrade Job would fail and the logs would show:
```
Checking cluster versions                                   
This utility cannot be used to downgrade to older major PostgreSQL versions.
Failure, exiting
```
However, the current code should never allow that to happen.

internal/postgres/reconcile.go
tjmoore4: :+1: I moved the script to a new function and added a shellcheck test.

config/manager/manager.yaml
tjmoore4: I believe the goal was to show the two most recent versions, since the PG version values are included. As for the new `RELATED_IMAGE_PGUPGRADE` value, that can use the latest version because the previous versions are included to facilitate the various upgrades.

internal/controller/postgrescluster/instance.go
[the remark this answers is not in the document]
andrewlecuyer: How about using `ConditionPGUpgradeCompleted` with a `false` status?

```
                        meta.SetStatusCondition(&cluster.Status.Conditions, metav1.Condition{
				ObservedGeneration: cluster.GetGeneration(),
				Type:               ConditionPGUpgradeCompleted,
				Status:             metav1.ConditionFalse,
				Reason:             "PGUpgradeFailed",
				Message:            "pg_upgrade failed",
			})
```

We can then get rid of `ConditionPGUpgradeFailed`.

Also, is the `ConditionPGUpgradeProgressing` condition ever removed?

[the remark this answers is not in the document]
tjmoore4: That works for me, I've removed `ConditionPGUpgradeFailed` and set `false` where appropriate.

Yes, that condition is removed when the two `Completed` Conditions are set, at line 1635.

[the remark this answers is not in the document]
andrewlecuyer: Ah I see it now, thanks!!

internal/controller/postgrescluster/instance.go
```diff
@@ -1341,3 +1384,361 @@ func (r *Reconciler) reconcileInstanceCertificates(
...
+			Status:             metav1.ConditionTrue,
+			Reason:             ReasonReadyForUpgrade,
+			Message:            "Upgrading cluster postgres major version",
+		})
+		// the cluster is no longer bootstrapped
+		cluster.Status.Patroni.SystemIdentifier = ""
+		// the upgrade will change the contents of the database, so the pgbouncer and exporter hashes
+		// are no longer valid
+		cluster.Status.Proxy.PGBouncer.PostgreSQLRevision = ""
+		cluster.Status.Monitoring.ExporterConfiguration = ""
```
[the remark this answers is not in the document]
tjmoore4: After discussing with the team, for now we'll plan to leave the status as is. Depending on the particulars of the custom sql configured, it may not be desirable to automatically run again.

internal/naming/names.go
```diff
@@ -40,6 +40,9 @@ const (
 	// ContainerPGBouncerConfig is the name of a container supporting PgBouncer.
 	ContainerPGBouncerConfig = "pgbouncer-config"
 
+	// ContainerPGUpgrade is the name of a container running pg_upgrade.
+	ContainerPGBUpgrade = "pgupgrade"
```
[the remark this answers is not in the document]
tjmoore4: Updated, thanks for the catch!

internal/postgres/reconcile_test.go
```diff
@@ -542,3 +543,187 @@ supplementalGroups:
...
+
+func TestUpgradeJob(t *testing.T) {
+	cluster := &v1beta1.PostgresCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "testcluster",
+			Namespace: "testnamespace",
+		},
+	}
+	cluster.Spec.ImagePullPolicy = corev1.PullAlways
+	cluster.Spec.PostgresVersion = 12
```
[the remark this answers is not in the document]
tjmoore4: For this test, the values don't matter that much as long as the 'from' value is less than the version we're going to. With that said, I think this is ok to leave as is because we really shouldn't need to adjust it in the future.

config/manager/manager.yaml
andrewlecuyer: Can we update any `crunchy-upgrade` tags per sc-12985?  E.g.

```
registry.developers.crunchydata.com/crunchydata/crunchy-upgrade:crunchy-upgrade:ubi8-5.1.0-0
```

We would also want to do the same for `installers/olm/config/redhat/related-images.yaml ` below.

internal/postgres/reconcile.go
andrewlecuyer: Sounds good, thank you!

config/manager/manager.yaml
tjmoore4: Sure thing, I'll make those updates.

internal/controller/postgrescluster/instance.go
```diff
@@ -1341,3 +1380,382 @@ func (r *Reconciler) reconcileInstanceCertificates(
...
+				}
+				if len(observedInstances.byName) != 0 {
+					return true, nil
+				}
+
+				// After all observed instances are deleted, delete endpoints to
+				// remove the old DCS information.
+				if len(endpoints) > 0 {
+					for i := range endpoints {
+						if err := r.Client.Delete(ctx, &endpoints[i]); client.IgnoreNotFound(err) != nil {
```
andrewlecuyer, line 1430: :pick: Feels like endpoint deletion might be better aligned under `prepareForUpgrade()` rather than this function.

internal/controller/postgrescluster/instance.go
```diff
@@ -1341,3 +1380,382 @@ func (r *Reconciler) reconcileInstanceCertificates(
...
+
+				// determine the startup InstanceSet
+				var spec *v1beta1.PostgresInstanceSetSpec
+				for i := range instanceSets {
+					if instanceSets[i].Name == cluster.Status.StartupInstanceSet {
+						spec = &instanceSets[i]
+					}
+				}
+				if spec == nil {
+					return false, errors.Errorf("Unable to determine startup instance set.")
```
andrewlecuyer, line 1452: :pick: Can use `errors.New()` here since there are no arguments.

andrewlecuyer (approved):

PR #2856, opened 2021-11-16
Use a different structure to accumulate pgBackRest options

cbandy:
Future work will need to set one of the pgBackRest options that take multiple values. Document these options and how pgBackRest merges them across configuration files, environment variables, and command-line arguments.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Other


internal/pgbackrest/options.go
```diff
@@ -0,0 +1,96 @@
...
+
+// Set replaces the values associated with key with the single value. The key
+// is case-sensitive.
+func (ms iniMultiSet) Set(key, value string) {
+	ms[key] = []string{value}
+}
+
+// Value returns empty string or the first value associated with the given key
+// if it is present in ms.
+func (ms iniMultiSet) Value(key string) string {
```
andrewlecuyer, line 62: :thinking: Returning the first value in this function feels a bit funny, especially if there are multiple values.  Not sure what the right answer is - maybe passing in an index or something along those lines?

Or maybe we should simply have the `Values()` function, and the client can simply inspect the slice contents as needed?

andrewlecuyer (approved):
Looks good.

Feel free to merge this one right away if needed to pave the way for the TLS story.

internal/pgbackrest/options.go
```diff
@@ -0,0 +1,96 @@
...
+			}
+		}
+	}
+	return b.String()
+}
+
+// Add associates value with key, appending it to any values already associated
+// with key. The key is case-sensitive.
+func (ms iniMultiSet) Add(key, value string) {
+	ms[key] = append(ms[key], value)
```
andrewlecuyer, line 51: :thinking: Should we care about duplicates when adding?

jmckulk (approved):
One comment but looks good

internal/pgbackrest/options.go
cbandy: This is for options that can only be set once. [net/http.Header](https://pkg.go.dev/net/http#Header) is a similar type; it uses "Get" for this.

Over in [postgres.ParameterSet](https://github.com/CrunchyData/postgres-operator/blob/master/internal/postgres/parameters.go#L93), we use "Value" for this.

I haven't needed to read _any_ values outside of testing, so I'll drop this method to avoid naming it. 😄

internal/pgbackrest/options.go
cbandy: I just checked, and pgBackRest allows duplicate values in at least one case.

andrewlecuyer: Ah very interesting ... thanks for confirming.

internal/pgbackrest/options.go
andrewlecuyer: Thanks for the reference to Header package! ... some other things in there have me thinking about naming a bit more (e.g. I see they have an `Add()` and a `Set()`).

Sounds good regarding dropping till we need it tho! :+1:

internal/pgbackrest/config.go
```diff
@@ -258,136 +265,101 @@ mv "${pgdata}" "${pgdata}_bootstrap"`
...
 
-	// set the default stanza name
-	pgBackRestConfig["stanza"]["name"] = DefaultStanzaName
-
-	// set global settings, which includes all repos
-	pgBackRestConfig["global"]["log-path"] = defaultLogPath
 	for _, repo := range repos {
-
-		repoConfigs := make(map[string]string)
+		global.Set(repo.Name+"-path", defaultRepo1Path+repo.Name)
```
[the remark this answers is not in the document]
cbandy: I agree. I started touching this and it grew a little bit bigger than belongs in this PR, probably. I'm gonna table what I have and open another PR later to address/discuss this.

internal/pgbackrest/options.go
cbandy: Added to the doc in [e5f9d2c](https://github.com/CrunchyData/postgres-operator/compare/912f6c652aa820b7dc85a9e2218353b8409ab817..e5f9d2c32a83c59c31ab391fae9e94ab35d952f7).

internal/pgbackrest/options.go
cbandy: Done in [e5f9d2c](https://github.com/CrunchyData/postgres-operator/compare/912f6c652aa820b7dc85a9e2218353b8409ab817..e5f9d2c32a83c59c31ab391fae9e94ab35d952f7).

With `Values` being the only way to read, I noticed that there was no way to set all the values for a key at once. I changed `Set` to take zero or more values.

PR #2866, opened 2021-11-21
Small cleanups to PKI tests

cbandy:
These two commits:
 - Add `Equal` methods to `pki.Certificate` and `pki.PrivateKey` so they can be used by `assert.DeepEqual` in tests.
 - Remove some Kubernetes API calls from `TestReconcileCerts`.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

### DIFF
--- a/internal/pki/pki.go
+++ b/internal/pki/pki.go
@@ -1,0 +1,31 @@
+/*
+ Copyright 2021 Crunchy Data Solutions, Inc.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package pki
+
+import "bytes"
+
+// Equal reports whether c and other have the same value.
+func (c Certificate) Equal(other Certificate) bool {
+	return bytes.Equal(c.Certificate, other.Certificate)
+}
+
+// Equal reports whether k and other have the same value.
+func (k PrivateKey) Equal(other PrivateKey) bool {
+	if k.PrivateKey == nil || other.PrivateKey == nil {
+		return k.PrivateKey == other.PrivateKey
+	}
+	return k.PrivateKey.Equal(other.PrivateKey)
+}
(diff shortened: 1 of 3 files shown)


internal/pki/pki.go
```diff
@@ -0,0 +1,31 @@
...
+
+// Equal reports whether c and other have the same value.
+func (c Certificate) Equal(other Certificate) bool {
+	return bytes.Equal(c.Certificate, other.Certificate)
+}
+
+// Equal reports whether k and other have the same value.
+func (k PrivateKey) Equal(other PrivateKey) bool {
+	if k.PrivateKey == nil || other.PrivateKey == nil {
+		return k.PrivateKey == other.PrivateKey
```
jkatz, line 28: 🤔 Do we want to evaluate `nil == nil` to `true`?

jkatz (approved):

internal/pki/pki.go
cbandy: `nil` happens in the zero value for `PrivateKey`, and I think we do want any value to equal itself.

jkatz: :+1:

internal/pki/pki_test.go
```diff
@@ -27,6 +27,38 @@ import (
...
+	root := NewRootCertificateAuthority()
+	assert.NilError(t, root.Generate())
+	assert.DeepEqual(t, root.PrivateKey, root.PrivateKey)
+
+	assert.Assert(t, !root.PrivateKey.Equal(zero))
+	assert.Assert(t, !zero.Equal(*root.PrivateKey))
+
+	other := NewRootCertificateAuthority()
+	assert.NilError(t, other.Generate())
+	assert.Assert(t, !root.PrivateKey.Equal(*other.PrivateKey))
```
benjaminjb, line 59: Could we add test cases where the Certificates / PrivateKeys were equal?

cbandy: That's L32, L36, L48, and L52. I'll rearrange to show that the `Equal` method works with and without `DeepEqual`.

Without the `Equal` method, `assert.DeepEqual` tries to call [`ecdsa.PrivateKey.Equal`](https://pkg.go.dev/crypto/ecdsa#PrivateKey.Equal) which panics when the receiver is nil (our zero value).

On other PrivateKey values, without the `Equal` method, `assert.DeepEqual` fails saying "cannot handle unexported field at {*pki.PrivateKey}.marshalECPrivateKey".

benjaminjb: Ah, ok, I think this comment helps explain that, so maybe this comment is all you need?

cbandy: Amended in af07b40d0.

benjaminjb (approved):

PR #2872, opened 2021-11-23
Reduce parameters of generateBackupJobSpecIntent

cbandy:
While looking at all the calls to `pgbackrest.AddConfigsToPod`, I noticed calls to `generateBackupJobSpecIntent` were passing the same values. This passes the repo object instead.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Other

### DIFF
--- a/internal/controller/postgrescluster/pgbackrest.go
+++ b/internal/controller/postgrescluster/pgbackrest.go
@@ -653,11 +653,16 @@ func (r *Reconciler) generateRepoVolumeIntent(postgresCluster *v1beta1.PostgresC
 }
 
 // generateBackupJobSpecIntent generates a JobSpec for a pgBackRest backup job
-func generateBackupJobSpecIntent(postgresCluster *v1beta1.PostgresCluster, selector,
-	containerName, repoName, serviceAccountName, configName string,
+func generateBackupJobSpecIntent(postgresCluster *v1beta1.PostgresCluster,
+	repo v1beta1.PGBackRestRepo, serviceAccountName string,
 	labels, annotations map[string]string, opts ...string) (*batchv1.JobSpec, error) {
 
-	repoIndex := regexRepoIndex.FindString(repoName)
+	selector, containerName, err := getPGBackRestExecSelector(postgresCluster, repo)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	repoIndex := regexRepoIndex.FindString(repo.Name)
 	cmdOpts := []string{
 		"--stanza=" + pgbackrest.DefaultStanzaName,
 		"--repo=" + repoIndex,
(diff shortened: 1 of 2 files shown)


internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -2003,11 +2009,12 @@ func (r *Reconciler) reconcileManualBackup(ctx context.Context,
 func (r *Reconciler) reconcileReplicaCreateBackup(ctx context.Context,
 	postgresCluster *v1beta1.PostgresCluster, instances *observedInstances,
 	replicaCreateBackupJobs []*batchv1.Job,
-	serviceAccount *corev1.ServiceAccount, configHash, replicaCreateRepoName string) error {
+	serviceAccount *corev1.ServiceAccount, configHash string,
+	replicaCreateRepo v1beta1.PGBackRestRepo) error {
```
benjaminjb, line 2013: I almost find it disturbing that you changed this func's signature, but because you changed `reconcileRepos` signature as well, you don't have to change where you invoke `reconcileReplicaCreateBackup` on  line 1322

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -2111,7 +2112,7 @@ func (r *Reconciler) reconcileReplicaCreateBackup(ctx context.Context,
 		//   made in the spec (specifically a change to the config for an external repo).  Delete
 		//   and recreate the Job with proper hash per the current config.
 		if failed || replicaCreateRepoChanged ||
-			(job.GetAnnotations()[naming.PGBackRestCurrentConfig] != configName) ||
+			(job.GetAnnotations()[naming.PGBackRestCurrentConfig] != containerName) ||
```
benjaminjb, line 2115: To double-check: in the old version, we were checking that this didn't equal either `pgbackrest.CMInstanceKey` or `pgbackrest.CMRepoKey`, but now it looks like we're checking that this isn't equal to either `naming.PGBackRestRepoContainerName` or `naming.ContainerDatabase` -- which works because we're setting that `annotation[naming.PGBackRestCurrentConfig]` to equal one of those container names.

It looks like that annotation label is only ever set and checked in this file, but is there a chance it's being used elsewhere?

benjaminjb (approved):
LGTM (after some digging in the code to make sure I could follow what was happening), and I'm _pretty_ sure that annotation change is fine -- I really couldn't find it anywhere else.

internal/controller/postgrescluster/pgbackrest.go
cbandy: Ah, right. The variable storing the result of the former and passing to the latter was already called `replicaCreateRepo`. It used to be a string/name and now it is a "full" repo spec.

internal/controller/postgrescluster/pgbackrest.go
cbandy: Right. This line is paired with L2153.

These two lines (and their tests) are the only place the annotation is used.

ConfigHash is also used on restore jobs. 🌱 Probably these two annotations/ideas need to be combined.

PR #2882, opened 2021-11-26
Remove unused pgBackRest configuration

cbandy:
A few functions were exported from our internal pgbackrest package but they were never called. This removes them and their constants. The remaining functions in this file do not make any API calls, so remove them from the tests as well.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

### DIFF
--- a/internal/pgbackrest/config_test.go
+++ b/internal/pgbackrest/config_test.go
@@ -90,8 +90,6 @@ func TestPGBackRestConfiguration(t *testing.T) {
 	var cmInitial *corev1.ConfigMap
 	// the returned configmap
 	var cmReturned corev1.ConfigMap
-	// pod spec for testing projected volumes and volume mounts
-	pod := &corev1.PodSpec{}
 
 	testInstanceName := "test-instance-abc"
 	testRepoName := "repo-host"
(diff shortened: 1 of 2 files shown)


jkatz (approved):

internal/pgbackrest/config_test.go
```diff
@@ -23,284 +20,161 @@ import (
...
-			}
-
-			err := testClient.Get(context.Background(), objectKey, &cmReturned)
-
-			assert.NilError(t, err)
-		})
-
-		// finally, verify initial and returned match
-		assert.Assert(t, reflect.DeepEqual(cmInitial.Data, cmReturned.Data))
+	configmap := CreatePGBackRestConfigMapIntent(cluster,
```
benjaminjb, line 66: :pick: I'm 50/50 here, but I feel like this would be a good place to start a `t.Run` section: everything that comes before is setup that is used by the two test cases; everything that comes after are the two cases: "standard in every way", and "custom metadata."

benjaminjb (approved):
LGTM

internal/pgbackrest/config_test.go
cbandy: Amended in [`2b0a8e015`](https://github.com/CrunchyData/postgres-operator/compare/2c4048ec8410e68e476b13997acf673ae5ab6b8b..2b0a8e015a4a54bfc5c745f533531228d3bdae10). I named the subtest `DedicatedRepoHost` because that is what is really being tested here. The two subtests now make a copy of `cluster` because they are independent.

PR #2889, opened 2021-11-30
Set Default Archive Timeout

jmckulk:
When archiving is enabled by PGO, an archive_timeout setting should also
be configured by default. More specifically, we should default to 60
seconds, as recommended by the PostgreSQL documentation as a
"reasonable" value for this setting:

https://www.postgresql.org/docs/current/runtime-config-wal.html#GUC-ARCHIVE-TIMEOUT

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

**Other Information**:
[sh-12876]

### DIFF
--- a/internal/pgbackrest/postgres.go
+++ b/internal/pgbackrest/postgres.go
@@ -30,6 +30,9 @@ func PostgreSQL(
 	if outParameters.Mandatory == nil {
 		outParameters.Mandatory = postgres.NewParameterSet()
 	}
+	if outParameters.Default == nil {
+		outParameters.Default = postgres.NewParameterSet()
+	}
 
 	// Send WAL files to all configured repositories when not in recovery.
 	// - https://pgbackrest.org/user-guide.html#quickstart/configure-archiving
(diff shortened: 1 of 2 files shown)


internal/pgbackrest/postgres.go
```diff
@@ -39,6 +42,11 @@ func PostgreSQL(
 	outParameters.Mandatory.Add("archive_mode", "on")
 	outParameters.Mandatory.Add("archive_command", archive)
 
+	// PostgreSQL documentation considers an archive_timeout of around a minute
+	// to be reasonable:
+	// - https://www.postgresql.org/docs/current/runtime-config-wal.html#GUC-ARCHIVE-TIMEOUT
+	outParameters.Default.Add("archive_timeout", "60")
```
jkatz, line 48:
```suggestion
	// archive_timeout is used to determine at what point a WAL file is switched,
	// if the WAL archive has not reached its full size in # of transactions
	// (16MB). This has ramifications for log shipping, i.e. it ensures a WAL file
	// is shipped to an archive every X seconds to help reduce the risk of data
	// loss in a disaster recovery scenario. For standby servers that are not
	// connected using streaming replication, this also ensures that new data is
	// available at least once a minute.
	//
	// PostgreSQL documentation considers an archive_timeout of 60 seconds to be
	// reasonable. There are cases where you may want to set archive_timeout to 0,
	// for example, when the remote archive (pgBackRest repo) is unavailable; this
	// is to prevent WAL accumulation on your primary.
	// - https://www.postgresql.org/docs/current/runtime-config-wal.html#GUC-ARCHIVE-TIMEOUT
	outParameters.Default.Add("archive_timeout", "60")
```

internal/pgbackrest/postgres.go
```diff
@@ -39,6 +42,11 @@ func PostgreSQL(
 	outParameters.Mandatory.Add("archive_mode", "on")
 	outParameters.Mandatory.Add("archive_command", archive)
 
+	// PostgreSQL documentation considers an archive_timeout of around a minute
+	// to be reasonable:
```
cbandy: :pick: The link is good, but I think we should call out why we're setting it.

> Send WAL files to repositories even when there are only a few transactions.

cbandy (approved):

tjmoore4 (approved):
:+1: Looks good to me.

internal/pgbackrest/postgres.go
```diff
@@ -39,6 +42,21 @@ func PostgreSQL(
...
+	// loss in a disaster recovery scenario. For standby servers that are not
+	// connected using streaming replication, this also ensures that new data is
+	// available at least once a minute.
+	//
+	// PostgreSQL documentation considers an archive_timeout of 60 seconds to be
+	// reasonable. There are cases where you may want to set archive_timeout to 0,
+	// for example, when the remote archive (pgBackRest repo) is unavailable; this
+	// is to prevent WAL accumulation on your primary.
+	// - https://www.postgresql.org/docs/current/runtime-config-wal.html#GUC-ARCHIVE-TIMEOUT
+	outParameters.Default.Add("archive_timeout", "60s")
```
jkatz, line 58:
```suggestion
	outParameters.Default.Add("archive_timeout", "60")
```

I don't believe we add the `s`? The PG docs reference above say to use an integer.

jkatz (changes requested):

jkatz (approved):

benjaminjb (approved):

PR #2898, opened 2021-12-04
Make Upgrade Check URL Configurable

andrewlecuyer:
Makes it possible to configure the URL for the endpoint used to check for PGO upgrades.

**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

The upgrade check URL is not configurable.

The upgrade check URL is configurable.

[sc-13171]

**Other Information**:

N/A

### DIFF
--- a/cmd/postgres-operator/main.go
+++ b/cmd/postgres-operator/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 
 import (
 	"context"
+	"errors"
 	"os"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


benjaminjb (approved):
LGTM, I just have a question about whether we want to put something in `make deploy-dev` for the new env var? 

I don't think it's necessary because if someone wants to turn on upgrade checking in that case they already need to set an env var to turn it on, so no big difficulty in setting two env vars.

andrewlecuyer:
@benjaminjb my thought for now was that we can simply set these environment variables in our local environments when using `deploy-dev`.  This will allow us to make explicit decisions around how we configure these settings, and get a better idea of how we want to set these settings for development (i.e. when using `deploy-dev`) moving forward.

We can then consider future updates to the `Makefile` accordingly.

PR #2931, opened 2021-12-16
Always close HTTP response bodies

cbandy:
Not doing so can leave connections open or other resources unclaimed.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Bug fix

### DIFF
--- a/internal/upgradecheck/http.go
+++ b/internal/upgradecheck/http.go
@@ -18,7 +18,7 @@ package upgradecheck
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"time"
 
@@ -77,8 +77,6 @@ func init() {
 func checkForUpgrades(log logr.Logger, versionString string, backoff wait.Backoff,
 	crclient crclient.Client, ctx context.Context, cfg *rest.Config,
 	isOpenShift bool) (message string, err error) {
-	var res *http.Response
-	var bodyBytes []byte
 	var headerPayloadStruct *clientUpgradeData
 
 	// Guard against panics within the checkForUpgrades function to allow the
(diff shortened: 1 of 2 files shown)


internal/upgradecheck/http.go
```diff
@@ -107,38 +105,40 @@ func checkForUpgrades(log logr.Logger, versionString string, backoff wait.Backof
...
-	if err == nil {
-		return string(bodyBytes), nil
+	// We received responses, but none of them were 200 OK.
+	if err == nil && status != http.StatusOK {
+		err = fmt.Errorf("received StatusCode %d", status)
 	}
 
-	return "", err
+	// TODO: Parse response and log info for user on potential upgrades
+	return string(bodyBytes), err
```
benjaminjb, line 141: Just to talk it out, when we get here, our potential results are

(any of the retries calls successfully)
* a successful call: string(bodyBytes) is the response, err is nil
* a successful call with a ReadAll error: string(bodyBytes) is probably "", err is the ReadAll error

(all the retries fail to get a 200 response)
* a non-errored, non-200 call: string(bodyBytes) is the response of the non-200 call, err is a message we set on l 137 abt the non-200 status
* error: string(bodyBytes) is "", err is the error message

benjaminjb (approved):
LGTM, just that one code comment update

internal/upgradecheck/http.go
cbandy: Agreed. There are two assignments to `err` inside the backoff callback/closure and those will be returned when `ExponentialBackoff` returns. `bodyBytes` is set right before `ExponentialBackoff` returns and will contain whatever we got from the read.

It seemed a little odd ignoring the return value from `ExponentialBackoff`, but we're not interested in knowing if backoff expired or not. Both before and now, we're interested in the values from the HTTP call.

benjaminjb (approved):

PR #2932, opened 2021-12-16
Do Not Remove Certain Status for Major PG Upgrade

andrewlecuyer:
When preparing a cluster for a major PG upgrade, only the `repos` section of the pgBackRest status is removed, while the `PostgresDataInitialized` condition (if present) is now also preserved.  This ensures that only the status and conditions that are pertinent to the upgrade process are removed, and all other status is preserved.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

The entire pgBackRest status, as well as the `PostgresDataInitialized` condition, is removed when preparing for a major PG upgrade.

[sc-13310]

Only the `repos` section of the pgBackRest status is removed when preparing for a major PG upgrade, while the `PostgresDataInitialized` condition is also preserved.

**Other Information**:

N/A

### DIFF
--- a/internal/controller/postgrescluster/instance_test.go
+++ b/internal/controller/postgrescluster/instance_test.go
@@ -2587,8 +2587,8 @@ func TestPrepareForUpgrade(t *testing.T) {
 				ObservedGeneration: cluster.GetGeneration(),
 				Type:               ConditionPostgresDataInitialized,
 				Status:             metav1.ConditionTrue,
-				Reason:             "PGUpgradeComplete",
-				Message:            "pg_upgrade completed successfully",
+				Reason:             "test",
+				Message:            "test",
 			})
 
 			testResources := tc.createResources(t, cluster)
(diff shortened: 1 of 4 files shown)


cbandy (approved):

tjmoore4 (approved):

internal/controller/postgrescluster/instance_test.go
```diff
@@ -2587,8 +2587,8 @@ func TestPrepareForUpgrade(t *testing.T) {
 				ObservedGeneration: cluster.GetGeneration(),
 				Type:               ConditionPostgresDataInitialized,
 				Status:             metav1.ConditionTrue,
-				Reason:             "PGUpgradeComplete",
-				Message:            "pg_upgrade completed successfully",
+				Reason:             "test",
```
jmckulk, line 2590: test?

jmckulk (approved):

internal/controller/postgrescluster/instance_test.go
andrewlecuyer: Just putting a random value in there to indicate that this condition is for test purposes only, and we don't care about the reason and message.  Also, the existing reason and message in this test is never actually utilized with `ConditionPostgresDataInitialized`, so I figured I'd change it to avoid confusion.

PR #2939, opened 2021-12-20
Update Kuttl tests to connect via Kube Job

jmckulk:
Connecting to the cluster locally with automated tests can be finicky;
This change uses a Kube job that can access internally (using the
service to connect and mount the user secret as env variables).

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

**Other Information**:
[sc-13231]

### DIFF
--- a/testing/kuttl/e2e/cluster-start/00-assert.yaml
+++ b/testing/kuttl/e2e/cluster-start/00-assert.yaml
@@ -17,3 +17,8 @@ metadata:
     postgres-operator.crunchydata.com/pgbackrest-backup: replica-create
 status:
   succeeded: 1
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: cluster-start-primary
--- a/testing/kuttl/e2e/cluster-start/01-psql-connect.yaml
+++ b/testing/kuttl/e2e/cluster-start/01-psql-connect.yaml
@@ -1,10 +1,26 @@
-apiVersion: kuttl.dev/v1beta1
-kind: TestStep
-commands:
-- script: CLUSTER=cluster-start $HOME/postgres-operator/testing/kuttl/script/port-forward.sh
-  background: true
-# this sleep is needed when running in cloud environments to ensure connection-psql.sh
-# will not fail when attempting the port-forward and has suffient time to connect, may 
-# not be needed when running locally
-- script: sleep 10
-- script: CLUSTER=cluster-start $HOME/postgres-operator/testing/kuttl/script/connect-psql.sh
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: psql-connect
+spec:
+  template:
+    spec:
+      restartPolicy: "OnFailure"
+      containers:
+        - name: psql
+          image: registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-14.1-1
+          command:
+            - "bash"
+            - "-c"
+            - "psql -c 'select version();'"
+          env:
+          - name: PGHOST
+            valueFrom: { secretKeyRef: { name: cluster-start-pguser-cluster-start, key: host } }
+          - name: PGPORT
+            valueFrom: { secretKeyRef: { name: cluster-start-pguser-cluster-start, key: port } }
+          - name: PGDATABASE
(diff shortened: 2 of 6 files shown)


testing/kuttl/e2e/cluster-start/00-assert.yaml
```diff
@@ -17,3 +17,8 @@ metadata:
     postgres-operator.crunchydata.com/pgbackrest-backup: replica-create
 status:
   succeeded: 1
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: cluster-start-primary
```
benjaminjb, line 24: So we assert that the service exists because our later test will use that service to connect. Just anecdotally, did the test ever fail for you without this?

testing/kuttl/e2e/cluster-start/01-psql-connect.yaml
```diff
@@ -1,10 +1,26 @@
...
-- script: CLUSTER=cluster-start $HOME/postgres-operator/testing/kuttl/script/connect-psql.sh
\ No newline at end of file
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: psql-connect
+spec:
+  template:
+    spec:
+      restartPolicy: "OnFailure"
```
benjaminjb, line 8: Are we sure we want a restart of this?

testing/kuttl/e2e/cluster-start/01-psql-connect.yaml
```diff
@@ -1,10 +1,26 @@
...
+      containers:
+        - name: psql
+          image: registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-14.1-1
+          command:
+            - "bash"
+            - "-c"
+            - "psql -c 'select version();'"
```
benjaminjb, line 15: What do you think about a bash command to get the version and compare it against the version we expect? Unnecessary and overcomplicated? (That's what I'm beginning to think, but just want to have the discussion.)

testing/kuttl/e2e/cluster-start/00-assert.yaml
jmckulk: Nope, it hasn't failed for me without. I think waiting for the initial backup to complete means that the service should always exist.

testing/kuttl/e2e/cluster-start/01-psql-connect.yaml
jmckulk: I think so. If we attempt to connect and it fails for some reason the job can retry (assuming we don't hit a testing timeout).

testing/kuttl/e2e/cluster-start/01-psql-connect.yaml
jmckulk: We can for sure add checks like that. In this case, I don't think it is necessary. If the command can't connect or can't run `select version();` it will fail and the job will fail.

benjaminjb (approved):

testing/kuttl/e2e/cluster-start/01-psql-connect.yaml
```diff
@@ -1,10 +1,26 @@
...
+kind: Job
+metadata:
+  name: psql-connect
+spec:
+  template:
+    spec:
+      restartPolicy: "OnFailure"
+      containers:
+        - name: psql
+          image: registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-14.1-1
```
cbandy, line 11: 🤔 No variables so I suppose no way to have this match/use the same image as the cluster?

cbandy (approved):
LGTM. Sometimes it's not obvious to me what the assertion is saying. It might be more obvious when I get familiar with kuttl. It might be good to put YAML comments that say what/why is being checked.

testing/kuttl/e2e/cluster-start/01-psql-connect.yaml
jmckulk: Nope, not as far as I am aware.

PR #2944, opened 2021-12-22
Allow passwords to be generated without symbols

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature

Passwords often contain special characters that sometimes don't survive transformation through layers of configuration.

Passwords can be configured to contain only letters and numerals on a per-user basis.

**Other Information**:

Issue: [sc-12862]


docs/content/tutorial/user-management.md
```diff
@@ -88,7 +88,7 @@ This will create a Secret of the pattern `<clusterName>-pguser-postgres` that co
 
 ## Deleting a User
 
-As mentioned earlier, PGO does not let you delete a user automatically: if you remove the user from the spec, it will still exist in your cluster. To remove a user and all of its objects, as a superuser you will need to run [`DROP OWNED`](https://www.postgresql.org/docs/current/sql-drop-owned.html) in each database the user has objects in, and [`DROP ROLE`](https://www.postgresql.org/docs/current/sql-droprole.html)
+As mentioned earlier, PGO does not delete users automatically: after you remove the user from the spec, it will still exist in your cluster. To remove a user and all of its objects, as a superuser you will need to run [`DROP OWNED`](https://www.postgresql.org/docs/current/sql-drop-owned.html) in each database the user has objects in, and [`DROP ROLE`](https://www.postgresql.org/docs/current/sql-droprole.html)
```
cbandy, line 91: 📝 "As mentioned earlier" … I didn't see this mentioned earlier. Perhaps I missed it or this is leftover from being in another section of the docs. This is a tutorial page; details are in `/architecture/user-management.md`.

docs/content/architecture/user-management.md
```diff
@@ -48,18 +48,37 @@ Users and databases can be customized in the `spec.users` section of the custom
...
 - Role attributes are not automatically dropped if you remove them. You will have to set the inverse attribute to drop them (e.g. `NOSUPERUSER`).
 - The special `postgres` user can be added as one of the custom users; however, the privileges of the users cannot be adjusted.
 
 For specific examples for how to manage users, please see the [user and database management]({{< relref "tutorial/user-management.md" >}}) section of the [tutorial]({{< relref "tutorial/_index.md" >}}).
 
+## Generated Passwords
+
+PGO generates a random password for each Postgres user it creates. Postgres allows almost any character
+in its passwords, but your application may have stricter requirements. When you need a password without
+any special characters set `spec.users.password.type` for that user to `AlphaNumeric`.
```
cbandy, line 62: ⛏️ "characters set" sounds like "character set". 🤔 Is there supposed to be a comma here?

docs/content/tutorial/user-management.md
jkatz: So it should be safe to ~As mentioned earlier,~

docs/content/tutorial/user-management.md
```diff
@@ -100,11 +100,9 @@ DROP ROLE rhino;
 
 Note that you may need to run `DROP OWNED BY rhino CASCADE;` based upon your object ownership structure -- be very careful with this command!
 
-Once you have removed the user in the database, you can remove the user from the custom resource.
-
 ## Deleting a Database
 
-As mentioned earlier, PGO does not let you delete a database automatically: if you remove all instances of the database from the spec, it will still exist in your cluster. To completely remove the database, you must run the [`DROP DATABASE`](https://www.postgresql.org/docs/current/sql-dropdatabase.html)
+As mentioned earlier, PGO does not delete databases automatically: after you remove all instances of the database from the spec, it will still exist in your cluster. To completely remove the database, you must run the [`DROP DATABASE`](https://www.postgresql.org/docs/current/sql-dropdatabase.html)
```
jkatz, line 105:
```suggestion
PGO does not delete databases automatically: after you remove all instances of the database from the spec, it will still exist in your cluster. To completely remove the database, you must run the [`DROP DATABASE`](https://www.postgresql.org/docs/current/sql-dropdatabase.html)
```

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -205,7 +205,7 @@ func (r *Reconciler) reconcileMonitoringSecret(
 	intent.Data = make(map[string][]byte)
 
 	if len(existing.Data["password"]) == 0 || len(existing.Data["verifier"]) == 0 {
-		password, err := util.GeneratePassword(util.DefaultGeneratedPasswordLength)
+		password, err := util.GenerateASCIIPassword(32)
```
jkatz, line 208: I'd prefer we increment the `util.DefaultGeneratedPasswordLength` constant vs. hard coding it to 32 everywhere.

jkatz (commented):
Looks pretty good overall.

internal/controller/postgrescluster/pgmonitor.go
tjmoore4, line 208: :thought_balloon:  Would we ever want to use  the `AlphaNumeric` password option here? I don't believe this password would be used anywhere else, unlike the Postgres password, so I think this makes sense.

internal/pgbouncer/postgres.go
```diff
@@ -214,7 +214,7 @@ func generatePassword() (plaintext, verifier string, err error) {
 	// - https://www.pgbouncer.org/config.html#authentication-file-format
 	// - https://github.com/pgbouncer/pgbouncer/issues/508#issuecomment-713339834
 
-	plaintext, err = util.GeneratePassword(32)
+	plaintext, err = util.GenerateASCIIPassword(32)
```
tjmoore4, line 217: :thought_balloon:  Same general thought as mentioned for pgMonitor. I'm thinking this is fine because this password will not be used anywhere else, but I want to make sure that's the case.

internal/util/secrets_test.go
```diff
@@ -15,50 +13,139 @@ package util
...
+	})
+}
+
+func TestGenerateAlphaNumericPassword(t *testing.T) {
+	for _, length := range []int{0, 1, 2, 3, 5, 20, 200} {
+		password, err := GenerateAlphaNumericPassword(length)
+
+		assert.NilError(t, err)
+		assert.Equal(t, length, len(password))
+		assert.Assert(t, cmp.Regexp(`^[A-Za-z0-9]*$`, password))
```
tjmoore4, line 65: :+1: I like this test. I think it does a reasonable job of checking the `AlphaNumberic` password generation feature, and it seems to be in line with what I'm seeing other places.

tjmoore4 (approved):

internal/controller/postgrescluster/pgmonitor.go
cbandy: The named constant is used for both spec users and our internal/system users. We know what passwords are tolerable for the tools of the internal users. A single constant for all of them doesn't seem right, but we can address that later.

I'll revert this.

docs/content/tutorial/user-management.md
benjaminjb: As long as you're touching this, want to break it up into <100 char lines? (And I like TJ's change.)

internal/util/secrets.go
```diff
@@ -15,78 +13,78 @@ package util
...
+	}
+
+	size := big.NewInt(int64(len(class)))
 
-		password[i] = char
-		i++
+	return func() (byte, error) {
+		i, err := rand.Int(random, size)
+		if err != nil {
+			i = new(big.Int)
```
benjaminjb, line 61: So if rand.Int errors, we always return the 0th element of the class?

internal/controller/postgrescluster/pgmonitor.go
cbandy: Maybe if something changes in the exporter or how it is configured. We've managed with an ASCII password so far.

internal/util/secrets.go
cbandy: Yes. 🤔 This was from an early attempt at dealing with errors. The random source is injected now and there's a test for errors. I'll change this to `return 0, err`.

docs/content/architecture/user-management.md
```diff
@@ -48,18 +48,37 @@ Users and databases can be customized in the `spec.users` section of the custom
...
+- To prevent accidental data loss, PGO does not automatically drop users. We will see how to drop a user below.
+- Similarly, to prevent accidental data loss PGO does not automatically drop databases. We will see how to drop a database below.
 - Role attributes are not automatically dropped if you remove them. You will have to set the inverse attribute to drop them (e.g. `NOSUPERUSER`).
 - The special `postgres` user can be added as one of the custom users; however, the privileges of the users cannot be adjusted.
 
 For specific examples for how to manage users, please see the [user and database management]({{< relref "tutorial/user-management.md" >}}) section of the [tutorial]({{< relref "tutorial/_index.md" >}}).
 
+## Generated Passwords
+
+PGO generates a random password for each Postgres user it creates. Postgres allows almost any character
```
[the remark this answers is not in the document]
cbandy: I see what you mean. My intuition is to describe the default/automatic cases then describe customizations as "when none of that works for you, then…"

How about a sentence at the end of this paragraph pointing to "custom". Roughly,

1. We generate passwords (with symbols).
2. We can generate without symbols.
3. You can set the whole thing.

docs/content/tutorial/user-management.md
jkatz: I'm generally not in favor of the line length restrictions on the docs. I know this is a debatable issue, but it makes it hard for one who is rapidly adding and editing documentation.

docs/content/architecture/user-management.md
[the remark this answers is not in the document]
cbandy: How about this?

> PGO generates a random password for each Postgres user it creates. Postgres allows almost any character
> in its passwords, but your application may have stricter requirements. To have PGO generate a password
> without special characters, set the `spec.users.password.type` field for that user to `AlphaNumeric`.
> For complete control over a user's password, see the [custom passwords](#custom-passwords) section.

internal/util/secrets.go
```diff
@@ -15,78 +13,78 @@ package util
...
-	generatedPasswordLength, err := strconv.Atoi(configuredPasswordLength)
-	// if there is an error...set it to a default
-	if err != nil {
-		generatedPasswordLength = DefaultGeneratedPasswordLength
-	}
+// policyASCII is the list of acceptable characters from which to generate an
+// ASCII password.
+const policyASCII = `` +
+	`()*+,-./` + `:;<=>?@` + `[]^_` + `{|}` +
+	`ABCDEFGHIJKLMNOPQRSTUVWXYZ` +
```
benjaminjb, line 81: Do you think it's worthwhile to define AlphaNumeric as a constant, then `+` it here and use in in the var randomAlphanumeric? Like, what does that save? Typing it out twice? It doesn't seem like something we need to worry about typing out multiple times in the future (esp since we can test with a regex). So, probably not worth it, but just wanted to mention it.

benjaminjb (approved):
👍 from me (those other things I brought up are really not necessary fixes, imho)

docs/content/architecture/user-management.md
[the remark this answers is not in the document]
benjaminjb: I like your first version and disagree with the grammar checker :) It makes more sense to me to say "For this effect, do X" -- but I guess I can see the grammar checker's point too. I think a sentence like that answers my worries about people who skim too much/don't know what section they are in.

internal/util/secrets.go
cbandy: In my mind, these two lists/sets are separate and unrelated. Sharing a constant would indicate to me that we intentionally want this one to be a superset of the other.

docs/content/tutorial/user-management.md
cbandy: I went with @jkatz's suggestion to strike "as mentioned earlier". I'm leaving the line length alone for now, too.

internal/controller/postgrescluster/pgmonitor.go
tjmoore4: Sure, if we don't have a clear case, I'm fine for leaving this as is for now.

Issue #2949, opened 2021-12-25
Allow underscore in identifiers, e.g.,`spec.users.name`


kriswuollett:
## Overview

Currently the underscore character, `_`, is not allowed in a username due to the [reUsers regex](https://github.com/CrunchyData/postgres-operator/blob/99fc484e6a459c07ebb26c60c27f89b2f702b1df/internal/controller/postgrescluster/postgres.go#L334), currently `^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`. As can be seen a dash, `-`, is currently not allowed. This is bothersome in postgres both because `_` is allowed as well to be an [identifier](https://www.postgresql.org/docs/14/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS), as well as the use of `-` requires the name token to be a quoted identifier when writing scripts anyhow. The error message does not say why, nor is there any documentation attached to the regex constant for the same.

## Use Case

Allow the use of underscore in postgres identifiers where appropriate.

## Environment

- Kubernetes v1.22.4
- Terraform v1.0.4
- Docker Desktop 4.3.1 macOS
- Postgres 14

### Error Encountered

```
Error: spec.users.name
│ 
│   with [REDACTED...].kubernetes_manifest.database,
│   on  [REDACTED...], in resource "kubernetes_manifest" "database":
│   50: resource "kubernetes_manifest" "database" {
│ 
│ Invalid value: "core_user_admin": spec.users.name in body should match '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
```

### Assumptions

I'm assuming the `reUsers` regex was created with the allowable characters for a Kubernetes `metadata.name` was taken in mind ([RFC 1123](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names)) for generated things like the `db1-pguser-db1-admin`. Its unfortunate to apply restrictions of the wrapper infrastructure to the wrapped service.

But since the `Secret` already has labels like the following already, the human operator should be able to search for:

```
postgres-operator.crunchydata.com/cluster: db1
postgres-operator.crunchydata.com/pguser: db1-admin
postgres-operator.crunchydata.com/role: pguser
```

[...]

PR #2950, opened 2021-12-25
Mention RFC 1123 label in user name error messages

kriswuollett:
**Checklist:**

**Type of Changes:**

 - [ ] New feature
 - [] Bug fix
 - [x] Documentation
 - [ ] Testing enhancement
 - [x] Other

Current error messages with naming of a user do not give a precise reason. In particular this can be an issue for `user.name` since it does not allow `_` which is permitted in the underlying postgres server. This relates to reducing investigation time if  #2949's error is encountered.

The error message also mentions that the name uses the RFC 1123 label constraints

### DIFF
--- a/internal/controller/postgrescluster/postgres.go
+++ b/internal/controller/postgrescluster/postgres.go
@@ -339,13 +339,13 @@ func (r *Reconciler) reconcilePostgresUserSecrets(
 		if n := len(cluster.Name); n > 63 {
 			allErrors = append(allErrors,
 				field.Invalid(path, cluster.Name,
-					fmt.Sprintf("should be at most %d chars long", 63)))
+					fmt.Sprintf("should be at most %d chars long (a RFC 1123 label)", 63)))
 		}
 		// See v1beta1.PostgresRoleSpec validation markers.
 		if !reUser.MatchString(cluster.Name) {
 			allErrors = append(allErrors,
 				field.Invalid(path, cluster.Name,
-					fmt.Sprintf("should match '%s'", reUser)))
+					fmt.Sprintf("should match '%s' (a RFC 1123 label)", reUser)))
 		}
 
 		if len(allErrors) > 0 {


internal/controller/postgrescluster/postgres.go
```diff
@@ -339,13 +339,13 @@ func (r *Reconciler) reconcilePostgresUserSecrets(
 		if n := len(cluster.Name); n > 63 {
 			allErrors = append(allErrors,
 				field.Invalid(path, cluster.Name,
-					fmt.Sprintf("should be at most %d chars long", 63)))
+					fmt.Sprintf("should be at most %d chars long (a RFC 1123 label)", 63)))
 		}
 		// See v1beta1.PostgresRoleSpec validation markers.
```
kriswuollett, line 344: Is there a URL we could put here for v1beta1.PostgresRoleSpec? Was not able to find it during a quick try to do so -- not sure if it was a GitHub search limitation.

jkatz: I'd be inclined to point at Kubernetes naming conventions (i.e. https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names). @cbandy what are your thoughts?

cbandy: These messages only occur when the _cluster_ name is not a valid user name. These lines do not affect the `user.name` field of the spec where we are using Kubernetes' OpenAPI validation.

We call out this limitation in the field description. The [API reference](https://access.crunchydata.com/documentation/postgres-operator/v5/references/crd/#postgresclusterspecusersindex) in the docs might be somewhere to link.

https://github.com/CrunchyData/postgres-operator/blob/99fc484e6a459c07ebb26c60c27f89b2f702b1df/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgres_types.go#L21-L23
https://github.com/CrunchyData/postgres-operator/blob/99fc484e6a459c07ebb26c60c27f89b2f702b1df/config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml#L7562-L7569


As I recall, these messages are intentionally similar to the ones that Kubernetes emits through OpenAPI validation.

cbandy: Unfortunately, `kubectl explain` shows only type, optional/required, and the description.

```console
$ kubectl explain postgrescluster.spec.users
…
FIELDS:
   …

   name <string> -required-
     The name of this PostgreSQL user. The value may contain only lowercase
     letters, numbers, and hyphen so that it fits into Kubernetes metadata.

   …
```

andrewlecuyer:
As described by @cbandy, the error message that has been changed in this PR is related to the _cluster_ name, and _not_ `user.name`.  Therefore, since this change would have no affect on https://github.com/CrunchyData/postgres-operator/issues/2949 as described, proceeding with closing this PR.

If you still have any additional questions or issues related to `user.name` validation, please do not hesitate to reach out.

PR #2960, opened 2022-01-03
KUTTL test: basic pgbouncer connection

benjaminjb:
Add a basic pgbouncer connection KUTTL test:
* create postgrescluster
* connect with required certs

**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

**Other Information**:
Issue [sc-13323]

### DIFF
--- a/testing/kuttl/e2e/pgbouncer/00-assert.yaml
+++ b/testing/kuttl/e2e/pgbouncer/00-assert.yaml
@@ -1,0 +1,15 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: pgbouncer-test
+status:
+  instances:
+    - name: instance1
+      readyReplicas: 2
+      replicas: 2
+      updatedReplicas: 2
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: pgbouncer-test-pgbouncer
--- a/testing/kuttl/e2e/pgbouncer/00-cluster.yaml
+++ b/testing/kuttl/e2e/pgbouncer/00-cluster.yaml
@@ -1,0 +1,29 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: pgbouncer-test
+spec:
+  postgresVersion: 14
+  instances:
+    - name: instance1
+      replicas: 1
+      dataVolumeClaimSpec:
+        accessModes:
+        - "ReadWriteOnce"
+        resources:
+          requests:
+            storage: 1Gi
+  backups:
+    pgbackrest:
+      repos:
+      - name: repo1
+        volume:
+          volumeClaimSpec:
+            accessModes:
+            - "ReadWriteOnce"
+            resources:
+              requests:
+                storage: 1Gi
+  proxy:
+    pgBouncer:
+      replicas: 1
--- a/testing/kuttl/e2e/pgbouncer/01-psql-connect.yaml
+++ b/testing/kuttl/e2e/pgbouncer/01-psql-connect.yaml
@@ -1,0 +1,38 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: psql-connect
+spec:
+  template:
+    spec:
+      backoffLimit: 3
+      restartPolicy: "OnFailure"
+      containers:
+        - name: psql
+          image: registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-14.1-0
+          command:
+            - "bash"
+            - "-c"
+            - "psql -c 'select version();'"
+          env:
+          - name: PGSSLMODE
+            value: verify-full
+          - name: PGSSLROOTCERT
+            value: "/tmp/certs/ca.crt"
+          - name: PGHOST
+            valueFrom: { secretKeyRef: { name: pgbouncer-test-pguser-pgbouncer-test, key: pgbouncer-host } }
+          - name: PGPORT
+            valueFrom: { secretKeyRef: { name: pgbouncer-test-pguser-pgbouncer-test, key: pgbouncer-port } }
+          - name: PGDATABASE
+            valueFrom: { secretKeyRef: { name: pgbouncer-test-pguser-pgbouncer-test, key: dbname } }
+          - name: PGUSER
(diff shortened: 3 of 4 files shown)


config/manager/manager.yaml
```diff
@@ -19,19 +19,19 @@ spec:
         - name: CRUNCHY_DEBUG
           value: "true"
         - name: RELATED_IMAGE_POSTGRES_13
-          value: "registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-13.5-1"
+          value: "registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-13.5-0"
```
cbandy, line 22: 🔧 These should stay the same.

testing/kuttl/e2e/pgbouncer/00-cluster.yaml
```diff
@@ -0,0 +1,40 @@
...
+        volume:
+          volumeClaimSpec:
+            accessModes:
+            - "ReadWriteOnce"
+            resources:
+              requests:
+                storage: 1Gi
+  proxy:
+    pgBouncer:
+      image: registry.developers.crunchydata.com/crunchydata/crunchy-pgbouncer:centos8-1.16-0
```
cbandy, line 29: ❓ Does related image work here? If so, we should omit the image here.

testing/kuttl/e2e/pgbouncer/00-assert.yaml
```diff
@@ -0,0 +1,24 @@
...
+      updatedReplicas: 2
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  labels:
+    postgres-operator.crunchydata.com/cluster: pgbouncer-test
+    postgres-operator.crunchydata.com/pgbackrest-backup: replica-create
+status:
+  succeeded: 1
```
cbandy, line 19: ❓ I didn't expect this job to have any effect on PgBouncer. What does its success mean here?

testing/kuttl/e2e/pgbouncer/01-psql-connect.yaml
```diff
@@ -0,0 +1,49 @@
...
+            privileged: false
+            readOnlyRootFilesystem: true
+            runAsNonRoot: true
+          image: registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-14.1-0
+          command:
+            - "bash"
+            - "-c"
+            # Getting this chown/chmod correct was easier to do with copies of the certs
+            # and in this main container rather than an init
+            - "mkdir -p /pgdata/certs && cp /tmp/certs/* /pgdata/certs && chown postgres /pgdata/certs/* && chmod 600 /pgdata/certs/* && psql -c 'select version();'"
```
cbandy, line 23: 🌱 The `install` command can do most of this in a single step: `-D` to create the target directories and `-m` to set the permissions. In this case, when globbing the source files, probably use `-t /pgdata/certs`.

`chown` is not usually necessary because the created files are owned by the container user.

testing/kuttl/e2e/pgbouncer/00-cluster.yaml
```diff
@@ -0,0 +1,40 @@
...
+      replicas: 1
+      affinity:
```
cbandy, line 31: 🤔 Why specify affinity when `replicas: 1` ? Is there supposed to be an assertion about it?

benjaminjb: No, just copy-paste from the -examples repo (except I tuned the replicas down to `1` -- I could bump it back up, but the pattern we're setting seems to be around focused tests, i.e., we need a test just for affinity...)

testing/kuttl/e2e/pgbouncer/00-assert.yaml
benjaminjb: In the other tests we've used this assert to mean that the cluster is up and running, though taking a step back I wonder if this is necessary/the best way to check that status.

config/manager/manager.yaml
benjaminjb: I thought we had turned some of these back to `-0` for testing because the `-1` images weren't released?

benjaminjb: Reverted

testing/kuttl/e2e/pgbouncer/01-psql-connect.yaml
```diff
@@ -0,0 +1,48 @@
...
+            - "install -D /tmp/certs/ca.crt /tmp/certs/tls.crt /tmp/certs/tls.key -t /pgdata/certs -m 600 && psql -c 'select version();'"
+          env:
+          - name: PGSSLMODE
+            value: verify-full
+          - name: PGSSLCERT 
+            value: "/pgdata/certs/tls.crt"
+          - name: PGSSLKEY
+            value: "/pgdata/certs/tls.key"
+          - name: PGSSLROOTCERT
+            value: "/pgdata/certs/ca.crt"
```
cbandy, line 31: 🔧 PgBouncer is not configured to verify client certificates, so there's no need to present one here. To successfully `verify-full` the client needs only the CA, and that file doesn't need any special permissions.

I believe we can run `psql` without copying the certificate files.

```suggestion
          - name: PGSSLROOTCERT
            value: "/tmp/certs/ca.crt"
```

benjaminjb: Ah, ok, tested and verified, so I could make this test simpler

cbandy (approved):

testing/kuttl/e2e/pgbouncer/00-cluster.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: pgbouncer-test
+spec:
+  postgresVersion: 14
+  instances:
+    - name: instance1
+      replicas: 2
```
jmckulk, line 9: Why have two replicas :question: It doesn't look like it is necessary as part of the pgBouncer checks. Can we reduce it to 1 to make the tests a little lighter?

jmckulk (approved):

PR #2962, opened 2022-01-04
Do not inject service environment variables

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Bug fix

A service starting with `pgbackrest` in the same namespace caused warnings in `pgbackrest` commands:

```
PGBACKREST_WRONG_HA_SERVICE_HOST=10.43.228.168

WARN: environment contains invalid option 'wrong-ha-service-host'
```

All pods are created with `spec.enableServiceLinks: false`. No such environment variables and no more warnings.

**Other Information**:

I verified with the attached Kyverno policy. It's not wired up to anything directly at this time.

Issue: [sc-13361]

### DIFF
--- a/internal/controller/postgrescluster/instance.go
+++ b/internal/controller/postgrescluster/instance.go
@@ -1270,6 +1270,11 @@ func generateInstanceStatefulSetIntent(_ context.Context,
 	// - https://docs.k8s.io/tasks/configure-pod-container/share-process-namespace/
 	sts.Spec.Template.Spec.ShareProcessNamespace = initialize.Bool(true)
 
+	// Patroni calls the Kubernetes API and pgBackRest may interact with a cloud
+	// storage provider. Use the instance ServiceAccount and automatically mount
+	// its Kubernetes credentials.
+	// - https://cloud.google.com/kubernetes-engine/docs/concepts/workload-identity
+	// - https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
 	sts.Spec.Template.Spec.ServiceAccountName = instanceServiceAccountName
 
 	sts.Spec.Template.Spec.SecurityContext = postgres.PodSecurityContext(cluster)
--- a/internal/controller/postgrescluster/instance.go
+++ b/internal/controller/postgrescluster/instance.go
@@ -1277,6 +1277,11 @@ func generateInstanceStatefulSetIntent(_ context.Context,
 	// - https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
 	sts.Spec.Template.Spec.ServiceAccountName = instanceServiceAccountName
 
+	// Disable environment variables for services other than the Kubernetes API.
+	// - https://docs.k8s.io/concepts/services-networking/connect-applications-service/#accessing-the-service
+	// - https://releases.k8s.io/v1.23.0/pkg/kubelet/kubelet_pods.go#L553-L563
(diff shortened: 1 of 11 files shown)


internal/controller/postgrescluster/instance.go
```diff
@@ -1270,8 +1270,19 @@ func generateInstanceStatefulSetIntent(_ context.Context,
...
+	// storage provider. Use the instance ServiceAccount and automatically mount
+	// its Kubernetes credentials.
+	// - https://cloud.google.com/kubernetes-engine/docs/concepts/workload-identity
+	// - https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
 	sts.Spec.Template.Spec.ServiceAccountName = instanceServiceAccountName
 
+	// Disable environment variables for services other than the Kubernetes API.
+	// - https://docs.k8s.io/concepts/services-networking/connect-applications-service/#accessing-the-service
+	// - https://releases.k8s.io/v1.16.0/pkg/kubelet/kubelet_pods.go#L523-L533
+	// - https://releases.k8s.io/v1.23.0/pkg/kubelet/kubelet_pods.go#L553-L563
```
tjmoore4, line 1283: :thinking: I see the two links above are from different releases, but they are otherwise the same. Why do we need both? Is it related to the surrounding code?

internal/controller/postgrescluster/pgbouncer.go
```diff
@@ -323,65 +323,19 @@ func (r *Reconciler) reconcilePGBouncerService(
...
 }
 
-// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get
-// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=create;delete;patch
-
-// reconcilePGBouncerDeployment writes the Deployment that runs PgBouncer.
-func (r *Reconciler) reconcilePGBouncerDeployment(
-	ctx context.Context, cluster *v1beta1.PostgresCluster,
+// generatePGBouncerDeployment returns an appsv1.Deployment that runs PgBouncer pods.
+func (r *Reconciler) generatePGBouncerDeployment(
```
tjmoore4, line 327: :+1: Nice refactor

tjmoore4 (approved):

benjaminjb (approved):
LGTM -- as near as I can tell, this is only preventing the service from auto-mounting env vars, but now I'm curious how serviceaccounts mount env vars as in the EKS IAM case

internal/controller/postgrescluster/instance.go
tjmoore4: Ah, I see. Maybe just a note that this field has been available since 1.13 or similar.

To the second find, I'm not sure if that's worth documenting here as it isn't really related to the update, but I suppose it couldn't hurt to have it either.

cbandy:
> LGTM -- as near as I can tell, this is only preventing the service from auto-mounting env vars, but now I'm curious how serviceaccounts mount env vars as in the EKS IAM case.

EKS has a [webhook that mutates Pods](https://github.com/aws/amazon-eks-pod-identity-webhook/blob/a65cc3d9c61cf6fc43f0f985818c474e0867d786/deploy/mutatingwebhook.yaml).

- https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html#pod-configuration

internal/controller/postgrescluster/instance.go
cbandy: I've removed the link to v1.16 and mentioned v1.13 in the commit message.

PR #2966, opened 2022-01-06
Postgres Major Upgrade E2E

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

Adds a KUTTL E2E test to verify the Postgres major upgrade feature
works as expected. A standard Postgres 13 PostgresCluster is created
and validated then upgraded to Postgres 14 and validated again. The
expected pgBackRest backup Jobs and the pg_upgrade Job are also
checked to ensure they complete successfully.

This update also fixes a syntax error in the KUTTL
README. The flag should be `--suppress-log` not
`--suppress-logs`.

**Other Information**:
Issue: [sc-13182]

### DIFF
--- a/testing/kuttl/e2e/major-upgrade/02-assert.yaml
+++ b/testing/kuttl/e2e/major-upgrade/02-assert.yaml
@@ -1,0 +1,30 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: major-upgrade
+spec:
+  postgresVersion: 14
+status:
+  instances:
+    - name: instance1
+      readyReplicas: 1
+      replicas: 1
+      updatedReplicas: 1
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  labels:
+    postgres-operator.crunchydata.com/cluster: major-upgrade
+    postgres-operator.crunchydata.com/pgupgrade: ""
+status:
+  succeeded: 1
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  labels:
+    postgres-operator.crunchydata.com/cluster: major-upgrade
+    postgres-operator.crunchydata.com/pgbackrest-backup: replica-create
+status:
+  succeeded: 1
--- a/testing/kuttl/e2e/major-upgrade/README.md
+++ b/testing/kuttl/e2e/major-upgrade/README.md
@@ -1,0 +1,5 @@
+### Postgres major upgrade test
+* 00: Create a standard Postgres 13 cluster; verify that the cluster instance is created as expected and that the initial backup job completed successfully.
+* 01: Run postgres-version-check.sh, which verifies the expected Postgres version is returned.
+* 02: Update the spec from step 00 to upgrade from 13 to 14; verify that the instance is ready, the version on the spec is correct, and both the backup and upgrade Jobs have completed successfully.
+* 03: Rerun postgres-version-check.sh to verify the expected Postgres version is returned.
--- a/testing/kuttl/scripts/postgres-version-check.sh
+++ b/testing/kuttl/scripts/postgres-version-check.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+EXPECTED_VERSION=$1
+
+CLUSTER=${CLUSTER:-default}
+
+POSTGRES_POD=$(kubectl -n ${NAMESPACE} get pod --selector=postgres-operator.crunchydata.com/cluster=${CLUSTER},postgres-operator.crunchydata.com/instance-set=instance1 --no-headers -o custom-columns=":metadata.name")
+
+# Grab the Postgres major version from `version()` output to compare to the provided value
+VERSION=$(kubectl -n ${NAMESPACE} exec -i ${POSTGRES_POD} -c database -- psql -tc "SELECT version();" |
+    xargs |
+    cut -d' ' -f 2 |
+    cut -d. -f 1)
+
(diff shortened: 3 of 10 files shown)


testing/kuttl/e2e/major-upgrade/01--postgres-connect.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kuttl.dev/v1beta1
```
benjaminjb, line 1: 💭 but also ⛏️  and :thinking_dinosaur: This may be a trivial issue, but in my conception, we have two types of files: files to arrange the situation (create this cluster, update this cluster), and files to test this situation (make sure the cluster exists, make sure the cluster doesn't exist). 

To me, a script that checks the version is a test file, so should be `01-postgres-connect` rather than `01--postgres-connect`. What do you think?

testing/kuttl/e2e/major-upgrade/02-assert.yaml
```diff
@@ -0,0 +1,30 @@
...
+  succeeded: 1
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  labels:
+    postgres-operator.crunchydata.com/cluster: major-upgrade
+    postgres-operator.crunchydata.com/pgbackrest-backup: replica-create
+status:
+  succeeded: 1
```
benjaminjb, line 30: Is this the same replica-create that you tested in 00-assert.yaml? Do we need this check?

testing/kuttl/e2e/major-upgrade/README.md
```diff
@@ -0,0 +1,6 @@
+### Postgres major upgrade test
+update this....
```
benjaminjb, line 2: ⛏️

testing/kuttl/scripts/postgres-version-check.sh
```diff
@@ -0,0 +1,19 @@
...
+
+EXPECTED_VERSION=$1
+
+CLUSTER=${CLUSTER:-default}
+
+POSTGRES_POD=$(kubectl -n ${NAMESPACE} get pod --selector=postgres-operator.crunchydata.com/cluster=${CLUSTER},postgres-operator.crunchydata.com/instance-set=instance1 --no-headers -o custom-columns=":metadata.name")
+
+# Grab the Postgres major version from `version()` output to compare to the provided value
+VERSION=$(kubectl -n ${NAMESPACE} exec -i ${POSTGRES_POD} -c database -- psql -c "SELECT version();" |
+    grep PostgreSQL |
```
benjaminjb, line 11:
```suggestion
VERSION=$(kubectl -n ${NAMESPACE} exec -i ${POSTGRES_POD} -c database -- psql -tc "SELECT version();" |
```
`-t` flag will drop the header of the response (tuple only)

benjaminjb (approved):
A few questions, but looks good overall!

jmckulk (approved):

testing/kuttl/e2e/major-upgrade/02-assert.yaml
tjmoore4: No, there should be another backup Job after the upgrade, so this is testing a separate step.

testing/kuttl/e2e/major-upgrade/README.md
tjmoore4: Got it, thanks!

testing/kuttl/scripts/postgres-version-check.sh
tjmoore4: :+1: Works for me, updated.

testing/kuttl/e2e/major-upgrade/01--postgres-connect.yaml
tjmoore4: I went back and forth a bit about this, and I think I could go either way. I'll go ahead and remove the extra `-` since I noticed that's where we've  already landed.

testing/kuttl/e2e/major-upgrade/02-assert.yaml
benjaminjb: But the labels here are the same as in 00-assert lines 18-19, so would we see 2 successful pods from this job?

tjmoore4: There shouldn't be. The upgrade utilizes the shutdown process, so the pods are removed before the upgrade Job is run. In testing, I only saw one backup Pod at a time.

testing/kuttl/scripts/postgres-version-check.sh
```diff
@@ -0,0 +1,19 @@
...
+
+EXPECTED_VERSION=$1
+
+CLUSTER=${CLUSTER:-default}
+
+POSTGRES_POD=$(kubectl -n ${NAMESPACE} get pod --selector=postgres-operator.crunchydata.com/cluster=${CLUSTER},postgres-operator.crunchydata.com/instance-set=instance1 --no-headers -o custom-columns=":metadata.name")
+
+# Grab the Postgres major version from `version()` output to compare to the provided value
+VERSION=$(kubectl -n ${NAMESPACE} exec -i ${POSTGRES_POD} -c database -- psql -tc "SELECT version();" |
+    grep PostgreSQL |
```
benjaminjb, line 11: Can cut this line?

tjmoore4: Yes, it can be. I'll update this.

PR #2974, opened 2022-01-10
Envtest fixes

jmckulk:
A few changes to get envtests running against an existing GKE cluster
- When checking for the deletion timestamp it is possible for the item to
not be found (i.e. is already deleted). Ignore not found errors
- Import GCP client auth package
- We no longer write external IPs for the primary cluster service. This
test can be removed.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

**Other Information**:
[sc-13356]

### DIFF
--- a/internal/controller/postgrescluster/apply_test.go
+++ b/internal/controller/postgrescluster/apply_test.go
@@ -341,56 +341,6 @@ func TestServerSideApply(t *testing.T) {
 		}
 	})
 
-	t.Run("ServiceSpec", func(t *testing.T) {
-		constructor := func(name string) *corev1.Service {
-			var service corev1.Service
-			service.SetGroupVersionKind(corev1.SchemeGroupVersion.WithKind("Service"))
-			service.Namespace, service.Name = ns.Name, name
-			service.Spec.Ports = []corev1.ServicePort{{
-				Port: 9999, Protocol: corev1.ProtocolTCP,
-			}}
-			return &service
-		}
-
-		reconciler := Reconciler{Client: cc, Owner: client.FieldOwner(t.Name())}
-
-		// Start with fields filled out.
-		intent := constructor("change-to-zero")
-		intent.Spec.ExternalIPs = []string{"10.9.8.7", "192.0.2.10"}
-
-		// Create the StatefulSet.
-		before := intent.DeepCopy()
-		assert.NilError(t,
-			cc.Patch(ctx, before, client.Apply, client.ForceOwnership, reconciler.Owner))
-
-		// Change fields to zero.
-		intent.Spec.ExternalIPs = nil
-
-		// client.Apply cannot correct it in old versions of Kubernetes.
-		after := intent.DeepCopy()
-		assert.NilError(t,
-			cc.Patch(ctx, after, client.Apply, client.ForceOwnership, reconciler.Owner))
-
-		switch {
-		case serverVersion.LessThan(version.MustParseGeneric("1.18.19")):
-
-			// - https://pr.k8s.io/101179
-			assert.Assert(t, !equality.Semantic.DeepEqual(
-				after.Spec.ExternalIPs,
-				intent.Spec.ExternalIPs),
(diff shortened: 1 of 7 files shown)


internal/upgradecheck/header_test.go
```diff
@@ -34,7 +34,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/version"
+
+	// Google Kubernetes Engine / Google Cloud Platform authentication provider
 	"k8s.io/client-go/discovery"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
cbandy, line 40: ⛏️ Something moved the comment away from the import. I remember something irritating about this... a linter does the wrong thing, maybe?

internal/controller/postgrescluster/postgres_test.go
```diff
@@ -414,8 +414,14 @@ volumeMode: Filesystem
...
-					assert.NilError(t, tClient.Get(ctx, key, fetched))
-					assert.Assert(t, fetched.DeletionTimestamp != nil, "expected deleted")
+					err = tClient.Get(ctx, key, fetched)
+					assert.NilError(t, client.IgnoreNotFound(err))
+					// If we have a NotFound error the item has been deleted
+					if !apierrors.IsNotFound(err) {
+						// If we don't have a NotFound error, the item still exists
+						// check that the item is marked for deletion
+						assert.Assert(t, fetched.DeletionTimestamp != nil, "expected deleted")
+					}
```
cbandy, line 424: 🤔 I thought we patched this already, but I guess not.

The checks for `IgnoreNotFound` and `IsNotFound` are a little confusing. Is the following any better? It is more clear to me that we look in `fetched` when it is populated.

```suggestion
					if err := tClient.Get(ctx, key, fetched); err == nil {
						assert.Assert(t, fetched.DeletionTimestamp != nil, "expected deleted")
					} else {
						assert.Assert(t, apierrors.IsNotFound(err), "expected NotFound, got %v", err)
					}
```

internal/controller/postgrescluster/apply_test.go
```diff
@@ -341,56 +341,6 @@ func TestServerSideApply(t *testing.T) {
...
-
-		switch {
-		case serverVersion.LessThan(version.MustParseGeneric("1.18.19")):
-
-			// - https://pr.k8s.io/101179
-			assert.Assert(t, !equality.Semantic.DeepEqual(
-				after.Spec.ExternalIPs,
-				intent.Spec.ExternalIPs),
-				"expected https://issue.k8s.io/89273, got %v",
-				after.Spec.ExternalIPs)
```
cbandy, line 382: 🤔 We should probably remove the code that this is testing, too.

https://github.com/CrunchyData/postgres-operator/blob/a7ce6ee7389967d02f7346e10e2f980b8a5cc21f/internal/controller/postgrescluster/apply.go#L170-L174

internal/upgradecheck/header_test.go
jmckulk: Yeah I had them together at the top and the linter moved the import without the comment. If I put the comment under the discovery import they seem to stick together

internal/controller/postgrescluster/apply_test.go
jmckulk: Oh, I thought that was already done. I can remove

internal/controller/postgrescluster/postgres_test.go
jmckulk: I think that is more clear :+1:

cbandy (approved):

benjaminjb (approved):

jmckulk:
Just to note; there are still a few flakes in the tests but this fixes a few issues.

PR #2977, opened 2022-01-12
Add end-to-end tests for clone and restore

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

```
--- PASS: kuttl/harness/pgbackrest-restore (180.39s)
```

**Other Information**:

Issue: [sc-12476]
Issue: [sc-13129]
Issue: [sc-13322]

### DIFF
--- a/testing/kuttl/e2e/pgbackrest-restore/07--update-cluster.yaml
+++ b/testing/kuttl/e2e/pgbackrest-restore/07--update-cluster.yaml
@@ -1,0 +1,24 @@
+---
+# Update the cluster with PostgreSQL parameters that require attention during recovery.
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: original
+  labels: { postgres-operator-test: kuttl }
+spec:
+  postgresVersion: 14
+  patroni:
+    dynamicConfiguration:
+      postgresql:
+        parameters:
+          max_connections: 1000
+  instances:
+    - dataVolumeClaimSpec: { accessModes: [ReadWriteOnce], resources: { requests: { storage: 1Gi } } }
+  backups:
+    pgbackrest:
+      manual:
+        repoName: repo1
+      repos:
+        - name: repo1
+          volume:
+            volumeClaimSpec: { accessModes: [ReadWriteOnce], resources: { requests: { storage: 1Gi } } }
--- a/testing/kuttl/e2e/pgbackrest-restore/07--update-cluster.yaml
+++ b/testing/kuttl/e2e/pgbackrest-restore/07--update-cluster.yaml
@@ -14,6 +14,7 @@ spec:
           max_connections: 1000
   instances:
     - dataVolumeClaimSpec: { accessModes: [ReadWriteOnce], resources: { requests: { storage: 1Gi } } }
+      replicas: 2
   backups:
     pgbackrest:
       manual:
--- a/testing/kuttl/e2e/pgbackrest-restore/14--lose-data.yaml
+++ b/testing/kuttl/e2e/pgbackrest-restore/14--lose-data.yaml
@@ -1,0 +1,55 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      PRIMARY=$(
+        kubectl get pod --namespace "${NAMESPACE}" \
+          --output name --selector '
+            postgres-operator.crunchydata.com/cluster=original,
+            postgres-operator.crunchydata.com/role=master'
+      )
+      OBJECTIVE=$(
+        kubectl exec --namespace "${NAMESPACE}" "${PRIMARY}" \
+          -- psql -qAt --command 'SELECT clock_timestamp()'
+      )
+
+      # Store the recovery objective for later steps.
+      kubectl annotate --namespace "${NAMESPACE}" postgrescluster/original \
+        "testing/objective=${OBJECTIVE}"
+
+      # A reason to restore. Wait for the change to be sent to the WAL archive.
+      kubectl exec --namespace "${NAMESPACE}" "${PRIMARY}" \
+        -- psql -qb original --set ON_ERROR_STOP=1 \
+          --command 'DROP TABLE important' \
+          --command "SELECT pg_stat_reset_shared('archiver')" \
+          --command '
+            DO $$
+            BEGIN
+              LOOP
+                PERFORM pg_switch_wal() FROM pg_stat_archiver WHERE archived_count = 0;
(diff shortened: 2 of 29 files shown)


testing/kuttl/e2e/pgbackrest-restore/16--check-data.yaml
```diff
@@ -0,0 +1,100 @@
...
+            - --set=ON_ERROR_STOP=1
+            - --command
+            - |
+              DO $$$$
+              DECLARE
+                restored jsonb;
+              BEGIN
+                ASSERT pg_is_in_recovery(), 'expected replica';
+                -- only users with "pg_read_all_settings" role may examine "primary_conninfo"
+                -- ASSERT current_setting('primary_conninfo') <> '', 'expected streaming';
```
benjaminjb, line 94: 💭 So could we test this with a `kubectl exec ... -- psql` command?

testing/kuttl/e2e/pgbackrest-restore/07--update-cluster.yaml
```diff
@@ -0,0 +1,25 @@
+---
```
benjaminjb, line 1: 💭 Do we know which of the 07 files will get run first? Is it always in alphabetic order (followed by the assert/error.yaml files)?

benjaminjb (approved):
LGTM, a few questions, and a few things for me to look up about postgres (mostly around pg_stat_archiver). Not sure if this needs a readme -- it's a long process, but the steps are detailed enough to follow.

testing/kuttl/e2e/pgbackrest-restore/07--update-cluster.yaml
cbandy: `TestStep` commands always [run first](https://github.com/kudobuilder/kuttl/blob/d2cd0d5ec86a758ca4a13db032a89ae214d3d524/pkg/test/step.go#L411) (say [the docs](https://kuttl.dev/docs/testing/reference.html#teststep)).

Also, despite KEP 4, there can be only one `TestStep` per step.

testing/kuttl/e2e/pgbackrest-restore/16--check-data.yaml
cbandy: We could. Should I add it?

There are a lot of "shell" steps in this test case, and sometimes it felt like cheating (?) ... All the data _could_ be checked by the superuser, but it seems more appropriate to come at it from the user account.

🤔 I'm thinking now it might be good to have a line in `14--lose-data` that shows the replica has lost data, too. (One more script that comes at it as the superuser.)

benjaminjb: Oh, that (showing the replica has lost data) makes a lot of sense; and yeah, as someone who does a lot of manual testing as both regular and super-user, I support checking as much from the regular user as possible. 

I'm not even sure I'd argue _for_ a special `kubectl exec` to check the `primary_conninfo`, I just thought I'd mentton it as it's pretty easy to do with KUTTL as a script (rather than job). That said, maybe it will save us a headache later if we make these tests  check as many elements as is reasonable? Any thoughts?

testing/kuttl/e2e/pgbackrest-restore/07--update-cluster.yaml
benjaminjb: Ah, ok, as long as it's deterministic, I don't know that we need to separate them out with different numbers. 

(Also do people write teststeps that do deletes, then commands, then applies? I wonder when that would make sense/be more legible or maintainable vs. just separating it out over different index numbers.)

testing/kuttl/e2e/pgbackrest-restore/16--check-data.yaml
cbandy: Done. Added the check to `14--lose-data` and added a new `17--check-replication` that verifies the settings that are visible only to a superuser.

testing/kuttl/e2e/pgbackrest-restore/07--update-cluster.yaml
cbandy: > (Also do people write teststeps that do deletes, then commands, then applies? I wonder when that would make sense/be more legible or maintainable vs. just separating it out over different index numbers.)

I can't imagine they do. I suspect it's more that assertions should happen after everything else. "Everything else" gets lumped into one thing with a deterministic order.

I've wondered a bit about deletes. The default for new things (and definitely CRs) is [background deletion](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#cascading-deletion). Kuttl docs say it waits, but for what? Does it always use foreground deletion?

testing/kuttl/e2e/pgbackrest-restore/14--lose-data.yaml
```diff
@@ -0,0 +1,55 @@
...
+      )
+
+      kubectl exec --stdin --namespace "${NAMESPACE}" "${REPLICA}" \
+        -- psql -qb original --set ON_ERROR_STOP=1 \
+          --file=- <<'SQL'
+        DO $$
+        BEGIN
+          ASSERT to_regclass('important') IS NULL, 'expected no table';
+          PERFORM * FROM information_schema.tables WHERE table_name = 'important';
+          ASSERT NOT FOUND, 'expected no table';
```
benjaminjb, line 53: For my own education, is there a case where line 51 would pass, but 53 would fail?

benjaminjb (approved):

testing/kuttl/e2e/pgbackrest-restore/14--lose-data.yaml
cbandy: Not that I know of. I did both because they both seem kinda obscure. `to_regclass` is the most direct but the least obvious to me.

I chose "dropped the wrong table" as a typical data-loss scenario but writing a positive assertion that it happened is awkward. I considered also changing from `DROP` to `DELETE` so that these assertions could be "no rows in table". I think the full assertion here would become: (no regclass)

```plpgsql
PERFORM * FROM important;
ASSERT NOT FOUND, 'expected no data';
```

Would that be preferable?

benjaminjb: Speaking of "obscure," I definitely had to go look up `to_regclass`, but I think this test is good.

(I'm trying to remember whether I have more often erased rows I didn't mean to or dropped whole tables by accident, and it's probably 50/50 in PG. (In Mongo, it's much more often to hear someone say, "oh no, I erased how many documents?") So I think dropping the table is a good test scenario.)

PR #2980, opened 2022-01-13
E2E Test for GSSAPI Authentication With PG

andrewlecuyer:
Adds an end-to-end test to verify the ability to properly configure PostgreSQL for GSSAPI authentication.  This is specifically done using Kerberos, i.e. a ticket issued by a Kerberos KDC server is utilized to log into PostgreSQL when running the test.

Please note that this specific test has been added to a `e2e-other` directory, since additional configuration and infrastructure (e.g. a a running Kerberos Key Distribution Center) must be in place in order to successfully run the test.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

There is no E2E test for authenticating into PG using GSSAPI.

[sc-13319]

There is now an E2E test for authenticating into PG using GSSAPI.

**Other Information**:

N/A

### DIFF
--- a/testing/kuttl/e2e-other/gssapi/00-krb5-keytab.yaml
+++ b/testing/kuttl/e2e-other/gssapi/00-krb5-keytab.yaml
@@ -1,0 +1,4 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- command: kubectl exec -n krb5 -it krb5-kdc-0 -- /krb5-scripts/krb5.sh "${NAMESPACE}"
--- a/testing/kuttl/e2e-other/gssapi/00-assert.yaml
+++ b/testing/kuttl/e2e-other/gssapi/00-assert.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: krb5
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: krb5-keytab
--- a/testing/kuttl/e2e-other/gssapi/01-assert.yaml
+++ b/testing/kuttl/e2e-other/gssapi/01-assert.yaml
@@ -1,0 +1,15 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: gssapi
+status:
+  instances:
+    - name: instance1
+      readyReplicas: 1
+      replicas: 1
+      updatedReplicas: 1
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: gssapi-primary
--- a/testing/kuttl/e2e-other/gssapi/01-cluster.yaml
+++ b/testing/kuttl/e2e-other/gssapi/01-cluster.yaml
@@ -1,0 +1,41 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: gssapi
+spec:
+  config:
+    files:
+    - secret:
+        name: krb5-keytab
+    - configMap:
+        name: krb5
+  patroni:
+    dynamicConfiguration:
+      postgresql:
+        pg_hba:
+        - host postgres postgres 0.0.0.0/0 scram-sha-256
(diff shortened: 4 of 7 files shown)


tjmoore4 (approved):

testing/kuttl/e2e-other/gssapi/00-krb5-keytab.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- command: kubectl exec -n krb5 -it krb5-kdc-0 -- /krb5-scripts/krb5.sh "${NAMESPACE}"
```
jmckulk, line 4: I wonder if we could pull this step out of the tests. Could we just rely on the assert to ensure that secrets, configmaps, services, deployments, etc exist and have the correct info? :thinking:

jmckulk (approved):

testing/kuttl/e2e-other/gssapi/00-krb5-keytab.yaml
andrewlecuyer: Not only does this script create the ConfigMaps and Secrets required for this test, it also populates the Kerberos database itself with the principals needed to run the test.  And since once of those principal's relates to the primary service that is used to connect to the DB (e.g. `postgres/gssapi-primary.${NAMESPACE}.svc.cluster.local@PGO.CRUNCHYDATA.COM`), we're forced to provide a namespace name when the test is run (which is not known until the test is run considering Kuttl generates the namespace).  Therefore, I think we're stuck initiating some of the setup from Kuttl to provide any test-run specific details (namespace names, Pod names, etc.).

However, the above is assuming we're using a namespace that has been randomly generated by Kuttl ...  maybe instead this should be a test where we're opinionated about the name of namespace in favor of keeping _all_ Kerberos setup/configuration out of the test (while also removing dependencies on a specific Kerberos container/implementation and keeping the Kerberos setup/configuration the same/consistent across all test runs)?

cbandy (approved):

testing/kuttl/e2e-other/gssapi/00-krb5-keytab.yaml
benjaminjb: I wonder if there's an easy way to set the namespace (and build it if it doesn't exist) just for one test, or would this require a separate `test-kuttl.yaml` file for this one test? (That's not a huge amount of overhead, I just wonder what best practice is.)

https://kuttl.dev/docs/testing/test-environments.html#environment-setup

benjaminjb: That said, what first creates this `krb5-kdc` pod with that script on it?

benjaminjb (approved):
Just some questions that you already have about the infrastructure that needs to be setup before the test, but I like putting this in an `e2e-other` folder.

testing/kuttl/e2e-other/gssapi/00-krb5-keytab.yaml
jmckulk: I don't think it would be hard to have a set namespace for a single test :thinking: 

maybe that is a future improvement

andrewlecuyer: @benjaminjb our CI infrastructure will be responsible for deploying the Kerberos KDC server a required for this this test, which is specifically run inside of a `krb5-kdc-0` Pod.  Instructions for deploying the KDC sever (and therefore creating the `krb5-kdc-0` Pod) can be found in CrunchyData/priv-all-pgo-ci#2.

@benjaminjb @jmckulk I really like the idea of hardcoding a namespace as a way to move more of the setup/configuration outside of the test (and into the CI Job).  This would remove any coupling to a specific KDC implementation/deplyoment, while also keeping configuration/setup the same for various tests runs.  Therefore, definitely something to consider for a future iteration of this test.

PR #2981, opened 2022-01-13
Move some test helpers to an internal package

cbandy:
Multiple tests and packages have the same logic or helper functions. Move some of them into a common internal package and improve them somewhat.

- A function to find the `shellcheck` executable or skip the test. The function looks up the executable once per package rather than once per test.
- A single implementation of the ubiquitous `marshalMatches`.
- A function that shows a diff when one multi-line string does not contain another.
- Improved documentation for `cmp.DeepEqual` and `cmp.Regexp`.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement


jmckulk (approved):
lgtm

internal/testing/cmp/cmp.go
```diff
@@ -0,0 +1,76 @@
...
+type Comparison = gotest.Comparison
+
+// Contains succeeds if item is in collection. The collection may be a string,
+// map, slice, or array. See [gotest.tools/v3/assert/cmp.Contains].
+func Contains(collection, item interface{}) Comparison {
+	cString, cStringOK := collection.(string)
+	iString, iStringOK := item.(string)
+
+	if cStringOK && iStringOK {
+		if strings.Contains(cString, "\n") || strings.Contains(iString, "\n") {
```
benjaminjb, line 35: Ah, so if this is a single line string, we just use the line 51 implementation of Contains, yeah?

internal/testing/cmp/cmp.go
```diff
@@ -0,0 +1,76 @@
...
+
+	gocmp "github.com/google/go-cmp/cmp"
+	gotest "gotest.tools/v3/assert/cmp"
+	"sigs.k8s.io/yaml"
+)
+
+type Comparison = gotest.Comparison
+
+// Contains succeeds if item is in collection. The collection may be a string,
+// map, slice, or array. See [gotest.tools/v3/assert/cmp.Contains].
```
benjaminjb, line 29: I would add here the comment in the PR re: why this change for multiline strings

internal/testing/require/exec.go
```diff
@@ -0,0 +1,66 @@
...
+			if err != nil {
+				result = func(t testing.TB) string {
+					t.Helper()
+					t.Skipf("requires %q executable", name)
+					return ""
+				}
+			} else if info, err := cmd.CombinedOutput(); err != nil {
+				result = func(t testing.TB) string {
+					t.Helper()
+					assert.NilError(t, err, "%q\n%s", cmd.Args, info)
```
benjaminjb, line 53: Could we call t.Fatal directly rather than `assert.NilError` on an error? (Not sure this would actually be preferable.)

benjaminjb (approved):
"A single implementation of the ubiquitous `marshalMatches`."
💯 

A couple of questions, but no blockers, really nice to see this, thank you!

internal/pgbackrest/helpers_test.go
```diff
@@ -16,20 +16,10 @@
...
-	"gotest.tools/v3/assert/cmp"
-
-	// Google Kubernetes Engine / Google Cloud Platform authentication provider
-	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
-	"sigs.k8s.io/yaml"
+	"github.com/crunchydata/postgres-operator/internal/testing/cmp"
 )
 
 // marshalMatches converts actual to YAML and compares that to expected.
 func marshalMatches(actual interface{}, expected string) cmp.Comparison {
```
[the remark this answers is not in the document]
cbandy: Looks like a dozen times versus just once.

```
● ag marshalMatches -- internal/pgbackrest
internal/pgbackrest/helpers_test.go
22:// marshalMatches converts actual to YAML and compares that to expected.
23:func marshalMatches(actual interface{}, expected string) cmp.Comparison {

internal/pgbackrest/reconcile_test.go
162:            assert.Assert(t, marshalMatches(result.Containers, `
194:            assert.Assert(t, marshalMatches(out.Volumes, `
219:            assert.Assert(t, marshalMatches(out.Volumes, `
247:            assert.Assert(t, marshalMatches(out.Volumes, `
292:            assert.Assert(t, marshalMatches(result.Containers, `
319:            assert.Assert(t, marshalMatches(out.Volumes, `
365:            assert.Assert(t, marshalMatches(result.Containers, `
392:            assert.Assert(t, marshalMatches(out.Volumes, `
462:            assert.Assert(t, marshalMatches(out.Containers, `
538:            assert.Assert(t, marshalMatches(out.Volumes, `
594:            assert.Assert(t, marshalMatches(out.Containers, `
663:            assert.Assert(t, marshalMatches(out.Volumes, `
```

```
● ag MarshalMatches -- internal/pgbackrest
internal/pgbackrest/helpers_test.go
24:     return cmp.MarshalMatches(actual, expected)

internal/pgbackrest/rbac_test.go
51:     assert.Assert(t, cmp.MarshalMatches(permissions, `
```

internal/controller/postgrescluster/helpers_test.go
```diff
@@ -67,11 +65,7 @@ func init() {
 
 // marshalMatches converts actual to YAML and compares that to expected.
 func marshalMatches(actual interface{}, expected string) cmp.Comparison {
```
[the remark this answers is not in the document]
cbandy: This one is used over 40 times. I wanted to avoid the conflicts that would come from such a big change. From the commit message:

> Replacing every call to "marshalMatches" is very likely to conflict with other work, so leave those names in place for now.

```
● ag -c marshalMatches -- internal/controller/postgrescluster
internal/controller/postgrescluster/volumes_test.go:3
internal/controller/postgrescluster/cluster_test.go:10
internal/controller/postgrescluster/topology_test.go:1
internal/controller/postgrescluster/pgbackrest_test.go:2
internal/controller/postgrescluster/pgbouncer_test.go:6
internal/controller/postgrescluster/postgres_test.go:2
internal/controller/postgrescluster/helpers_test.go:2
internal/controller/postgrescluster/instance_test.go:8
internal/controller/postgrescluster/pgadmin_test.go:9
internal/controller/postgrescluster/patroni_test.go:3
```

internal/testing/cmp/cmp.go
cbandy: Right!

internal/testing/cmp/cmp.go
cbandy: 🤔 "why this change" ... Do I understand correctly: The godoc here should say that this function does something for multi-line strings.

I'm having a hard time deciding what to add. [gotest.tools/v3/assert/cmp](https://pkg.go.dev/gotest.tools/v3/assert/cmp) doesn't talk about what is in its failure messages.

> Contains succeeds if item is in collection. The collection may be a string, map, slice, or array. See [gotest.tools/v3/assert/cmp.Contains].
>
> When item is missing from collection and both are multi-line strings, … the failure looks like a diff. (?)

benjaminjb: Yeah, I was thinking of something like the PR comment re: “shows a diff when one multi-line string does not contain another.”

(but also definitely not a sticking point, just something I thought of reading the PR comment and the code comments.)

internal/testing/require/exec.go
cbandy: Certainly. I was deferring to `gotest` to inspect and format the error. I believe that amounts to:

```go
t.Fatalf("error is not nil: %s: %q\n%s", err.Error(), cmd.Args, info)
```

- https://github.com/gotestyourself/gotest.tools/blob/v3.1.0/internal/assert/assert.go#L51-L53
- https://github.com/gotestyourself/gotest.tools/blob/v3.1.0/internal/assert/assert.go#L114

Is that preferable?

benjaminjb: No, I think it’s fine, I was just curious why you chose to assert.NilError. I really only noticed because the code comment specifically calls out t.Fatal

cbandy: I agree, that isn't obvious. I'll add a comment here about it.

internal/testing/cmp/cmp.go
cbandy: Done in amended https://github.com/CrunchyData/postgres-operator/pull/2981/commits/ea63ca00f3dcbaf0b8c770a7a409cbc4ee13e8c4.

internal/testing/require/exec.go
cbandy: Done in amended [`e4f3f71`](https://github.com/CrunchyData/postgres-operator/pull/2981/commits/e4f3f71f73622e0307cb166f9070856bbcb729f9).

benjaminjb (approved):
Thx for the extra comments!

PR #2984, opened 2022-01-14
envsubst for KUTTL

benjaminjb:
Add a /source/ folder with an example KUTTL test 
and make targets used to create generated KUTTL tests with envsubst
and run those generated tests

**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

A potential model for further parameterizing tests with envsubst.

Files generated with 
`KUTTL_PSQL_VERSION=13 KUTTL_PSQL_IMAGE=registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-13.5-0 make generate-kuttl`
`KUTTL_PSQL_VERSION=14 KUTTL_PSQL_IMAGE=registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-14.1-0 make generate-kuttl`

**Other Information**:
Issue [sc-13318]


testing/kuttl/kuttl-test-generated.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestSuite
+testDirs:
+- testing/kuttl/generated/13
```
jmckulk, line 4: would we want multiple directories with the generated tests :question:  I was thinking we would just have one set of tests committed. If you wanted to test a different version than the one committed you would generate the files with the make target and a new set of versions/images/etc :thought_balloon: We maybe we could even `gitignore` the files so changes aren't annoying in git?

testing/kuttl/source/cluster-start/01-psql-connect.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: Job
+metadata:
+  name: psql-connect
+spec:
+  template:
+    spec:
+      restartPolicy: "OnFailure"
+      containers:
+        - name: psql
+          image: ${KUTTL_PSQL_IMAGE}
```
jmckulk, line 11: I wonder if we could point to the latest upgrade image or even build our testing image? All we need in the container are the clients, psql, pgbackrest, krb5, etc. :thinking:

jmckulk (commented):
Should this have makefile changes?

testing/kuttl/kuttl-test-generated.yaml
benjaminjb: (a) multiple directories -- I don't think so, and I've redone the makefile* so we're adding tests into one `generated` folder, so we don't have to keep changing this file as we add new versions

* Argh, there should've been a makefile added
* 
(b) I definitely can see the appeal of not adding versions_supported * files_needed_to_test number of files each time we add an e2e test. Heck, I'm not sure how you've been using KUTTL, but I've been doing single test runs, so maybe we leave the generated folder completely gitignore'd and the user can generate the tests/versions they want to test? (Meanwhile our CI is running all of them.)

testing/kuttl/source/cluster-start/01-psql-connect.yaml
benjaminjb: But wouldn't we want the psql client to be the same version as the PG pod?

testing/kuttl/kuttl-test-generated.yaml
benjaminjb: I am adding tests in the `generated` folder, but more just to see what they would be like.

testing/kuttl/source/cluster-start/01-psql-connect.yaml
jmckulk: Not necessarily, I think Postgres allows you to use a newer client version with an older server. 

https://www.postgresql.org/docs/14/app-psql.html
> psql works best with servers of the same or an older major version

testing/kuttl/kuttl-test-generated.yaml
jmckulk: To point (b) - I like the idea of not having anything committed a lot. You would have to run the generate target every time you checkout a new version to ensure you have the latest changes. I guess that could lead to testing with the wrong generated files if I forget :thinking: 

What are your thoughts on generating the files as part of the `make check-kuttl` target? 
```
.PHONY: check-kuttl
check-kuttl: generate-kuttl
	${PGO_KUBE_CLIENT} ${KUTTL_TEST} --config testing/kuttl/kuttl-test.yaml
```
The workflow could be 1) make changes to the template 2) generate and run the tests. This would ensure that changes are always made and work in the template. You could still run tests separately with the make target `KUTTL_TEST='kuttl test --test <test-name>' make check-kuttl`.

cbandy: > I like the idea of not having anything committed a lot.

> What are your thoughts on generating the files as part of the `make check-kuttl` target?

This would be my preference. The separate `generate-kuttl` target (that `check-kuttl` invokes) also accommodates those that run `kuttl` outside of make.

Makefile
```diff
@@ -209,6 +209,38 @@ check-kuttl:
 	${PGO_KUBE_CLIENT} ${KUTTL_TEST} \
 		--config testing/kuttl/kuttl-test.yaml
 
+# KUTTL_ params -- could be left out entirely and just set as env vars
+KUTTL_PSQL_VERSION ?= 14
+# TODO: should the test image here be set according to the KUTTL_PSQL_VERSION
+KUTTL_PSQL_IMAGE ?= registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-14.1-0
```
cbandy, line 215: 👍🏻 I think a "psql client image" variable is the right approach.

Makefile
```diff
@@ -209,6 +209,38 @@ check-kuttl:
 	${PGO_KUBE_CLIENT} ${KUTTL_TEST} \
 		--config testing/kuttl/kuttl-test.yaml
 
+# KUTTL_ params -- could be left out entirely and just set as env vars
+KUTTL_PSQL_VERSION ?= 14
```
cbandy, line 213: ⛏️ Higher up we use `PGO_PG_VERSION` for this. Perhaps `KUTTL_PG_VERSION` here? When I see `PSQL` I think of the client specifically.

Makefile
```diff
@@ -209,6 +209,38 @@ check-kuttl:
...
+			mkdir -p testing/kuttl/generated/$(KUTTL_PSQL_VERSION)-$${SRCDIR}; \
+	done;
+	for FILENAME in $(shell find testing/kuttl/source -regex ".*\.yaml" -print); do \
+		KUTTL_PSQL_VERSION=$(KUTTL_PSQL_VERSION) \
+			KUTTL_PSQL_IMAGE=$(KUTTL_PSQL_IMAGE) \
+			envsubst < $${FILENAME} \
+				> testing/kuttl/generated/$(KUTTL_PSQL_VERSION)-$$(basename $$(dirname $${FILENAME}))/$$(basename $${FILENAME}); \
+	done
+
+# TODO: Generate upgrade KUTTL tests might require different handling
```
cbandy, line 235: 🤔 It seems like upgrade tests need two versions. Can we have two variables? `_VERSION` and `_UPGRADE_FROM_VERSION` ?

Makefile
```diff
@@ -209,6 +209,38 @@ check-kuttl:
...
+# KUTTL_ params -- could be left out entirely and just set as env vars
+KUTTL_PSQL_VERSION ?= 14
+# TODO: should the test image here be set according to the KUTTL_PSQL_VERSION
+KUTTL_PSQL_IMAGE ?= registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-14.1-0
+
+# TODO: When I attempted to loop through the directories and find the yaml files,
+# due to how makefile expands variables, each file would end up in each directory.
+# So this implementation loops through to create the folders under generated
+# and then loops through all the yaml files, using the dirname/basename to place
+# the generated file in the correct folder
```
cbandy, line 221: 🤔 It definitely feels like Make built-ins could be used to get there, but I don't know it well enough to suggest a solution.

I don't think we need separate directories for separate versions if that helps. I think we can probably get by with a single source directory and a single target directory. When we want to test a different version, we invoke `make` again with different arguments. Like,

```shell
make generate-kuttl check-kuttl PG_VERSION=10
make generate-kuttl check-kuttl PG_VERSION=11
```

If that is the case, then patsubst or something might work.
`testing/kuttl/source/*/*.yaml --> testing/kuttl/dest/*/*.yaml`

testing/kuttl/kuttl-test-generated.yaml
benjaminjb: OK, I added the generate to the test command; this way you can just run test and it will generate the files (for version 14 only) OR you can run generate separately if you need to check the generated files or produce test files for another version.

And added a .gitignore to prevent any files from going to /generated

testing/kuttl/source/cluster-start/01-psql-connect.yaml
benjaminjb: Just tested and can confirm that psql 14 can work with PG 13 servers. But after this morning's discussion, the decision was made not to create a testing image; the upgrade image might be worth considering, but for now I think it might be useful to leave this as a env var.

Makefile
benjaminjb: I was thinking our procedure would be 
```
generate 13
generate 14
test 13 + 14
```
but I am curious about the idea of our CI running something more like
```
test 13 (generates, then tests)
test 14 (generates, then tests)
```
The only slight disadvantage that I see here is around what behavior we want to see when a single test in `test 13` fails. Like, if I run KUTTL on 13 + 14, I'll get all the failures/successes -- what behavior do we want to see if we run one test after the other? That's probably something we can discuss and tweak in our CI, i.e., we probably want the failures to log and then to go on to the `test 14` command so that we can see a full report of successful/failed tests after they all run.

That said, I'm looking at this make command now and really wanting to simplify it a bit, so let me follow this thread a little.

testing/kuttl/kuttl-test-generated.yaml
benjaminjb: (and just to be explicit, you can also run the test make target with the alternate env vars, which should generate and test the files in one go)

Makefile
benjaminjb: tested and it looks fine to me.

Makefile
benjaminjb: 8 hours later, this Make target looks cleaner, but also I am tired. (I mean, I did also make sure a 4yo was fed, bathed, and slept, so it wasn't constant 8 hours, but still: tired.)

cbandy: > The only slight disadvantage that I see here is around what behavior we want to see when a single test in `test 13` fails. Like, if I run KUTTL on 13 + 14, I'll get all the failures/successes -- what behavior do we want to see if we run one test after the other? That's probably something we can discuss and tweak in our CI

Yeah, we can leave that to something else. It can run the two versions in a single job, separate jobs, sequentially, or in parallel, etc.

testing/kuttl/generated/.gitignore
```diff
@@ -0,0 +1,3 @@
+*
+*/
+!.gitignore
```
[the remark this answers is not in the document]
jmckulk: Does that mean we could force add initial files then ignore any future changes:question: :thinking: if so we would have to force add any changes from new generated tests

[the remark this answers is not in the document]
cbandy: > Does that mean we could force add initial files then ignore any future changes❓

Nope. Git stops ignoring them once they are added. https://git-scm.com/docs/gitignore

Makefile
```diff
@@ -209,6 +209,33 @@ check-kuttl:
...
+KUTTL_PSQL_CLIENT_IMAGE ?= registry.developers.crunchydata.com/crunchydata/crunchy-postgres:centos8-14.1-0
+
+# TODO: When I attempted to loop through the directories and find the yaml files,
+# due to how makefile expands variables, each file would end up in each directory.
+# So this implementation loops through to create the folders under generated
+# and then loops through all the yaml files, using a `sed` substition to place them
+.PHONY: generate-kuttl
+generate-kuttl:
+	for SRCDIR in $(shell ls testing/kuttl/source); do \
+		mkdir -p testing/kuttl/generated/$${SRCDIR}; \
```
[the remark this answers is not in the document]
benjaminjb: Makes sense, I'll add it to clean, but I also think it might make sense to add a line here cleaning up the folder because it seems like something we wouldn't want to let users (us) forget to do.

Makefile
```diff
@@ -209,6 +209,33 @@ check-kuttl:
...
+	done;
+	for FILEPATH in $(wildcard testing/kuttl/source/*/*.yaml); do \
+		NEW_FILEPATH=`echo $${FILEPATH} | sed 's/source/generated/'`; \
```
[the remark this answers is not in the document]
benjaminjb: I spent an inordinate amount of time on this PR going "huh, why isn't this text operation working?"

Makefile
```diff
@@ -209,6 +209,33 @@ check-kuttl:
 	${PGO_KUBE_CLIENT} ${KUTTL_TEST} \
 		--config testing/kuttl/kuttl-test.yaml
 
+# Set reasonable default for KUTTL_ env vars for templating
```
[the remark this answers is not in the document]
benjaminjb: Yeah, definitely worth considering once we hit some threshold of env vars: add a values file, have the user update the values file, then apply. (Very much like the workflow of Helm...)

Makefile
```diff
@@ -209,6 +209,33 @@ check-kuttl:
...
+	for FILEPATH in $(wildcard testing/kuttl/source/*/*.yaml); do \
+		NEW_FILEPATH=`echo $${FILEPATH} | sed 's/source/generated/'`; \
+		KUTTL_PG_VERSION=$(KUTTL_PG_VERSION) \
+			KUTTL_PG_UPGRADE_FROM_VERSION=$(KUTTL_PG_UPGRADE_FROM_VERSION) \
+			KUTTL_PSQL_CLIENT_IMAGE=$(KUTTL_PSQL_IMAGE) \
+			envsubst < $${FILEPATH} > $${NEW_FILEPATH}; \
+	done
+
+.PHONY: check-generated-kuttl
+check-generated-kuttl: generate-kuttl
```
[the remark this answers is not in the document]
benjaminjb: Yeah, we could have check-kuttl call generate-kuttl and run the tests both in `e2e` and in `generated`. (Eventually, if we go forward with this, we probably want to alter and move all the existing tests to `/source` and only check `generated`, yeah?)

For now I was just keeping them separate during the transition, so that we would be sure of what tests we were running.

testing/kuttl/generated/.gitignore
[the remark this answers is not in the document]
benjaminjb: Ah, I think I got this precise form from some git recipe; but if we're going to be removing the `generated` folder, I should add some line in our main .gitignore. 

That still leaves the problem of people forcing tracking, but is that something we need to or can guard against?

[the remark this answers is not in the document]
cbandy: Sorry, I don't think forcing tracking is a problem. It's one way to commit a `.gitignore` that happens to match/ignore itself.

Makefile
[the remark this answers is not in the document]
jmckulk: Awesome, I saw a mix of the two and wasn't sure what was going on. Transition does make sense, thanks for clarifying.

.gitignore
```diff
@@ -1,3 +1,4 @@
 .DS_Store
 /vendor/
 tools
+testing/kuttl/generated/
```
cbandy, line 4: ⛏️ This matches correctly, but do you mind adding a slash `/` at the front of the pattern? It reminds me that this pattern is anchored to the directory of the gitignore file. 🙇🏻 

⛏️ Please also include the trailing newline in this and other files.

cbandy (approved):

jmckulk (approved):

.gitignore
benjaminjb: Added / and newline (found a few extra files that needed them)

Makefile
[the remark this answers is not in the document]
jmckulk: Figuring out how to implement this in Jenkins could be interesting. :thinking: Currently, we are creating a matrix with only the os and pg_version are passed to the kuttl job. We may need to figure out how to decide what image and upgrade version based on a single version env var. Probably fine in this case, but I wonder what we would do for non-version based env vars :thought_balloon:

Anyway, that is a thought for another day :seedling:

testing/kuttl/generated/.gitignore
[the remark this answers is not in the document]
benjaminjb: ah, I see, I didn't know that. Not sure which way I like better, putting a `!` in the .gitignore or just force-adding a file...

PR #2988, opened 2022-01-20
Clean KUTTL

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

Tests do not all adhere to some common grammar. (Actually, since I was the one who pushed for using the `--` format to help order files AND since my tests didn't have that, I thought I should do something about that.)

1) prefer `psql` to `bash` in commands
2) regularize indices of tests
3) regularize `--` for file ordering
4) add pgconnect timeout to all PG connect pods

**Other Information**:


testing/kuttl/e2e/pgbouncer/00-assert.yaml
```diff
@@ -5,9 +5,9 @@ metadata:
 status:
   instances:
     - name: instance1
-      readyReplicas: 2
-      replicas: 2
-      updatedReplicas: 2
+      readyReplicas: 1
+      replicas: 1
+      updatedReplicas: 1
```
cbandy, line 10: 📝 #2983 changes the cluster to have `replicas: 2`. The tests don't use multiple replicas right now, so maybe this is best. 🌱 I suspect we'll include something with read-replicas eventually here.

cbandy (approved):
I don't think we need to start every test case at `00`. Having some space between numbers is fine and can allow for adding/injecting things without shifting all the files around.

#2977 starts at `01` and I'm (selfishly) okay with that.

I added the TLS steps starting at `10` in #2983 in case there were some other "basics" we wanted to add ahead of them.

testing/kuttl/e2e/pgbouncer/00-assert.yaml
benjaminjb: OK, made that change. (Also reverted the index change.)

PR #2995, opened 2022-01-26
Custom configuration for pgAdmin

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature

No pgAdmin settings can be altered via the PostgresCluster spec.

Arbitrary settings can be specified and are passed to pgAdmin during startup. The settings are validated only as key/value pairs where the value can be any JSON or YAML: strings, numbers, booleans, arrays, etc.

This is the second place in the spec where we've accepted arbitrary key/value pairs with potential nesting, the other being Patroni dynamic configuration. We had worked around a limitation in `controller-tools` with this kind of schema, but recent versions support it better. This PR bumps `controller-tools` and adds a type that both sections of the spec can use to generate the CRD and marshal and unmarshal correctly.

**Other Information**:

Issue: [sc-13471]


config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
```diff
@@ -5839,15 +5839,15 @@ spec:
...
-                              zone1 | zone2 | zone3 | | P P P |   P   |   P   | If
-                              WhenUnsatisfiable is set to DoNotSchedule, incoming
+                              schedule the pod in any location, but giving higher
+                              precedence to topologies that would help reduce the
+                              skew. A constraint is considered "Unsatisfiable" for
+                              an incoming pod if and only if every possible node assigment
+                              for that pod would violate "MaxSkew" on some topology.
+                              For example, in a 3-zone cluster, MaxSkew is set to
+                              1, and pods with the same labelSelector spread as 3/1/1:
+                              | zone1 | zone2 | zone3 | | P P P |   P   |   P   |
```
cbandy, line 5849: 📝 This is a particularly long and difficult [description to read](https://access.crunchydata.com/documentation/postgres-operator/v5/references/crd/#postgresclusterspecinstancesindextopologyspreadconstraintsindex) in plaintext. We might want to consider amending this field like we've done with some others.

testing/kuttl/e2e/pgadmin/02--check-settings.yaml
```diff
@@ -0,0 +1,25 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  # Log the amount of space on the startup volume.
```
benjaminjb, line 5: We don't want to compare this to what it should be, just log it for the user?

config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
benjaminjb: :+1: Do you have an example of a field we've amended?

benjaminjb (approved):
LGTM, just a few clarifying questions really

config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
cbandy: We removed a `TODO` buried within volume projections upstream: https://github.com/CrunchyData/postgres-operator/commit/28c2995189e63044d02432d56dd2f104c4181a2b

internal/controller/postgrescluster/pgadmin_test.go
```diff
@@ -298,62 +389,31 @@ func TestReconcilePGAdminStatefulSet(t *testing.T) {
...
+
+		// Containers and Volumes should be populated.
+		assert.Assert(t, len(template.Spec.Containers) != 0)
+		assert.Assert(t, len(template.Spec.InitContainers) != 0)
+		assert.Assert(t, len(template.Spec.Volumes) != 0)
+
+		// Ignore Containers and Volumes in the comparison below.
+		template.Spec.Containers = nil
+		template.Spec.InitContainers = nil
+		template.Spec.Volumes = nil
```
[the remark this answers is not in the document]
cbandy: Precisely. This function needs to populate those things but their exact contents aren't decided here. I'm not totally satisfied with the division between these two packages, and this pattern might make more/less sense for other components/packages.

internal/pgadmin/config.go
```diff
@@ -42,4 +50,95 @@ const (
...
+				},
+			},
+		},
+	}
+}
+
+// startupCommand returns an entrypoint that prepares the filesystem for pgAdmin.
+func startupCommand() []string {
+	// pgAdmin reads from the following file by importing its public names.
+	// Make sure to assign only to variables that begin with underscore U+005F.
```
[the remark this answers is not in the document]
cbandy: Yes. `_conf`, `_data`, and `_f` so far. I assumed, incorrectly, that `f` would be okay, but later I could see that it was public. I used `import config; help(config)` iirc.

testing/kuttl/e2e/pgadmin/02--check-settings.yaml
cbandy: 🤔 My concern is that `sizeLimit: 32Ki` is too small despite my understanding. The result looks like this:

```
Filesystem     1K-blocks  Used Available Use% Mounted on
tmpfs                 32     4        28  13% /etc/pgadmin
```

docs/content/architecture/pgadmin4.md
```diff
@@ -73,6 +73,20 @@ The operator will synchronize users defined in the spec (e.g., in `spec.users`)
 deployment. Any user created in the database without being defined in the spec will not be
 synchronized.
 
+## Custom Configuration
+
+You can adjust some pgAdmin settings through the
+[`userInterface.pgAdmin.config`]({{< relref "/references/crd#postgresclusterspecuserinterfacepgadminconfig" >}})
+field. For example, set `SHOW_GRAVATAR_IMAGE` to disable automatic profile pictures:
```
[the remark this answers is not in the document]
cbandy: I was concerned about explaining this as well. The values pass through a few layers before becoming Python constants, but thankfully YAML has us covered. YAML `False` becomes JSON `false` which becomes Python `False` again. Similarly, a YAML number becomes a JSON number becomes a Python number. (Use quotes if you want something to be a string, etc.) In short, the values make sense in their context.

cbandy:
Rebased.

config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
benjaminjb: Ah, I see now, yeah, I wouldn't describe that as a blocker, but I agree that it might be easier if we can link to something else that describes MaxSkew (and maybe WhenUnsatisfiable, that's another big chunk of topology-related text that's hard to read in this form).

docs/content/architecture/pgadmin4.md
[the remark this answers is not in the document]
benjaminjb: TIL YAML canonically reads "yes", "no", "on", and "off" as boolean 
https://github.com/yaml/yaml-spec/blob/dae1085b93210a1a6fe978121106f34719f5e9b0/spec/1.2/docbook/bool.dbk#L138

benjaminjb (approved):

testing/kuttl/e2e/pgadmin/02--check-settings.yaml
cbandy: After some more testing, I'm confident we should expect 4KiB. Added an assertion.

testing/kuttl/e2e/pgadmin/02--check-settings.yaml
```diff
@@ -0,0 +1,26 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  # Log the amount of space on the startup volume. Assert that 4KiB are used.
```
benjaminjb, line 5:
```suggestion
  # Log the amount of space on the startup volume. Assert that 4KiB are used.
  # `df` output looks like:
  # ```
  # Filesystem     1K-blocks  Used Available Use% Mounted on
  # tmpfs            2015812     4   2015808   1% /etc/pgadmin
  # ```
```
I'm not crazy about how this ^ looks, but it might be helpful if we ever need to revisit what we're printing. (Maybe?)

benjaminjb (approved):

PR #2997, opened 2022-01-26
INFO Log with header information

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Add header information as to existing logs, updated to INFO level
from DEBUG, and update related Go tests to match changes.

**Other Information**:
Issue: [sc-13581]

### DIFF
--- a/internal/upgradecheck/http.go
+++ b/internal/upgradecheck/http.go
@@ -75,7 +75,7 @@ func init() {
 
 func checkForUpgrades(ctx context.Context, versionString string, backoff wait.Backoff,
 	crclient crclient.Client, cfg *rest.Config,
-	isOpenShift bool) (message string, err error) {
+	isOpenShift bool) (message string, header string, err error) {
 	var headerPayloadStruct *clientUpgradeData
 
 	// Guard against panics within the checkForUpgrades function to allow the
@@ -137,7 +137,7 @@ func checkForUpgrades(ctx context.Context, versionString string, backoff wait.Ba
 	}
 
 	// TODO: Parse response and log info for user on potential upgrades
-	return string(bodyBytes), err
+	return string(bodyBytes), req.Header.Get(clientHeader), err
 }
 
 // CheckForUpgradesScheduler invokes the check func when the operator starts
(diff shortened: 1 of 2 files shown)


internal/upgradecheck/http.go
```diff
@@ -136,8 +136,12 @@ func checkForUpgrades(ctx context.Context, versionString string, backoff wait.Ba
 		err = fmt.Errorf("received StatusCode %d", status)
 	}
 
+	if len(req.Header[clientHeader]) > 0 {
+		header = req.Header[clientHeader][0]
+	}
```
cbandy, line 141: ⛏️ `http.Header` has methods for this kind of thing.

```suggestion
	header = req.Header.Get(clientHeader)
```

internal/upgradecheck/http_test.go
```diff
@@ -67,6 +68,14 @@ func TestCheckForUpgrades(t *testing.T) {
 	ctx := logging.NewContext(context.Background(), logging.Discard())
 	cfg := &rest.Config{}
 
+	checkData := func(header string) {
+		data := clientUpgradeData{}
+		err := json.Unmarshal([]byte(header), &data)
+		assert.NilError(t, err)
```
cbandy, line 74: 🔧 `checkData` is called from within subtests, so it should take a `*testing.T` for the asserts here.

🤔 I forget exactly what happens when asserting on the wrong `T`, but it makes it hard to diagnose the error. (It panics maybe?)

internal/upgradecheck/http.go
tjmoore4: :+1: updated

internal/upgradecheck/http_test.go
tjmoore4: Good note, I've updated the test accordingly and put a link to [the go docs](https://pkg.go.dev/github.com/stretchr/testify/assert#hdr-Assertions) that talk a bit about why `T` is passed.

benjaminjb (approved):
LGTM. Just a question (and a scope-creep question at that) but are there any other changes we should make to logging in this package?

internal/upgradecheck/http_test.go
cbandy: `assert` here is [gotest.tools/v3/assert](https://pkg.go.dev/gotest.tools/v3/assert) so we should probably link to that, if it explains.

When it's the wrong `T`, it prints `test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test` with none of the messages from the assert package.

https://go.dev/play/p/gIVKDrd9rTR

tjmoore4: :thinking: Looking through that link and the code a bit, I can't find anything that exactly points that out. I'll update the note and remove the link.

tjmoore4:
> LGTM. Just a question (and a scope-creep question at that) but are there any other changes we should make to logging in this package?

Nothing that stood out to me at this point. Was there anything in particular you were thinking?

cbandy (approved):

andrewlecuyer (approved):

PR #3006, opened 2022-01-31
KUTTL: Add scaledown test

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

The current `TestScaleDown` test was flaky.

Replace the existing flaky scaledown test with a KUTTL test that covers the same functionality.

**Other Information**:
Issue [sc-13118]


internal/controller/postgrescluster/cluster_test.go
```diff
@@ -47,6 +47,8 @@ import (
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
+func Int32(v int32) *int32 { return &v }
```
jmckulk, line 50: I think we can use `initialize.Int32()` here, maybe?

https://github.com/CrunchyData/postgres-operator/blob/master/internal/initialize/primitives.go#L29

benjaminjb: Ah, ok, I removed and replaced, thanks for the reference

jmckulk (approved):
LGTM! The only thing I'm wondering is if it would be worthwhile to check the primary in every test. I know we aren't in the old tests, I forget why at this point but seems like it could be nice to have/easy to copy over.

testing/kuttl/e2e/scaledown/01-errors.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: scaledown
+status:
+  instances:
+    - name: instance1
```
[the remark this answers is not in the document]
benjaminjb: I added the assert & the errors files because I liked the explicitness of saying, for instance, "we want to find this pod, but we don't want this other pod to exist"; but now that I'm looking at it again, I think anything I wanted to be explicit about is already captured by the asserts on the cluster (i.e., the `status.instances` asserts).

benjaminjb:
> LGTM! The only thing I'm wondering is if it would be worthwhile to check the primary in every test. I know we aren't in the old tests, I forget why at this point but seems like it could be nice to have/easy to copy over.

I was curious about why it was only checking primary in one of the three test cases, but the KUTTL test failed when I annotated for role and expected the master to remain the master.

PR #3009, opened 2022-02-01
Adjust user management for pgAdmin v4.30

cbandy:
The User and Server models changed as pgAdmin v4.21 introduced support for LDAP authentication. pgAdmin v4.30 introduced support for Kerberos authentication.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Other

**Other Information**:

Issue: [sc-13413]

### DIFF
--- a/internal/pgadmin/users.go
+++ b/internal/pgadmin/users.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/crunchydata/postgres-operator/internal/logging"
 	"github.com/crunchydata/postgres-operator/internal/naming"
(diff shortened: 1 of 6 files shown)


internal/pgadmin/users.go
```diff
@@ -57,103 +75,100 @@ spec = importlib.util.find_spec('.pgadmin', (
...
         if not line.strip():
             continue
 
         data = json.loads(line)
+        address = data['username'] + '@pgo'
         user = (
-            db.session.query(User).filter_by(email=data['username']).first() or
-            User(email=data['username'])
+            db.session.query(User).filter_by(username=address).first() or
+            User()
```
ValClarkson, line 134: ❓ would the user ever put in an email address on their own? just wondering if we need to check for the @ first before just padding it with @pgo

ValClarkson (approved):
looks good to me just that one question

internal/pgadmin/users.go
cbandy: I hadn't considered that. We create pgAdmin users from the list of Postgres users in `spec.users` where we don't allow `@`: (I forgot about this and tried it out.)

```
The PostgresCluster "mycluster" is invalid: spec.users.name: Invalid value: "human@computer": spec.users.name in body should match '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
```

I tried it in the guts pgAdmin anyway and found that we can write it, but the user isn't able to log in. They get an error: `Email/Username is not valid`

```python
>>> user = User()
>>> user.auth_source = INTERNAL
>>> user.email = user.username = 'human@some@where'
>>> user.password = 'passwordpassword'
>>> user.active = True
>>> user.roles = db.session.query(Role).filter_by(name='User').all()
>>> user.masterpass_check = 'any'
>>> user.verify_and_update_password(user.password)
True
>>> db.session.add(user)
>>> db.session.commit()
```

So, for now, at least, I think we can ignore it. Validation blocks it, and bypassing validation won't crash PGO or block reconciles.

ValClarkson: oh cool, it's caught in the spec 👍

PR #3010, opened 2022-02-02
Generate a coverage report during GitHub actions

cbandy:
This uses GitHub artifacts and the reports provided by the built-in `go tool cover` command. Functions without complete coverage are printed to the GitHub action logs, and a full HTML report is uploaded as an artifact. The total percent of statements covered appears on the workflow run summary as an annotation.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

**Other Information**:

Artifacts and annotations appear on the [summary page](https://github.com/CrunchyData/postgres-operator/actions/runs/1787439361) after the workflow finishes.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -31,8 +31,18 @@ jobs:
       - uses: actions/setup-go@v2
         with: { go-version: 1.x }
       - run: go mod download
-      - env: { KUBERNETES: "${{ matrix.kubernetes }}" }
-        run: ENVTEST_K8S_VERSION="${KUBERNETES#default}" make check-envtest
+      - run: ENVTEST_K8S_VERSION="${KUBERNETES#default}" make check-envtest
+        env:
+          KUBERNETES: "${{ matrix.kubernetes }}"
+          GO_TEST: go test --coverprofile 'envtest.coverage' --coverpkg ./internal/...
+
+      # Upload coverage to GitHub
+      - run: gzip envtest.coverage
+      - uses: actions/upload-artifact@v2
+        with:
+          name: "kubernetes-api=${{ matrix.kubernetes }}"
+          path: envtest.coverage.gz
+          retention-days: 1
 
   kubernetes-k3d:
     if: "${{ github.repository == 'CrunchyData/postgres-operator' }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -31,8 +31,18 @@ jobs:
...
+          KUBERNETES: "${{ matrix.kubernetes }}"
+          GO_TEST: go test --coverprofile 'envtest.coverage' --coverpkg ./internal/...
+
+      # Upload coverage to GitHub
+      - run: gzip envtest.coverage
+      - uses: actions/upload-artifact@v2
+        with:
+          name: "kubernetes-api=${{ matrix.kubernetes }}"
+          path: envtest.coverage.gz
+          retention-days: 1
```
cbandy, line 45: These coverage profiles are large and only uploaded as artifacts so the final job can download them to generate reports. The minimum possible retention is one day.

.github/workflows/test.yaml
```diff
@@ -84,5 +94,48 @@ jobs:
...
+          sed -e '/100.0%/d' -e "s,$(go list -m)/,," total-coverage.txt
+          echo "::notice title=Total Coverage::$(awk 'END { print $3 }' total-coverage.txt)"
+
+      # Upload coverage to GitHub
+      - run: gzip total-coverage.html
+      - uses: actions/upload-artifact@v2
+        with:
+          name: coverage-report
+          path: total-coverage.html.gz
+          retention-days: 15
```
cbandy, line 141: I chose to upload the HTML report as an artifact because it shows which lines lack coverage. This can go away if the plaintext, function-level report is enough. The default retention of 90 days seems too long, so I chose something shorter here.

.github/workflows/test.yaml
```diff
@@ -84,5 +94,48 @@ jobs:
...
+      # 100% coverage, and upload a complete HTML report.
+      - name: Generate report
+        run: |
+          gunzip --keep download/*/*.gz
+          ( sed -e '1q' download/*/*.coverage
+            tail -qn +2 download/*/*.coverage ) > total.coverage
+          go tool cover --func total.coverage -o total-coverage.txt
+          go tool cover --html total.coverage -o total-coverage.html
+          sed -e '/100.0%/d' -e "s,$(go list -m)/,," total-coverage.txt
+          echo "::notice title=Total Coverage::$(awk 'END { print $3 }' total-coverage.txt)"
```
cbandy, line 133: 🤔 The last line of the text report clarifies what the total percentage is about, `total:	(statements)	89.7%`. We probably should keep that context so the annotation reads: `::notice title=Total Coverage::89.7% (statements)`.

```suggestion
          echo "::notice title=Total Coverage::$(awk 'END { print $3 " " $2 }' total-coverage.txt)"
```

.github/workflows/test.yaml
jmckulk: Is there any limit to how much we can store or for how long?

jmckulk (approved):
lgtm

.github/workflows/test.yaml
cbandy: The docs say artifacts "use storage space" and that Actions are "free for public repositories". The docs about billing for Actions talk about what's free for private repositories.

- https://docs.github.com/en/actions/advanced-guides/storing-workflow-data-as-artifacts
- https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions

So, I don't see any hard limits. Perhaps charges will accrue at "$0.25 USD per GB of storage per month."

cbandy: Sorry, the docs _are_ clear about retention. Up to 90 days for public repos.

.github/workflows/test.yaml
```diff
@@ -31,8 +31,18 @@ jobs:
       - uses: actions/setup-go@v2
         with: { go-version: 1.x }
       - run: go mod download
-      - env: { KUBERNETES: "${{ matrix.kubernetes }}" }
-        run: ENVTEST_K8S_VERSION="${KUBERNETES#default}" make check-envtest
+      - run: ENVTEST_K8S_VERSION="${KUBERNETES#default}" make check-envtest
+        env:
+          KUBERNETES: "${{ matrix.kubernetes }}"
+          GO_TEST: go test --coverprofile 'envtest.coverage' --coverpkg ./internal/...
```
benjaminjb, line 37: Not a blocker, but I wonder if it would be useful to generate test coverage report locally to see how changes we've made might affect the coverage? If so, we could change the GO_TEST in the makefile to this, yeah?

benjaminjb (approved):
LGTM, thanks for going over this yesterday

.github/workflows/test.yaml
cbandy: The `check` targets already pass the `-cover` flag which prints the percentage per package.

https://github.com/CrunchyData/postgres-operator/blob/1c3cc3597c95de3b4265a70a5982ecfc2f768ca5/Makefile#L196

```shell-session
$ make check
…
ok      github.com/crunchydata/postgres-operator/internal/config        (cached)        coverage: 96.3% of statements
…
ok      github.com/crunchydata/postgres-operator/internal/initialize    (cached)        coverage: 100.0% of statements
ok      github.com/crunchydata/postgres-operator/internal/kubeapi       (cached)        coverage: 100.0% of statements
…
```

🤔 We could generate the profile every time, but turning the profile into a meaningful report requires more tools, like `go tool cover` in this PR.

When I want to look at coverage while working on something, I usually use my editor to show it. VS Code [highlights covered lines](https://code.visualstudio.com/docs/languages/go#_test), vim-go has `:GoCoverage`, other editors probably have similar.

PR #3017, opened 2022-02-04
Remove deprecated test helpers

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

### DIFF
--- a/internal/controller/postgrescluster/helpers_test.go
+++ b/internal/controller/postgrescluster/helpers_test.go
@@ -208,35 +208,6 @@ func testCluster() *v1beta1.PostgresCluster {
 	return cluster.DeepCopy()
 }
 
-// setupTestEnv configures and starts an EnvTest instance of etcd and the Kubernetes API server
-// for test usage, as well as creates a new client instance.
-//
-// Deprecated: use setupKubernetes instead.
-func setupTestEnv(t *testing.T,
-	_ string) (*envtest.Environment, client.Client, *rest.Config) {
-
-	testEnv := &envtest.Environment{
-		CRDDirectoryPaths: []string{filepath.Join("..", "..", "..", "config", "crd", "bases")},
-	}
-	cfg, err := testEnv.Start()
-	if err != nil {
-		t.Fatal(err)
-	}
-	t.Log("DEPRECATED: Use setupKubernetes instead of setupTestEnv.")
-	t.Log("Test environment started")
-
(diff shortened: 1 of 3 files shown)


internal/controller/postgrescluster/helpers_test.go
```diff
@@ -107,6 +108,23 @@ func setupKubernetes(t testing.TB) (*envtest.Environment, client.Client) {
 		kubernetes.Lock()
 		defer kubernetes.Unlock()
 
+		if t.Failed() {
```
jmckulk, line 111: nice :+1:

jmckulk (approved):
LGTM

internal/controller/postgrescluster/helpers_test.go
```diff
@@ -107,6 +108,23 @@ func setupKubernetes(t testing.TB) (*envtest.Environment, client.Client) {
...
+
+				type shaped map[string]corev1.NamespaceStatus
+				result := make([]shaped, len(namespaces.Items))
+
+				for i, ns := range namespaces.Items {
+					result[i] = shaped{ns.Labels["postgres-operator-test"]: ns.Status}
+				}
+
+				formatted, _ := yaml.Marshal(result)
+				t.Logf("Test Namespaces:\n%s", formatted)
```
benjaminjb, line 124: OK, so any time we call `setupNamespace`, we add the label `"postgres-operator-test": t.Name`, so a failed test will print `t.Name: ns.Status` here for every namespace before it gets cleaned up (in the `setupNamespace` cleanup) (cite: https://github.com/CrunchyData/postgres-operator/blob/master/internal/controller/postgrescluster/helpers_test.go#L130-L141 )

❓ But `setupKubernetes` is getting called by multiple tests and the kubernetes.env.Config is getting set once and shared, so would this return the state of all namespaces, not just the state of the namespaces related to the failed test?

cbandy: Right. The list of namespaces is a proxy to the list of running tests.

I'm finding that when a test fails for me locally, there's usually a lot of objects still waiting to be cleaned up. Here, for example, there are 13 namespaces from _other_ tests still being cleaned up. Those that are reporting delays say they are waiting on 22 pods and 2 PVCs.

https://github.com/CrunchyData/postgres-operator/runs/5061064961?check_suite_focus=true#step:7:119

benjaminjb (approved):

PR #3019, opened 2022-02-04
Copy pgBackRest configuration resources when performing a restore

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

This commit copies pgBackRest configuration and related ConfigMaps
and Secrets used by the source cluster as VolumeProjections when
bootstrapping a new cluster using pgBackRest restore. The
configuration copied are given a unique name to avoids conflicts
with any other clusters and is removed once cluster bootstrap is
complete. If the ConfigMap or Secret VolumeProjections are configured
as "optional" in the source cluster, those resources may be absent
without any error or failure when the copying is attempted.

**Other Information**:
Issue: [sc-13601]

### DIFF
--- a/internal/controller/postgrescluster/pgbackrest_test.go
+++ b/internal/controller/postgrescluster/pgbackrest_test.go
@@ -2217,6 +2217,254 @@ func TestReconcileCloudBasedDataSource(t *testing.T) {
 	}
 }
 
+func TestCopyConfigurationResources(t *testing.T) {
+	_, tClient := setupKubernetes(t)
+	ctx := context.Background()
+	require.ParallelCapacity(t, 2)
+
+	r := &Reconciler{Client: tClient, Owner: client.FieldOwner(t.Name())}
+
+	ns1 := setupNamespace(t, tClient)
+	ns2 := setupNamespace(t, tClient)
+
+	secret := func(testNum string) *corev1.Secret {
+		return &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "source-secret" + testNum,
+				Namespace: ns1.Name,
+			},
+		}
+	}
+
+	configMap := func(testNum string) *corev1.ConfigMap {
+		return &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "source-configmap" + testNum,
+				Namespace: ns1.Name,
+			},
+		}
+	}
+
+	clusterUID := "hippouid"
+
+	sourceCluster := func(testNum string) *v1beta1.PostgresCluster {
+		return &v1beta1.PostgresCluster{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "source-cluster" + testNum,
+				Namespace: ns1.Name,
+				UID:       types.UID(clusterUID),
+			},
+			Spec: v1beta1.PostgresClusterSpec{
+				PostgresVersion: 13,
+				Image:           "example.com/crunchy-postgres-ha:test",
+				InstanceSets: []v1beta1.PostgresInstanceSetSpec{{
+					Name: "instance1",
+					DataVolumeClaimSpec: corev1.PersistentVolumeClaimSpec{
+						AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteMany},
+						Resources: corev1.ResourceRequirements{
+							Requests: corev1.ResourceList{
+								corev1.ResourceStorage: resource.MustParse("1Gi"),
+							},
+						},
+					},
+				}},
+				Backups: v1beta1.Backups{
+					PGBackRest: v1beta1.PGBackRestArchive{
+						Configuration: []corev1.VolumeProjection{{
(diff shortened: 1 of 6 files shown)


internal/controller/postgrescluster/pgbackrest_test.go
```diff
@@ -2217,6 +2217,263 @@ func TestReconcileCloudBasedDataSource(t *testing.T) {
...
+	ctx, cancel := setupManager(t, tEnv.Config, func(mgr manager.Manager) {
+		r = &Reconciler{
+			Client:   tClient,
+			Recorder: mgr.GetEventRecorderFor(ControllerName),
+			Tracer:   otel.Tracer(ControllerName),
+			Owner:    ControllerName,
+		}
+	})
+
+	t.Cleanup(func() { teardownManager(cancel, t) })
```
benjaminjb, line 2234: ❓ Skimming our tests, it looks like `setupManager` is only being used in a few places, where otherwise we're mostly setting the reconciler directly (and getting ctx from `ctx := context.Background()` -- is there something else that this formulation get us? (It also looks like `setupKubernetes` creates the PG scheme for the client.)

internal/controller/postgrescluster/pgbackrest_test.go
```diff
@@ -2217,6 +2217,263 @@ func TestReconcileCloudBasedDataSource(t *testing.T) {
...
+		if err := tClient.Create(ctx, nc); err != nil {
+			t.Fatal(err)
+		}
+
+		assert.NilError(t, r.copyConfigurationResources(ctx, nc, sc))
+
+		assert.NilError(t, checkSecret(secret.Name+"-restorecopy-0", ns2.Name))
+		assert.Check(t, apierrors.IsNotFound(checkConfigMap(configMap.Name+"-restorecopy-1", ns2.Name)))
+	})
+	t.Run("Secret and ConfigMap configured, both optional", func(t *testing.T) {
```
benjaminjb, line 2458: Nice series of test cases :+1:

benjaminjb (approved):
LGTM overall, one bigger question, one question where I think I answered my own question.

Basically, what this does is, after copying over the configmap and secret we always expect, is it checks for any other configuration secrets/configmaps to copy over, updating the sourcecluster (a copy) with those names so that we can append those volumes to the restore pod. Did I miss anything?

internal/controller/postgrescluster/pgbackrest_test.go
tjmoore4: This could potentially be slimmed down a bit, but I went with `setupManager` as it was a standard way we provide both  `ctx` and `cancel`. For this test I was particularly concerned about proper cleanup, so I wanted to make sure to properly define the `teardownManager` function. Do you think it's worth reworking the setup for this test?

benjaminjb: Hmmm, I'm torn: I think `cancel` is only being used in the teardown, so if we never setup the manager, we never need `cancel` to tear it down, yeah? And`setupManager` is only being used in the pgbackrest tests, and I'm not sure why it survived here.

Here's the flow-chart of my thoughts: would it be a pain/do we have the time now to try this without `setupManager`? If you can try it, I'd be curious if it works; if not, maybe make a `chore` ticket see if we can eliminate it at some point? (If the slimmed down test works, then you can add that note to the ticket -- because it seems like it would be scope creep to add "eliminating that test helper code" to this ticket, so I guess we need a ticket anyway?)

benjaminjb:
> > LGTM overall, one bigger question, one question where I think I answered my own question.
> > Basically, what this does is, after copying over the configmap and secret we always expect, is it checks for any other configuration secrets/configmaps to copy over, updating the sourcecluster (a copy) with those names so that we can append those volumes to the restore pod. Did I miss anything?
> 
> Nope, that's the essential approach. The only things I might call out is that optional CMs and Secrets can be missing and other projected volume types (i.e. downwardAPIs and serviceAccountTokens) must be handled manually. (I figure you read those comments, but doesn't hurt to mention again here)

👍 We probably will need to add some documentation around the stuff they'll need to handle manulally -- there's no mention of that in the ticket, maybe we need another ticket? (Seems low priority as I don't imagine people doing that a lot).

internal/controller/postgrescluster/pgbackrest_test.go
tjmoore4: Sure, it was easy enough to switch out the code. Local testing seems good, I just pushed the change to make sure all the actions pass as well. If this looks good to you, I think we can leave this change. I was also wondering about the the `Owner` value impacting things, but I'm now thinking that may not be an issue with how I'm actually setting ownership in the test.

benjaminjb (approved):

internal/controller/postgrescluster/pgbackrest_test.go
benjaminjb: Nice, thx!

PR #3027, opened 2022-02-09
Set resources on init container for restore pod

jmckulk:
All containers in the restore pod should have resources set. Use the
resources defined on the restore container to set resources on the
nss_wrapper init-container.

Issue: [sc-13689]

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Currently, the nss_wrapper init container on the restore pod does not have resources defined

**Other Information**:

Issue: [sc-13689]


internal/controller/postgrescluster/util_test.go
```diff
@@ -325,42 +341,44 @@ func TestAddNSSWrapper(t *testing.T) {
...
+				var actualUpdatedContainerCount int
+				for i, c := range template.Spec.Containers {
+					if c.Name == naming.ContainerDatabase ||
+						c.Name == naming.PGBackRestRepoContainerName ||
+						c.Name == naming.ContainerPGUpgrade ||
+						c.Name == naming.PGBackRestRestoreContainerName {
+						assert.DeepEqual(t, expectedEnv, c.Env)
+						actualUpdatedContainerCount++
+					} else {
+						assert.DeepEqual(t, beforeAddNSS[i], c)
```
cbandy, line 355: ❓ I think the contents of `beforeAddNSS` are modified along with `template.Spec.Containers`. Does this assertion pass if it is moved outside of the `else` branch? If so, use `DeepCopy()` around L340 to ensure that "before" values stay separate.

internal/controller/postgrescluster/util_test.go
```diff
@@ -325,42 +341,44 @@ func TestAddNSSWrapper(t *testing.T) {
...
-					assert.Assert(t, c.SecurityContext != &corev1.SecurityContext{})
-
-					for i, c := range template.Spec.Containers {
-						if c.Name == "database" || c.Name == "pgbackrest" {
-							assert.DeepEqual(t, c.Resources.Requests,
-								template.Spec.Containers[i].Resources.Requests)
-						}
+				// verify database and/or pgbackrest containers updated
+				assert.Equal(t, actualUpdatedContainerCount,
+					expectedUpdatedContainerCount(template))
```
cbandy, line 361: 🏕️ This would make more sense to me as a property of the test case. Is there something that prevents that?

```go
	tcName: "database container with pgbackrest sidecar",
	expected: 2,
	podTemplate: &corev1.PodTemplateSpec{Spec: corev1.PodSpec{
	…
```

```go
	// verify database and/or pgbackrest containers updated
	assert.Equal(t, actualUpdatedContainerCount, tc.expected)
```

cbandy (commented):
Your changes look good. I wonder if we should address some other things in this test while you're looking at it.

cbandy (approved):
LGTM with a squash.

PR #3052, opened 2022-02-22
Enable Support for Secrets for pgAdmin LDAP_BIND_PASSWORD

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [x] Documentation
 - [ ] Testing enhancement
 - [ ] Other

This update allows the use of a mounted Secret that contains
the desired LDAP_BIND_PASSWORD setting for pgAdmin which is used
for the pgAdmin login user when LDAP authentication is enabled.
This allows the password value to be encoded in a Secret rather
than stored in plaintext in a ConfigMap with the other pgAdmin
configuration settings.

This PR also includes documentation for mounting this Secret, as well
as other more generic Secrets and ConfigMaps used for pgAdmin configuration.

**Other Information**:
Issue: [sc-13712]

### DIFF
--- a/config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
+++ b/config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
@@ -8632,6 +8632,26 @@ spec:
                                   type: object
                               type: object
                             type: array
+                          ldapBindPassword:
+                            description: 'A Secret containing the value for the LDAP_BIND_PASSWORD
+                              setting. More info: https://www.pgadmin.org/docs/pgadmin4/latest/ldap.html'
+                            properties:
+                              key:
+                                description: The key of the secret to select from.  Must
+                                  be a valid secret key.
+                                type: string
+                              name:
+                                description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                                  TODO: Add other useful fields. apiVersion, kind,
+                                  uid?'
+                                type: string
+                              optional:
+                                description: Specify whether the Secret or its key
+                                  must be defined
+                                type: boolean
+                            required:
+                            - key
+                            type: object
                           settings:
                             description: 'Settings for the pgAdmin server process.
                               Keys should be uppercase and values must be constants.
--- a/internal/pgadmin/config.go
+++ b/internal/pgadmin/config.go
@@ -40,6 +40,10 @@ const (
 	dataVolume    = "pgadmin-data"
 	dataMountPath = "/var/lib/pgadmin"
 
+	// ldapPasswordPath is the path for mounting the LDAP Bind Password
+	ldapPasswordPath         = "~postgres-operator/ldap-bind-password" /* #nosec */
+	ldapPasswordAbsolutePath = configMountPath + "/" + ldapPasswordPath
+
 	// TODO(tjmoore4): The login and password implementation will be updated in
 	// upcoming enhancement work.
 
@@ -74,7 +78,7 @@ const (
 // podConfigFiles returns projections of pgAdmin's configuration files to
 // include in the configuration volume.
 func podConfigFiles(configmap *corev1.ConfigMap, spec v1beta1.PGAdminPodSpec) []corev1.VolumeProjection {
-	return append(append([]corev1.VolumeProjection{}, spec.Config.Files...),
+	config := append(append([]corev1.VolumeProjection{}, spec.Config.Files...),
 		[]corev1.VolumeProjection{
 			{
 				ConfigMap: &corev1.ConfigMapProjection{
@@ -90,6 +94,31 @@ func podConfigFiles(configmap *corev1.ConfigMap, spec v1beta1.PGAdminPodSpec) []
 				},
 			},
 		}...)
+
+	// To enable LDAP authentication for pgAdmin, various LDAP settings must be configured.
+	// While most of the required configuration can be set using the 'settings'
+	// feature on the spec (.Spec.UserInterface.PGAdmin.Config.Settings), those
+	// values are stored in a ConfigMap in plaintext.
+	// As a special case, here we mount a provided Secret containing the LDAP_BIND_PASSWORD
+	// for use with the other pgAdmin LDAP configuration.
+	// - https://www.pgadmin.org/docs/pgadmin4/latest/config_py.html
+	// - https://www.pgadmin.org/docs/pgadmin4/development/enabling_ldap_authentication.html
+	if spec.Config.LDAPBindPassword != nil {
+		config = append(config, corev1.VolumeProjection{
+			Secret: &corev1.SecretProjection{
+				LocalObjectReference: spec.Config.LDAPBindPassword.LocalObjectReference,
+				Optional:             spec.Config.LDAPBindPassword.Optional,
+				Items: []corev1.KeyToPath{
+					{
+						Key:  spec.Config.LDAPBindPassword.Key,
+						Path: ldapPasswordPath,
+					},
+				},
+			},
+		})
+	}
+
+	return config
 }
 
 // startupCommand returns an entrypoint that prepares the filesystem for pgAdmin.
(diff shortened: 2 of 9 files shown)


pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgadmin_types.go
```diff
@@ -52,6 +52,12 @@ type PGAdminPodSpec struct {
 	// +optional
 	Config PGAdminConfiguration `json:"config,omitempty"`
 
+	// The Secret containing the pgAdmin value for the LDAP_BIND_PASSWORD
+	// configuration setting which is used for the pgAdmin login user when LDAP
+	// authentication is enabled.
+	// +optional
+	LDAPBindPassword *corev1.SecretProjection `json:"ldapBindPassword,omitempty"`
```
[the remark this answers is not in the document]
tjmoore4: :thinking: That's a good thought. I'll try that out and see how things look.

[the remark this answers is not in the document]
tjmoore4: :question: I'm having a bit of trouble figuring out how we'd want to use a `SecretKeySelector` here. I'm currently using the Secret as a [VolumeSource](https://pkg.go.dev/k8s.io/api/core/v1#VolumeSource) and there doesn't seem to be a way to use a `SecretKeySelector` in the same way. I could use that as an [EnvVarSource](https://pkg.go.dev/k8s.io/api/core/v1#EnvVarSource), but that would store the value as an environment variable, which seems less than ideal. Is there another option I'm missing?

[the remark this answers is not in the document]
cbandy: Have you tried adding a SecretProjection around here:

https://github.com/CrunchyData/postgres-operator/blob/e7c62c89ef371b839a09566a15058f1da1046a4c/internal/pgadmin/config.go#L78-L80

I suspect we can do like this:

```go
corev1.VolumeProjection{
{
	Secret: &corev1.SecretProjection{
		LocalObjectReference: spec.LDAPBindPassword.LocalObjectReference,
		Optional:             spec.LDAPBindPassword.Optional,
		Items: []corev1.KeyToPath{
			{
				Key:  spec.LDAPBindPassword.Key,
				Path: constantHere,
			},
		},
	},
}
```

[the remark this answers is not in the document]
tjmoore4: I'm tracking now, thanks. That works as expected, the code, tests and docs are updated accordingly.

config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
```diff
@@ -8771,6 +8771,26 @@ spec:
...
+                          LDAP_BIND_PASSWORD configuration setting which is used for
+                          the pgAdmin login user when LDAP authentication is enabled.
+                        properties:
+                          key:
+                            description: The key of the secret to select from.  Must
+                              be a valid secret key.
+                            type: string
+                          name:
+                            description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                              TODO: Add other useful fields. apiVersion, kind, uid?'
```
cbandy, line 8785: 📝 We have to run `hack/create-todo-patch.sh` to remove these `TODO`. It looks like we haven't done it in a while, so it will pull in more than just this one occurrence. It's a little fiddly; I think it has to run between two `make generate` or something?

🏕️ If you want to try it now, go for it, otherwise, I can take care of it in a separate PR.

internal/pgadmin/config.go
```diff
@@ -40,6 +40,10 @@ const (
 	dataVolume    = "pgadmin-data"
 	dataMountPath = "/var/lib/pgadmin"
 
+	// ldapPasswordMountPath is the path for mounting the LDAP Bind Password
+	ldapPasswordPath         = "~postgres-operator/ldap-bind-password" /* #nosec */
```
cbandy, line 44: ⛏️ Comment is `ldapPasswordMountPath` while variable is `ldapPasswordPath`.

internal/pgadmin/config.go
```diff
@@ -121,8 +150,25 @@ with open('` + settingsAbsolutePath + `') as _f:
...
 `
+	// Set pgAdmin's LDAP_BIND_PASSWORD setting if the value was provided via Secret.
+	// As this assignment happens after any values provided via the 'Settings'
+	// ConfigMap loaded above, this value will overwrite any previous configuration
+	// of LDAP_BIND_PASSWORD (that is, last write wins).
+	var addPassword string
+	if ldapBindPassword != nil {
+
+		addPassword = `import os
+if os.path.isfile('` + ldapPasswordAbsolutePath + `'):
```
cbandy, line 161: 👍🏻 `os.path.isfile` here makes sense in case the secret ref has `optional: true`.

❓ Is it easier to always have this Python code? Can we drop the `if ldapBindPassword != nil`, the append, etc?

internal/pgadmin/reconcile.go
```diff
@@ -301,8 +301,46 @@ func Pod(
...
+	// While most of the required configuration can be set using the 'settings'
+	// feature on the spec (.Spec.UserInterface.PGAdmin.Config.Settings), those
+	// values are stored in a ConfigMap in plaintext.
+	// As a special case, here we mount a provided Secret containing the LDAP_BIND_PASSWORD
+	// for use with the other pgAdmin LDAP configuration.
+	// - https://www.pgadmin.org/docs/pgadmin4/latest/config_py.html
+	// - https://www.pgadmin.org/docs/pgadmin4/development/enabling_ldap_authentication.html
+	// if inCluster.Spec.UserInterface.PGAdmin.LDAPBindPassword != nil {
+
+	// 	ldapPasswordVolumeMount := corev1.VolumeMount{
```
cbandy, line 318: 🔧 Please remove rather than comment out.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgadmin_types.go
```diff
@@ -52,6 +52,12 @@ type PGAdminPodSpec struct {
 	// +optional
 	Config PGAdminConfiguration `json:"config,omitempty"`
 
+	// The Secret containing the pgAdmin value for the LDAP_BIND_PASSWORD
+	// configuration setting which is used for the pgAdmin login user when LDAP
+	// authentication is enabled.
+	// +optional
+	LDAPBindPassword *corev1.SecretKeySelector `json:"ldapBindPassword,omitempty"`
```
cbandy, line 59: 👍🏻 This seems like the "right" combo of camel-case. 😵

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgadmin_types.go
```diff
@@ -52,6 +52,12 @@ type PGAdminPodSpec struct {
 	// +optional
 	Config PGAdminConfiguration `json:"config,omitempty"`
 
+	// The Secret containing the pgAdmin value for the LDAP_BIND_PASSWORD
+	// configuration setting which is used for the pgAdmin login user when LDAP
+	// authentication is enabled.
```
cbandy, line 57: ⛏️ Not all LDAP situations require this setting. I think we can keep it simple and link out to the docs.

```suggestion
	// A Secret containing the value for the LDAP_BIND_PASSWORD setting.
	// More info: https://www.pgadmin.org/docs/pgadmin4/latest/ldap.html
```

internal/pgadmin/reconcile.go
tjmoore4: Good catch, thanks. Removed.

internal/pgadmin/config.go
tjmoore4: :+1: updated.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgadmin_types.go
tjmoore4: :+1: works for me, updated.

internal/pgadmin/config.go
tjmoore4: Yes, testing this out everything worked as expected, code is updated.

config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
tjmoore4: I gave it a shot, but I think I have the wrong version of `yq`. I'll leave that to you, if you don't mind :slightly_smiling_face:

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgadmin_types.go
```diff
@@ -52,6 +52,11 @@ type PGAdminPodSpec struct {
 	// +optional
 	Config PGAdminConfiguration `json:"config,omitempty"`
 
+	// A Secret containing the value for the LDAP_BIND_PASSWORD setting.
+	// More info: https://www.pgadmin.org/docs/pgadmin4/latest/ldap.html
+	// +optional
+	LDAPBindPassword *corev1.SecretKeySelector `json:"ldapBindPassword,omitempty"`
```
cbandy, line 58: 🤔 I just noticed. Should this be in the `config` section; in the `PGAdminConfiguration` struct?

cbandy (approved):
LGTM. I noticed only here at the very end where the field is in the spec. 🙇🏻

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgadmin_types.go
tjmoore4: :thinking: Originally, I was thinking we might want to keep this setting off on its own since it was more of a special case than the more general configuration settings, but I do think putting it in `PGAdminConfiguration` might make more sense. I'll work on that change.

cbandy (approved):

PR #3080, opened 2022-03-08
Change "status.instances.replicas" to include all pods

cbandy:
We added this field while implementing rolling updates and chose to populate it with semantics similar to the appsv1.Deployment controller which ignores terminating Pods. However, our controller does *not* ignore terminating PostgreSQL Pods because they are still writing to disk and still participating in HA quorum.

So, this field now reflects the number of observed Pods, regardless of their state. This is the same approach taken by the appsv1.StatefulSet controller. When this field is zero, we can be relatively certain that all PostgreSQL processes have stopped and disks are idle.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Bug fix

Terminating instances are not counted in PostgresCluster status.

Terminating instances are counted in PostgresCluster status.

**Other Information**:

Issue: [sc-13696]
See: https://issue.k8s.io/107920
See: e0a885ccb842a735fa007604576a2c31a80b2f7a

### DIFF
--- a/internal/controller/postgrescluster/delete_test.go
+++ b/internal/controller/postgrescluster/delete_test.go
@@ -28,7 +28,6 @@ import (
 
 	"go.opentelemetry.io/otel"
 	"gotest.tools/v3/assert"
-	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


tjmoore4 (approved):
:+1: Looks good to me.

internal/controller/postgrescluster/delete_test.go
```diff
@@ -412,27 +402,20 @@ func TestReconcilerHandleDeleteNamespace(t *testing.T) {
...
-
-		ready = 0
-		for i := range instances {
-			ready += instances[i].Status.ReadyReplicas
-		}
-		return ready >= 2, nil
-	}), "expected 2 instances to be ready, got:\n%v", ready)
+	// Wait until instances are healthy.
+	if ready := int32(0); !assert.Check(t,
+		wait.Poll(time.Second, Scale(time.Minute), func() (bool, error) {
```
cbandy, line 407: These `if !Check(…) { FailNow() }` look awkward. I put some logging ahead of the `FailNow` while debugging flakes, but that code was grotty and not ready to commit.

PR #3092, opened 2022-03-15
Little fixes

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

`make install` might error because the CRD is big; pgbackrest's exec had an inaccurate comment

Fix comment; add server-side to `make install`

**Other Information**:
[n/a]


Makefile
```diff
@@ -95,7 +95,7 @@ deletenamespaces:
 
 # Install the postgrescluster CRD
 install:
-	$(PGO_KUBE_CLIENT) apply -k ./config/crd
+	$(PGO_KUBE_CLIENT) apply --server-side=true --force-conflicts -k ./config/crd
```
cbandy, line 98: This is because of the tracking annotation that gets too big? When is `--force-conflicts` required?

⛏️ If it works without `=true`, I'd like to drop that part.

```suggestion
	$(PGO_KUBE_CLIENT) apply --server-side --force-conflicts -k ./config/crd
```

benjaminjb: Tested and made changes -- 
1) dropped the `true`, 
2) verified that multiple deploys would error without `--force-conflicts` and added a not about what I was seeing.

PR #3125, opened 2022-04-01
Add CRD Update Guidance for the Helm Installer

andrewlecuyer:
Adds instructions to the upgrade section of the Helm install guide for manually updating any PGO CRDs before upgrading PGO via Helm.  This is specifically due to a design decision on Helm v3 in which CRDs are only applied by Helm when running `helm install`.

https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

[sc-14156]


docs/content/installation/helm.md
```diff
@@ -90,13 +90,26 @@ PGO will check for updates upon startup and once every 24 hours. Any errors in c
...
+Once PGO has been installed, it can then be upgraded using the `helm upgrade` command.
+However, before running the `upgrade` command, any Custom Resource Definitions (CRDs) must first be
+manually updated (this is specifically due to a
+[design decision in Helm v3](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/),
+in which any CRDs in the Helm chart are only applied when using the `helm install` command).
+
+Therefore, to first update the PGO CRDs, run the following command:
+
+```shell
+kubectl apply -n <namespace> -f helm/install/crds
```
cbandy, line 102: 🔧 CRDs have no namespace.

```suggestion
kubectl apply -f helm/install/crds
```

<br/>

🤔 I want to suggest `--server-side` so the last-applied-configuration annotation doesn't get too big, but I encountered conflicts with helm:

```console
$ kubectl apply --server-side -f helm/install/crds
error: Apply failed with 3 conflicts: conflicts with "helm" using apiextensions.k8s.io/v1:
- .metadata.annotations.controller-gen.kubebuilder.io/version
- .metadata.labels.app.kubernetes.io/version
- .spec.versions

Please review the fields above--they currently have other managers. Here
are the ways you can resolve this warning:
* If you intend to manage all of these fields, please re-run the apply
  command with the `--force-conflicts` flag.
* If you do not intend to manage all of the fields, please edit your
  manifest to remove references to the fields that should keep their
  current managers.
* You may co-own fields by updating your manifest to match the existing
  value; in this case, you'll become the manager if the other manager(s)
  stop managing the field (remove it from their configuration).
See https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts
```

docs/content/installation/helm.md
```diff
@@ -90,13 +90,26 @@ PGO will check for updates upon startup and once every 24 hours. Any errors in c
 
 ## Upgrade and Uninstall
 
-And once PGO has been installed, it can then be upgraded and uninstalled using applicable `helm`
-commands:
+Once PGO has been installed, it can then be upgraded using the `helm upgrade` command.
+However, before running the `upgrade` command, any Custom Resource Definitions (CRDs) must first be
+manually updated (this is specifically due to a
+[design decision in Helm v3](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/),
```
cbandy, line 96: ⛏️ This link describes how to design a chart with CRDs. One of these seems like a better fit for chart users:

- https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds
  Describes CRDs a bit and how it works when there's a custom resource in the templates.

- https://helm.sh/docs/topics/charts/#limitations-on-crds
  Gets right to it and says what Helm does _not_ do; upgrades nor uninstalls.

I like the latter, but maybe that's just me.

```suggestion
[design decision in Helm v3](https://helm.sh/docs/topics/charts/#limitations-on-crds),
```

cbandy, line 96: ⛏️ This link describes how to design a chart with CRDs. One of these seems like a better fit for chart users:

- https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds
  Describes CRDs a bit and how it works when there's a custom resource in the templates.

- https://helm.sh/docs/topics/charts/#limitations-on-crds
  Gets right to it and says what Helm does _not_ do; upgrades nor uninstalls.

I like the latter, but maybe that's just me.

```suggestion
[design decision in Helm v3](https://helm.sh/docs/topics/charts/#limitations-on-crds),
```

docs/content/installation/helm.md
```diff
@@ -90,13 +90,43 @@ PGO will check for updates upon startup and once every 24 hours. Any errors in c
...
 helm upgrade <name> -n <namespace> helm/install
 ```
 
+To uninstall PGO, remove all your PostgresCluster objects, then use the `helm uninstall` command:
+
 ```shell
 helm uninstall <name> -n <namespace>
 ```
+
+Helm [leaves the CRDs][helm-crd-limits] in place. You can remove them with `kubectl delete`:
```
andrewlecuyer, line 128: :100:

docs/content/installation/helm.md
```diff
@@ -90,13 +90,43 @@ PGO will check for updates upon startup and once every 24 hours. Any errors in c
...
+kubectl apply --server-side --force-conflicts -f helm/install/crds
+```
+
+Then, perform the upgrade using Helm:
 
 ```shell
 helm upgrade <name> -n <namespace> helm/install
 ```
 
+To uninstall PGO, remove all your PostgresCluster objects, then use the `helm uninstall` command:
```
andrewlecuyer, line 122: Is is necessary to remove all PostgresCluster objects?

In the past I've def seen requests to uninstall PGO, but leave all clusters in place.

cbandy: 🤔 If the controller goes away first, nothing will handle the finalizers on those PostgresClusters.

So it's a question, I guess, of whether someone wants to give up completely on PGO or just undo something done by Helm?

andrewlecuyer: Agreed.  Looking back (I distinctly remember this coming up for the Ansible installer for v4 in the past), we've seen it both ways - i.e. some users looking for _everything_ to be installed, and others looking for just the PGO Deployment to go away (but not the PG clusters).

I'm good with assuming `uninstall` means _everything_ for now, especially considering the finalizers :+1:

docs/content/installation/helm.md
andrewlecuyer: Agree that other link gets right to the point.  I'm good with updating :+1:

andrewlecuyer: Link has been updated.

cbandy (approved):

PR #3136, opened 2022-04-05
Document some fields to use for pgAdmin LDAP and Kerberos

cbandy:
The CRD reference links to pgAdmin's overall configuration, but nothing specific about LDAP or Kerberos, which are now supported.

**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [x] Documentation

**Other Information**:

Issue: [sc-13978]

### DIFF
--- a/docs/content/architecture/pgadmin4.md
+++ b/docs/content/architecture/pgadmin4.md
@@ -87,12 +87,66 @@ field. For example, set `SHOW_GRAVATAR_IMAGE` to `False` to disable automatic pr
           SHOW_GRAVATAR_IMAGE: False
 ```
 
-You can also add a Secret containing the pgAdmin `LDAP_BIND_PASSWORD` through the
-[`userInterface.pgAdmin.config.ldapBindPassword`]
-({{< relref "/references/crd#postgresclusterspecuserinterfacepgadminconfigldapbindpassword" >}})
-field. This is one of the configuration settings needed to enable LDAP authentication
-for pgAdmin and is handled separately from the other pgAdmin settings to allow for
-proper storage of the sensitive value in a Secret rather than a ConfigMap.
+You can also mount files to `/etc/pgadmin/conf.d` inside the pgAdmin container using
+[projected volumes](https://kubernetes.io/docs/concepts/storage/projected-volumes/).
+The following mounts `useful.txt` of Secret `mysecret` to `/etc/pgadmin/conf.d/useful.txt`:
+
+```yaml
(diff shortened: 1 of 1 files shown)


docs/content/architecture/pgadmin4.md
```diff
@@ -87,40 +87,73 @@ field. For example, set `SHOW_GRAVATAR_IMAGE` to `False` to disable automatic pr
...
+        - secret:
+            name: mysecret
+            items:
+            - key: some.keytab
+```
+
+### LDAP Configuration
+
+You can configure pgAdmin to [authenticate its users using LDAP](https://www.pgadmin.org/docs/pgadmin4/latest/ldap.html)
+passwords. In addition to setting `AUTHENTICATION_SOURCES` and `LDAP_SERVER_URI`, you need to
```
andrewlecuyer, line 133: :question: I assume `LDAP_AUTO_CREATE_USER` needs to be enabled since there currently isn't a way to login as admin and manually create/provision users?

andrewlecuyer (approved):

docs/content/architecture/pgadmin4.md
cbandy: Precisely. We'd need to allow admin user(s) or provide some other way to specify pgAdmin users on our spec.

docs/content/architecture/pgadmin4.md
```diff
@@ -87,40 +87,73 @@ field. For example, set `SHOW_GRAVATAR_IMAGE` to `False` to disable automatic pr
...
-          - secret:
-              name: mysecret
-          - configMap:
-              name: myconfigmap
-              optional: false
+        - secret:
+            name: mysecret
+            items:
+            - key: some.keytab
+```
```
tjmoore4, line 128: :wrench: I believe you will need to also mount the krb.conf file, which is expected to be at [`/etc/pgadmin/conf.d/krb5.conf`](https://github.com/CrunchyData/postgres-operator/blob/master/internal/pgadmin/reconcile.go#L236)

tjmoore4, line 128: :wrench: I believe you will need to also mount the krb.conf file, which is expected to be at [`/etc/pgadmin/conf.d/krb5.conf`](https://github.com/CrunchyData/postgres-operator/blob/master/internal/pgadmin/reconcile.go#L236)

tjmoore4 (approved):
:+1: Looks good to me

PR #3168, opened 2022-04-25
Convert container security context test to kuttl

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

This creates a Kuttl test to verify a PostgresCluster's containers'
security context settings which replaces the current EnvTest
implementation.

**Other Information**:
Issue: [sc-14262]

### DIFF
--- a/testing/kuttl/e2e/security-context/00--cluster.yaml
+++ b/testing/kuttl/e2e/security-context/00--cluster.yaml
@@ -1,0 +1,26 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: security-context
+  labels: { postgres-operator-test: kuttl }
+spec:
+  postgresVersion: ${KUTTL_PG_VERSION}
+  instances:
+    - name: instance1
+      replicas: 1
+      dataVolumeClaimSpec: { accessModes: [ReadWriteOnce], resources: { requests: { storage: 1Gi } } }
+  backups:
+    pgbackrest:
+      repos:
+      - name: repo1
+        volume:
+          volumeClaimSpec: { accessModes: [ReadWriteOnce], resources: { requests: { storage: 1Gi } } }
+  proxy:
+    pgBouncer:
+      replicas: 1
+  userInterface:
+    pgAdmin:
+      dataVolumeClaimSpec: { accessModes: [ReadWriteOnce], resources: { requests: { storage: 1Gi } } }
+  monitoring:
+    pgmonitor:
+      exporter: {}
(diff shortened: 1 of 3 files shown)


testing/kuttl/e2e/security-context/00--cluster.yaml
```diff
@@ -0,0 +1,23 @@
...
+      repos:
+      - name: repo1
+        volume:
+          volumeClaimSpec: { accessModes: [ReadWriteOnce], resources: { requests: { storage: 1Gi } } }
+  proxy:
+    pgBouncer:
+      replicas: 1
+  userInterface:
+    pgAdmin:
+      dataVolumeClaimSpec: { accessModes: [ReadWriteOnce], resources: { requests: { storage: 1Gi } } }
```
jmckulk, line 23: newline :pick:

testing/kuttl/e2e/security-context/00-assert.yaml
```diff
@@ -0,0 +1,180 @@
...
+kind: Pod
+metadata:
+  labels:
+    postgres-operator.crunchydata.com/cluster: security-context
+    postgres-operator.crunchydata.com/data: postgres
+    postgres-operator.crunchydata.com/instance-set: instance1
+    postgres-operator.crunchydata.com/patroni: security-context-ha
+    postgres-operator.crunchydata.com/role: master
+spec:
+  containers:
```
jmckulk, line 50: This looks like it will get all of the pods and containers but I wonder if there is a more generic way to do it. Could you have one assert that only checked for pods with the cluster label in metadata, then leave out the container name :recycle: ?
```
  containers:
  - securityContext:
      allowPrivilegeEscalation: false
      privileged: false
      readOnlyRootFilesystem: true
      runAsNonRoot: true
```

jmckulk (commented):
Should we add in the exporter container? I know the story says move over current tests but I would think the exporter is low hanging fruit

testing/kuttl/e2e/security-context/00--cluster.yaml
tjmoore4: :+1: updated.

testing/kuttl/e2e/security-context/00-assert.yaml
tjmoore4: I know that's more or less how the existing test does things, but I think there may be value in explicitly checking each container in case a container were to somehow be missing.

Otherwise, I'm a little unclear how best to do that with kuttl. I don't believe the above works as expected. I think we'd need to capture and loop through each container name, but I'm not aware with a good way to do that with kuttl. Any thoughts on that?

jmckulk: I do not. I tried out a couple of things and I couldn't get it working. This seems like the way to go :+1:

jmckulk (approved):
Sweet :+1: thanks for adding that

PR #3169, opened 2022-04-25
Replace TestReconcilerHandleDelete With Kuttl Test

benjaminjb:
**Checklist:**
   - [ ] 
**Type of Changes:**
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

TestReconcilerHandleDelete was prone to flakes when run with `envtest-existing`

Replace TestReconcilerHandleDelete with a KUTTL test with matching functionality.

**Other Information**:
Issue [sc-14264]


testing/kuttl/e2e/delete/13-delete-cluster-and-check.yaml
```diff
@@ -0,0 +1,37 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  # Delete all previous events
+  - script: |
+      kubectl delete --namespace="${NAMESPACE}" events --all
```
tjmoore4, line 7: :question:  I'm curious about what happens when all the tests are run in the same namespace. I don't think we currently do a lot of (any?) event checking with our other tests, but I could see this being an issue in the future. Did you happen to run the tests in a single namespace during your testing?

testing/kuttl/e2e/delete/02-errors.yaml
```diff
@@ -0,0 +1,30 @@
...
+---
+apiVersion: v1
+kind: Service
+labels:
+    postgres-operator.crunchydata.com/cluster: delete
+---
+apiVersion: v1
+kind: Secret
+labels:
+    postgres-operator.crunchydata.com/cluster: delete
```
tjmoore4, line 30: :pick: missing newline

testing/kuttl/e2e/delete/14-errors.yaml
```diff
@@ -0,0 +1,30 @@
...
+---
+apiVersion: v1
+kind: Service
+labels:
+    postgres-operator.crunchydata.com/cluster: delete-switchover
+---
+apiVersion: v1
+kind: Secret
+labels:
+    postgres-operator.crunchydata.com/cluster: delete-switchover
```
tjmoore4, line 30: :pick: missing newline

testing/kuttl/e2e/delete/14-errors.yaml
```diff
@@ -0,0 +1,30 @@
...
+---
+apiVersion: v1
+kind: Service
+labels:
+    postgres-operator.crunchydata.com/cluster: delete-switchover
+---
+apiVersion: v1
+kind: Service
```
tjmoore4, line 23: :thinking: Do we need `Service` twice here?

testing/kuttl/e2e/delete/22-errors.yaml
```diff
@@ -0,0 +1,30 @@
...
+---
+apiVersion: v1
+kind: Service
+labels:
+    postgres-operator.crunchydata.com/cluster: delete-not-running
+---
+apiVersion: v1
+kind: Secret
+labels:
+    postgres-operator.crunchydata.com/cluster: delete-not-running
```
tjmoore4, line 30: :pick: missing newline

testing/kuttl/e2e/delete/22-errors.yaml
```diff
@@ -0,0 +1,30 @@
...
+---
+apiVersion: v1
+kind: Service
+labels:
+    postgres-operator.crunchydata.com/cluster: delete-not-running
+---
+apiVersion: v1
+kind: Service
```
tjmoore4, line 23: :thinking: Same question as above, do we need both `Service` references?

testing/kuttl/e2e/delete/21-errors.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: delete-not-running
+status:
+  instances:
+    - name: instance1
+      readyReplicas: 1
```
tjmoore4, line 8: :thinking: Should this error check happen before the delete step? Is the purpose here to validate the cluster never reached a running state, or is it checking something else?

testing/kuttl/e2e/delete/README.md
```diff
@@ -0,0 +1,21 @@
+### Delete test
```
tjmoore4, line 1: :100: Appreciate the README here

tjmoore4 (commented):
Looks good overall, a few questions and nitpicks.

testing/kuttl/e2e/delete/13-delete-cluster-and-check.yaml
benjaminjb: My initial thought here was 
a) most of the time, we run several tests at once, but in different namespaces -- I think actually that some of our other KUTTL tests would fail if we ran several tests at once in the same namespace just because our selectors aren't specific (though I don't think that's true of all the tests)
b) when we do run in the same namespace (which is how I used to test in AWS, since I needed a secret for pulling images), I would make sure to turn it to run only a single test at a time.

That said, I think I can obviate the need for this by getting the events for the specific pod and then comparing the created timestamps...

testing/kuttl/e2e/delete/21-errors.yaml
benjaminjb: It's to validate that it never runs, yeah. 

(That said, I put this in to make sure that the KUTTL test matches the functionality of the test I'm removing, but I'm still confused -- as I noted on our spreadsheet -- what this test case is really supposed to be testing. Like: maybe there was a possibility that an unrecoverable cluster would retain its finalizer and so never be deleted? I mean, I guess that could happen. I just am not 100% sure.)

testing/kuttl/e2e/delete/13-delete-cluster-and-check.yaml
benjaminjb: Changed to avoid the need for this, which I think is better testing practice regardless.

testing/kuttl/e2e/delete/21-errors.yaml
tjmoore4: I suppose part of what I was wondering was if `21-errors.yaml` should effectively move up in the ordering, perhaps to `20-errors.yaml`, so that error check is run before the `delete` step. Otherwise, I think this test makes sense as it covers what was in the original EnvTest (at least implicitly).

tjmoore4 (approved):
:+1: This looks good with the changes. The only thing I think might be worth trying is the ordering of the one cluster check (outlined below), but I don't think that's necessarily a blocker.

PR #3173, opened 2022-04-27
Replace TestReconcileCerts owner reference test with Kuttl test

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

This creates a Kuttl test to verify proper owner references are
set on the root CA certificate secret which replaces the current
EnvTest implementation.

**Other Information**:
Issue: [sc-14269]

### DIFF
--- a/internal/controller/postgrescluster/pki_test.go
+++ b/internal/controller/postgrescluster/pki_test.go
@@ -25,16 +25,13 @@ import (
 	"reflect"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/pkg/errors"
 	"gotest.tools/v3/assert"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/crunchydata/postgres-operator/internal/naming"
@@ -43,6 +40,11 @@ import (
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
+// TestReconcileCerts tests the proper reconciliation of the root ca certificate
+// secret, leaf certificate secrets and the updates that occur when updates are
+// made to the cluster certificates generally. For the removal of ownership
+// references and deletion of the root CA cert secret, a separate Kuttl test is
+// used due to the need for proper garbage collection.
 func TestReconcileCerts(t *testing.T) {
 	// Garbage collector cleans up test resources before the test completes
 	if strings.EqualFold(os.Getenv("USE_EXISTING_CLUSTER"), "true") {
@@ -137,38 +139,6 @@ func TestReconcileCerts(t *testing.T) {
 			}
 		})
 
-		t.Run("remove owner references after deleting first cluster", func(t *testing.T) {
-
-			if !strings.EqualFold(os.Getenv("USE_EXISTING_CLUSTER"), "true") {
-				t.Skip("requires a running garbage collection controller")
-			}
-
-			err = tClient.Get(ctx, client.ObjectKeyFromObject(cluster1), cluster1)
-			assert.NilError(t, err)
-
-			err = tClient.Delete(ctx, cluster1)
-			assert.NilError(t, err)
-
-			err = wait.Poll(time.Second/2, Scale(time.Second*15), func() (bool, error) {
-				err := tClient.Get(ctx, client.ObjectKeyFromObject(rootSecret), rootSecret)
-				return len(rootSecret.ObjectMeta.OwnerReferences) == 1, err
-			})
-			assert.NilError(t, err)
-
-			assert.Check(t, len(rootSecret.ObjectMeta.OwnerReferences) == 1, "owner reference not removed")
-
-			expectedOR := metav1.OwnerReference{
-				APIVersion: "postgres-operator.crunchydata.com/v1beta1",
-				Kind:       "PostgresCluster",
-				Name:       "hippocluster2",
-				UID:        cluster2.UID,
-			}
-
-			if len(rootSecret.ObjectMeta.OwnerReferences) > 0 {
-				assert.Equal(t, rootSecret.ObjectMeta.OwnerReferences[0], expectedOR)
-			}
-		})
-
 		t.Run("root certificate is returned correctly", func(t *testing.T) {
 
 			fromSecret, err := getCertFromSecret(ctx, tClient, naming.RootCertSecret, namespace, "root.crt")
@@ -200,28 +170,6 @@ func TestReconcileCerts(t *testing.T) {
 
 			// check that the returned cert matches the cert from the secret
 			assert.DeepEqual(t, fromSecret, returnedRoot.Certificate)
-		})
-
-		t.Run("root CA secret is deleted after final cluster is deleted", func(t *testing.T) {
-
-			if !strings.EqualFold(os.Getenv("USE_EXISTING_CLUSTER"), "true") {
-				t.Skip("requires a running garbage collection controller")
-			}
-
-			err = tClient.Get(ctx, client.ObjectKeyFromObject(cluster2), cluster2)
-			assert.NilError(t, err)
-
-			err = tClient.Delete(ctx, cluster2)
-			assert.NilError(t, err)
-
-			err = wait.Poll(time.Second/2, Scale(time.Second*15), func() (bool, error) {
-				if err := tClient.Get(ctx,
-					client.ObjectKeyFromObject(rootSecret), rootSecret); apierrors.ReasonForError(err) == metav1.StatusReasonNotFound {
-					return true, err
-				}
-				return false, nil
-			})
-			assert.Assert(t, apierrors.IsNotFound(err))
 		})
 
 	})
--- a/testing/kuttl/e2e/root-cert-ownership/01-check-owners.yaml
+++ b/testing/kuttl/e2e/root-cert-ownership/01-check-owners.yaml
@@ -1,0 +1,15 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  # Get a list of the current owners of the root ca cert secret and verify that
+  # both owners are listed.
+  - script: |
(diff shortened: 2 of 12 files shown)


internal/controller/postgrescluster/pki_test.go
```diff
@@ -44,10 +40,6 @@ import (
 )
 
 func TestReconcileCerts(t *testing.T) {
```
benjaminjb, line 42: Would it be a pain to write a comment for this test stating
* what the test cases are here
* what test case is being covered in the kuttl test <insert path to kuttl test folder here>
?

internal/controller/postgrescluster/pki_test.go
```diff
@@ -44,10 +40,6 @@ import (
...
-	// Garbage collector cleans up test resources before the test completes
-	if strings.EqualFold(os.Getenv("USE_EXISTING_CLUSTER"), "true") {
-		t.Skip("USE_EXISTING_CLUSTER: Test fails due to garbage collection")
-	}
```
benjaminjb, line 50: I think we want this still so that this test won't run at all in our existing cluster tests (and then later on, we _should_ have a ticket for removing all these guards and removing all notion of existing cluster tests from our makefile, github actions, etc.).

testing/kuttl/e2e/root-cert-ownership/01-check-owners.yaml
```diff
@@ -0,0 +1,16 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  # Get a list of the current owners of the root ca cert secret and verify that
+  # both owners are listed.
+  - script: |
+      CURRENT_OWNERS=$(kubectl --namespace="${NAMESPACE}" get secret \
+      pgo-root-cacert -o jsonpath='{.metadata.ownerReferences[*].name}')
```
benjaminjb, line 9: I think it's easier to read if we indent this line that continues the line above.

testing/kuttl/e2e/root-cert-ownership/01-check-owners.yaml
```diff
@@ -0,0 +1,16 @@
...
+      echo "$CURRENT_OWNERS"
```
benjaminjb, line 10: Don't need this in the test, do we? Hmmm, you know, now that I'm thinking about it, I've been removing all these echoes from my tests, but if they're useful for debugging maybe they're also useful in testing?

(I'd still remove it for now since that's mostly our style, right? But might be worth a chat.)

testing/kuttl/e2e/root-cert-ownership/03-check-owners.yaml
```diff
@@ -0,0 +1,19 @@
...
+      sleep 2 # this sleep allows time for the owner reference list to be updated
+      CURRENT_OWNERS=$(kubectl --namespace="${NAMESPACE}" get secret \
+      pgo-root-cacert -o jsonpath='{.metadata.ownerReferences[*].name}')
+      echo "$CURRENT_OWNERS"
+      if [[ "$CURRENT_OWNERS" == *"owner1"* ]]; then
+          echo "first if"
+          exit 1
+      fi
+      if [[ "$CURRENT_OWNERS" != *"owner2"* ]]; then
+          echo "second if"
```
benjaminjb, line 17: What error does the test print without the echoes here? Do we need to put an echo in each `if` to get a clear understanding why it failed?

testing/kuttl/e2e/root-cert-ownership/03-check-owners.yaml
```diff
@@ -0,0 +1,19 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  # Get a list of the current owners of the root ca cert secret and verify that
+  # owner1 is no longer listed and owner2 is found.
+  - script: |
+      sleep 2 # this sleep allows time for the owner reference list to be updated
+      CURRENT_OWNERS=$(kubectl --namespace="${NAMESPACE}" get secret \
+      pgo-root-cacert -o jsonpath='{.metadata.ownerReferences[*].name}')
```
benjaminjb, line 10: Indent for readability?

testing/kuttl/e2e/root-cert-ownership/05--check-secret.yaml
```diff
@@ -0,0 +1,26 @@
...
+  # and 'owner2' are not listed.
+  # If there are no other PostgresClusters in the namespace, the 'pgo-root-cacert'
+  # secret should be deleted.
+  - script: |
+      NUM_CLUSTERS=$(kubectl --namespace="${NAMESPACE}" get postgrescluster --output name | wc -l)
+      ROOT_SECRET=$(kubectl --namespace="${NAMESPACE}" get secret pgo-root-cacert --output name | wc -l)
+      if [[ "$NUM_CLUSTERS" != 0 ]]; then
+          CURRENT_OWNERS=$(kubectl --namespace="${NAMESPACE}" get secret \
+          pgo-root-cacert -o jsonpath='{.metadata.ownerReferences[*].name}')
+          echo "$CURRENT_OWNERS"
```
benjaminjb, line 15: Indent, remove echo

testing/kuttl/e2e/root-cert-ownership/05--check-secret.yaml
```diff
@@ -0,0 +1,26 @@
...
+          if [[ "$CURRENT_OWNERS" == *"owner1"* ]]; then
+              exit 1
+          fi
+          if [[ "$CURRENT_OWNERS" == *"owner1"* ]]; then
+              exit 1
+          fi
+      else
+          if [[ "$ROOT_SECRET" != 0 ]]; then
```
benjaminjb, line 23: Since we only use ROOT_SECRET here, I wonder if you could move line 11's kubectl call into this `else` loop. (You could put it just as it is above line 23 -- usually if I'm just using a var once, I'll want to cut it out, but for readability, `[[ "$ROOT_SECRET" != 0 ]]` might be preferable to ``[[ $(kubectl --namespace="${NAMESPACE}" get secret pgo-root-cacert --output name | wc -l) != 0 ]]`!

testing/kuttl/e2e/root-cert-ownership/README.md
```diff
@@ -0,0 +1,23 @@
...
+- 02: Delete the first cluster, assert that the second cluster and the root cert
+secret are still present and that the first cluster is not present
+- 03: Check that the secret shows the second cluster as an owner but does not show
+the first cluster as an owner
+
+#### Delete the second cluster and verify the root certificate secret ownership
+
+- 04: Delete the second cluster, assert that both clusters are not present
+- 05: Check the number of clusters in the namespace. If there are any remaining
+clusters, ensure that neither the first or second cluster are listed as owners.
```
benjaminjb, line 21:
```suggestion
clusters, ensure that secret shows neither the first nor second cluster as an owner.
```
The sentence didn't originally say what they should be not listed as owners of, so I just reworded it using the way you mentioned it in lines 08 and 14-5.

benjaminjb (commented):
Functionality good, just a few cleanup and discussion questions.

internal/controller/postgrescluster/pki_test.go
tjmoore4: Generally, I don't think we comment the test functions, but it may be worth adding a note in this case since we're pulling out a subset of the original test. I'll see what I can come up with.

internal/controller/postgrescluster/pki_test.go
tjmoore4: Yeah, I considered this in general, but neglected to consider our github actions, that may be an issue. I'll add this back and put a note on the current cleanup story (let me know if you think it should be split out in a separate one).

testing/kuttl/e2e/root-cert-ownership/01-check-owners.yaml
tjmoore4: They are useful for debugging, but I think you're correct on the general style we've been using thus far. I'll remove them for now, we can always revisit this later and approach things more holistically.

testing/kuttl/e2e/root-cert-ownership/03-check-owners.yaml
tjmoore4: Per above, I'm removing the echo output. These in particular probably should've been removed or updated for more clarity.

testing/kuttl/e2e/root-cert-ownership/05--check-secret.yaml
tjmoore4: Agreed on moving to the 'else'. I originally just grouped this with `NUM_CLUSTERS` to make it a bit cleaner, but it technically doesn't need to be called for the 'else' case, so it's worth moving. Also agreed on not removing ROOT_SECRET all together, I find that statement a bit tricky to parse at first glance.

internal/controller/postgrescluster/pki_test.go
benjaminjb: I think it would be nice to do it all in one ticket and then when someone asks we can say "envtest-existing? doesn't sound familiar."

testing/kuttl/e2e/root-cert-ownership/05--check-secret.yaml
```diff
@@ -0,0 +1,26 @@
...
+  # secret should be deleted.
+  - script: |
+      NUM_CLUSTERS=$(kubectl --namespace="${NAMESPACE}" get postgrescluster --output name | wc -l)
+      if [[ "$NUM_CLUSTERS" != 0 ]]; then
+          CURRENT_OWNERS=$(kubectl --namespace="${NAMESPACE}" get secret \
+            pgo-root-cacert -o jsonpath='{.metadata.ownerReferences[*].name}')
+          if [[ "$CURRENT_OWNERS" == *"owner1"* ]]; then
+              exit 1
+          fi
+          if [[ "$CURRENT_OWNERS" == *"owner1"* ]]; then
```
benjaminjb, line 17:
```suggestion
          if [[ "$CURRENT_OWNERS" == *"owner2"* ]]; then
```

tjmoore4: Nice catch! Updated, thanks!

benjaminjb (approved):

internal/controller/postgrescluster/pki_test.go
benjaminjb: Actually, now I'm second-guessing myself about these skips if `USE_EXISTING_CLUSTER` is true: 
* on one hand, we will not be using an existing cluster, so it makes sense to remove these guard-rails
* on the other hand, if someone does run the test with USE_EXISTING_CLUSTER=true, these guard-rails will prevent the tests from erroring out when there's no real error

What do y'all think?

PR #3193, opened 2022-05-09
Enable seccomp on containers

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other -- security enhancement

As of Kubernetes v1.19, SecurityContext has a seccompProfile field that can be set to RuntimeDefault to limit syscalls. We were not using that.

This PR adds that setting to the containers in order to
(a) limit syscalls from PGO-managed containers, while
(b) not preventing users from using other tools involving sidecars, etc.

**Other Information**:
Issue [sc-11286]

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -38,4 +38,6 @@ spec:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
           runAsNonRoot: true
+          seccompProfile:
+            type: RuntimeDefault
       serviceAccountName: pgo
--- a/internal/initialize/security_test.go
+++ b/internal/initialize/security_test.go
@@ -97,8 +97,11 @@ func TestRestrictedSecurityContext(t *testing.T) {
 				"Containers must be required to run as non-root users.")
 		}
 
-		assert.Assert(t, sc.SeccompProfile == nil,
-			"The RuntimeDefault seccomp profile must be required, or allow specific additional profiles.")
+		if assert.Check(t, sc.SeccompProfile != nil) {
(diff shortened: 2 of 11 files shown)


internal/initialize/security_test.go
```diff
@@ -97,7 +97,7 @@ func TestRestrictedSecurityContext(t *testing.T) {
 				"Containers must be required to run as non-root users.")
 		}
 
-		assert.Assert(t, sc.SeccompProfile == nil,
+		assert.Assert(t, sc.SeccompProfile != nil,
 			"The RuntimeDefault seccomp profile must be required, or allow specific additional profiles.")
```
cbandy, line 101: 🔧 We should check the value, too. I'm pretty sure it's possible to compare to an untyped literal here.

The message of other assertions appears to quote the standard. I think the following makes it clear which rule we're targeting and that we're compliant:

```suggestion
		if assert.Check(t, sc.SeccompProfile != nil) {
			assert.Assert(t, *sc.SeccompProfile.Type == "RuntimeDefault",
				"Seccomp profile must be explicitly set to one of the allowed values.")
		}
```

config/manager/manager.yaml
```diff
@@ -38,4 +38,6 @@ spec:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
           runAsNonRoot: true
+          seccompProfile:
+            type: RuntimeDefault
```
cbandy, line 42: 👍🏻

cbandy (changes requested):

cbandy (approved):

PR #3217, opened 2022-05-18
Add more assertions to delete and switchover tests

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

The switchover test occasionally fails when it does not find pods that have changed their role. It's possible that we aren't detecting the change because we aren't effectively recording the role before the switchover.

Wait for the expected labels before using a selector to record their original values.

As a follow-up to c13154e93977c09e03a3f046eae4493a54bbb135, this also checks that Patroni endpoints are removed during cluster deletion.

### DIFF
--- a/testing/kuttl/e2e/delete/02-errors.yaml
+++ b/testing/kuttl/e2e/delete/02-errors.yaml
@@ -9,6 +9,12 @@ kind: StatefulSet
 labels:
     postgres-operator.crunchydata.com/cluster: delete
 ---
+# Patroni DCS objects are not owned by the PostgresCluster.
+apiVersion: v1
+kind: Endpoints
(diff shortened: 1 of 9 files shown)


testing/kuttl/e2e/delete/02-errors.yaml
```diff
@@ -10,6 +10,11 @@ labels:
     postgres-operator.crunchydata.com/cluster: delete
 ---
 apiVersion: v1
+kind: Endpoints
```
jmckulk, line 13: :pick: a comment saying why we care about the endpoint(s) would be helpful

jmckulk (approved):

PR #3226, opened 2022-05-24
Simplify the PKI implementation

cbandy:
The original implementation dynamically assigns functions that return errors so we can swap them under test. Errors from these calls are wrapped in sentinels so they can be identified at runtime. In practice, however, these errors are never examined.

Fields on the "Certificate" and "PrivateKey" types are no longer exported, making them opaque to consumers except for the PEM marshaling methods. These methods are now verified using OpenSSL.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Bug fix
 - [x] Other

Certificates and private keys are regenerated with their subject does not match the controller's intent. They are no longer regenerated when their Secret cannot be read from Kubernetes.

**Other Information**:

Issue: [sc-14620]


internal/pki/encoding_test.go
```diff
@@ -15,340 +13,182 @@ package pki
...
-		}
+		// The "openssl x509" command parses X.509 certificates.
+		cmd := exec.Command(openssl, "x509",
+			"-in", certFile, "-inform", "PEM", "-noout", "-text")
 
-		if certificate != nil {
-			t.Fatalf("expected certificate to be nil")
-		}
+		output, err := cmd.CombinedOutput()
+		assert.NilError(t, err, "%q\n%s", cmd.Args, output)
```
benjaminjb, line 99: Are we printing the cmd.Args here to make sure the `certFile` is what we expect it to be or for some other reason?

cbandy: To show the command that failed. The `openssl pkey` on my Mac, for example, doesn't support the `-check` argument. Removing the skip in that similar test fails like this:

```
--- FAIL: TestPrivateKeyTextMarshaling (0.01s)
    --- FAIL: TestPrivateKeyTextMarshaling/ReadByOpenSSL (0.01s)
        encoding_test.go:158: using "/usr/bin/openssl"
            LibreSSL 2.8.3
            built on: date not available
            platform: information not available
            options:  bn(64,64) rc4(16x,int) des(idx,cisc,16,int) blowfish(idx)
            compiler: information not available
            OPENSSLDIR: "/private/etc/ssl"
        --- FAIL: TestPrivateKeyTextMarshaling/ReadByOpenSSL/Check (0.01s)
            encoding_test.go:187: assertion failed: error is not nil: exit status 1: ["/usr/bin/openssl" "pkey" "-check" "-in" "/var/folders/bw/_dxdzsmx6h1ffcgp6rvfh55c0000gn/T/TestPrivateKeyTextMarshalingReadByOpenSSL1838345719/001/key.pem" "-inform" "PEM" "-noout" "-text"]
                Unknown cipher check
                Usage pkey [options]
                where options are
                -in file        input file
                -inform X       input format (DER or PEM)
                -passin arg     input file pass phrase source
                -outform X      output format (DER or PEM)
                -out file       output file
                -passout arg    output file pass phrase source
```

internal/pki/pki_test.go
```diff
@@ -15,32 +13,63 @@ package pki
...
+	assert.Assert(t, zero.DNSNames() == nil)
+}
+
+func TestCertificateHasSubject(t *testing.T) {
+	zero := Certificate{}
+
+	// The zero value has no subject.
+	for _, cn := range []string{"", "any"} {
+		for _, dns := range [][]string{nil, {}, {"any"}} {
+			assert.Assert(t, !zero.hasSubject(cn, dns), "for (%q, %q)", cn, dns)
```
benjaminjb, line 54: This feels like it is written to test a variety of cn and dns, but won't `hasSubject` set ok == false early?

benjaminjb (commented):
Finished first review of the pki package, had a few comments I wanted to share before I go over the usage of the pki package elsewhere (which is mostly straightforward substitution of the old funcs with the new funcs + params, so not anticipating much difficulty there).

benjaminjb (commented):
Finished first pass through, 👍

internal/pki/pki_test.go
cbandy: Yes, it does. Any arguments have the same outcome.

internal/controller/postgrescluster/patroni.go
```diff
@@ -352,22 +352,16 @@ func (r *Reconciler) reconcileReplicationSecret(
 	err := errors.WithStack(client.IgnoreNotFound(
 		r.Client.Get(ctx, client.ObjectKeyFromObject(existing), existing)))
 
-	clientLeaf := pki.NewLeafCertificate("", nil, nil)
-	clientLeaf.DNSNames = []string{postgres.ReplicationUser}
-	clientLeaf.CommonName = clientLeaf.DNSNames[0]
+	leaf := &pki.LeafCertificate{}
+	_ = leaf.Certificate.UnmarshalText(existing.Data[naming.ReplicationCert])
+	_ = leaf.PrivateKey.UnmarshalText(existing.Data[naming.ReplicationPrivateKey])
```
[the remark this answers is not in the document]
cbandy: You've hit on it. It kinda doesn't matter what's not-good in the Secret; if it doesn't contain a certificate we want, then we generate a replacement.

> OK, I've talked myself into it, but tell me if I'm missing anything.

Did you convince yourself this is okay as-is, or that we should do something more with the errors?

🔧 You're right that it is unusual to throw away errors. Any time it happens there should be a comment explaining why it seemed safe at the time.

internal/pki/pki.go
```diff
@@ -15,17 +15,199 @@
...
+}
+
+// hasSubject checks that c has these values in its subject.
+func (c Certificate) hasSubject(commonName string, dnsNames []string) bool {
+	ok := c.x509 != nil &&
+		c.x509.Subject.CommonName == commonName &&
+		len(c.x509.DNSNames) == len(dnsNames)
+
+	for i := range dnsNames {
+		ok = ok && c.x509.DNSNames[i] == dnsNames[i]
```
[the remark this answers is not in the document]
benjaminjb: Ah, right, I ran a test that skipped checking if the lengths were the same, so `ok` was true and I ran into that index error. I don't think it's necessary to rewrite this.

benjaminjb (approved):

internal/pki/pki_test.go
cbandy: When I remove `c.x509 != nil` from the lines you highlighted, this test panics and other tests pass. So this test is the only time `hasSubject` is called with a zero value.

🤔 That means the other exported functions (probably) don't hit this path, but it feels irresponsible to use the pointer without first checking if it is nil.

I think both the check and the test should stay. You're right that the one line is covered regardless of the arguments passed in, but this seems like a quick enough way to show that in the test. Should there be fewer assertions? Drop the loops and do just a couple?

```go
assert.Assert(t, !zero.hasSubject("", nil))
assert.Assert(t, !zero.hasSubject("any", []string{"thing"}))
```

internal/pki/common.go
```diff
@@ -15,40 +13,94 @@ package pki
...
+
+	parsed, _ := x509.ParseCertificate(bytes)
+	return parsed, err
+}
+
+func generateRootCertificate(
+	privateKey *ecdsa.PrivateKey, serialNumber *big.Int,
+) (*x509.Certificate, error) {
+	const rootCommonName = "postgres-operator-ca"
+	const rootExpiration = time.Hour * 24 * 365 * 10
```
benjaminjb, line 82: Wait, sorry, just double-checking -- 10 years for the root and a year for the leaf?

cbandy: Yep, same as before.

https://github.com/CrunchyData/postgres-operator/blob/d0c3dae8236c7b7dd40f7003b21ebd9359871d92/internal/pki/root.go#L39-L41

internal/controller/postgrescluster/patroni.go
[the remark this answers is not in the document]
cbandy: Added comments in https://github.com/CrunchyData/postgres-operator/pull/3226/commits/ca5461e2a04a4038bd03caec3f978b80a113c842.

internal/controller/postgrescluster/pki.go
```diff
@@ -55,22 +55,17 @@ func (r *Reconciler) reconcileRootCertificate(
...
+	// Unmarshal and validate the stored root. These first errors can be ignored
+	// because they result in an invalid root which is then correctly regenerated.
+	root := &pki.RootCertificateAuthority{}
+	_ = root.Certificate.UnmarshalText(existing.Data[keyCertificate])
+	_ = root.PrivateKey.UnmarshalText(existing.Data[keyPrivateKey])
+
+	if err == nil && !pki.RootIsValid(root) {
+		root, err = pki.NewRootCertificateAuthority()
 		err = errors.WithStack(err)
 	}
```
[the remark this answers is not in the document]
cbandy: Might as well.

PR #3229, opened 2022-05-26
Rotate leaf cert before expiration

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [x] New feature
 - [ ] Bug fix
 - [x] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Certs renew when expired

* automates cert renewal before expiry -- go with 2/3rd lifespan as per cert-manager
* update docs
* use helper func for shellcheck in one test that wasn't updated

**Other Information**:
Issue [sc-11173]


tjmoore4 (commented):
This looks good. One very small doc change that I think we'll want to make and a comment that could be addressed.

docs/content/tutorial/administrative-tasks.md
```diff
@@ -35,10 +35,30 @@ To turn a Postgres cluster that is shut down back on, you can set `spec.shutdown
...
+### Manually Triggering a Certificate Rotation
+
+If you want to rotate a single client certificate, you can regenerate the certificate
+of an existing cluster by deleting the `tls.key` field from its certificate Secret.
 
+If you want to rotate all the client certificates for all the cluster, you can do so by
+deleting the root certificate Secret `pgo-root-cacert` or by deleting the `root.crt` field
+from that secret. Once PGO goes through a reconcile loop, it will regenerate that Secret
+and use that Secret's certificate to generate new client certificates. If following this
+procedure, it may be necessary to trigger a reconcile loop by adding an annotation to a cluster.
```
cbandy, line 56:
```suggestion
and use that Secret's certificate to generate new client certificates. When following this
procedure, it may be necessary to trigger a reconcile loop by adding an annotation to a cluster.
```

docs/content/tutorial/administrative-tasks.md
```diff
@@ -55,7 +75,7 @@ When changing the PostgreSQL certificate authority, make sure to update
 PGO automatically notifies PgBouncer when there are changes to the contents of
 PgBouncer certificate Secrets. Recent PgBouncer versions load those changes
 without downtime, but versions prior to 1.16.0 need to be restarted manually.
-There are a few ways to do it:
+There are a few ways to restart PgBouncer if necessary to reload Secrets:
```
cbandy, line 78:
```suggestion
There are a few ways to restart an old PgBouncer to reload Secrets:
```

internal/pki/pki.go
```diff
@@ -194,9 +197,24 @@ func (root *RootCertificateAuthority) leafIsValid(leaf *LeafCertificate) bool {
...
 	return ok
 }
 
+// isBeforeRenewalTime checks if the result of `currentTime`
+// is after the default renewal time of
+// 1/3rds before the certificate's expiry
+func isBeforeRenewalTime(before, after time.Time) bool {
+	renewalDuration := after.Sub(before) / renewalRatio
+	renewalTime := after.Add(-1 * renewalDuration)
+	return currentTime().Before(renewalTime)
```
cbandy, line 215: 🌱 Knowing that we read from a cache, I wonder if we should do something somewhere to avoid generating certificates when the cache is stale.

Have you ever seen a leaf get generated "extra" times while testing?

cbandy (approved):

internal/pki/pki.go
benjaminjb: I have not seen extra generating of the leaf cert, but here's the situation where I didn't see one: 

I had a print line in here at one time, and a setup where most of my certs were year-long, and one cert was valid for 3 hours (starting an hour ago, so I expected a regen in one hour); when that hour was up and I had triggered a reconcile, I only saw this give a `false` value once for that one cert.

This does make me want to put in an event when regenerating certs though, so that we could see if/when certs get regenerated multiple times before the cache is updated.

PR #3234, opened 2022-06-01
Change the volume claim test into a unit test

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

`TestPersistentVolumeClaimLimitations` requires a running Kubernetes cluster and flakes during PR checks.

`TestHandlePersistentVolumeClaimError` uses an event recorder and mocked API errors to verify the correct events and conditions are set by `handlePersistentVolumeClaimError`.

**Other Information**:

Issue: [sc-14270]

### DIFF
--- a/internal/testing/events/recorder.go
+++ b/internal/testing/events/recorder.go
@@ -1,0 +1,110 @@
+/*
+ Copyright 2022 Crunchy Data Solutions, Inc.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+*/
+
+package events
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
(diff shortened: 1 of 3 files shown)


internal/testing/events/recorder.go
```diff
@@ -0,0 +1,110 @@
...
+			Type:                eventtype,
+		})
+	}
+
+	return &recorder
+}
+
+var _ record.EventRecorder = (*Recorder)(nil)
+
+func (*Recorder) AnnotatedEventf(object runtime.Object, annotations map[string]string, eventtype, reason, messageFmt string, args ...interface{}) {
```
tjmoore4, line 98: :thinking: Why do we need this method if it results in a panic? Is it just to match the [record package](https://pkg.go.dev/k8s.io/client-go/tools/record#FakeRecorder.AnnotatedEventf)?

tjmoore4 (approved):
:+1: Looks good. No blockers, one question.

internal/testing/events/recorder.go
cbandy: Yep. [EventRecorder](https://pkg.go.dev/k8s.io/client-go/tools/record#EventRecorder) is the interface from the record package, and it's what [controller-runtime uses](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.1/pkg/cluster#Cluster).

The [future](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/383-new-event-api-ga-graduation) is the [events package](https://pkg.go.dev/k8s.io/client-go/tools/events#EventRecorder), which has a simpler API (and implementation!). That future has no "annotated event" concept, so we should stay away from it now.

PR #3235, opened 2022-06-01
Use timeline as status to prevent multiple failovers

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Switchover was not idempotent

Use timeline in status to prevent multiple failovers.

**Other Information**:
Issue [sc-14610]


internal/patroni/api.go
```diff
@@ -177,3 +178,24 @@ func (exec Executor) RestartPendingMembers(ctx context.Context, role, scope stri
...
+	var stdout, stderr bytes.Buffer
+
+	// The following exits zero when it is able to read the DCS and communicate
+	// with the Patroni HTTP API. It prints the result of calling "GET /cluster"
+	// - https://github.com/zalando/patroni/blob/v2.1.1/patroni/ctl.py#L849
+	err := exec(ctx, nil, &stdout, &stderr,
+		"patronictl", "list")
+
+	regEx := regexp.MustCompile(`\|\s+Leader\s+\|\s+running\s+\|\s+(\d+)\s+\|`)
+	timeline := regEx.FindStringSubmatch(stdout.String())
```
benjaminjb, line 194: Couple notes here:

(a) I'm swallowing stderr here because I've not seen anything interesting, but maybe if it has something we should exit like an err?
(b) If we have any use for `patronictl list`, then we might just want to pass stdout through and leave the parsing for the calling func?

testing/kuttl/e2e/delete/13-delete-cluster-and-check.yaml
```diff
@@ -19,6 +19,11 @@ commands:
           --output=jsonpath={.items..metadata.name}
       )
 
+      echo "DELETE: Found primary ${PRIMARY} and replica ${REPLICA} pods"
+
+      if [ -z "$PRIMARY" ]; then exit 1; fi      
+      if [ -z "$REPLICA" ]; then exit 1; fi
```
benjaminjb, line 25: I actually saw this test pass with replica being not found!

internal/controller/postgrescluster/patroni.go
```diff
@@ -499,6 +512,44 @@ func (r *Reconciler) reconcilePatroniSwitchover(ctx context.Context,
...
+	// We return from here without calling for an explicit requeue, but since we're updating
+	// the object, we will reconcile this again for the actual switchover/failover action.
+	if statusTimeline == nil || (statusTimeline != nil && *statusTimeline == 0) {
+		log.V(1).Info("Setting ExpectedTimeline to current timeline")
+		cluster.Status.Patroni.ExpectedTimeline = &timeline
+		return err
+	}
+
+	// If the `ExpectedTimeline` field does not match the current timeline as reported by Patroni,
+	// then we assume a switchover has been completed, and we have reached this point because the
```
tjmoore4, line 544: :thinking: I'm not sure I'm following why it's safe to make this assumption (although I'm not saying it isn't!). I'm mainly curious if there could be a scenario where the current timeline doesn't match but the switchover has not completed.

tjmoore4 (commented):
This looks good overall, a couple of questions.

tjmoore4 (approved):
:+1: LGTM; new tests looks good.

internal/patroni/api.go
```diff
@@ -177,3 +184,39 @@ func (exec Executor) RestartPendingMembers(ctx context.Context, role, scope stri
...
+		"patronictl", "list", "-f", "json")
+	if err != nil {
+		return 0, err
+	}
+
+	if stderr.String() != "" {
+		return 0, errors.New(stderr.String())
+	}
+
+	members := []PatroniMember{}
```
cbandy, line 208: 📝 `PatroniMember` doesn't have to be defined as an exported type. It can be here for this one JSON payload.

```go
	var members []struct {
		Role     string
		State    string
		Timeline int32 `json:"TL"`
	}
	err = json.Unmarshal(stdout.Bytes(), &members)
```

internal/patroni/api.go
```diff
@@ -177,3 +184,39 @@ func (exec Executor) RestartPendingMembers(ctx context.Context, role, scope stri
...
+// Returns `0` as the int32 if it runs into errors or cannot find
+// a running Leader pod to get the up-to-date timeline from
+func (exec Executor) GetTimeline(ctx context.Context) (int32, error) {
+	var stdout, stderr bytes.Buffer
+
+	// The following exits zero when it is able to read the DCS and communicate
+	// with the Patroni HTTP API. It prints the result of calling "GET /cluster"
+	// - https://github.com/zalando/patroni/blob/v2.1.1/patroni/ctl.py#L849
+	err := exec(ctx, nil, &stdout, &stderr,
+		"patronictl", "list", "-f", "json")
```
cbandy, line 199: ⛏️ I prefer long arguments for clarity.

```suggestion
		"patronictl", "list", "--format", "json")
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/patroni_types.go
```diff
@@ -121,4 +121,8 @@ type PatroniStatus struct {
 	// Tracks the execution of the switchover requests.
 	// +optional
 	Switchover *string `json:"switchover,omitempty"`
+
+	// Tracks the current timeline during switchovers
+	// +optional
+	ExpectedTimeline *int32 `json:"expectedTimeline,omitempty"`
```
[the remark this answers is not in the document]
benjaminjb: TJ had a similar question about "expected" and in fact in an earlier draft I was using "expectedswitchover" or maybe the other way around, but I'm v open to changing the name.

testing/kuttl/e2e/delete/13-delete-cluster-and-check.yaml
benjaminjb: Oh, yeah, as far as I could tell by watching it -- and mind, this was before the fix was implemented but after these echoes were added -- we had a double-switchover, so step 12 asserts that the switchover happened, another switchover happened but didn't complete, so we had a primary and no replica, and the test passed.

Just playing around locally, but why does ```if [[ "$(date)" < "" ]]; then echo hello; fi``` not echo?

internal/patroni/api.go
```diff
@@ -177,3 +184,39 @@ func (exec Executor) RestartPendingMembers(ctx context.Context, role, scope stri
 
 	return err
 }
+
+// GetTimeline gets the patronictl status and returns the timeline,
+// currently the only information required by PGO.
+// Returns `0` as the int32 if it runs into errors or cannot find
+// a running Leader pod to get the up-to-date timeline from
+func (exec Executor) GetTimeline(ctx context.Context) (int32, error) {
```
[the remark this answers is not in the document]
benjaminjb: Test added

internal/controller/postgrescluster/patroni.go
```diff
@@ -499,6 +512,44 @@ func (r *Reconciler) reconcilePatroniSwitchover(ctx context.Context,
...
+	// cache does not yet have the updated `cluster.Status.Patroni.Switchover` field.
```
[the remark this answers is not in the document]
benjaminjb: This is the part that I am most leery about, and it's really around that "assume", but like you point out (and considering that timelines are increasing, like we talked about yesterday):

(a) a user annotates a cluster; 
(b) we get the timeline from patroni and add that to the cluster; 
(c) next time we check patroni, if it doesn't match what have on the cluster, e.g., what we we got from patroni at a previous time, then a timeline increase has happened between (b) and (c), which means a switchover _of some sort_ has occurred, yeah?

So we know a switchover occurred between the user asking for one and now; whether that timeline change was us, however, is not entirely clear in this situation. That's the part that I'm most worried about, but that's also the part that I think we have to assume for now because if we don't, then we're back at "single triggers causing multiple failovers."

internal/controller/postgrescluster/patroni.go
```diff
@@ -525,8 +576,12 @@ func (r *Reconciler) reconcilePatroniSwitchover(ctx context.Context,
 	if err = errors.WithStack(err); err == nil && !success {
 		err = errors.New("unable to switchover")
 	}
+
+	// If we've reached this point, a switchover has successfully been triggered
+	// and we set the status accordingly.
 	if err == nil {
 		cluster.Status.Patroni.Switchover = initialize.String(annotation)
+		cluster.Status.Patroni.ExpectedTimeline = nil
```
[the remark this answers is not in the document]
benjaminjb: Gonna hold off on addressing this until we get a chance to sync up on the assumptions of L 548

testing/kuttl/e2e/delete/13-delete-cluster-and-check.yaml
cbandy: The comparison uses collation which can be a bit mysterious, but I suspect it comes down to the lengths. Probably "blank" is less than "not-blank", so everything is greater than the empty string.

```sh-session
$ if [[ "aa" < "a" ]]; then echo hello; fi
$ if [[ "aa" < "aa" ]]; then echo hello; fi
$ if [[ "aa" < "aaa" ]]; then echo hello; fi
hello
```

internal/controller/postgrescluster/patroni_test.go
```diff
@@ -444,12 +444,20 @@ func TestReconcilePatroniSwitchover(t *testing.T) {
 	require.ParallelCapacity(t, 0)
 
 	var called, failover, callError, callFails bool
+	var timelineCallNoLeader, timelineCall bool
+	var nil32 *int64
```
cbandy, line 448: ⛏️ This variable name is odd now that the type is int64.

internal/controller/postgrescluster/patroni_test.go
```diff
@@ -516,41 +524,51 @@ func TestReconcilePatroniSwitchover(t *testing.T) {
...
+			check   func(*testing.T, error, *v1beta1.PostgresCluster)
 		}{
 			{
 				desc:    "Switchover not enabled",
 				enabled: false,
-				check: func(t *testing.T, err error) {
+				check: func(t *testing.T, err error, cluster *v1beta1.PostgresCluster) {
 					assert.NilError(t, err)
+					assert.Equal(t, cluster.Status.Patroni.SwitchoverTimeline, nil32)
+					assert.Equal(t, cluster.Status.Patroni.Switchover, nilString)
```
cbandy, line 535: 🤔 I don't think `assert.Equal` shows the values behind non-nil pointers. When I [try it](https://go.dev/play/p/N1rXlmK9Wfc) I see memory locations instead:

```
assertion failed: 0xc000188300 (*string) != <nil> (*string)
assertion failed: 0xc000188300 (*string) != 0xc000188310 (*string)
```

⛏️ `{address} != {address}` doesn't help me understand what happened or implement a fix. These nil checks can be simpler with the nil constant.

```go
assert.Assert(t, cluster.Status.Patroni.Switchover == nil)
// assertion failed: cluster.Status.Patroni.Switchover is not nil
```

📝 I forget what property(s) of Go allows comparison to nil like this. Nil is untyped. Pointers are comparable. Nil is assignable to any pointer.

internal/patroni/api.go
```diff
@@ -177,3 +178,42 @@ func (exec Executor) RestartPendingMembers(ctx context.Context, role, scope stri
 
 	return err
 }
+
+// GetTimeline gets the patronictl status and returns the timeline,
+// currently the only information required by PGO.
+// Returns `0` as the int32 if it runs into errors or cannot find
```
cbandy, line 184: ⛏️ 
```suggestion
// Returns zero if it runs into errors or cannot find
```

internal/controller/postgrescluster/patroni.go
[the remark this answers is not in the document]
cbandy: > then a timeline increase has happened between (b) and (c), which means a switchover _of some sort_ has occurred, yeah?

That's my thinking, right.

> So we know a switchover occurred between the user asking for one and now; whether that timeline change was us, however, is not entirely clear in this situation. That's the part that I'm most worried about, but that's also the part that I think we have to assume for now because if we don't, then we're back at "single triggers causing multiple failover."

The fields on the spec describe a desire to change the leader of the cluster. I don't think it matters what _caused_ the leader to change; it only matters that it changed. We're altering the criteria from "we called patronictl" to "we see a Postgres promotion."

That's where it gets fuzzy, I suppose. We still have "if called successfully, set done" and I'm not opposed to keeping that (especially in a bug fix). One of those criteria/transitions is the "main" one and the other is some approximation. I can't decide now which deserves what role. No changes are needed, IMO.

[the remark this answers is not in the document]
benjaminjb: I was thinking about adding something to the documentation about "always check your instance to make sure the post-fail/switchover state is what you want it to be" but that's already there with comments about checking the labels after a fail/switchover. 

(And I've tried to add code comments about the logic of the behavior, but is it enough?)

PR #3248, opened 2022-06-07
Add support for feature gates

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Adds a feature gate capability to PGO by leveraging the relevant
Kubernetes packages. This will allow users to enable or disable
certain features by setting the "PGO_FEATURE_GATES" environment
variable to a list similar to "feature1=true,feature2=false,..."
in the PGO Deployment.

**Other Information**:
Issue [sc-14488]

### DIFF
--- a/cmd/postgres-operator/main.go
+++ b/cmd/postgres-operator/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/crunchydata/postgres-operator/internal/controller/runtime"
 	"github.com/crunchydata/postgres-operator/internal/logging"
 	"github.com/crunchydata/postgres-operator/internal/upgradecheck"
+	"github.com/crunchydata/postgres-operator/internal/util"
 )
 
 var versionString string
@@ -51,6 +52,10 @@ func initLogging() {
 }
 
 func main() {
+	// Set any supplied feature gates; panic on any unrecognized feature gate
+	err := util.AddAndSetFeatureGates(os.Getenv("PGO_FEATURE_GATES"))
+	assertNoError(err)
+
 	otelFlush, err := initOpenTelemetry()
 	assertNoError(err)
 	defer otelFlush()
--- a/internal/util/README.md
+++ b/internal/util/README.md
@@ -1,0 +1,109 @@
+<!--
+ Copyright 2021 - 2022 Crunchy Data Solutions, Inc.
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+
+## Feature Gates
+
+Feature gates allow users to enable or disable
+certain features by setting the "PGO_FEATURE_GATES" environment
+variable to a list similar to "feature1=true,feature2=false,..."
+in the PGO Deployment.
+
+This capability leverages the relevant Kubernetes packages. Documentation and
+code implementation examples are given below.
+
+- Documentation:
+  - https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
+
+- Package Information:
+  - https://pkg.go.dev/k8s.io/component-base@v0.20.1/featuregate
+
+- Adding the feature gate key:
+  - https://releases.k8s.io/v1.20.0/pkg/features/kube_features.go#L27
+
+- Adding the feature gate to the known features map:
+  - https://releases.k8s.io/v1.20.0/pkg/features/kube_features.go#L729-732
+
+- Adding features to the featureGate
+  - https://releases.k8s.io/v1.20.0/staging/src/k8s.io/component-base/featuregate/feature_gate.go#L110-L111
+
+- Setting the feature gates
+  - https://releases.k8s.io/v1.20.0/staging/src/k8s.io/component-base/featuregate/feature_gate.go#L105-L107
+
+## Developing with Feature Gates in PGO
+
+To add a new feature gate, a few steps are required. First, in
+`internal/util/features.go`, you will add a feature gate key name. As an example,
+for a new feature called 'FeatureName', you would add a new constant and comment
+describing what the feature gate controls at the top of the file, similar to
+```
+// Enables FeatureName in PGO
+FeatureName featuregate.Feature = "FeatureName"
+```
+
+Next, add a new entry to the `pgoFeatures` map
+```
+var pgoFeatures = map[featuregate.Feature]featuregate.FeatureSpec{
+    FeatureName: {Default: false, PreRelease: featuregate.Alpha},
+}
+```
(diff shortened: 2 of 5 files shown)


cmd/postgres-operator/main.go
```diff
@@ -51,6 +52,11 @@ func initLogging() {
 }
 
 func main() {
+	// Set any supplied feature gates; panic on any unrecognized feature gate
```
tjmoore4, line 55: :memo: I put this first in `main` because misconfigurations will result in an abort, so I think it makes sense to fail quickly. That said, is there any reason this should go further down?

internal/util/features.go
```diff
@@ -0,0 +1,70 @@
+/*
```
tjmoore4, line 1: :memo: I felt that `util` was a reasonable location for this functionality in our current architecture. I also considered creating a separate `feature` package. I ultimately settled on using `util` because it seemed like overkill to move this to a separate package, but I'm certainly up for thoughts/recommendations.

internal/util/features.go
```diff
@@ -0,0 +1,70 @@
...
+// // beta: v1.X
+// // Enables FeatureName
+// FeatureName featuregate.Feature = "FeatureName"
+//
+// - https://releases.k8s.io/v1.20.0/pkg/features/kube_features.go#L27
+//
+// Feature gates should be listed in alphabetical, case-sensitive
+// (upper before any lower case character) order.
+//
+// TODO(tjmoore4): Features will start here
```
tjmoore4, line 37: :memo: This `TODO` and the one below will go away once the first feature-gated feature is committed.

internal/util/features.go
```diff
@@ -0,0 +1,70 @@
...
+// - https://pkg.go.dev/k8s.io/component-base@v0.20.1/featuregate
+func AddAndSetFeatureGates(features string) error {
+	// Add PGO features to the featureGate
+	// - https://releases.k8s.io/v1.20.0/staging/src/k8s.io/component-base/featuregate/feature_gate.go#L110-L111
+	if err := feature.DefaultMutableFeatureGate.Add(pgoFeatures); err != nil {
+		return fmt.Errorf("unable to add PGO features to the featureGate. %w", err)
+	}
+
+	// Set the feature gates from environment variable config
+	// - https://releases.k8s.io/v1.20.0/staging/src/k8s.io/component-base/featuregate/feature_gate.go#L105-L107
```
benjaminjb, line 65: I'm not sure if this requires more documentation -- classic example of "I had trouble, but maybe that's just me", as we've talked about -- but I had to review all the links in this page to make sense of what's going on / what a feature gate will look like.

cmd/postgres-operator/main.go
```diff
@@ -51,6 +52,11 @@ func initLogging() {
...
+	testValue := os.Getenv("PGO_FEATURE_GATES")
+	err := util.AddAndSetFeatureGates(testValue)
```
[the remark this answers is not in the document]
tjmoore4: ah, no! I actually don't think we need that variable at all, I'll probably just remove it.

internal/util/features_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	t.Run("One feature gate set", func(t *testing.T) {
+		err := AddAndSetFeatureGates("TestGate1=true")
+		assert.NilError(t, err)
+	})
+
+	t.Run("Two feature gates set", func(t *testing.T) {
+		err := AddAndSetFeatureGates("TestGate1=true,TestGate3=true")
+		assert.NilError(t, err)
+	})
```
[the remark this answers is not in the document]
tjmoore4: I looked into it, but I don't think there's a good way to test deeper into the functionality. As you mentioned, the package tests are already pretty verbose (e.g. [Set's test](https://github.com/kubernetes/kubernetes/blob/v1.20.0/staging/src/k8s.io/component-base/featuregate/feature_gate_test.go#L232)).

internal/util/features.go
tjmoore4: It may still be worth adding to what's in the comments, but it's worth noting I'm working the initial follow-on stories as well, so I'll be adding real examples in the near feature. I am definitely happy to demo as well.

internal/util/features.go
tjmoore4: Understood. As I mentioned above, I'll have some examples in follow on work and the plan is to add additional documentation with those. It still may be worth expanding on the documentation here though. Might be worth having the demo first, then we can decide what makes sense.

internal/util/features.go
cbandy: 🌱 There's a PR upstream to rearrange their similar file: https://pr.k8s.io/109865

internal/util/features.go
```diff
@@ -0,0 +1,70 @@
...
+// add the default PGO features to the featureGate and then set the values provided
+// via the 'PGO_FEATURE_GATES' environment variable. This function expects a string
+// like feature1=true,feature2=false,...
+//
+// - https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
+// - https://pkg.go.dev/k8s.io/component-base@v0.20.1/featuregate
+func AddAndSetFeatureGates(features string) error {
+	// Add PGO features to the featureGate
+	// - https://releases.k8s.io/v1.20.0/staging/src/k8s.io/component-base/featuregate/feature_gate.go#L110-L111
+	if err := feature.DefaultMutableFeatureGate.Add(pgoFeatures); err != nil {
```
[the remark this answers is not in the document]
tjmoore4: :+1: added one new variable, `DefaultMutableFeatureGate` that removes the need for the `k8s.io/apiserver` module. At this point, I don't believe we need a `featuregate.FeatureGate` instance, but please let me know if I misunderstood. In any case, I think we could potentially add one later.

internal/util/README.md
```diff
@@ -0,0 +1,100 @@
...
+```
+Note that for more than one feature, this variable accepts a comma delimited
+list, e.g.
+```
+PGO_FEATURE_GATES="FeatureName=true,FeatureName2=true,FeatureName3=true"
+```
+
+While `PGO_FEATURE_GATES` does not have to be set, please note that the features
+ must be defined before use, otherwise PGO deployment will fail with the
+ following message
```
benjaminjb, line 96: extra spaces at the beginning of the lines

benjaminjb (approved):
makes sense to me, and if it's still working with the upcoming PR on the sidecar gate 👍

internal/util/README.md
```diff
@@ -0,0 +1,100 @@
...
+where `FeatureName` is the constant defined previously, `Default: false` sets the
+default behavior and `PreRelease: featuregate.Alpha`. The possible `PreRelease` 
+values are `Alpha`, `Beta`, `GA` and `Deprecated`.
+
+- https://pkg.go.dev/k8s.io/component-base@v0.20.1/featuregate#pkg-constants
+
+Features should generally be disabled by default. By Kubernetes convention, `Alpha`
+features have almost always been disabled by default. `Beta` features were
+generally enabled by default prior to Kubernetes 1.24, and are now generally
+disabled. For consistency with the broadest range of our supported Kubernetes
```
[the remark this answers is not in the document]
tjmoore4: :+1: Thanks for the clarification, I'll update the README to better reflect what you mentioned and add the command line reference link and a note about to keep an eye on any changes going forward.

cbandy (approved):

PR #3258, opened 2022-06-13
Change use_pg_rewind for PG10

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

`use_pg_rewind` is set to true for all PG versions

* don't use use_pg_rewind for pg10
* update KUTTL test to reinit pg10 for PITR

**Other Information**:
Issue [sc-12408]


testing/kuttl/e2e/pgbackrest-restore/15--in-place-pitr.yaml
```diff
@@ -20,3 +20,13 @@ commands:
       # Annotate the cluster to trigger the restore.
       kubectl annotate --namespace="${NAMESPACE}" postgrescluster/original \
         'postgres-operator.crunchydata.com/pgbackrest-restore=one'
+      
+      # This has worked locally, but should we wait for the pod to get stuck at 3/4 containers ready?
```
cbandy, line 24: 🤔 I think the check against `KUTTL_PG_VERSION` is enough to tell us that we _should_ do the reinit. _When_ to try the `kubectl exec` ... hmm.

I think technically we need only the `database` container to be "running". That's kind of awkward to get from the status: the container status state `running` field is populated. 😖 Something like this JSON path?

```
{.status.containerStatuses[?(@.name=="database")].state.running}
```

testing/kuttl/e2e/pgbackrest-restore/15--in-place-pitr.yaml
```diff
@@ -20,3 +20,13 @@ commands:
...
+      # For PG10, we need to run a patronictl reinit for the replica
+      if [[ ${KUTTL_PG_VERSION} == 10 ]]; then
+        REPLICA=$(kubectl get pods --namespace "${NAMESPACE}" \
+          --selector=postgres-operator.crunchydata.com/cluster=original,postgres-operator.crunchydata.com/data=postgres,postgres-operator.crunchydata.com/role!=master \
+          --output=jsonpath={.items..metadata.name})
+        
+        yes | kubectl exec --namespace "${NAMESPACE}" "${REPLICA}" -- patronictl reinit original-ha "${REPLICA}" --force
```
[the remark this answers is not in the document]
benjaminjb: It does not, removing now.

testing/kuttl/e2e/pgbackrest-restore/15--in-place-pitr.yaml
```diff
@@ -20,3 +20,13 @@ commands:
       # Annotate the cluster to trigger the restore.
       kubectl annotate --namespace="${NAMESPACE}" postgrescluster/original \
         'postgres-operator.crunchydata.com/pgbackrest-restore=one'
+      
+      # This has worked locally, but should we wait for the pod to get stuck at 3/4 containers ready?
+      # For PG10, we need to run a patronictl reinit for the replica
+      if [[ ${KUTTL_PG_VERSION} == 10 ]]; then
+        REPLICA=$(kubectl get pods --namespace "${NAMESPACE}" \
+          --selector=postgres-operator.crunchydata.com/cluster=original,postgres-operator.crunchydata.com/data=postgres,postgres-operator.crunchydata.com/role!=master \
```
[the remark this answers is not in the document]
benjaminjb: That looks better to me, changing now

internal/patroni/config.go
```diff
@@ -263,7 +263,7 @@ func DynamicConfiguration(
 	postgresql["pg_hba"] = hba
 
 	// TODO(cbandy): explain this.
```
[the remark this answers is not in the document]
benjaminjb: Added

internal/patroni/config_test.go
```diff
@@ -120,7 +120,7 @@ func TestDynamicConfiguration(t *testing.T) {
 				"postgresql": map[string]interface{}{
 					"parameters":    map[string]interface{}{},
 					"pg_hba":        []string{},
-					"use_pg_rewind": true,
+					"use_pg_rewind": false,
```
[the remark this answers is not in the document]
benjaminjb: I agree -- I added a test case specifically for pg10 (there are two tests, I added that case to both, easy to remove them when we drop pg10), and some logic in the test for setting PG version to 14 if not set.

testing/kuttl/e2e/pgbackrest-restore/15--in-place-pitr.yaml
benjaminjb: I was testing this in GKE and found the pod took longer to terminate and restart than otherwise, so I added a bunch of waiting/checking, and I don't like it so much -- feels brittle. Are there any other ways that I can see that a postgrescluster restore is in progress and a pod exists?

cbandy: `patronictl reinit` takes the member name as an argument, which means it can see/find that member. Can we run it from the primary? If so, we can wait for that to be ready and exec there?

```
kubectl wait pod … --for condition=Ready --timeout=90s
```

I believe the restore is finished when PostgresCluser has the `PostgresDataInitialized` condition.

benjaminjb: Just ran some experiments to double-check, and here's my findings:

The first problem that I ran into was that the pre-PITR pods can still exist after the annotation triggers the PITR. So if we annotate and then immediately wait for the pod to be ready, we're likely to hit the pre-restore pod, which is not the pod we want to reinit.

I've cleaned this up with a few waits but it doesn't work: 
* first, we wait for the postgrescluster to have `status.pgbackrest.restore.finished`=false -- but the `restore` field doesn't exist, so it just errors out on that. (See https://github.com/kubernetes/kubectl/issues/1204 )
* second, we wait for the pod container `database` to be running, but we can't wait for an object that doesn't exist yet (https://github.com/kubernetes/kubernetes/issues/83242), so I'm back to an explicit loop.

benjaminjb: Yeah, kept running into some problems when trying to wait for conditions, and this is meant to be a temporary solution for a temporary situation (ha!), so I'll stick with some of the original way that worked and leave this note for later if we want to clean this up.

cbandy (approved):

PR #3259, opened 2022-06-14
Use ReadWriteOnce through documentation

cbandy:
It is the most commonly supported access mode, and using different modes in different places can be confusing.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Documentation

**Other Information**:

Issue: [sc-14874]

### DIFF
--- a/docs/content/tutorial/customize-cluster.md
+++ b/docs/content/tutorial/customize-cluster.md
@@ -189,7 +189,7 @@ spec:
     - name: instance
       walVolumeClaimSpec:
         accessModes:
-        - "ReadWriteMany"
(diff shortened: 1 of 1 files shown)


tjmoore4 (approved):
:+1: That makes sense to me.

:thinking: It's interesting that this is the only place we use `ReadWriteMany`. It seems it just happens to be set that way. I see nothing indicating it was set that way for any particular reason.

:seedling: Outside of scope for this PR, but perhaps it would be worth adding a short paragraph and a link to the relevant Kubernetes documentation regarding access modes in general?

benjaminjb (approved):

PR #3261, opened 2022-06-15
Allow Custom Sidecars for pgBouncer Pods

### LINKED ISSUE #2618: adding additional container to pgbouncer v5
## Overview
As of now, there is no option in the postgrescluster crd to add sidecars for metrics inside the pgbouncer pods. e.g. https://github.com/prometheus-community/pgbouncer_exporter
If would be nice to have the option to attach sidecars including stuff like env/valueFrom or resources.
While doing so, it would make sense to extend the container capabilities of postgres instances, where currently the only option is to add a monitoring image.

## Use Case
Monitoring pgbouncer alongside postgres instances is important to fine tune settings and identifying issues.

## Desired Behavior
A standard way of specifying additional container to pgbouncer

## Environment

Tell us about your environment:

Please provide the following details:

- Platform: `Kubernetes`, `GKE`
- Platform Version: `1.19.x`
- PGO Image Tag: `ubi8-5.0.1-0`
- Postgres Version `13`
- Number of Postgres clusters:  soon >50

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

This update allows you to configure custom sidecar Containers for
your pgBouncer Pods. To use this feature, currently in `Alpha`,
you will need to enable it via the relevant PGO feature gate.
This is done by setting the `PGO_FEATURE_GATES` environment
variable on the PGO Deployment to

'PGO_FEATURE_GATES="PGBouncerSidecars=true'

**Other Information**:
Issue: [sc-14727]
This commit fixes #2618

### DIFF
--- a/config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
+++ b/config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
@@ -5473,8 +5473,8 @@ spec:
                           type: object
                       type: object
                     containers:
-                      description: Defines custom sidecars for the PostgreSQL instance
-                        pods
+                      description: Custom sidecars for PostgreSQL instance pods. Changing
+                        this value causes PostgreSQL to restart.
                       items:
                         description: A single application container that you want
                           to run within a pod.
--- a/docs/content/references/crd.md
+++ b/docs/content/references/crd.md
@@ -3718,7 +3718,7 @@ Resource requirements for a sidecar container
       </tr><tr>
         <td><b><a href="#postgresclusterspecinstancesindexcontainersindex">containers</a></b></td>
         <td>[]object</td>
-        <td>Defines custom sidecars for the PostgreSQL instance pods</td>
+        <td>Custom sidecars for PostgreSQL instance pods. Changing this value causes PostgreSQL to restart.</td>
         <td>false</td>
       </tr><tr>
         <td><b><a href="#postgresclusterspecinstancesindexmetadata">metadata</a></b></td>
--- a/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
+++ b/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
@@ -428,7 +428,8 @@ type PostgresInstanceSetSpec struct {
 	// +optional
 	Affinity *corev1.Affinity `json:"affinity,omitempty"`
 
-	// Defines custom sidecars for the PostgreSQL instance pods
+	// Custom sidecars for PostgreSQL instance pods. Changing this value causes
+	// PostgreSQL to restart.
 	// +optional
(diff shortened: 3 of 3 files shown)


pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbouncer_types.go
```diff
@@ -73,6 +73,10 @@ type PGBouncerPodSpec struct {
 	// +optional
 	Config PGBouncerConfiguration `json:"config,omitempty"`
 
+	// Defines custom sidecars for the pgBouncer pods
```
cbandy, line 76: 🔧 Capitalize `PgBouncer`.
⛏️ Other fields in this section are phrased differently.

🤔 Changes here become changes to the pod spec which causes a rollout, right? If so, we should mention PgBouncer restart.

```suggestion
	// Custom sidecars for a PgBouncer pod.
	// Changes ... cause PgBouncer to restart.
	// Changing this ... causes PgBouncer to restart.
```

There's a variety of this phrase. Singular, plural, noun, gerund... 🤷🏻

build/crd/condition.yaml
```diff
@@ -14,3 +14,11 @@
     Localhost - a profile defined in a file on the node should be used.
     RuntimeDefault - the container runtime default profile should be used.
     Unconfined - no profile should be applied.
+- op: add
+  path: "/spec/versions/0/schema/openAPIV3Schema/properties/spec/properties/proxy/properties/pgBouncer/properties\
+    /containers/items/properties/securityContext/properties/seccompProfile/properties/type/description"
```
cbandy, line 19: ⛏️ This operation and the one above it seem a bit out of place in a file named `condition.yaml`.

cbandy (approved):

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbouncer_types.go
tjmoore4: :+1: Updated as recommended, and I picked what seemed to be the most common option for the rollout warning.

build/crd/condition.yaml
tjmoore4: I don't think the other files we use for similar purpose are great fits either. The only other options that come to mind are to either rename this file or add a new one, but I'm not sure I'm thrilled with those either. 
I think for this commit, I'd like to leave these here, but perhaps we can discuss a refactor later.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbouncer_types.go
tjmoore4: :thinking: This same behavior is true of the Postgres instance sidecars as well. I'm going to update that comment as well (as a separate commit).

PR #3279, opened 2022-06-27
Fix 'GCS' Typo in Azure Storage Blob section

brandonavant:
The Azure Storage Blob section contains the following sentence:

"Similar to the above, setting up backups in Azure Blob Storage requires a few additional modifications to your custom resource spec and the use of a Secret to protect your GCS credentials."

it should read:

"Similar to the above, setting up backups in Azure Blob Storage requires a few additional modifications to your custom resource spec and the use of a Secret to protect your Azure Storage credentials."

**Checklist:**

 - [ x] Have you added an explanation of what your changes do and why you'd like them to be included?

**Type of Changes:**

 - [ x] Documentation

### DIFF
--- a/docs/content/tutorial/backups.md
+++ b/docs/content/tutorial/backups.md
@@ -220,7 +220,7 @@ Watch your cluster: you will see that your backups and archives are now being st
 
 ## Using Azure Blob Storage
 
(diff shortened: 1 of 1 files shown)


andrewlecuyer (approved):

andrewlecuyer:
Thank you @brandonavant for your contribution!

PR #3282, opened 2022-06-29
minor typo in doc

andyli:
I guess `Any example recovery target...` should be `An example recovery target...`.


andrewlecuyer (approved):

andrewlecuyer:
Thank you @andyli!

PR #3297, opened 2022-07-11
Allow NodePort Port to be Specified via the PostgresCluster Spec

### LINKED ISSUE #3008: Statically assign NodePort service ports
## Overview

I would like to be able to define the port value for any `NodePort` service statically. In the case where we integrate with an external service (such as a hardware load balancer) that sends traffic to the services running in a Kube cluster, if we have to rebuild the `postgrescluster` from scratch, or recreate the service, it assigns a new `NodePort`, requiring either patching of the service or reconfiguration of the integrating external service.

## Use Case

We run two `postgresclusters` in different Kubernetes cluster in different datacenters, to serve a database to a single service that also runs in both in an HA manner. We pipe traffic to the current active cluster by way of a hardware load balancer, which talks to the clusters via the `pgbouncer` node port service. If we have to rebuild the clusters, the service gets created fresh with a different NodePort. Right now we are patching it post deployment, but being able to statically set this NodePort value would eliminate the need for this step.

## Desired Behavior

When I define that a service should be of type `NodePort`, I should also be able to statically set the value I want to use for said NodePort.

## Environment

- Platform: `Kubernetes`
- Platform Version: `1.18.18`
- PGO Image Tag: `ubi8-5.0.4-0`
- Postgres Version `13`
- Storage: `s3`
- Number of Postgres clusters: `2`

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Cannot statically assign NodePort service ports.

This update allows a specific NodePort port to be specified for
the primary Postgres, pgBouncer and pgAdmin services via the
PostgresCluster spec.

Note this is used when type is NodePort or LoadBalancer only.
Setting this value when using the 'ClusterIP' type will result
in an error. The specified value must be also be in-range and
not currently in use or the operation will fail. If unspecified,
a port will be allocated if this Service requires one as before.

**Other Information**:
Resolves #3008
Issue: [sc-14918]


internal/controller/postgrescluster/patroni_test.go
```diff
@@ -146,10 +160,47 @@ ownerReferences:
 
 			service, err := reconciler.generatePatroniLeaderLeaseService(cluster)
 			assert.NilError(t, err)
-			alwaysExpect(t, service)
+			alwaysExpect(t, service, false)
```
cbandy, line 163: ⛏️ Same here; no argument, assert the ports.

internal/controller/postgrescluster/patroni_test.go
```diff
@@ -146,10 +160,47 @@ ownerReferences:
...
+
+	typesAndPort := []struct {
+		Description string
+		Type        string
+		NodePort    *int32
+		Expect      func(testing.TB, *corev1.Service)
+	}{
+		{Description: "ClusterIP with Port 32000", Type: "ClusterIP",
+			NodePort: initialize.Int32(32000), Expect: func(t testing.TB, service *corev1.Service) {
+				assert.Equal(t, service.Spec.Type, corev1.ServiceTypeClusterIP)
```
cbandy, line 176: 🔧 This is never called.

🤔 Perhaps the expectation should be `Expect      func(testing.TB, *corev1.Service, error)` then these cases show when there's an error or not.

```suggestion
				assert.Assert(t, service == nil)
				assert.ErrorContains(t, err, "NodePort cannot be set with type ClusterIP on pg2-ha")
```

The other cases can do `alwaysExpect(t, service)` and assert the ports here, too.

internal/controller/postgrescluster/pgbouncer_test.go
```diff
@@ -40,7 +41,13 @@ func TestGeneratePGBouncerService(t *testing.T) {
 	_, cc := setupKubernetes(t)
 	require.ParallelCapacity(t, 0)
 
-	reconciler := &Reconciler{Client: cc}
+	// reconciler := &Reconciler{Client: cc}
```
cbandy, line 44: ⛏️ leftover (and a couple lines below)

internal/controller/postgrescluster/patroni.go
```diff
@@ -246,21 +247,37 @@ func (r *Reconciler) generatePatroniLeaderLeaseService(
...
+			if service.Spec.Type == corev1.ServiceTypeClusterIP {
+				// The NodePort can only be set when the Service type is NodePort or
+				// LoadBalancer. However, due to a known issue prior to Kubernetes
+				// 1.20, we clear these errors during our apply. To preserve the
+				// appropriate behavior, we log an Event and return a customer Error.
+				// TODO(tjmoore4): Once Validation Rules are available, this check
+				// and event could potentially be removed in favor of that validation
+				r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "MisconfiguredClusterIP",
+					"NodePort cannot be set with type ClusterIP on %s", service.Name)
+				return nil, fmt.Errorf("NodePort cannot be set with type ClusterIP on %s", service.Name)
```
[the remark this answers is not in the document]
tjmoore4: I think that's a good idea, I've updated that language for all three services. I do think our style is likely a bit inconsistent, but more information is (usually) better.

internal/controller/postgrescluster/patroni.go
```diff
@@ -246,21 +247,37 @@ func (r *Reconciler) generatePatroniLeaderLeaseService(
...
 		Name:       naming.PortPostgreSQL,
 		Port:       *cluster.Spec.Port,
 		Protocol:   corev1.ProtocolTCP,
 		TargetPort: intstr.FromString(naming.PortPostgreSQL),
-	}}
+	}
+
+	if spec := cluster.Spec.Service; spec != nil {
+		service.Spec.Type = corev1.ServiceType(spec.Type)
+		if spec.NodePort != nil {
```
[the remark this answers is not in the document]
tjmoore4: wfm :+1:

internal/controller/postgrescluster/patroni_test.go
```diff
@@ -91,7 +105,7 @@ ownerReferences:
 	t.Run("NoServiceSpec", func(t *testing.T) {
 		service, err := reconciler.generatePatroniLeaderLeaseService(cluster)
 		assert.NilError(t, err)
-		alwaysExpect(t, service)
+		alwaysExpect(t, service, false)
```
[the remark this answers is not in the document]
tjmoore4: That works for me. I went back and forth on the `if` and how that impacted readability. If this seems clearer, happy to make the change.

internal/controller/postgrescluster/patroni.go
```diff
@@ -246,21 +247,37 @@ func (r *Reconciler) generatePatroniLeaderLeaseService(
...
+	} else {
+		service.Spec.Type = corev1.ServiceType(spec.Type)
+		if spec.NodePort != nil {
+			if service.Spec.Type == corev1.ServiceTypeClusterIP {
+				// The NodePort can only be set when the Service type is NodePort or
+				// LoadBalancer. However, due to a known issue prior to Kubernetes
+				// 1.20, we clear these errors during our apply. To preserve the
+				// appropriate behavior, we log an Event and return an error.
+				// TODO(tjmoore4): Once Validation Rules are available, this check
+				// and event could potentially be removed in favor of that validation
```
jmckulk, line 272: nice comment :+1:

jmckulk (approved):

cbandy (approved):

PR #3327, opened 2022-07-29
Set runAsNonRoot at the container-level only

cbandy:
We satisfy Kubernetes' Restricted Pod Security policy by setting `runAsNonRoot` for all our containers, so setting it on the pod is redundant.

A new step in `kuttl/e2e/security-context` verifies things using the [Require runAsNonRoot](https://kyverno.io/policies/pod-security/restricted/require-run-as-nonroot/require-run-as-nonroot/) Kyverno policy when the [Kyverno CLI](https://kyverno.io/docs/kyverno-cli/) is available.

**Checklist:**

 
**Type of Changes:**

 

 - [x] Bug fix
 - [x] Testing enhancement

Sidecars that need to run as root cannot be injected into any of our Pods.

Sidecars can be added or injected with any privileges or security context settings.

**Other Information**:

- https://groups.google.com/a/crunchydata.com/g/postgres-operator/c/8Y3viamQ80w/m/eYbTKBb_CQAJ

Issue: [sc-15204]

### DIFF
--- a/internal/initialize/security.go
+++ b/internal/initialize/security.go
@@ -19,14 +19,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// RestrictedPodSecurityContext returns a v1.PodSecurityContext with safe defaults.
-// See https://docs.k8s.io/concepts/security/pod-security-standards/
-func RestrictedPodSecurityContext() *corev1.PodSecurityContext {
+// PodSecurityContext returns a v1.PodSecurityContext with some defaults.
+func PodSecurityContext() *corev1.PodSecurityContext {
 	onRootMismatch := corev1.FSGroupChangeOnRootMismatch
 	return &corev1.PodSecurityContext{
-		// Fail to start a container if its image runs as UID 0 (root).
-		RunAsNonRoot: Bool(true),
-
 		// If set to "OnRootMismatch", if the root of the volume already has
 		// the correct permissions, the recursive permission change can be skipped
 		FSGroupChangePolicy: &onRootMismatch,
--- a/internal/initialize/security_test.go
+++ b/internal/initialize/security_test.go
@@ -24,8 +24,12 @@ import (
 	"github.com/crunchydata/postgres-operator/internal/initialize"
 )
 
-func TestRestrictedPodSecurityContext(t *testing.T) {
-	psc := initialize.RestrictedPodSecurityContext()
+func TestPodSecurityContext(t *testing.T) {
+	psc := initialize.PodSecurityContext()
+
+	if assert.Check(t, psc.FSGroupChangePolicy != nil) {
+		assert.Equal(t, string(*psc.FSGroupChangePolicy), "OnRootMismatch")
+	}
 
 	// Kubernetes describes recommended security profiles:
 	// - https://docs.k8s.io/concepts/security/pod-security-standards/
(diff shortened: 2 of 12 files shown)


internal/initialize/security.go
```diff
@@ -19,14 +19,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// RestrictedPodSecurityContext returns a v1.PodSecurityContext with safe defaults.
-// See https://docs.k8s.io/concepts/security/pod-security-standards/
-func RestrictedPodSecurityContext() *corev1.PodSecurityContext {
+// PodSecurityContext returns a v1.PodSecurityContext with some defaults.
+func PodSecurityContext() *corev1.PodSecurityContext {
```
benjaminjb, line 23: Hmmm, yeah, I suppose this SecurityContext is no longer restricting anything... but we still need this separate from `postgres.PodSecurityContext` because of the extra things we do with that PodSecurityContext. (This comment is just me thinking out loud.)

internal/initialize/security_test.go
```diff
@@ -47,9 +51,9 @@ func TestRestrictedPodSecurityContext(t *testing.T) {
 	// > operators and developers of security-critical applications, as well as
 	// > lower-trust users.
 	t.Run("Restricted", func(t *testing.T) {
-		if assert.Check(t, psc.RunAsNonRoot != nil) {
-			assert.Assert(t, *psc.RunAsNonRoot == true,
-				"Containers must be required to run as non-root users.")
+		if assert.Check(t, psc.RunAsNonRoot == nil) {
```
benjaminjb, line 54: But if, through some regression, `psc.RunAsNonRoot != nil`, would we catch it?

testing/policies/kyverno/kustomization.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+bases:
+  - https://github.com/kyverno/policies/pod-security/restricted
```
benjaminjb, line 5: ❓ This seems to assert that all the containers are restricted to run as nonRoot OR the pod is restricted -- do I have that right?

benjaminjb: Also I think I need to brush up on Kyverno because I'm a little confused about this log:

```
    logger.go:42: 11:53:43 | security-context/10-10--kyverno | I0731 11:53:43.469837    7232 evaluate.go:63] EngineValidate "msg"="a condition failed in 'all' block" "kind"="StatefulSet" "name"="security-context-pgadmin" "namespace"="kuttl-test-select-rooster" "policy"="restrict-volume-types" "rule"="autogen-restricted-volumes" "condition"={"key":["name","emptyDir","name","persistentVolumeClaim","projected","name","emptyDir","name","emptyDir","name"],"operator":"AnyNotIn","value":["name","configMap","csi","downwardAPI","emptyDir","ephemeral","persistentVolumeClaim","projected","secret",""]}
    logger.go:42: 11:53:43 | security-context/10-10--kyverno | pass: 207, fail: 0, warn: 0, error: 0, skip: 830
```

A condition failed, but `fail: 0`... Is that what they're talking about when they log the note that I should log at level 5?

benjaminjb (approved):
No blockers, just some questions, mostly around testing.

internal/initialize/security_test.go
cbandy: Yes. [assert.Check](https://pkg.go.dev/gotest.tools/v3/assert#hdr-Assert_and_Check) fails the test in that case.

testing/policies/kyverno/kustomization.yaml
cbandy: > ❓ This seems to assert that all the containers are restricted to run as nonRoot OR the pod is restricted -- do I have that right?

Yep, that's right. [This policy](https://kyverno.io/policies/pod-security/restricted/require-run-as-nonroot/require-run-as-nonroot/) is also how they explain when/why to use [anyPattern](https://kyverno.io/docs/writing-policies/validate/#anypattern).

> I'm a little confused about this log

I noticed that, too, but haven't opened an issue. I looked some time ago, and my memory is that they added some logging deep somewhere that doesn't know when the expression is negated or not. 🤔 But I see `AnyNotIn` right there...

"Condition failed" is definitely confusing next to the overall "pass / fail" message.

cbandy: I mentioned it on Kyverno \#4074, and looks like they have an idea for a fix.

PR #3362, opened 2022-08-23
Update runtime-controller

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

runtime-controller is old

* update runtime-controller
* adjust logger
* adjust envtest
* adjust tests

**Other Information**:
Issue [sc-12818]


internal/controller/postgrescluster/controller.go
```diff
@@ -204,11 +204,7 @@ func (r *Reconciler) Reconcile(
 		})
 		return patchClusterStatus()
 	} else {
-		// Avoid a panic! Fixed in Kubernetes v1.21.0 and controller-runtime v0.9.0-alpha.0.
```
benjaminjb, line 207: If we're supporting K8s 1.20, then maybe we should put these guards back

tjmoore4 (commented):
I think this looks good overall, but there are a few things I want to make sure I understand before approving.

internal/controller/postgrescluster/pgbackrest_test.go
```diff
@@ -344,6 +344,7 @@ restartPolicy: Always
 schedulerName: default-scheduler
 securityContext:
   fsGroup: 26
+  fsGroupChangePolicy: OnRootMismatch
```
benjaminjb, line 347: Double-check this -- why here now?

config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
```diff
@@ -613,11 +618,82 @@ spec:
                                                     are ANDed.
                                                   type: object
                                               type: object
+                                            namespaceSelector:
```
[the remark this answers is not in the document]
benjaminjb: I didn't find any issues, in the sense that all the tests passed, both automated and manual. 

Regarding these changes, there were a lot between 1.20 and 1.24; `namespaceSelector` for instance was added to affinity in k8s 1.21+, and we've got several of for different pod/job fields in our spec. (`minDomain` got added to `TopologySpreadConstraint`, etc.)

internal/kubeapi/patch.go
```diff
@@ -117,7 +117,7 @@ func (patch *JSON6902) Replace(path ...string) func(value interface{}) *JSON6902
 func (patch JSON6902) Bytes() ([]byte, error) { return patch.Data(nil) }
 
 // Data returns the JSON representation of patch.
-func (patch JSON6902) Data(runtime.Object) ([]byte, error) { return json.Marshal(patch) }
+func (patch JSON6902) Data(client.Object) ([]byte, error) { return json.Marshal(patch) }
```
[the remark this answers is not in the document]
benjaminjb: We discussed this in a meeting, but in short, `r.patch` takes a `client.Patch` param, and that `client.Patch` interface has a `Data` method that takes a `client.Object` now rather than a `runtime.Object` -- though `client.Object` actually includes `runtime.Object`; ergo, changing `runtime.Object` => `client.Object` doesn't really require many other changes from us and is required for our JSON6902 type to satisfy the `client.Patch` interface

https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Patch
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Object
https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#Object

internal/controller/postgrescluster/controller.go
benjaminjb: OK, I've run check-envtest-existing and check-kultt pointed at 4.6 and had no problems related to this issue. (There were other problems, but I didn't see any panics.)

tjmoore4: That's good :+1:. Might be worth discussing, but if it passes in OCP 4.6, we should be in good shape since 'vanilla kube' should be 1.21+ at this point (per the [docs](https://access.crunchydata.com/documentation/postgres-operator/v5/)).

internal/controller/postgrescluster/pgbackrest_test.go
tjmoore4: :thinking: That's a good question, the change to `OnRootMismatch` should have required this update previously, iiuc. Might be worth discussing with the team.

cbandy: L187 skips when `strings.EqualFold(os.Getenv("USE_EXISTING_CLUSTER"), "true")`, so this has only been running in local envtest. CI for that has been pinned to a specific version for a while.

I think this is a candidate for splitting into "generate" methods.

internal/controller/postgrescluster/apply_test.go
```diff
@@ -147,6 +147,9 @@ func TestServerSideApply(t *testing.T) {
 	})
 
 	t.Run("StatefulSetPodTemplate", func(t *testing.T) {
+		if serverVersion.LessThan(version.MustParseGeneric("1.22.01")) {
+			t.Skip("Statefulset.Status added field in 1.22.01, cannot test `patch` in earlier versions")
```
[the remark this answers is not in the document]
benjaminjb: Yes, that version number is strange, and I don't love skipping tests in two supported version (k8s 20 and 21), but: looks like in api 0.24.2, the `availableReplicas` field is optional, but not `omitemtpy`, so we are sending a `0`; the error we get here is from line 174-5 below (the `before` is the sts intent):
```
assert.NilError(t, cc.Patch(ctx, before, client.Apply, client.ForceOwnership, reconciler.Owner))
```
And the error we get is
```
assertion failed: error is not nil: failed to create typed patch object: .status.availableReplicas: field not declared in schema
```

internal/controller/postgrescluster/pgbackrest_test.go
benjaminjb: Right, k8s 1.19 was the only version where this wasn't true and the only version being tested in our GH test action

internal/controller/postgrescluster/apply_test.go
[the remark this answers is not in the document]
cbandy: > the error we get here is from line 174-5 below (the `before` is the sts intent):
> 
> ```
> assert.NilError(t, cc.Patch(ctx, before, client.Apply, client.ForceOwnership, reconciler.Owner))
> ```

I see. So the ... plain/upstream apply can't setup the test.

🤔 The TODO you've added a few lines down is fair as well. We don't need this code since we've stopped supporting 1.18. I'll try something out.

internal/controller/postgrescluster/pgbackrest_test.go
```diff
@@ -1332,21 +1333,24 @@ func TestReconcileManualBackup(t *testing.T) {
...
 						Type: condition, Reason: "testing", Status: status})
 				}
-				assert.NilError(t, tClient.Create(ctx, postgresCluster))
+
 				t.Cleanup(func() {
 					// Remove finalizers, if any, so the namespace can terminate.
 					assert.Check(t, client.IgnoreNotFound(
 						tClient.Patch(ctx, postgresCluster, client.RawPatch(
 							client.Merge.Type(), []byte(`{"metadata":{"finalizers":[]}}`)))))
 				})
```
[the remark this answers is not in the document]
benjaminjb: Yeah, I don't see that finalizer being removed in the other case, and when I removed the cleanup, this ran fine.

(That cleanup was introduced here: https://github.com/CrunchyData/postgres-operator/pull/2667/files#diff-ed04828a528013a5d994920780056ad14460ba0e8136ef0b8c6e5e16d68c7ace)

internal/controller/postgrescluster/apply.go
```diff
@@ -58,16 +56,6 @@ func (r *Reconciler) apply(ctx context.Context, object client.Object) error {
...
-	case *appsv1.StatefulSet:
-		applyPodTemplateSpec(patch,
-			actual.Spec.Template, intent.(*appsv1.StatefulSet).Spec.Template,
-			"spec", "template")
-
-	case *batchv1.Job:
-		applyPodTemplateSpec(patch,
-			actual.Spec.Template, intent.(*batchv1.Job).Spec.Template,
-			"spec", "template")
-
```
benjaminjb, line 70: so SSA now works with this specific patch?

internal/controller/postgrescluster/apply_test.go
```diff
@@ -159,48 +159,24 @@ func TestServerSideApply(t *testing.T) {
...
-				intent.Spec.Template.Spec.SecurityContext)
+			assert.NilError(t,
+				cc.Patch(ctx, upstream, client.Apply, client.ForceOwnership, reconciler.Owner))
 		}
 
-		// Our apply method corrects it.
-		again := intent.DeepCopy()
+		// Our apply method generates the correct apply-patch.
+		again := constructor("status-local")
 		assert.NilError(t, reconciler.apply(ctx, again))
```
benjaminjb, line 179: Just to walk through this,
(a) we try to client.Patch, which fails below 1.22, and succeeds above that version
(b) we try to use our `apply`, which with a statefulset now passes it through to reconciler.patch which passes it to client.Patch

But we're not setting/checking the SecurityContext in either case (because in both cases it will be there)?

internal/logging/logr.go
```diff
@@ -59,3 +53,53 @@ func FromContext(ctx context.Context) logr.Logger {
 
 	return log
 }
+
+// sink implements logr.LogSink using two function pointers.
```
benjaminjb, line 57: I also am not sure about this -- why include fnError and fnInfo?

internal/logging/logr.go
```diff
@@ -59,3 +53,53 @@ func FromContext(ctx context.Context) logr.Logger {
...
+	names     []string
+	values    []interface{}
+
+	// TODO(cbandy): add names or frame to the functions below.
+
+	fnError func(error, string, ...interface{})
+	fnInfo  func(int, string, ...interface{})
+}
+
+var _ logr.LogSink = (*sink)(nil)
```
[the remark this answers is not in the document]
benjaminjb: How did we decide to use Zap elsewhere?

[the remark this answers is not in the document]
benjaminjb: I'm not sure I am either, but 

* the LogSink interface includes all the public funcs below (Enabled, Init, et al. : https://github.com/go-logr/logr/blob/a49c9717d660988d99b6f931e67c49745caacbcd/logr.go#L416
* using function pointers in our sink struct makes it easy to swap out the actual funcs (good for testing)
* I'm not actually sure that `WithName` is being used anywhere
* actual implementation of fnError and fnInfo is handled by the `Logrus` func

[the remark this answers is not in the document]
benjaminjb: (It is interesting that we have this sort of set up as if we were going to have log levels, but essentially we just have a bool for debug on/off in main.go.)

(Also, this logging is 100% from your branch, so I'm glad it looks good :) )

internal/controller/postgrescluster/apply_test.go
cbandy: > Just to walk through this,
> (a) we try to client.Patch, which fails below 1.22, and succeeds above that version
> (b) we try to use our `apply`, which with a statefulset now passes it through to reconciler.patch which passes it to client.Patch

Right. `client.Patch` marshals the `StatefulSet` struct which lacks an `omitempty` and fails on certain server versions. Our method does more than one marshal to produce an apply-patch without that field.

> But we're not setting/checking the SecurityContext in either case (because in both cases it will be there)?

The diff makes it look like one test changed, but it's really a new test. This isn't about the security context anymore.

internal/logging/logr.go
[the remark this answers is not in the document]
cbandy: > * I'm not actually sure that `WithName` is being used anywhere

We don't call it but other things do, e.g. https://github.com/kubernetes-sigs/controller-runtime/blob/v0.12.3/pkg/manager/manager.go#L355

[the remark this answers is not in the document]
cbandy: > How did we decide to use Zap elsewhere?

I think we used it because controller-runtime provides some [glue/enhancements](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/log/zap/zap.go) for it, and/or it's part of the Kubebuilder [project template](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v4/main.go#L32).

internal/controller/postgrescluster/apply_test.go
cbandy: `upstream`, `local`, `again`... If there are comments or names that can be improved to make this clear, go for it!

cbandy (approved):

PR #3371, opened 2022-09-04
Add support for building docker images based on rocky8 and rocky9

jelmer:
Add support for building docker images based on rocky8.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,12 @@ ifeq ("$(PGO_BASEOS)", "rocky8")
     DOCKERBASEREGISTRY=docker.io/rockylinux/rockylinux:
     PACKAGER=dnf
 endif
+ifeq ("$(PGO_BASEOS)", "rocky9")
+    BASE_IMAGE_OS=9
+    DOCKERBASEREGISTRY=docker.io/rockylinux/rockylinux:
+    PACKAGER=dnf
+endif
+
 
 DEBUG_BUILD ?= false
 GO ?= go
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,11 @@ ifeq ("$(PGO_BASEOS)", "ubi8")
     DOCKERBASEREGISTRY=registry.access.redhat.com/
     PACKAGER=microdnf
 endif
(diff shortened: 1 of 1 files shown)


andrewlecuyer:
@jelmer thanks for you submission!

Can you provide additional details about your need and/or want for Rocky?

Right now PGO supports the Red Hat Universal Base Image (UBI), and accordingly there should not be any restrictions building the `postgres-operator` container.  Also, we do have support for UBI 9 on our roadmap.

jelmer:
Hi Andrew,

I'm using this image as the base image to build an image that includes the debversion extension, which needs to be linked against apt and dpkg. See my dockerfile here: https://salsa.debian.org/jelmer/janitor.debian.net/-/blob/master/Dockerfile_postgres

I'd be open to using ubi (or something else), but don't think I have access to all the right dnf repositories.

andrewlecuyer:
It looks like you are specifically referencing the build for the Crunchy Postgres image, which is maintained in the Crunchy [Container Suite repository](https://github.com/CrunchyData/crunchy-containers) (and not this repository).  Therefore, your changes here would not impact builds of the `crunchy-postgres` container.

Do you have a specific need to build any of the containers maintained in this repository using Rocky?  For instance, do you have a need to build the `postgres-operator` image itself with Rocky?  You should be able to fully build this container using UBI 8, but feel free to reach out if you're running into any issues.

pere3:
Hi Andrew and jelmer,

i've a PR opened for rocky8 support and some additional work for simplifying pg_extensions installation in the crunchy-containers repo https://github.com/CrunchyData/crunchy-containers/pull/1475

i've been using this changes to build myself rocky8-based images of every postgres version since 10

maybe we can work something out of this PR for community to support pg_extensions and rocky, since not everybody have access for redhat repos

jelmer:
Yeah, I don't technically needed this repository to be rocky as well - but it's nice for consistency. Obviously there is a bit of an extra maintenance burden on the project if this is merged. 

Since creating this PR I've migrated to another postgres operator, but happy to leave this open if others are still interested.

PR #3400, opened 2022-10-03
PGO updates pgnodemx/pg_stat_statements

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Users reported that an updated image wouldn't trigger an update of monitoring extensions.

This changes that behavior by

* adding the monitor and pg image tags to the revision hash,
* adding update lines to the pgmonitor enable action,
* adding a two-step reconciliation process to this.

The two-step reconciliation handles a timing issue where a hash difference would kick off this action before the pod terminated.

Note: this _only_ targets these two extensions as updating other extensions should probably be under the user's power.

**Other Information**:
Issue: [sc-14476]

### DIFF
--- a/internal/controller/postgrescluster/pgmonitor.go
+++ b/internal/controller/postgrescluster/pgmonitor.go
@@ -78,16 +78,24 @@ func (r *Reconciler) reconcilePGMonitorExporter(ctx context.Context,
 		writableInstance *Instance
 		writablePod      *corev1.Pod
 		setup            string
+		pgImageSHA       string
+		exporterImageSHA string
 	)
 
 	// Find the PostgreSQL instance that can execute SQL that writes to every
 	// database. When there is none, return early.
-
 	writablePod, writableInstance = instances.writablePod(naming.ContainerDatabase)
 	if writableInstance == nil || writablePod == nil {
 		return nil
 	}
 
+	// For the writableInstance found above
+	// 1) make sure the `exporter` container is running
+	// 2) get and save the imageIDs for the `exporter` and `database` containers, and
+	// 3) exit early if we can't get the ImageID of either of those containers.
+	// We use these ImageIDs in the hash we make to see if the operator needs to rerun
+	// the `EnableExporterInPostgreSQL` funcs; that way we are always running
+	// that function against an updated and running pod.
 	if pgmonitor.ExporterEnabled(cluster) {
 		running, known := writableInstance.IsRunning(naming.ContainerPGMonitorExporter)
 		if !running || !known {
@@ -97,11 +105,15 @@ func (r *Reconciler) reconcilePGMonitorExporter(ctx context.Context,
 
 		for _, containerStatus := range writablePod.Status.ContainerStatuses {
 			if containerStatus.Name == naming.ContainerPGMonitorExporter {
-				setup = containerStatus.ImageID
+				exporterImageSHA = containerStatus.ImageID
+			}
+			if containerStatus.Name == naming.ContainerDatabase {
+				pgImageSHA = containerStatus.ImageID
 			}
 		}
-		if setup == "" {
-			// Could not get exporter container imageID
+
+		// Could not get container imageIDs
+		if exporterImageSHA == "" || pgImageSHA == "" {
 			return nil
 		}
 	}
@@ -127,11 +139,13 @@ func (r *Reconciler) reconcilePGMonitorExporter(ctx context.Context,
 		) error {
 			_, err := io.Copy(hasher, stdin)
 			if err == nil {
-				_, err = fmt.Fprint(hasher, command)
+				// Use command and image tag in hash to execute hash on image update
+				_, err = fmt.Fprint(hasher, command, pgImageSHA, exporterImageSHA)
 			}
 			return err
 		})
 	})
+
 	if err != nil {
 		return err
 	}
@@ -149,7 +163,6 @@ func (r *Reconciler) reconcilePGMonitorExporter(ctx context.Context,
 		}
 
 		// Apply the necessary SQL and record its hash in cluster.Status
-
 		if err == nil {
 			err = action(ctx, func(_ context.Context, stdin io.Reader,
 				stdout, stderr io.Writer, command ...string) error {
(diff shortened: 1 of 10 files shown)


internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -97,10 +99,14 @@ func (r *Reconciler) reconcilePGMonitorExporter(ctx context.Context,
...
-				setup = containerStatus.ImageID
+				exporterImageSHA = containerStatus.ImageID
+			}
+			if containerStatus.Name == naming.ContainerDatabase {
+				pgImageSHA = containerStatus.ImageID
 			}
 		}
-		if setup == "" {
+
+		if exporterImageSHA == "" {
```
cbandy, line 109: 🤔 `pgImageSHA` goes into the hash as well. Should proceed only when we have both (i.e. return early when we lack either)?

⛏️ It might be good to explain what's going on in this block and/or why; around L90, maybe.

benjaminjb: My thought process on not checking that was that if we don't have the pg image, it wouldn't be a writable pod, but now thinking about it, I like the idea of the explicit check. Will change that and add an explanation.

testing/kuttl/e2e-other/exporter-upgrade/00--cluster.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - name: repo1
+        volume:
+          volumeClaimSpec:
+            accessModes:
+            - "ReadWriteOnce"
+            resources:
+              requests:
+                storage: 1Gi
+  monitoring:
+    pgmonitor:
```
jmckulk, line 28: Should we be defining the exporter image explicitly here too :question:

testing/kuttl/e2e-other/exporter-upgrade/00-assert.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: exporter
+status:
+  instances:
+    - name: instance1
+      readyReplicas: 1
+      replicas: 1
+      updatedReplicas: 1
```
jmckulk, line 10: Should we check that the exporter status is set? Maybe we should be doing this in the `e2e` test :thinking:

testing/kuttl/e2e-other/exporter-upgrade/02--cluster.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: exporter
+spec:
+  postgresVersion: 14
+  image: registry.developers.crunchydata.com/crunchydata/crunchy-postgres:ubi8-14.5-1
```
jmckulk, line 7: I think we should remove the defined image and let `RELATED_IMAGES` take over. Assuming we can remove things from the spec with KUTTL :thinking: :pick:

testing/kuttl/e2e-other/exporter-upgrade/03--check-exporter.yaml
```diff
@@ -0,0 +1,21 @@
...
+          --output name --selector '
+            postgres-operator.crunchydata.com/cluster=exporter,
+            postgres-operator.crunchydata.com/role=master'
+      )
+
+      # Get errors from the exporter
+      ERR=$(kubectl logs --namespace "${NAMESPACE}" "${PRIMARY}" -c exporter | grep -e "Error running query on database")
+      ERR_COUNT=$(echo "$ERR" | wc -l)
+
+      if [[ "$ERR_COUNT" -gt 2 ]]; then
```
jmckulk, line 18: Why are we checking if the count is greater than 2 here, instead of expecting it to be 0 :question: 

Edit: Oh, I see the Readme; maybe throw a comment here? :pick:

testing/kuttl/e2e-other/exporter-upgrade/README.md
```diff
@@ -0,0 +1,15 @@
...
+This starts up a cluster with a purposely outdated extension. Because we want a particular
+extension, the image used here is hard-coded and not publicly available.
+
+02-03
+The cluster is updated, again with a hardcoded image. Due to the change made in
+https://github.com/CrunchyData/postgres-operator/pull/3400, this should no longer produce errors.
+However, what has been observed is that there are occasionally a few errors logged
+before the extension is updated. So this checks that there are no more than 2 errors, since that
+was the observed number during manual tests of the check; and since without the fix,
+there were more errors observed.
```
jmckulk, line 15: I do not understand. What are the 2 errors that we expect to find?

testing/kuttl/e2e-other/exporter-upgrade/00--cluster.yaml
benjaminjb: That's an interesting question... we are looking for a particular interaction here (between the exporter image and two different pgnodemx versions in two different pg images), and as long as we're hard-coding one image, yeah, I think we might as well.

testing/kuttl/e2e-other/exporter-upgrade/00-assert.yaml
benjaminjb: I think we run into the problem we've had with KUTTL where checking anything in an array becomes tricky if we only care about one thing. (Like the `exporter` entry in the `containerStatuses`.) The 01--check-exporter.yaml would fail if the exporter wasn't working, and at least in the `e2e`, that exporter will be working by the time the backup is finished.

(Here, because I'm using an outdated image, the backup isn't finishing, which is why I pulled that part out of 00-assert and why I put a retry in 001-check-exporter.)

testing/kuttl/e2e-other/exporter-upgrade/02--cluster.yaml
benjaminjb: For now, and with the understanding that this test is odd because the image is hard-coded, I would rather leave it hard-coded just so we know what the error we expect will be. (Or to be totally honest: I feel like this test isn't going to be run as is very often, but may serve as a blueprint for a better version and/or as documentation for what we might expect.)

testing/kuttl/e2e-other/exporter-upgrade/README.md
benjaminjb: Updated

jmckulk (approved):

PR #3407, opened 2022-10-06
Custom TLS for Exporter (Encryption Only)

jmckulk:
With this change we allow users to bring custom certificates and enable TLS for the exporter. This will be an opt-in feature, PGO will not automatically generate certs like it does for some other features.

You can enable TLS by using the following spec fields:
```
spec:
  monitoring:
    pgmonitor:
      exporter:
        customTLSSecret:
          name: hippo.tls
```
Once TLS is enabled in the exporter, you can configure your Prometheus instance to scrape over https.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

**Other Information**:
[sc-11974]


pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -678,13 +678,17 @@ type PGMonitorSpec struct {
...
+	// the customization of PostgreSQL Exporter queries. If a "queries.yml" file is detected in
 	// any volume projected using this field, it will be loaded using the "extend.query-path" flag:
 	// https://github.com/prometheus-community/postgres_exporter#flags
 	// Changing the values of field causes PostgreSQL and the exporter to restart.
 	// +optional
 	Configuration []corev1.VolumeProjection `json:"configuration,omitempty"`
 
+	// Projected secret containing custom TLS certificates to encrypt output from the exporter
+	// web server
+	CustomTLSSecret *corev1.SecretProjection `json:"customTLSSecret,omitempty"`
```
cbandy, line 690: 🔧 `omitEmpty` implies `+optional`, but we should include the marker so people don't have to know that.

testing/kuttl/e2e/exporter/10--tls-cluster.yaml
```diff
@@ -0,0 +1,40 @@
...
+  tls.key: LS0tLS1CRUdJTiBFQyBQUklWQVRFIEtFWS0tLS0tCk1IY0NBUUVFSUlFRlpFS1ZIZ3YyT25uZkljaXlRUFlzbzBvalBVN3NIVS9KUFI3TE5CWERvQW9HQ0NxR1NNNDkKQXdFSG9VUURRZ0FFNHgzdDRuMUpHWnVDMHdHNFpjU3hUYnZRMUh5aElrRWdFTzh3SXVOMG1tMDE0WmY3SFFjWQo0VmdPWGlzM0FGaEU0T0lidUNLR0pXMkQ3Y3pkclB0eGt3PT0KLS0tLS1FTkQgRUMgUFJJVkFURSBLRVktLS0tLQo=
```
cbandy, line 9: ❓ How'd you generate these? When do they expire? 🤔 The answers should probably be written near here, too. (comment, readme, whatever.)

testing/kuttl/e2e/exporter/10-assert.yaml
```diff
@@ -0,0 +1,29 @@
...
+  labels:
+    postgres-operator.crunchydata.com/cluster: exporter-tls
+    postgres-operator.crunchydata.com/pgbackrest-backup: replica-create
+status:
+  succeeded: 1
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: exporter-tls-primary
```
cbandy, line 24: 🤔 I don't see this Service being used (port, DNS, etc) in later steps. What does the assertion mean here?

testing/kuttl/e2e/exporter/10-assert.yaml
```diff
@@ -0,0 +1,29 @@
...
+      updatedReplicas: 1
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  labels:
+    postgres-operator.crunchydata.com/cluster: exporter-tls
+    postgres-operator.crunchydata.com/pgbackrest-backup: replica-create
+status:
+  succeeded: 1
```
cbandy, line 19: 🤔 I don't see any backups being used in later steps. What does the assertion mean here?

testing/kuttl/e2e/exporter/12-assert.yaml
```diff
@@ -0,0 +1,15 @@
...
+  instances:
+    - name: instance1
+      readyReplicas: 1
+      replicas: 1
+      updatedReplicas: 1
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: exporter-tls-primary
```
cbandy, line 15: 🤔 Same question/thought about this Service assertion.

testing/kuttl/e2e/exporter/13--check-exporter.yaml
```diff
@@ -0,0 +1,49 @@
...
+      PRIMARY=$(
+        kubectl get pod --namespace "${NAMESPACE}" \
+          --output name --selector '
+            postgres-operator.crunchydata.com/cluster=exporter-tls,
+            postgres-operator.crunchydata.com/role=master'
+      )
+
+      # Wait some time so that the instance pod is redeployed
+      # TODO: automate this wait
+      sleep 30
```
cbandy, line 16: 🌱 Hmm. The status we have is the revision hash. My only idea atm is to wait for that to change. That requires storing the old value (before 12--disable-tls.yaml) somewhere (annotation?) then looping here with sleep until the status `!=` the old value.

cbandy (commented):
I've looked at the tests but not the implementation.

testing/kuttl/e2e/exporter/10-assert.yaml
jmckulk: I thought these were good general checks to ensure that the cluster had started successfully. I don't think they are specifically necessary so could be removed

testing/kuttl/e2e/exporter/10-assert.yaml
jmckulk: Again, I have left these in as general "cluster is stable" checks

testing/kuttl/e2e/exporter/13--check-exporter.yaml
jmckulk: Would the revision hash help here? I need the instance pod to be rebuilt so that exporter is restarted without tls. Wouldn't the revision hash be updated when the StatefulSet is updated? Would that have a concept of if the pod has stopped/restarted?

internal/controller/postgrescluster/pgmonitor_test.go
```diff
@@ -578,3 +599,121 @@ func TestReconcilePGMonitorSecret(t *testing.T) {
...
+    sources:
+    - secret:
+        name: custom-exporter-certs
+- configMap:
+    name: test-exporter-web-config
+  name: web-config
+		`))
+
+		// Is the exporter container in position 0?
+		assert.Assert(t, template.Spec.Containers[0].Name == naming.ContainerPGMonitorExporter)
```
tjmoore4, line 697: :question: Are we just verifying position here before running the subsequent tests or is there another reason for this check?

internal/naming/names.go
```diff
@@ -343,6 +343,16 @@ func MonitoringUserSecret(cluster *v1beta1.PostgresCluster) metav1.ObjectMeta {
 	}
 }
 
+// ExporterWebConfigMap returns ObjectMeta necessary to lookup and create the
+// exporter web configmap. This configmap is used to ocnfigure the exporter
```
tjmoore4, line 347: :pick: 
```suggestion
// exporter web configmap. This configmap is used to configure the exporter
```

tjmoore4 (commented):
Looks good overall. A couple of questions and nitpicks.

testing/kuttl/e2e/exporter/10--tls-cluster.yaml
cbandy: I forget how to generate certs using `openssl` or any other tool, so those steps would help me in the future. All certs expire _eventually_; we'll need to regenerate compatible ones then. If it's "far enough" out we can punt. 🤷🏻

testing/kuttl/e2e/exporter/13--check-exporter.yaml
jmckulk: I spent a lot of time attempting to `kubectl wait` for the pod to be not ready and then ready again, but it got to the point where I just wanted/needed to be done with it. There is definitely room for improvement

internal/controller/postgrescluster/pgmonitor_test.go
jmckulk: Yup, just checking the position.

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -321,9 +323,130 @@ func addPGMonitorExporterToInstancePodSpec(
...
+	for i, container := range template.Spec.Containers {
+		if container.Name == naming.ContainerPGMonitorExporter {
+			exporterContainer = &template.Spec.Containers[i]
+			found = true
+		}
+	}
+
+	if found &&
+		pgmonitor.ExporterEnabled(cluster) &&
+		(cluster.Spec.Monitoring.PGMonitor.Exporter.CustomTLSSecret != nil) {
```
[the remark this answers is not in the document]
jmckulk: It is redundant, but considering we are checking pointers, I thought it would be good to check the values in the function

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -321,9 +323,130 @@ func addPGMonitorExporterToInstancePodSpec(
...
+		// TODO (jmckulk): params for paths and such
+		certVolume := corev1.Volume{Name: "exporter-certs"}
```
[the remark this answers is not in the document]
tjmoore4: If we're going to continue refactoring at a later point, that certainly makes sense.

 If I'm remembering correctly, throughout the code we have plain strings for things that are only ever used once, so anything in that category is probably fine. For the rest, we can leave as is for now (perhaps with a TODO?) or move them to the top of the file for a bit more clarity (like we do [here](https://github.com/CrunchyData/postgres-operator/blob/master/internal/controller/postgrescluster/pgbackrest.go#L107-L112)). 

Regardless, this isn't a blocker. I'm fine with any approach that makes sense to you.

PR #3417, opened 2022-10-13
Adds Sleep to Cluster Annotation Test [sc-15828]

tony-landreth:
The requireClusterReady helper in suite_helpers_test.go does not always wait until the cluster is truly ready. Consequently, 'on_update' in cluster_annotation_test.go flakes, attempting to run update on a pod that isn't ready.

[sc-15828]

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

**Other Information**:

### DIFF
--- a/testing/pgo_cli/cluster_annotation_test.go
+++ b/testing/pgo_cli/cluster_annotation_test.go
@@ -216,6 +216,9 @@ func TestClusterAnnotation(t *testing.T) {
 						requirePgBouncerReady(t, namespace(), test.testName, (2 * time.Minute))
 					}
 
+					// Allow pods time to reach ready status before running update.
+					time.Sleep(time.Minute + 30*time.Second)
+
 					updateCMD := []string{"update", "cluster", test.testName, "-n", namespace(), "--no-prompt"}
 					updateCMD = append(updateCMD, test.addFlags...)
 					output, err = pgo(updateCMD...).Exec(t)


testing/pgo_cli/cluster_annotation_test.go
```diff
@@ -216,6 +216,9 @@ func TestClusterAnnotation(t *testing.T) {
 						requirePgBouncerReady(t, namespace(), test.testName, (2 * time.Minute))
 					}
 
+					// Allow pods time to reach ready status before running update.
+					time.Sleep(time.Minute + 30*time.Second)
```
tjmoore4, line 220: :question: I may be missing something about this particular case, but I'm wondering if, instead of adding the sleep here, we could either lengthen the time given to the `requireClusterReady` function above? Alternatively, if that function isn't waiting correctly, perhaps we could refactor it a bit?

tony-landreth: Good question. Unfortunately, lengthening the time given to `requireClusterReady` doesn't solve the problem, because the condition it uses to determine readiness isn't truly readiness. It looks for label to be applied, which will happen before the pod is fully ready. The upshot is that extending the wait time doesn't help, because `requireClusterReady` will happily signal that the cluster is ready when it isn't. I started digging into whether a refactoring was worth it -- trying to identify a more accurate readiness condition -- but the cost/benefit came down on the side of just using Sleep.

tjmoore4: :+1: Understood, thanks for the explanation. If the sleep is sufficient and the other function isn't a good option, we can definitely go with this approach.

tjmoore4 (approved):

Issue #3286, opened 2022-07-04
Support IPv6 for backrest (mTLS)


gkolvo:
## Overview

Add support for pure IPv6 environments.

## Use Case

Now pgBackRest 2.38 listens on the first address returned by ```getaddrinfo()```. For pure IPv6 the first address is always IPv4.
```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
```

## Desired Behavior

I would like to have possibility to customize ```tls-server-address``` and set it to ```--tls-server-address=::1``` 

Now it is hardcoded to ```global.Set("tls-server-address", "0.0.0.0")```

https://github.com/CrunchyData/postgres-operator/blob/6feb0ab6a28aafde023abd41ae2474323b7d043b/internal/pgbackrest/config.go#L439-L458

The proposal is to make this parameter configurable somewhere.

## Environment

Tell us about your environment:

Please provide the following details:

- Platform: (`Kubernetes`)
- Platform Version: (e.g. `1.21.7`)
- PGO Image Tag: (e.g. `ubi8-5.1.0-0`)
- Postgres Version (e.g. `13`)
- Storage: (e.g. `rook-cephfs`)
- Number of Postgres clusters: (`2`)

## Additional Information
Using global variables I get:
**pgbackrest ERROR: [031]: option 'tls-server-address' cannot be set multiple times**

cbandy:
I've opened a request at https://github.com/pgbackrest/pgbackrest/issues/1841.

andrewlecuyer:
@gkolvo can you try the following workaround to configure the pgBackRest TLS server for IPv6?

First, create a custom pgBackRest configuration file (e.g. `pgbr-custom.conf`) containing the following:
```conf
[global:server]
tls-server-address = ::

[global:server-ping]
tls-server-address = localhost
```

Which you can then turn into a ConfigMap:
```bash
kubectl create cm --from-file=pgbr-custom.conf my-pgbr-config
```

And from there you can add that to your pgBackRest configuration in the PostgresCluster spec, e.g.:
```yaml
spec:
  backups:
    pgbackrest:
      configuration:
      - configMap:
[...]

Issue #1841, opened 2020-09-02
How can I run the pgo operator with a custom image?


IlyaNakhaichuk:
I have replaced the values ​​ccp_image_prefix and pgo_image_prefix in the helm chart
to your custom image, but I ran into the fact that when replacing ccp_image_prefix --- pgo-deploy does not start, when replacing pgo_image_prefix: ----- tries to download the image from registry.developers.crunchydata.com/crunchydata/custom_container:tag

Сould you help with this?

jkatz:
Can you please be more specific?

- Which images are customized?
- What are the customizations?
- How are they customized? Where are they derived from?
- What is the end goal?
- What is the configuration of the YAML chart?
- Are there any logs?

IlyaNakhaichuk:
Which images are customized?
FROM  registry.developers.crunchydata.com/crunchydata/crunchy-postgres-ha:centos7-12.3-4.4.0

What are the customizations?
Add to image wal2json12

How are they customized? Where are they derived from?
By adding layers to the container.

What is the end goal?
Database replication creation.

What is the configuration of the YAML chart?
`rbac:
    # rbac.create: if false RBAC resources should be in place
    create: true
    # rbac.useClusterAdmin: creates a ClusterRoleBinding giving cluster-admin to serviceAccount.name
    useClusterAdmin: true

# serviceAccount: settings for Service Account used by the deployer
serviceAccount:
    # serviceAccount.create: Whether to create a Service Account or not
    create: true
    # serviceAccount.name: The name of the Service Account to create or use
    name: ""

# =====================
# Configuration Options
# More info for these options can be found in the docs
# https://crunchydata.github.io/postgres-operator/latest/installation/configuration/
# =====================
archive_mode: "true"
archive_timeout: "60"
#------Bucket for test env:
backrest_aws_s3_bucket: "pgo"
#------Bucket for prod env:
#backrest_aws_s3_bucket: "pgo"
backrest_aws_s3_endpoint: "s3-eu-west-1.amazonaws.com"
#------New secret key test env:
#backrest_aws_s3_key: "$AWS_KEY_TEST"
#backrest_aws_s3_secret: "$AWS_SECRET_TEST"
#------New secret key prod:
backrest_aws_s3_key: 
backrest_aws_s3_secret: 
backrest_aws_s3_region: "eu-west-1"
backrest_aws_s3_uri_style: ""
backrest_aws_s3_verify_tls: "true"
backrest_port: ""
[...]

PR #3426, opened 2022-10-19
Ipv6 workaround for pgbackrest issue #1841

dsessler7:
These changes provide a workaround for an issue in pgBackRest that prevents it from working in an IPv6 environment. The workaround involves a user adding an annotation to the PostgresCluster. The changes in the code make the operator check for the existence of the annotation, and if present, set pgBackRest's `tls-server-address` to `::` rather than `0.0.0.0`.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

https://github.com/pgbackrest/pgbackrest/issues/1841
https://github.com/CrunchyData/postgres-operator/issues/3286

Currently, when PGO is used in an IPv6 environment, pgBackRest uses an IPv4 address, making it inaccessible and therefore cannot perform any backups.

The operator now checks for the existence of a specific annotation on the PostgresCluster (`postgres-operator.crunchydata.com/pgbackrest-ip-version: IPv6`) and if it is present it will set the pgBackRest `tls-server-address` to `::`, allowing it to communicate with the rest of the cluster and perform backups as expected.

**Other Information**:

[sc-15736]

### DIFF
--- a/docs/content/tutorial/backups.md
+++ b/docs/content/tutorial/backups.md
@@ -379,6 +379,19 @@ The full list of [pgBackRest configuration options](https://pgbackrest.org/confi
 
 [https://pgbackrest.org/configuration.html](https://pgbackrest.org/configuration.html)
 
+## IPv6 Support
+
+If you are running your cluster in an IPv6-only environment, you will need to add an annotation to your PostgresCluster so that PGO knows to set pgBackRest's `tls-server-address` to an IPv6 address. Otherwise, `tls-server-address` will be set to `0.0.0.0`, making pgBackRest inaccessible, and backups will not run. The annotation should be added as shown below:
+
+```yaml
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: hippo
+  annotations:
+    postgres-operator.crunchydata.com/pgbackrest-ip-version: IPv6
+```
+
 ## Next Steps
 
 We've now seen how to use PGO to get our backups and archives set up and safely stored. Now let's take a look at [backup management]({{< relref "./backup-management.md" >}}) and how we can do things such as set backup frequency, set retention policies, and even take one-off backups!
--- a/internal/pgbackrest/config.go
+++ b/internal/pgbackrest/config.go
@@ -472,12 +472,8 @@ func serverConfig(cluster *v1beta1.PostgresCluster) iniSectionSet {
 	// going to workaround the issue by allowing the user to add an annotation to
 	// enable IPv6. We will check for that annotation here and override the
 	// "tls-server-address" setting accordingly.
-	annotations := cluster.GetAnnotations()
-	if annotations != nil {
-		if ipVersion, exists := annotations[naming.PGBackRestIPVersion]; exists &&
-			strings.ToLower(ipVersion) == "ipv6" {
-			global.Set("tls-server-address", "::")
-		}
+	if strings.ToLower(cluster.Annotations[naming.PGBackRestIPVersion]) == "ipv6" {
+		global.Set("tls-server-address", "::")
 	}
 
 	// The client certificate for this cluster is allowed to connect for any stanza.
--- a/internal/pgbackrest/config.go
+++ b/internal/pgbackrest/config.go
@@ -472,7 +472,7 @@ func serverConfig(cluster *v1beta1.PostgresCluster) iniSectionSet {
 	// going to workaround the issue by allowing the user to add an annotation to
 	// enable IPv6. We will check for that annotation here and override the
 	// "tls-server-address" setting accordingly.
-	if strings.ToLower(cluster.Annotations[naming.PGBackRestIPVersion]) == "ipv6" {
+	if strings.EqualFold(cluster.Annotations[naming.PGBackRestIPVersion], "ipv6") {
 		global.Set("tls-server-address", "::")
 	}
 
--- a/internal/pgbackrest/config.go
+++ b/internal/pgbackrest/config.go
@@ -18,6 +18,7 @@ package pgbackrest
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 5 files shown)


internal/pgbackrest/config.go
```diff
@@ -465,6 +466,20 @@ func serverConfig(cluster *v1beta1.PostgresCluster) iniSectionSet {
...
+	// NOTE (dsessler7): As pointed out by Chris above, there is an issue in
+	// pgBackRest (#1841), where using a wildcard address to bind all addresses
+	// does not work in certain IPv6 environments. Until this is fixed, we are
+	// going to workaround the issue by allowing the user to add an annotation to
+	// enable IPv6. We will check for that annotation here and override the
+	// "tls-server-address" setting accordingly.
+	annotations := cluster.GetAnnotations()
+	if annotations != nil {
+		if ipVersion, exists := annotations[naming.PGBackRestIPVersion]; exists &&
+			strings.ToLower(ipVersion) == "ipv6" {
```
cbandy, line 478: 👍🏻 The docs all say (capital) `IPv6` and I'm fully on board with being case-insensitive here.

⛏️ If we're trying to do a case-insensitive comparison, we should go ahead and use the [function for that](https://pkg.go.dev/strings#EqualFold) `strings.EqualFold(ipVersion, "ipv6")`.

cbandy (approved):

internal/pgbackrest/config.go
```diff
@@ -465,6 +466,20 @@ func serverConfig(cluster *v1beta1.PostgresCluster) iniSectionSet {
...
+			global.Set("tls-server-address", "::")
+		}
+	}
```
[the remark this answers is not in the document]
dsessler7: I was definitely aware of the fact that in Go the zero value will be returned when attempting to retrieve a value that doesn't exist from a map, so I'm honestly not sure why I didn't use that functionality when retrieving the specific annotation and comparing its value to "ipv6". Guess it just slipped my mind.

However, in regards to retrieving the annotations themselves, I think I noticed the usage of the `GetAnnotations()` and `GetAnnotationsOrNil()` functions throughout the codebase and assumed, perhaps incorrectly, that one should be used when retrieving annotations and that if no annotations had been set that the return value would be `nil` (and attempting to retrieve a value from `nil` would throw an error). Hence the initial check for its existence.

I think I'm also slightly confused about the structure of the PostgresCluster struct. I would not have thought to access Annotations directly from the cluster since they reside inside the cluster's metadata. Also, when I look at the PostgresCluster struct I see ObjectMeta and TypeMeta, which further confuse me, but I also haven't taken the time to dig in and figure out the difference and how they each work.

internal/pgbackrest/config_test.go
```diff
@@ -337,3 +337,26 @@ log-level-stderr = error
 log-timestamp = n
 `)
 }
+
+func TestServerConfigIPv6(t *testing.T) {
+	cluster := &v1beta1.PostgresCluster{}
+	cluster.UID = "shoe"
+	annotations := map[string]string{}
+	annotations[naming.PGBackRestIPVersion] = "IPv6"
+	cluster.ObjectMeta.Annotations = annotations
```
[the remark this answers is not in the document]
dsessler7: I think I initially had code similar to yours and my linter was complaining for some reason, so I switched it to what I pushed. I agree that your code is easier to read.

internal/pgbackrest/config.go
dsessler7: I wasn't aware of that function, but good to know.

internal/pgbackrest/config.go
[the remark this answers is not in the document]
cbandy: > I think I'm also slightly confused about the structure of the PostgresCluster struct. I would not have thought to access Annotations directly from the cluster since they reside inside the cluster's metadata.

These are [embedded fields](https://go.dev/ref/spec#Struct_types) because they don't have a field name in front of the field type. The methods you're calling are actually being called on those.

https://github.com/CrunchyData/postgres-operator/blob/607c1b1ae80427cfdb93e96a54293446990c3eb2/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go#L596-L597

> Also, when I look at the PostgresCluster struct I see ObjectMeta and TypeMeta, which further confuse me, but I also haven't taken the time to dig in and figure out the difference and how they each work.

ObjectMeta is name, namespace, etc. TypeMeta is the GVK. The `json:",inline"` tag tells the JSON encoder/decoder to put... that field's fields... up one level.

- https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta
- https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#TypeMeta

cbandy (approved):

docs/content/tutorial/backups.md
```diff
@@ -379,6 +379,19 @@ The full list of [pgBackRest configuration options](https://pgbackrest.org/confi
...
+## IPv6 Support
+
+If you are running your cluster in an IPv6-only environment, you will need to add an annotation to your PostgresCluster so that PGO knows to set pgBackRest's `tls-server-address` to an IPv6 address. Otherwise, `tls-server-address` will be set to `0.0.0.0`, making pgBackRest inaccessible, and backups will not run. The annotation should be added as shown below:
+
+```yaml
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: hippo
+  annotations:
```
David-Angel, line 391: Pretty sure this should be a label.   (see labels vs annotations)

andrewlecuyer: We aligned with the [Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions) with the use of an annotation here:

> Use annotations to store API extensions that the controller responsible for the resource doesn't need to know about, experimental fields that aren't intended to be generally used API fields, etc. Beware that annotations aren't automatically handled by the API conversion machinery.

This fits under the category of an "API extension" that is not intended to be a "generally used" API field, as we intend this setting to no longer be needed once https://github.com/pgbackrest/pgbackrest/issues/1841 is implemented in pgBackRest.

David-Angel: good answer thanks  :D

PR #3434, opened 2022-10-24
Update pgBackRest repo option logic

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

When taking a backup, PGO tries to help by not allowing the user to pass the "--repo" option. However, the current method for catching this results in catching any option that begins with "--repo", which prevents users from passing in perfectly valid options.

This commit corrects the flag check to only block on exact matches of "--repo".

**Other Information**:
Issue: [sc-16128]

### DIFF
--- a/internal/controller/postgrescluster/pgbackrest.go
+++ b/internal/controller/postgrescluster/pgbackrest.go
@@ -1011,7 +1011,9 @@ func (r *Reconciler) reconcileRestoreJob(ctx context.Context,
 	for _, opt := range options {
 		var msg string
 		switch {
-		case strings.Contains(opt, "--repo"):
+		// Since '--repo' can be set with or without an equals ('=') sign, we check for both
+		// usage patterns.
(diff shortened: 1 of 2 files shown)


benjaminjb (approved):
Looks good -- is that test func all  for invalid options? Do we have a test for valid options and would it be worthwhile to add a `--repo1-xxx=yyy` option for that?

tjmoore4:
> Looks good -- is that test func all for invalid options? Do we have a test for valid options and would it be worthwhile to add a `--repo1-xxx=yyy` option for that?

It might potentially be worth adding checks for `--target-action` and `--link-map`, but I didn't really go too far into updating tests for all possible flags for this fix. Worth noting that the only place that required an update was the `TestReconcilePostgresClusterDataSource` function. If I start going down that path I think there are other enhancements that would be worth considering for the test overall, but I feel like this test change is appropriately scoped for the update.

Issue #3435, opened 2022-10-25
Services missing appProtocol definition


szelenka:
Please ensure you do the following when reporting a bug:

- [x] Provide a concise description of what the bug is.
- [x] Provide information about your environment.
- [x] Provide clear steps to reproduce the bug.
- [x] Attach applicable logs. Please do not attach screenshots showing logs unless you are unable to copy and paste the log data.
- [x] Ensure any code / output examples are [properly formatted](https://docs.github.com/en/github/writing-on-github/basic-writing-and-formatting-syntax#quoting-code) for legibility.

## Overview

Kubernetes v1.18+ introduced [Application Protocol](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol) for all Services. This is a simple field called `appProtocol` which identifies the application protocol being used on that service port.

The services created by the v5.2.0 PGO does not include these fields, which prevents other services looking for them to work with the Crunchy PGO. Such as Istio:
https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection

## Environment

Please provide the following details:

- Platform: Kubernetes
- Platform Version: 1.22
- PGO Image Tag: ubi8-5.2.0-0
- Postgres Version 14
- Storage: PVC

## Steps to Reproduce

1. Install latest Crunchy PGO 
2. Provision a new Instance
3. Inspect Services created for the Instance, notice the `port` is missing the `appProtocol` field.

### EXPECTED

1. When creating a service, I would expect the Crunchy PGO to populate the `appProtocol` field appropriately.

### ACTUAL

1. No `appProtocol` field is populated

## Logs
n/a

## Additional Information

n/a

PR #3436, opened 2022-10-25
adding appProtocol to all services

szelenka:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

When the operator generates a new service, it will add the `appProtocol` field on that service, to allow it to be used by services which look for `appProtocol`

**Other Information**:
 
Issue: #3435


internal/controller/postgrescluster/cluster_test.go
```diff
@@ -753,7 +754,8 @@ ownerReferences:
 	`))
 	assert.Assert(t, marshalMatches(service.Spec, `
 ports:
-- name: postgres
+- appProtocol: tcp
+  name: postgres
```
cbandy, line 758: ❓ Is `tcp` the correct value here? The Kubernetes [docs](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol) and API [reference](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec) seem to say these should be IANA service names. Based on that guidance I expect `appProtocol: postgresql` here. Would Istio behave correctly in that case?

📝 I've opened #3437 to add IANA constants in two packages.

internal/controller/postgrescluster/pgadmin_test.go
```diff
@@ -266,7 +266,8 @@ ownerReferences:
 		// Defaults to ClusterIP.
 		assert.Equal(t, service.Spec.Type, corev1.ServiceTypeClusterIP)
 		assert.Assert(t, marshalMatches(service.Spec.Ports, `
-- name: pgadmin
+- appProtocol: tcp
+  name: pgadmin
```
cbandy, line 270: ❓ pgAdmin is definitely speaking HTTP on this port. Should this be `appProtocol: http` as [registered](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=http) at IANA?

❓ In the future, when pgAdmin is deployed with TLS, will this need to change to `https` to work correctly with Istio?

internal/controller/postgrescluster/pgbouncer_test.go
```diff
@@ -175,7 +175,8 @@ ownerReferences:
 		// Defaults to ClusterIP.
 		assert.Equal(t, service.Spec.Type, corev1.ServiceTypeClusterIP)
 		assert.Assert(t, marshalMatches(service.Spec.Ports, `
-- name: pgbouncer
+- appProtocol: tcp
+  name: pgbouncer
```
cbandy, line 179: 📝 PgBouncer is speaking the PostgreSQL [protocol](https://www.postgresql.org/docs/current/protocol.html), so it should have the same `appProtocol` as the primary and replica Services.

internal/controller/postgrescluster/cluster.go
```diff
@@ -142,11 +142,13 @@ func (r *Reconciler) generateClusterPrimaryService(
 	service.Spec.ClusterIP = corev1.ClusterIPNone
 	service.Spec.Selector = nil
 
+	appProtocol := naming.AppProtocolTCP
 	service.Spec.Ports = []corev1.ServicePort{{
 		Name:       naming.PortPostgreSQL,
 		Port:       *cluster.Spec.Port,
 		Protocol:   corev1.ProtocolTCP,
 		TargetPort: intstr.FromString(naming.PortPostgreSQL),
+		AppProtocol: &appProtocol,
```
cbandy, line 151: ⛏️ We have a package that can turn expressions into pointers. I recommend `AppProtocol: initialize.String(…),` here.

https://github.com/CrunchyData/postgres-operator/blob/dd408ff606a3754bf008ef80e40400dbb9a95c51/internal/initialize/primitives.go#L34-L35

cbandy (commented):
Thanks for this! I've been curious about this API field since pgBackRest registered with the IANA at the beginning of this year.

❓ The Istio docs for [protocol selection](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/) seem to say that Istio will treat protocols it can't detect as TCP. What happens in Istio when all these Services are explicitly set to `tcp`?

I have a few more questions inline.

szelenka:
@cbandy when it's set explicitly to `tcp` nothing much happens TBH. It'll still default to plain TCP.

In the past this was the default we used to enable some of the features in Istio, and some tooling explicitly looks for appProtocol; such as [Kiali](https://kiali.io/docs/features/validations/#kia0602---port-appprotocol-must-follow-protocol-form).

Now that I know there's a IANA for postgres, it makes sense to use that instead. I'll work on updating them

internal/controller/postgrescluster/pgadmin_test.go
szelenka: will change to `http` for now.

RE: `https`, it depends. When traffic is identified as `https`, the Istio sidecar(may) not re-encrypt the traffic using mTLS configured in the service mesh. Essentially it'd just forward the traffic to the destination. So as long as `pgadmin` is doing the TLS termination/validation it should be fine.

However, then it could be a headache for whoever's managing the Istio mesh to assert all traffic in the cluster is satisfying certain encryption. But that'd be on them to document. IMO, I'd prefer to keep all traffic between Pods through Istio's mTLS.. though we don't really use `pgadmin` so I'll defer to someone else on that

internal/controller/postgrescluster/pgbouncer_test.go
szelenka: yep

internal/controller/postgrescluster/cluster.go
cbandy: #3437 merged, so this might be:
```suggestion
		AppProtocol: initialize.String(postgres.IANAServiceName),
```

szelenka: updated

szelenka:
I also updated the Endpoints, such that they'd copy over the appProtocol from the Service as well

szelenka:
@cbandy this is my first time contributing to this repo. Not sure what the next steps are after resolving the conversations, can you help guide me?

cbandy:
I've kicked off some PR checks. I expect an unrelated failure from `kubernetes-k3d (latest)` but the rest need to pass before we merge.

internal/naming/names.go
```diff
@@ -186,6 +186,11 @@ const (
 	RestoreConfigCopySuffix = "%s-restorecopy-%d"
 )
 
+const (
+	// AppProtocolTCP is the name of the appProtocol for the Service which use the http protocol
+	AppProtocolHTTP = "http"
```
cbandy, line 191: ⛏️ The comment says AppProtocolTCP but the name is AppProtocolHTTP.

```suggestion
	// AppProtocolHTTP is the name of the appProtocol for Services which use the HTTP protocol
	AppProtocolHTTP = "http"
```

cbandy (approved):
@szelenka Thank you for your patience! I have one nitpick on a doc comment.

GitHub didn't run any checks on your last push (not your fault). Do you mind squashing these commits, which will kick them off again?

internal/naming/names.go
szelenka: Made this change and attempted to squash the commits, running `make check-envtest` passes locally

cbandy:
I ran into trouble testing this on OCP 4.8. API calls succeed but the `appProtocol` field isn't saved in the API. The update event(s) trigger another reconcile which writes again which causes more update event(s) causing more reconciles ad nauseam.

szelenka:
I don't have access to an OCP cluster to debug on.

@cbandy is it just version 4.8, or does it happen on the later builds as well?

cbandy:
> @cbandy is it just version 4.8, or does it happen on the later builds as well?

I only tried on that one version. Looking at the dates, I probably tested on 4.8.51. The cluster updated to 4.8.52 two days after my last comment.

> This suggests OCP may have just gotten around to "GA" appProtocol in version 4.8.52? https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable-4.8/changelog.html

That lists all the changes since 4.7.0, so I think you're seeing the fact that OCP 4.8 is based on K8s 1.21. A more granular list is here: https://docs.openshift.com/container-platform/4.8/release_notes/ocp-4-8-release-notes.html#ocp-4-8-asynchronous-errata-updates

The API field was present/available. I was able to set it using `kubectl edit`. Maybe something with SSA?

Now that I think about it... the symptoms could also be from an old controller running. Maybe I had a dirty environment.

> I don't have access to an OCP cluster to debug on.

This is the cluster version today. Maybe you can see something similar on another distro?

```
Server Version: 4.8.52
Kubernetes Version: v1.21.11+5cc9227
```

szelenka:
I'm not having luck with getting an OpenShift cluster setup. 

I've gone through these steps to create a cluster on my Mac with a M1 chip:
https://console.redhat.com/openshift/create/local

But when attempting to run the PGO, it'll assert with:
```
exec /usr/local/bin/postgres-operator: exec format error
```

Rather than debugging OpenShift, would it be better to just use the `isOpenshift` method in the PGO to prevent adding the `appProtocol` field in OpenShift clusters?

Issue #3364, opened 2022-08-26
Bug(?): invalid number of restore Jobs found when attempting to reconcile a pgBackRest data source


Richard87:
Had some issues with restore jobs hanging, so I deleted the cluster related pods. Now nothing is working with this log from PGO:
```
time="2022-08-26T14:02:56Z" level=debug msg="debug flag set to true" file="cmd/postgres-operator/main.go:63" func=main.main version=5.1.0-0
I0826 14:02:57.301458       1 request.go:655] Throttling request took 1.034908398s, request: GET:https://10.56.0.1:443/apis/acme.cert-manager.io/v1beta1?timeout=32s
time="2022-08-26T14:02:58Z" level=info msg="metrics server is starting to listen" addr=":8080" file="sigs.k8s.io/controller-runtime@v0.8.3/pkg/log/deleg.go:130" func="log.(*DelegatingLogger).Info" version=5.1.0-0
time="2022-08-26T14:02:58Z" level=info msg="starting controller runtime manager and will wait for signal to exit" file="cmd/postgres-operator/main.go:84" func=main.main version=5.1.0-0
time="2022-08-26T14:02:58Z" level=info msg="upgrade checking enabled" file="cmd/postgres-operator/main.go:89" func=main.main version=5.1.0-0
time="2022-08-26T14:02:58Z" level=info msg="starting metrics server" file="sigs.k8s.io/controller-runtime@v0.8.3/pkg/manager/internal.go:385" func="manager.(*controllerManager).serveMetrics.func2" path=/metrics version=5.1.0-0
time="2022-08-26T14:02:58Z" level=info msg="Starting EventSource" file="sigs.k8s.io/controller-runtime@v0.8.3/pkg/internal/controller/controller.go:165" func="controller.(*Controller).Start.func1" reconciler group=postgres-operator.crunchydata.com reconciler kind=PostgresCluster source="kind source: /, Kind=" version=5.1.0-0
time="2022-08-26T14:02:58Z" level=info msg="Starting EventSource" file="sigs.k8s.io/controller-runtime@v0.8.3/pkg/internal/controller/controller.go:165" func="controller.(*Controller).Start.func1" reconciler group=postgres-operator.crunchydata.com reconciler kind=PostgresCluster source="kind source: /, Kind=" version=5.1.0-0
[...]

Issue #3058, opened 2022-02-24
Failed to migrate databases to pgo v5 with the same name but in different namespaces


radoslaw505:
## Overview

We have several namespaces with version 4 pgo clusters. Some of them have the same name. 
For example, let it be 2 namespaces - `sample1` and `sample2` in which a pgo cluster named `pgo-test-cluster` exists.
When I migrate the first cluster `pgo-test-cluster` in namsespace `sample1` with the instruction: [www](https://access.crunchydata.com/documentation/postgres-operator/v5/guides/v4tov5/), the pod `* -move- *` is properly created and the cluster is initialized.
When migrating second cluster `pgo-test-cluster` from namespace `sample2` the pod's * -move- * DOES not been created, which leads to an error:
2022-02-24 13:11:41,659 INFO: Running custom bootstrap script: mv "/pgdata/pg13_bootstrap" "/pgdata/pg13"
mv: cannot stat '/pgdata/pg13_bootstrap': No such file or directory
2022-02-24 13:11:41,666 INFO: removing initialize key after failed attempt to bootstrap the cluster
and pod is in crash loop.

## Environment

- Platform: OpenShift
- Platform Version: 4.8.29
- PGO Image Tag: ubi8-13.5-0
- Postgres Version: 13
- Storage: CEPH

### REPRO

`pgo delete cluster pgo-test-cluster --keep-data --keep-backups -n sample2`

```cat <<EOF | oc apply -n sample2 -f -
apiVersion: postgres-operator.crunchydata.com/v1beta1
kind: PostgresCluster
metadata:
  name: pgo-test-cluster
spec:
  dataSource:
    volumes:
      pgDataVolume:
        pvcName: pgo-test-cluster
        directory: "pgo-test-cluster"
      pgBackRestVolume:
        pvcName: pgo-test-cluster-pgbr-repo
        directory: "pgo-test-cluster-backrest-shared-repo"
  image: registry.connect.redhat.com/crunchydata/crunchy-postgres:ubi8-13.5-0
  postgresVersion: 13 
  instances:
    - name: instance1
      replicas: 1
      resources:
        requests:
          cpu: 100m  
          memory: 256Mi
        limits:
          cpu: 700m
          memory: 512Mi
      dataVolumeClaimSpec:
        accessModes:
        - "ReadWriteOnce"
        resources:
          requests:
            storage: 1Gi  
  monitoring:
    pgmonitor:
      exporter:
[...]

PR #3438, opened 2022-10-27
Added namespace limiters to all client.List() calls in pgbackrest and…

dsessler7:
… volumes files in the controller. This change was mandated by [sc-16139] and we believe the changes, particularly in `volumes.go`, should also solve the problem seen by the customer in [sc-13871]. 

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

A couple customers have seen issues when doing restores and volume moves with multiple clusters having the same name but being in different namespaces.

**Other Information**:
Issue: [sc-16139] 
Issue: [sc-13871]
Issue: CrunchyData/postgres-operator#3364
Issue: CrunchyData/postgres-operator#3058

### DIFF
--- a/internal/controller/postgrescluster/pgbackrest.go
+++ b/internal/controller/postgrescluster/pgbackrest.go
@@ -789,6 +789,7 @@ func (r *Reconciler) observeRestoreEnv(ctx context.Context,
 
 	restoreJobs := &batchv1.JobList{}
 	if err := r.Client.List(ctx, restoreJobs, &client.ListOptions{
+		Namespace:     cluster.Namespace,
 		LabelSelector: naming.PGBackRestRestoreJobSelector(cluster.GetName()),
 	}); err != nil {
 		return nil, nil, errors.WithStack(err)
@@ -836,8 +837,9 @@ func (r *Reconciler) observeRestoreEnv(ctx context.Context,
 			selector := naming.PGBackRestRestoreConfigSelector(cluster.GetName())
 			restoreConfigMaps := &corev1.ConfigMapList{}
 			if err := r.Client.List(ctx, restoreConfigMaps, &client.ListOptions{
+				Namespace:     cluster.Namespace,
 				LabelSelector: selector,
-			}, client.InNamespace(cluster.Namespace)); err != nil {
+			}); err != nil {
 				return nil, nil, errors.WithStack(err)
 			}
 			for i := range restoreConfigMaps.Items {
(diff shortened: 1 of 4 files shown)


internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -790,7 +790,7 @@ func (r *Reconciler) observeRestoreEnv(ctx context.Context,
 	restoreJobs := &batchv1.JobList{}
 	if err := r.Client.List(ctx, restoreJobs, &client.ListOptions{
 		LabelSelector: naming.PGBackRestRestoreJobSelector(cluster.GetName()),
-	}); err != nil {
+	}, client.InNamespace(cluster.GetNamespace())); err != nil {
```
cbandy, line 793: ⛏️ The other hunks in this PR look like the following. Can we do the same here?

```diff
 	if err := r.Client.List(ctx, restoreJobs, &client.ListOptions{
+		Namespace: cluster.Namespace
 		LabelSelector: naming.PGBackRestRestoreJobSelector(cluster.GetName()),
 	}); err != nil {
```

cbandy (commented):
Please include `Issue: [sc-NNN]` at the bottom of the commit message.

Does this close some GH issues, too? Please include them with `Issue: CrunchyData/postgres-operator#NNN` as well.

benjaminjb (approved):
LGTM

internal/controller/postgrescluster/pgbackrest.go
dsessler7: Yeah, so, as I said before, we are passing in list options in a lot of different ways throughout the codebase. I can go change this particular one if you want, but there will still be many other `Client.List` calls out there that don't have all of their list options neatly packaged in a `ListOptions` literal... Would you like for me to update all of them as well?

cbandy: No need to change everything. If it is currently using the struct, set the field. If it is passing the option arguments, pass `client.InNamespace()`. This stuck out because it is mixing the two styles.

dsessler7: I hear ya. I was actually just copying what was done in two other `Client.List` calls later in the `observerRestoreEnv` method when I did it that way. I'll change all three.

cbandy: 🤔 I see. What you have here matches ~L840 and ~L850. Not a blocker, but if you don't mind changing those to also use the struct "style", that's my preference.

benjaminjb (approved):

bendilley:
When might we expect this fix to be released? I think I may have run into a related problem.

BTW, am I wrong in thinking that I can use a pgbackrest backup from a non-PGO instance as a dataSource for a new PGO cluster?

PR #3445, opened 2022-11-01
Migration assistance

benjaminjb:

**Type of Changes:**
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

Migration (from Bitnami image + Helm chart) ran into problems and needed manual remediation.

This automates the manual remediation for that or similar cases

**Other Information**:
Issue: [sc-11804]
Issue: [sc-15909]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 .DS_Store
 /vendor/
 tools
-/testing/kuttl/e2e-generated/
-/testing/kuttl/e2e-generated-other/
+/testing/kuttl/e2e-generated*/
--- a/internal/postgres/config.go
+++ b/internal/postgres/config.go
@@ -38,6 +38,15 @@ const (
 		` stat -Lc '%A %4u %4g %n' "$@";` +
 		` }`
 
+	// bashRecreateDirectory is a Bash function that moves the contents of an
+	// existing directory into a newly created directory of the same name.
+	bashRecreateDirectory = `
+recreate() (
+  local tmp; tmp=$(mktemp -d -p "${1%/*}"); GLOBIGNORE='.:..'; set -x
+  chmod "$2" "${tmp}"; mv "$1"/* "${tmp}"; rmdir "$1"; mv "${tmp}" "$1"
+)
+`
+
 	// bashSafeLink is a Bash function that moves an existing file or directory
 	// and replaces it with a symbolic link.
 	bashSafeLink = `
@@ -205,7 +214,10 @@ func startupCommand(
 		// Function to log values in a basic structured format.
 		`results() { printf '::postgres-operator: %s::%s\n' "$@"; }`,
 
-		// Function to change a directory symlink while keeping the directory content.
+		// Function to change the owner of an existing directory.
+		strings.TrimSpace(bashRecreateDirectory),
+
+		// Function to change a directory symlink while keeping the directory contents.
 		strings.TrimSpace(bashSafeLink),
 
 		// Log the effective user ID and all the group IDs.
@@ -234,11 +246,27 @@ func startupCommand(
 
 		// PostgreSQL requires its directory to be writable by only itself.
 		// Pod "securityContext.fsGroup" sets g+w on directories for *some*
-		// storage providers.
+		// storage providers. Ensure the current user owns the directory, and
+		// remove group permissions.
 		// - https://www.postgresql.org/docs/current/creating-cluster.html
-		// - https://git.postgresql.org/gitweb/?p=postgresql.git;f=src/backend/utils/init/miscinit.c;hb=REL_13_0#l319
+		// - https://git.postgresql.org/gitweb/?p=postgresql.git;f=src/backend/postmaster/postmaster.c;hb=REL_10_0#l1522
+		// - https://git.postgresql.org/gitweb/?p=postgresql.git;f=src/backend/utils/init/miscinit.c;hb=REL_14_0#l349
 		// - https://issue.k8s.io/93802#issuecomment-717646167
-		`install --directory --mode=0700 "${postgres_data_directory}" ||`,
+		//
+		// When the directory does not exist, create it with the correct permissions.
+		// When the directory has the correct owner, set the correct permissions.
+		`if [[ ! -e "${postgres_data_directory}" || -O "${postgres_data_directory}" ]]; then`,
+		`install --directory --mode=0700 "${postgres_data_directory}"`,
+		//
+		// The directory exists but its owner is wrong. When it is writable,
+		// the set-group-ID bit indicates that "fsGroup" probably ran on its
+		// contents making them safe to use. In this case, we can make a new
+		// directory (owned by this user) and refill it.
+		`elif [[ -w "${postgres_data_directory}" && -g "${postgres_data_directory}" ]]; then`,
+		`recreate "${postgres_data_directory}" '0700'`,
+		//
+		// The directory exists, its owner is wrong, and it is not writable.
+		`else (halt Permissions!); fi ||`,
 		`halt "$(permissions "${postgres_data_directory}" ||:)"`,
 
 		// Create the pgBackRest log directory.
--- a/internal/postgres/config.go
+++ b/internal/postgres/config.go
@@ -26,6 +26,18 @@ import (
 )
 
 const (
+	// bashHalt is a Bash function that prints its arguments to stderr then
+	// exits with a non-zero status. It uses the exit status of the prior
+	// command if that was not zero.
+	bashHalt = `halt() { local rc=$?; >&2 echo "$@"; exit "${rc/#0/1}"; }`
+
+	// bashPermissions is a Bash function that prints the permissions of a file
+	// or directory and all its parent directories, except the root directory.
+	bashPermissions = `permissions() {` +
+		` while [[ -n "$1" ]]; do set "${1%/*}" "$@"; done; shift;` +
+		` stat -Lc '%A %4u %4g %n' "$@";` +
+		` }`
+
 	// bashSafeLink is a Bash function that moves an existing file or directory
 	// and replaces it with a symbolic link.
 	bashSafeLink = `
@@ -184,6 +196,12 @@ func startupCommand(
 	script := strings.Join([]string{
 		`declare -r expected_major_version="$1" pgwal_directory="$2" pgbrLog_directory="$3"`,
 
+		// Function to print the permissions of a file or directory and its parents.
+		bashPermissions,
+
+		// Function to print a message to stderr then exit non-zero.
+		bashHalt,
+
 		// Function to log values in a basic structured format.
 		`results() { printf '::postgres-operator: %s::%s\n' "$@"; }`,
 
@@ -198,14 +216,16 @@ func startupCommand(
 		// match the cluster spec.
 		`results 'postgres path' "$(command -v postgres)"`,
 		`results 'postgres version' "${postgres_version:=$(postgres --version)}"`,
-		`[[ "${postgres_version}" == *") ${expected_major_version}."* ]]`,
+		`[[ "${postgres_version}" =~ ") ${expected_major_version}"($|[^0-9]) ]] ||`,
+		`halt Expected PostgreSQL version "${expected_major_version}"`,
 
 		// Abort when the configured data directory is not $PGDATA.
 		// - https://www.postgresql.org/docs/current/runtime-config-file-locations.html
 		`results 'config directory' "${PGDATA:?}"`,
 		`postgres_data_directory=$([ -d "${PGDATA}" ] && postgres -C data_directory || echo "${PGDATA}")`,
 		`results 'data directory' "${postgres_data_directory}"`,
-		`[ "${postgres_data_directory}" = "${PGDATA}" ]`,
+		`[[ "${postgres_data_directory}" == "${PGDATA}" ]] ||`,
+		`halt Expected matching config and data directories`,
 
 		// Determine if the data directory has been prepared for bootstrapping the cluster
 		`bootstrap_dir="${postgres_data_directory}_bootstrap"`,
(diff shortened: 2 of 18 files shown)


internal/postgres/config.go
```diff
@@ -214,15 +244,39 @@ func startupCommand(
...
+
+		// The directory exists, its owner is wrong, and it is not writable.
+		`else (halt Permissions!); fi ||`,
+		`halt "$(permissions "${postgres_data_directory}" ||:)"`,
+
+		// Patroni will complain if there's no `postgresql.conf` file
+		// and PGDATA may be missing that file if this is a restored database
+		// where the conf file was kept elsewhere.
+		`[[ ! -f "${postgres_data_directory}/postgresql.conf" ]] &&`,
+		`touch "${postgres_data_directory}/postgresql.conf"`,
```
cbandy, line 274: ❓ I have a memory that Patroni very much wants a missing or empty PGDATA directory at bootstrap (or is it replica create?). Does this cause any trouble in those scenarios?

testing/kuttl/e2e/cluster-migrate/00--non-crunchy-cluster.yaml
```diff
@@ -0,0 +1,195 @@
...
+            failureThreshold: 6
+            initialDelaySeconds: 30
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 5
+            exec:
+              command:
+                - /bin/sh
+                - -c
+                - exec pg_isready -U "postgres" -h 127.0.0.1 -p 5432
```
cbandy, line 156: ⛏️ Can we change `127.0.0.1` to `localhost` here?

🌱 This IPv4 address (probably?) won't work in an IPv6-only environment.

Same a few lines down.

testing/kuttl/e2e/cluster-migrate/00--non-crunchy-cluster.yaml
```diff
@@ -0,0 +1,195 @@
...
+              mountPath: /bitnami/postgresql
+      volumes:
+        - name: dshm
+          emptyDir:
+            medium: Memory
+  volumeClaimTemplates:
+    - metadata:
+        name: data
+        # annotations:
+        #   volumeType: local
```
cbandy, line 189: 🤔 It's unfortunate that they didn't prefix their annotation with a domain... Perhaps it's innocuous enough to leave on all the time?

internal/postgres/config.go
benjaminjb: Ah, darn, yeah, we get an `data dir for the cluster is not empty, but system ID is invalid` error. So we only want to add this if we're bootstrapping from a preexisting datasource... or is even that too broad a generalization?

benjaminjb: (Or do we just want to document the manual fix?)

testing/kuttl/e2e/cluster-migrate/05--alter-pv-claim.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      VOLUME_NAME=$(kubectl get pvc --namespace "${NAMESPACE}" --output=jsonpath={.items..spec.volumeName})
+
+      kubectl patch pv "${VOLUME_NAME}" -p '{"spec":{"claimRef": null}}'
```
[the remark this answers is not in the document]
benjaminjb: Testing, I think this was a leftover from when I was trying something else, removing this whole step (and labeling the pv elsewhere)

testing/kuttl/e2e/cluster-migrate/00--non-crunchy-cluster.yaml
```diff
@@ -0,0 +1,195 @@
...
+                  - "default"
+                topologyKey: kubernetes.io/hostname
+              weight: 1
+      securityContext:
+        fsGroup: 1001
+      hostNetwork: false
+      hostIPC: false
+      containers:
+        - name: postgresql
+          image: docker.io/bitnami/postgresql:14.5.0-debian-11-r37
```
benjaminjb, line 109: If we want to run this in our Jenkins, with envsubst for different versions, we're going to have to figure out some way to parameterize this...

internal/postgres/config.go
cbandy: My first thought was the dataSource jobs, but perhaps we can move this somewhere below `// When the data directory is empty, there's nothing more to do.` a few lines down.

benjaminjb:
This KUTTL test only works in GKE, but we could alter it (dropping fsGroups) to get it to work in OpenShift. Easy enough to add additional make targets to our CI for GKE/Openshift, but having separate but parallel additional testing opens up some maintainability issues.

testing/kuttl/e2e-other-gke/cluster-migrate/11--delete.yaml
```diff
@@ -0,0 +1,7 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+delete:
+  - apiVersion: postgres-operator.crunchydata.com/v1beta1
+    kind: PostgresCluster
+    name: cluster-migrate
```
cbandy, line 7: 🤔 We don't usually delete. Is there a benefit to it?

testing/kuttl/e2e/cluster-migrate/00--non-crunchy-cluster.yaml
benjaminjb: I've taken this out -- when we can get a passing test with local-path-provisioner, we can revisit this.

testing/kuttl/e2e/cluster-migrate/02--create-data.yaml
```diff
@@ -0,0 +1,30 @@
...
+            - { name: PGCONNECT_TIMEOUT, value: '5' }
+            - { name: PGPASSWORD, valueFrom: { secretKeyRef: { name: non-crunchy-cluster, key: postgres-password } } }
+          command:
+            - psql
+            - --username=postgres
+            - --dbname=postgres
+            - --set=ON_ERROR_STOP=1
+            - --command
+            - |
+              CREATE TABLE IF NOT EXISTS important (data) AS VALUES ('treasure');
```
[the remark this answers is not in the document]
benjaminjb: Yeah, I believe that's the case. (The Crunchydata blog post about this change in PG 15 notes that this is for non-superusers.)

.gitignore
```diff
@@ -3,3 +3,4 @@
 tools
 /testing/kuttl/e2e-generated/
 /testing/kuttl/e2e-generated-other/
+/testing/kuttl/e2e-generated-other-gke/
```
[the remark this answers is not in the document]
benjaminjb: I'm removing `-other-gke` as a separate folder, but I'm expecting to separate out by folder in the future, so 👍

testing/kuttl/e2e-other-gke/cluster-migrate/04-errors.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: non-crunchy-cluster-0
```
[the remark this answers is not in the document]
benjaminjb: Don't quote me on this yet, but: I thought the way this would work is that each step # gets its own timeout, but it looks like if you have `00--delete` and `00-errors`, then both get the timeout length:

Delete's timeout: https://github.com/kudobuilder/kuttl/blob/63f81f8f6f40913631578b18fe3042ef26609aef/pkg/test/step.go#L155

Assert/error get run here with ctx with timeout:
https://github.com/kudobuilder/kuttl/blob/63f81f8f6f40913631578b18fe3042ef26609aef/pkg/test/utils/kubernetes.go#L1069-L1072

testing/kuttl/e2e-other-gke/cluster-migrate/07--set-collation.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      set -e
+      if [[ ${KUTTL_PG_VERSION} == 15 ]]; then
```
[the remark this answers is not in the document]
benjaminjb: A little future proofing never hurt.

(Smash cut to the future, everything on fire)

testing/kuttl/e2e-other-gke/cluster-migrate/08--alter-pv.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - script: |
+      SAVED_DATA=$(
+        kubectl get configmap persistent-volume-reclaim-policy --namespace "${NAMESPACE}" \
+          --output=jsonpath="{.data..['ORIGINAL_POLICY','VOLUME_NAME']}"
+      )
+
+      IFS=' '
+      read ORIGINAL_POLICY VOLUME_NAME <<< "${SAVED_DATA}"
+      
+      kubectl patch pv "${VOLUME_NAME}" -p '{"spec":{"persistentVolumeReclaimPolicy":"'${ORIGINAL_POLICY}'"}}'
```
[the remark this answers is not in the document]
benjaminjb: Worth adding to all uses of script and maybe adding to our best practices?

testing/kuttl/e2e-other-gke/cluster-migrate/09-assert.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: cluster-migrate
+status:
+  pgbackrest:
+    repos:
+    - bound: true
+      name: repo1
+      stanzaCreated: true
```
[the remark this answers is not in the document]
benjaminjb: This is a proxy for the PG cluster being ready (so the PV has been adopted, postgres has started up successfully, collation issues have been dealt with)

testing/kuttl/e2e-other-gke/cluster-migrate/11--delete.yaml
benjaminjb: This came up as a possible topic around our CI best practice: KUTTL will delete ns (unless set not to), but it doesn't consider deletion of namespace as part of the test, so it doesn't wait for the ns to be gone before starting the next test.

benjaminjb: That said, this delete was actually doing double-duty:
1) it's _maybe_ best practice for our CI system
2) I was deleting this cluster and making sure it was gone before manually deleting the PV. (I'm not manually deleting any more, since we switch the Reclaim policy back to what it was before.)

testing/kuttl/e2e-other-gke/cluster-migrate/04-errors.yaml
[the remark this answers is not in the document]
cbandy: > Do KUTTL error assertions wait for the entire step timeout?

Looking here, I think the answer is "no." When no `00-errors` match, `nil` bubbles up, and `"test step completed"`. KUTTL does not wait any time to find those.

📝 When we want to check that a field is not some value, we should be certain its object exists -- with an empty-ish `00-assert`, if need be.

- https://github.com/kudobuilder/kuttl/blob/63f81f8f6f40913631578b18fe3042ef26609aef/pkg/test/step.go#L438
- https://github.com/kudobuilder/kuttl/blob/63f81f8f6f40913631578b18fe3042ef26609aef/pkg/test/step.go#L385
- https://github.com/kudobuilder/kuttl/blob/63f81f8f6f40913631578b18fe3042ef26609aef/pkg/test/step.go#L310

testing/kuttl/e2e-other-gke/cluster-migrate/08--alter-pv.yaml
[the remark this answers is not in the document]
cbandy: Maybe so! I'm not opposed.

testing/kuttl/e2e-other-gke/cluster-migrate/04-errors.yaml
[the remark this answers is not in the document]
benjaminjb: Followup on this: Chris and I talked about it, because I suspect I misread his first comment: `asserts` steps will wait the entire time _until_ the condition is matched and then exit early. (This has bitten us before when we try to delete a pod to regen it, and assert on it existing -- well, the pod that's being terminated still exists, which matches the assert condition.)

Likewise, `errors` steps will exit early if the condition is negatively matched. (So if you `errors` for a pod existing, as soon as that pod doesn't exist, that satisfies the condition and exits. But if the pod did exist, we would wait the whole timeout length.)

testing/kuttl/e2e/cluster-migrate/00--non-crunchy-cluster.yaml
benjaminjb: (Solved in the Makefile for now.)

testing/kuttl/e2e-other-gke/cluster-migrate/08--alter-pv.yaml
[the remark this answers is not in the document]
benjaminjb: Will hold off making other changes for another PR

testing/kuttl/e2e-other-gke/cluster-migrate/11--delete.yaml
benjaminjb: After further discussion, I'm removing the delete for now: this will either need to be a convention we decide on (either always or never deleting), or possibly a change in KUTTL, which could be changed to wait for the ns to delete before marking the test as done.

PR #3464, opened 2022-11-17
Remove disable exporter tls test

jmckulk:
Checking that tls has been disabled on a cluster (where it was previously enabled) is difficult. This is because we need to wait for the instance pod to be redeployed without tls configuration. We are removing case from the kuttl test with plans to ensure we have the same coverage in go tests in the future.

[Issue: sc-16572]

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

**Other Information**:


cbandy:
Can/Should we also remove `12--disable-tls.yaml` ?

cbandy (approved):

PR #3469, opened 2022-11-19
Moving PG Major Upgrades API to postgres-operator repo.

dsessler7:
Some explanations:

With the `pgupgrade_types.go` moved over, the `controller-gen` generates both `pgupgrades` and `postgresclusters` yaml files. We now need to run two separate `kustomize` commands, so we need a way to differentiate between the two `kustomization.yaml` files. Creating two different directories made sense. The problem is that `kustomize` will not let you reference resources "up" the filesystem tree for security reasons. So, I couldn't keep the generated CRDs in the `./build/crd/generated` directory that they were initially in. I decided to generate the CRDs twice, placing one copy in the `./build/crd/postgresclusters/generated` directory and another in the `./build/crd/pgupgrades/generated` directory. It's a bit redundant, but it allows for keeping the API files together and keeping the `kustomize` stuff as it was.

I commented out the `[ "$$(ls -1 ./build/crd/generated)" = 'postgres-operator.crunchydata.com_postgresclusters.yaml' ]` line since the comment above it does say that it only makes sense when there is only one CRD file and we now have two CRDs.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [x] Documentation
 - [ ] Testing enhancement
 - [x] Other

[SC-16347]

**Other Information**:


docs/content/guides/major-postgres-version-upgrade.md
```diff
@@ -0,0 +1,179 @@
...
+
+Why do we need to add an annotation to the cluster if the PGUpgrade already has the cluster's name? This is another security mechanism--think of it as a two-key nuclear system: the `PGUpgrade` has to know which Postgres cluster to upgrade; and the Postgres cluster has to allow this upgrade to work on it.
+
+The annotation to add is `postgres-operator.crunchydata.com/allow-upgrade`, with the name of the `PGUpgrade` object as the value. So for our example above with a Postgres cluster named `hippo` and a `PGUpgrade` object named `hippo-upgrade`, we could annotate the cluster with the command
+
+```bash
+kubectl -n postgres-operator annotate postgrescluster hippo postgres-operator.crunchydata.com/allow-upgrade="hippo-upgrade"
+```
+
+To shutdown the cluster, edit the `spec.shutdown` field to false and reapply the spec with `kubectl`. For example, if you used the [tutorial]({{< relref "tutorial/_index.md" >}}) to [create your Postgres cluster]({{< relref "tutorial/create-cluster.md" >}}), you would run the following command:
```
cbandy, line 71: > To shutdown the cluster, edit the `spec.shutdown` field to false and reapply

❓To shutdown, the field should be `true`, right?

docs/content/guides/major-postgres-version-upgrade.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Step 5: Restart your Postgres cluster with the new version
+
+Once the upgrade process is complete, you can erase the `PGUpgrade` object, which will clean up any Jobs and Pods that were created during the upgrade. But as long as the process completed successfully, that `PGUpgrade` object will remain inert. If you find yourself needing to upgrade the cluster again, you will not be able to edit the existing `PGUpgrade` object with the new versions, but will have to create a new `PGUpgrade` object. Again, this is a safety mechanism to make sure that any PGUpgrade can only be run once.
+
+Likewise, you may remove the annotation on the Postgres cluster as part of the cleanup. While not necessary, it is recommended to leave your cluster without unnecessary annotations.
+
+To restart your newly upgraded Postgres cluster, you will have to update the `spec.postgresVersion` to the new version. You may also have to update the `spec.image` value to reflect the image you plan to use if that field is already filled in. Turn `spec.shutdown` to false, and PGO will restart your cluster:
```
cbandy, line 105: > Turn `spec.shutdown` to false, and PGO will restart your cluster

📝 This sounds right.

cbandy (approved):

PR #3473, opened 2022-11-23
Adding uniqueness to cluster names when testing service type changes …

dsessler7:
…to work around race condition that is causing these tests to flake.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Currently, our tests for reconciling different services intermittently flake, specifically when testing changing the service type. This is due to a race condition where sometimes a service from a prior test has not yet been deleted when another test runs, causing the comparison of the "before" and "after" IP addresses to be misaligned. This fix makes each cluster/service name unique so that different services' IP addresses won't ever be compared.

[sc-16571]

Tests should not flake due to this race condition anymore.

**Other Information**:

### DIFF
--- a/internal/controller/postgrescluster/patroni_test.go
+++ b/internal/controller/postgrescluster/patroni_test.go
@@ -283,10 +283,21 @@ func TestReconcilePatroniLeaderLease(t *testing.T) {
 	// CRD validation looks only at the new/incoming value of fields. Confirm
 	// that each ServiceType can change to any other ServiceType. Forbidding
 	// certain transitions requires a validating webhook.
+	serviceTypeChangeClusterCounter := 0
 	for _, beforeType := range serviceTypes {
 		for _, changeType := range serviceTypes {
 			t.Run(beforeType+"To"+changeType, func(t *testing.T) {
-				cluster := cluster.DeepCopy()
+				// Creating fresh clusters for these tests
+				cluster := testCluster()
+				cluster.Namespace = ns.Name
+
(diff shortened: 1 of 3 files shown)


internal/controller/postgrescluster/patroni_test.go
```diff
@@ -283,10 +283,21 @@ func TestReconcilePatroniLeaderLease(t *testing.T) {
...
-				cluster := cluster.DeepCopy()
+				// Creating fresh clusters for these tests
+				cluster := testCluster()
+				cluster.Namespace = ns.Name
+
+				// Note (dsessler): Adding a number to each cluster name to make cluster/service
+				// names unique to work around an intermittent race condition where a service
+				// from a prior test has not been deleted yet when the next test runs, causing
+				// the test to fail due to non-matching IP addresses.
+				cluster.Name += "-" + strconv.Itoa(serviceTypeChangeClusterCounter)
```
cbandy, line 298: 🌱 `fmt` is good for smooshing strings together. `%v` is _very_ forgiving. I use it to append sometimes:

```go
cluster.Name = fmt.Sprintf("%v-%v", cluster.Name, serviceTypeChangeClusterCounter)
// or
cluster.Name += fmt.Sprintf("-%d", serviceTypeChangeClusterCounter)
```

cbandy (approved):

Issue #3444, opened 2022-11-01
Failed pgbackrest jobs in k8


Suhogruz:
Today our k8 cluster restarted and backup schedule (once a day at 1 AM) didn't work we saw 7 failed pgbackrest jobs. Is there any way to set the number of backup jobs if the previous job failed and the interval between them ?

cbandy:
Scheduled backup jobs run using Kubernetes' [CronJob](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/cron-job-v1/) with the default settings for `failedJobsHistoryLimit` (1) and `successfulJobsHistoryLimit` (3).

What version of PGO and Kubernetes are you using?

Suhogruz:
Kubernetes 1.22, PGO 5.1.1

cbandy:
I see three successful and one failed in my test. Were you looking at pods perhaps?

```
$ kubectl get job
NAMESPACE     NAME                            COMPLETIONS   DURATION   AGE
cbandy        mycluster-repo1-full-27811770   0/1           167m       167m
cbandy        mycluster-repo1-full-27811935   1/1           30s        2m46s
cbandy        mycluster-repo1-full-27811936   1/1           33s        106s
cbandy        mycluster-repo1-full-27811937   1/1           31s        46s
```

```
$ kubectl get pod
NAMESPACE     NAME                                      READY   STATUS      RESTARTS      AGE
cbandy        mycluster-repo1-full-27811770-s8hxf       0/1     Error       0             174m
cbandy        mycluster-repo1-full-27811770-jrznk       0/1     Error       0             174m
cbandy        mycluster-repo1-full-27811770-s6dtj       0/1     Error       0             174m
cbandy        mycluster-repo1-full-27811770-97bq8       0/1     Error       0             174m
cbandy        mycluster-repo1-full-27811770-7gwws       0/1     Error       0             174m
cbandy        mycluster-repo1-full-27811770-4v8j9       0/1     Error       0             174m
cbandy        mycluster-repo1-full-27811770-mpgsn       0/1     Error       0             174m
cbandy        mycluster-repo1-full-27811942-mq5j4       0/1     Completed   0             2m40s
cbandy        mycluster-repo1-full-27811943-kc4wr       0/1     Completed   0             100s
```

cbandy:
[...]

PR #3475, opened 2022-11-25
Make the TTL of pgBackRest backups configurable

cbandy:
The default retention of one failed backup Job can leave a Job and its Pods in a failed state indefinitely. The TTL setting lets someone choose how long they want Jobs, Pods, and their logs to be available.

**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature

Scheduled backups keep one failed Job around so their Pods and logs can be reviewed by a human when no other logging infrastructure exists.

- CrunchyData/postgres-operator#3444

The [TTL of backup Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs) can be configured from the PostgresCluster spec. The minimum is one minute to avoid PGO racing with the TTL controller. This minimum may be expanded (i.e. lowered) in the future.

**Other Information**:

This field is functional in Kubernetes 1.21 and OpenShift 4.8 where the TTLAfterFinished feature gate is enabled by default.

The only [validation upstream](https://github.com/kubernetes/kubernetes/blob/v1.25.0/pkg/apis/batch/validation/validation.go#L181-L183) is the minimum.

Issue: [sc-14014]


internal/controller/postgrescluster/pgbackrest_test.go
```diff
@@ -2635,16 +2634,9 @@ volumes:
 	})
 
 	t.Run("PriorityClassName", func(t *testing.T) {
-		cluster := &v1beta1.PostgresCluster{
```
tjmoore4, line 2638: :pick: :pick: Part of me wants these refactors as a separate commit, but I should probably stifle that impulse (no change actually requested :stuck_out_tongue_closed_eyes:)

tjmoore4 (approved):

internal/controller/postgrescluster/pgbackrest_test.go
cbandy: 👍🏻 Understood. They felt small enough and related enough, but I can keep things like this separate next time. It's easier to squash during review than the reverse.

PR #3512, opened 2022-12-24
Integrate pg upgrades

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

Currently, the PGUpgrades controller runs in its own image.

[sc-16348]

With this change, a single postgres-operator binary will run/manage the controllers for both the PostgresCluster and PGUpgrade APIs.

**Other Information**:

### DIFF
--- a/internal/controller/pgupgrade/apply.go
+++ b/internal/controller/pgupgrade/apply.go
@@ -1,0 +1,164 @@
+// Copyright 2021 - 2022 Crunchy Data Solutions, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pgupgrade
+
+import (
+	"context"
+	"encoding/json"
+	"reflect"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// JSON6902 represents a JSON Patch according to RFC 6902; the same as
+// k8s.io/apimachinery/pkg/types.JSONPatchType.
+type JSON6902 []interface{}
+
+// NewJSONPatch creates a new JSON Patch according to RFC 6902; the same as
+// k8s.io/apimachinery/pkg/types.JSONPatchType.
+func NewJSONPatch() *JSON6902 { return &JSON6902{} }
+
+// escapeJSONPointer encodes '~' and '/' according to RFC 6901.
+var escapeJSONPointer = strings.NewReplacer(
+	"~", "~0",
+	"/", "~1",
+).Replace
+
+func (*JSON6902) pointer(tokens ...string) string {
+	var b strings.Builder
+
+	for _, t := range tokens {
+		_ = b.WriteByte('/')
+		_, _ = b.WriteString(escapeJSONPointer(t))
+	}
+
+	return b.String()
+}
+
+// Add appends an "add" operation to patch.
+//
+// > The "add" operation performs one of the following functions,
+// > depending upon what the target location references:
+// >
+// > o  If the target location specifies an array index, a new value is
+// >    inserted into the array at the specified index.
+// >
+// > o  If the target location specifies an object member that does not
+// >    already exist, a new member is added to the object.
+// >
+// > o  If the target location specifies an object member that does exist,
+// >    that member's value is replaced.
+//
+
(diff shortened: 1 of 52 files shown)


internal/controller/pgupgrade/apply.go
```diff
@@ -0,0 +1,164 @@
...
+	data, err := client.MergeFrom(zero.(client.Object)).Data(object)
+	apply := client.RawPatch(client.Apply.Type(), data)
+
+	// Send the apply-patch with force=true.
+	if err == nil {
+		err = r.patch(ctx, object, apply, client.ForceOwnership)
+	}
+
+	return err
+}
```
benjaminjb, line 164: This doesn't quite look like the `apply` that we have for either pgo or pgo-upgrade right now. Was there a decision made to change it?

benjaminjb (commented):
Overall looks ok, a few questions (mostly around `apply.go` right now, and also wanted to check: have you run make deploy-dev and the kuttl tests?

internal/controller/pgupgrade/apply.go
tony-landreth: It should be identical to pg-upgrade...

internal/controller/pgupgrade/apply.go
```diff
@@ -0,0 +1,164 @@
+// Copyright 2021 - 2022 Crunchy Data Solutions, Inc.
```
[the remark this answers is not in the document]
tony-landreth: Were you picturing a new package?

tony-landreth:
@benjaminjb We have run make deploy-dev and the kuttl tests.

internal/controller/pgupgrade/apply.go
dsessler7: It was copied directly from `pgo-upgrade`... the only change that was made is that the package was renamed

internal/controller/pgupgrade/apply.go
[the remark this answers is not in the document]
dsessler7: The `pgupgrade` and `postgrescluster` controllers each have their own reconciler struct upon which the `apply.go` methods (among other logic) are defined, so changing the code so that the controllers share some of that logic does not seem like a trivial code change, AFAICT. Also seems like it would require TODOs in a lot of places... :thinking:

internal/postgres/config_test.go
```diff
@@ -175,7 +175,8 @@ func TestBashRecreateDirectory(t *testing.T) {
 	cmd.Args = append(cmd.Args, "-ceu", "--",
 		bashRecreateDirectory+` recreate "$@"`, "-",
 		filepath.Join(dir, "d"), "0740")
-
+	// Set LANG=C to ensure that CombinedOutput is correctly ordered
+	cmd.Env = append(cmd.Env, "LANG=C")
```
cbandy, line 179: 🤔 I'm confident the test failed because an uninitialized `cmd.Env` is replaced with a copy of the environment of `go test`. The implementation doesn't care about the environment, only the test. We want the test to behave consistently regardless of the environment.

🤔 I see that `cmd.Env = []string{}` also causes the test to pass for me. I _think_ this is a more direct solution ("ignore the environment while testing") but I'm also finding it harder to describe in a comment. Maybe the following?

```go
	// The assertion below expects alphabetically sorted filenames.
	// Set an empty environment to always use the default/standard locale.
	cmd.Env = []string{}
```

internal/controller/pgupgrade/apply.go
benjaminjb: Nvm, I double-checked directly from the dsessler7 repo, I must've been comparing files from the wrong branches or something.

internal/controller/pgupgrade/apply.go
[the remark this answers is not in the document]
benjaminjb: Yeah, and the problem with TODOs is they are very easy to lose track of when planning sprints or thinking about priorities, so if we wanted to do this, we should make a ticket in our backlog about it.

But that's an IF, which is why I would consider throwing a TODO in here for now -- or maybe just discussing it here would be enough, so: 

when we first made the upgrade manager, this `apply.go` was largely taken from pgo; and now that we're combining the managers, we're moving this `apply.go` back into the repo where it came from. So to me, this redundancy is more bothersome than it might be from an other/objective POV. Certainly for this ticket, I think this is fine: let's make sure our upgrade works when it's part of this binary. But in the future, will this redundancy be aggravating (i.e., will it produce more work for us in the future if we have to make changes in two places?) or will this redundancy be fine? Saying it out loud, I'm sort of leaning towards doing nothing right now and waiting to see if it's a problem in the future. Any thoughts?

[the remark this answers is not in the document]
tony-landreth: @benjaminjb I lean toward the "do nothing for now" position, but I get where you're coming from. It might be clearer how to extract a layer that pgupgrade and postgrescluster rest on from an actual shotgun surgery experience.

benjaminjb (approved):

benjaminjb:
LGTM, don't forget to squash!

Issue #3194, opened 2022-05-09
Patroni dynamicConfiguration not working properly


alepefe95:
Hello guys,

I was exploring the possibility of using your new **5.1.0 postgres-operator** as the old version 4.7.5 is no longer working on a **k8s v1.22.9** due to a custom resource definition error.

One of our project's requirement is to use the **postgres 14** citus preloaded library.
*The citus library need to be the first of the directive **shared_preload_libraries** in the postgres configuration file.

In order to achieve my goal, I forked your repo [postgres-operator-examples](postgres-operator-examples) and modify you postgres.yaml resource definition.

```
apiVersion: postgres-operator.crunchydata.com/v1beta1
kind: PostgresCluster
...
spec:
....
  patroni:
    dynamicConfiguration:
      postgresql:
        # custom_conf : /opt/custom-postgresql.conf
        parameters:
          shared_preload_libraries: citus,pgaudit,pg_stat_statements,pgnodemx
          shared_buffers: 256MB
```

After applying the new configuration postgres doesnt boot as the citus extension was not the first library loaded.
The problem is that the value of 
`
shared_preload_libraries: citus,pgaudit,pg_stat_statements,pgnodemx
`

what really does is **concatenate pgaudit plus your custom value**, the result is:

`shared_preload_libraries = pgaudit, citus,pgaudit,pg_stat_statements,pgnodemx`
Thats why the postgres server boot fails.

I also tried defining a custom postgres configuration `custom_conf : /opt/custom-postgresql.conf` but It didnt work either.

Is there any way to solve this problem?

Thank you,
Alepefe

anthonator:
I'm running into this exact same situation. Need `citus` to be the first item in `shared_preload_libraries`.

@Alepefe did you have any luck with this?

Alepefe:
> Yes, but I had to create a customized docker image for the "postgres-operator"

Its a little tricky but you can:

1. Branch the operator's repository
2. Find the code that is causing the problem  (I think its a function in controller.go file)
3. Compile it
4. Create a new docker image

I can provide you with the docker image link if you want.

I'm having trouble when it comes to configuring pgbouncer's hba for citus.

When can help each other :)

anthonator:
@andrewlecuyer could you provide any color around why pgaudit is automatically added to `shared_preload_libraries`? Also, is there interest in changing this?

pere3:
[...]

PR #3530, opened 2023-01-13
Allow to control the order of shared_preload_libraries (citus)

ratranqu:
The citus extension needs to be loaded first when bringing up a cluster. The current implementation allows customising the extensions loaded, but extensions provided by the resource definition will be loaded after any internal ones.

Therefore, we updated the implementation such that the extension list order will be driven by the shared_preload_libraries list when present.

When your definition has a shared_preload_libraries list containing one of the internally loaded libraries, the extension order will change to take the order from the shared_preload_libraries list. ***This may change the behaviour of existing definitions when the shared_preload_libraries contains one of the internally loaded libraries***.

When there shared_preload_libraries list contains one or more or the internal extensions, the duplicate internal extensions will be ignored, which allows the user to fully define the order of loading of extensions.

Issue: https://github.com/CrunchyData/postgres-operator/issues/3194

### DIFF
--- a/internal/patroni/config.go
+++ b/internal/patroni/config.go
@@ -20,6 +20,8 @@ import (
 	"path"
 	"strings"
 
+	"golang.org/x/exp/slices"
+
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/yaml"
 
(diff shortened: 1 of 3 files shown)


cbandy:
🔧 This will need tests before it can merge. Perhaps here or a new dedicated one: https://github.com/CrunchyData/postgres-operator/blob/2171bfea00afb51ad7af52ff89e3173b0b9d5084/internal/patroni/config_test.go#L156

This looks like it will unblock the referenced issue, but I do want to point out that there is still no way to communicate "front of list" with this. Yes, setting `citus,pgaudit` indicates "citus before pgaudit", but that may not always resolve to the front of the list.

🤔 I also want to consider security boundaries here. One of the reasons PGO puts `pgaudit` early is to try to audit as much as possible. Prior to this PR, there is no way to disrupt the placement of `pgaudit`.

cbandy:
> that may not always resolve to the front of the list

Sorry, as written, it does. But I'm still thinking about `pgaudit` and "mandatory" and how we might want those to be "ahead" of things.

ratranqu:
@cbandy thx for the feedback. I have added two tests which cover ordering and the preventing of duplication. 

Furthermore, I realised that I misrepresented what is happening in the original PR description (will update that). Indeed, we ***only*** change the location of the internally loaded extensions ***if*** they are redefined in the shared_preload_libraries. Sorry about that, the change has been done a few weeks ago.

I think that might address part of your security concerns in so far as the default behaviour remains the same unless there is an explicit intent of changing the order of internally loaded extension.

ratranqu:
Is there anything I can help with to take this forward or is there no interest for this change at this stage?

tunatoksoz:
Also interested - looking to activate citus, like @ratranqu

tony-landreth:
@ratranqu Thank you for your PR. We have plans to provide a solution to specifically support Citus.

ratranqu:
> @ratranqu Thank you for your PR. We have plans to provide a solution to specifically support Citus.

Awesome! Any sort of timeframe? Happy to beta test if useful!

tunatoksoz:
Any guesses on the timeframe?

tunatoksoz:
@tony-landreth hey tony / crunchydata team - any guesses on timeframe? Really appreciate it.

cbandy:
We took a different approach in #3761 to specifically accommodate the `citus` library. Please let me know if that doesn't resolve things or if there are other extensions that need to be loaded a certain way. Thanks!

PR #3545, opened 2023-01-24
Stop using the k8s.io/utils module directly

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

### DIFF
--- a/internal/controller/pgupgrade/utils.go
+++ b/internal/controller/pgupgrade/utils.go
@@ -19,10 +19,10 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	"github.com/crunchydata/postgres-operator/internal/initialize"
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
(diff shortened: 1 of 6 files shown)


internal/controller/pgupgrade/utils.go
```diff
@@ -49,8 +49,8 @@ func (r *PGUpgradeReconciler) setControllerReference(
 			Kind:               "PGUpgrade",
 			Name:               owner.GetName(),
 			UID:                owner.GetUID(),
-			BlockOwnerDeletion: pointer.BoolPtr(true),
-			Controller:         pointer.BoolPtr(true),
+			BlockOwnerDeletion: initialize.Pointer(true),
```
benjaminjb, line 52: 🤔 would this complain... if it wasn't clear that we're returning a pointer to a bool?

internal/initialize/primitives.go
```diff
@@ -25,12 +25,25 @@ func ByteMap(m *map[string][]byte) {
...
+}
+
 // Int32 returns a pointer to v.
 func Int32(v int32) *int32 { return &v }
 
 // Int64 returns a pointer to v.
 func Int64(v int64) *int64 { return &v }
 
+// Pointer returns a pointer to v.
+func Pointer[T any](v T) *T { return &v }
```
benjaminjb, line 45: 🤔 Def not suggesting this now, but any reason to keep around non-generic funcs that do this same thing?

benjaminjb (approved):
Just some reminders for myself to go read up on generics

internal/initialize/primitives.go
cbandy: The functions for numeric types might be convenient with untyped constants; `initialize.Int32(0)` versus `initialize.Pointer(int32(0))`. The latter, "initialize pointer int32," sounds more clear to me at this moment.

https://go.dev/ref/spec#Constants
> The default type of an untyped constant is `bool`, `rune`, `int`, `float64`, `complex128` or `string` respectively, depending on whether it is a boolean, rune, integer, floating-point, complex, or string constant.

internal/controller/pgupgrade/utils.go
cbandy: Yes. The types in the struct are... established. When I change `true` to `9`, `go build` and `go test` tell me:

```
internal/controller/pgupgrade/utils.go:52:24: cannot use initialize.Pointer(9) (value of type *int) as type *bool in struct literal
```

Issue #3477, opened 2022-11-29
Not honoring hugepages setting during initdb causes DB crash


howels:
## Overview
On an OS where hugepages are enabled, if no hugepages resources are assigned in Kubernetes and the postgres instance is set to `hugepages = off` in the config then one would assume that the DB would not use hugepages.
However, because the initdb process uses `postgresql.conf.sample` or `postgresql.conf.template` instead of the actual specified configuration the applied setting is actually `hugepages = try` during initdb.
In these case the initdb phase will attempt to allocate hugepages which are available in the OS, but it will be denied access by Kubernetes and fail.

## Environment
- Environment: Rancher RKE2
- PGO: 5.2
- RKE2: 1.24/1.25 

## Steps to Reproduce
Run a DB with no hugepages resources assigned on a OS with hugepages and with `hugepages = off` set in postgres conf.

### Expected
When hugepages are disabled they should not be used, during initdb or otherwise. The inability to specify hugepages forces all DBs to use hugepages on any system where they are enabled, even if the hugepages are intended for other applications and not available to the DB Pods.

### Actual
Postgres crashes during initdb after speculatively attempting to access hugepages and being denied by Kubernetes policies.

Additional Information
Issue has persisted for some time and the implementation goes against the spirit of the configuration options provided. Exceedingly awkward workaround such as attempting to mount configmaps over `postgresql.conf.sample` are not suitable for ongoing long-term usage.

howels:
Note that we are aware of https://github.com/CrunchyData/postgres-operator/issues/3039 but this is a workaround not a solution. HugePages should only be used by apps which explicity require hugepages.

howels:
Bug also mentioned in https://github.com/CrunchyData/postgres-operator/issues/3421

David-Angel:
The workaround to enable hugepages isn't going to work when you are required to disable hugepages.
We have a requirement not to use hugepages which can't be disabled without altering crunchydata code.

This file needs to change to turn it off.
./usr/pgsql-14/share/postgresql.conf.sample

initdb uses that file instead of the standard config file.
It would be very nice if crunchydata exposed a way of doing this through the "postgrescluster".

howels:
Can we please get a response on this?

David-Angel:
Here is a possible fix:
https://github.com/CrunchyData/crunchy-containers/compare/master...David-Angel:crunchy-containers:patch-1

[...]

PR #3555, opened 2023-01-30
Update config.go

David-Angel:
Another solution to the huge_pages issue here:
https://github.com/CrunchyData/postgres-operator/issues/3477

A better solution would be to use what gets passed in for huge_pages and turn it off by default.

**Checklist:**

 
 - [N/A] Have you added automated tests?  

**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

https://github.com/CrunchyData/postgres-operator/issues/3477

It should not break anything.

**Other Information**:

### DIFF
--- a/internal/postgres/config.go
+++ b/internal/postgres/config.go
@@ -322,6 +322,21 @@ func startupCommand(
 		// - https://git.postgresql.org/gitweb/?p=postgresql.git;f=src/backend/access/transam/xlog.c;hb=REL_12_0#l5318
 		// TODO(cbandy): Remove this after 5.0 is EOL.
 		`rm -f "${postgres_data_directory}/recovery.signal"`,
+
+		// # # # # # # # # # # # # # # # # # # # # # # # # #
+		// # The "initdb" tool uses settings defined in the shared "postgresql.conf.sample" file preventing "huge_pages" from getting configured properly.
+		// # Remove the "huge_pages" configuration from "postgresql.conf.sample" as it may or may not exist.
+		// # Add "huge_pages" to the bottom of the page turned off.
+		// # This will not prevent "huge_pages" from being passed in and configured as it only changes what "initdb" is doing.
+		`echo "initdb fix for huge_pages."`,
+		`CSAMPLE="$(find /usr/ -type d -name 'pgsql-*')/share/postgresql.conf.sample"`,
+		`echo "$CSAMPLE"`,
+		`echo "before..."`,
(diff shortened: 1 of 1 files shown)


andrewlecuyer:
@David-Angel has this been tested successfully?

More specifically, I'd expect you to run into issues modifying the sample conf file due read-only root file system being enabled for the `securityContext`.

For instance, simply running the additional script you provided within a running PG instance Pod results in a read-only file system error:

```bash
$ kubectl exec -it hippo-instance1-kvr7-0 -c database -- bash
bash-4.4$ (
> echo "initdb fix for huge_pages."
> CSAMPLE="$(find /usr/ -type d -name 'pgsql-*')/share/postgresql.conf.sample"
> echo "$CSAMPLE"
> echo "before..."
> cat "$CSAMPLE" | grep huge
> awk -i inplace '($1 ~ /huge_pages/ || $1 ~ /#huge_pages/ || $2 ~ /huge_pagescat $CSAMPLE | grep huge/ || $2 ~ /#huge_pages/) { next } { print } END { print "huge_pages = off" }' ${CSAMPLE}
> echo "after..."
> cat "$CSAMPLE" | grep huge
> )
initdb fix for huge_pages.
/usr/pgsql-14/share/postgresql.conf.sample
before...
#huge_pages = try                       # on, off, or try
#huge_page_size = 0                     # zero for system default
awk: inplace:47: fatal: inplace_begin: mkstemp(`/usr/pgsql-14/share/postgresql.conf.sample.gawk.wtsq08') failed (Read-only file system)
after...
#huge_pages = try                       # on, off, or try
#huge_page_size = 0                     # zero for system default
```

David-Angel:
> o run into issues modifying the sample c


@andrewlecuyer 
I will need to open a new PR.
I've been testing this for 2 days and yes it will need to be altered.
Security for the init container will have to be adjusted to get this to function.

David-Angel:
A better solution is here:
https://github.com/CrunchyData/crunchy-containers/pull/1502

David-Angel:
This solution did not work as the change made in the init container did not persist in the database container.
I was able to modify the files by altering the security settings but the changes were lost when the database container started.
I resolved this by replacing the file with a custom docker build.

This is still the best solution if approved by the Crunchydata team.
https://github.com/CrunchyData/crunchy-containers/pull/1502

PR #3569, opened 2023-02-09
Add release-prep make target

tony-landreth:
Adds a make target, templates, script and yaml to consolidate image tag and version updates to source code, during the release process.

Issue: [sc-17508]

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

This PR introduces a script, some yaml, templates, and a make target for updating versions and image tags in source code and doc files. It also makes a minor correction to docs/config.toml.

### DIFF
--- a/hack/release-prep.yaml
+++ b/hack/release-prep.yaml
@@ -1,0 +1,154 @@
+# Instructions on how to use this file:
+#
+# This is a double-curly brace, "handlebars", style of templating file. When you
+# want to update a release tag, set the values in `release_tags:` below and they'll
+# be used by hack/update-release-parameters.sh to replace handlebars in template
+# files located in hack/release-templates.
+#
+# To add new handlebars, you can reuse the values in release_tags and update
+# the templates, or you can create new release_tags. In either case, you'll
+# need to list the release_tag key in the replacements array for the respective
+# update.
+#
+# For example, after adding a new handlebar to hack/release-templates/github-ISSUE_TEMPLATE-bug-report.md
+# e.g. blah_version:
+#  updates:
+#  - targetFile: .github/ISSUE_TEMPLATE/bug_report.md
+#    template: hack/release-templates/github-ISSUE_TEMPLATE-bug-report.md
+#    replacements:
+#    - ubi8_pgo_latest_image_tag
+#    - blah_version
+#
+# Note that the targetFile refers to the "real" file and "template" refers to
+# the document where you'll place handlebars to make new replacements possible.
+
+---
+release_tags:
+  # Versions
+  pgmonitor_version: v4.8.0
+  pgo_latest_version: 5.3.0
+  postgres_15_version: 15.1
+  postgres_14_version: 14.6
+  postgres_13_version: 13.9
+  postgres_12_version: 12.13
+  postgres_11_version: 11.18
+  postgres_exporter_version: 0.10.1
+  k8s_version_range: 1.22-1.25
+  oc_version_range: 4.8-4.11
+
+  # Image tags
+  ubi8_pgbackrest_tag: ubi8-2.41-2
+  ubi8_pgbouncer_tag: ubi8-1.17-5
+  ubi8_pgbouncer_component_tag: ubi8-1.17-1
+  ubi8_pgadmin4_tag: ubi8-4.30-8
+  ubi8_pgupgrade_tag: ubi8-5.3.0-0
+  ubi8_postgres_exporter_tag: ubi8-5.3.0-0
+  ubi8_postgres_15_tag: ubi8-15.1-0
+  ubi8_postgres_14_tag: ubi8-14.6-2
+  ubi8_postgres_14_component_tag: ubi8-14.6-3.1-2
+  ubi8_postgres_13_tag: ubi8-13.9-2
+  ubi8_postgres_gis_14_32_tag: ubi8-14.6-3.2-5.3.0-0
+  ubi8_postgres_gis_14_31_tag: ubi8-14.6-3.1-5.3.0-0
+  ubi8_postgres_gis_component_15_33_tag: ubi8-15.1-3.3-0
+  ubi8_postgres_gis_component_14_33_tag: ubi8-14.6-3.3-2
+  ubi8_postgres_gis_component_14_32_tag: ubi8-14.6-3.2-2
+  ubi8_postgres_gis_component_14_31_tag: ubi8-14.6-3.1-2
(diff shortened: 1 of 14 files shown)


hack/update-release-parameters.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
```
cbandy, line 2: ♻️ I recommend using `set -e` to bail when any single command fails. If it is okay for something to fail, `|| true` or '|| :` allows it.

♻️ I recommend using `set -u` to bail when an unset variable is used. For example, if someone forgets to pass an argument to the script it will exit non-zero with a message about needing `$1`.

```suggestion
#!/usr/bin/env bash
set -eu

```

hack/update-release-parameters.sh
```diff
@@ -0,0 +1,31 @@
...
+# Accept a file path for param yaml.
+releaseParams=$1
+
+# Get the paths to all files that can be automatically updated for a release.
+readarray updates < <(yq e -o=j -I=0 '.updates[]' $releaseParams)
```
cbandy, line 7: ♻️ With `set -e`, it is good to break pipelines into separate commands. With the way this is arranged now, the script will continue if `yq` fails.

Below, the first line will bail when `yq` does. The `<<<` redirection passes a single word to command's stdin.

```suggestion
updates=$(yq e -o=j -I=0 '.updates[]' $releaseParams)
readarray updates <<< "$updates"
```

hack/update-release-parameters.sh
```diff
@@ -0,0 +1,31 @@
...
+
+for update in "${updates[@]}"; do
+  targetFile=$(echo $update | yq -e '.targetFile')
+  templatePath=$(echo $update | yq -e '.template')
```
cbandy, line 11: ⛏️ These could use `<<<` redirection.

```bash
yq -e '.targetFile' <<< "$update" # or
<<< "$update" yq -e '.targetFile' # or
yq <<< "$update" -e '.targetFile'
```

hack/update-release-parameters.sh
```diff
@@ -0,0 +1,31 @@
...
+  echo -e 'Regenerating...' $targetFile
+  draft=$(cat "$templatePath")
+
+  readarray -d " " replacements < <(echo $(echo $update | yq e '.replacements[]'))
```
cbandy, line 15: 🤔 Why two `echo`s in a row? Maybe split into two commands again?

```bash
replacements=$(yq ...)
readarray -d " " replacements <<< "$replacements"
```

hack/update-release-parameters.sh
```diff
@@ -0,0 +1,31 @@
...
+for update in "${updates[@]}"; do
+  targetFile=$(echo $update | yq -e '.targetFile')
+  templatePath=$(echo $update | yq -e '.template')
+  echo -e 'Regenerating...' $targetFile
+  draft=$(cat "$templatePath")
+
+  readarray -d " " replacements < <(echo $(echo $update | yq e '.replacements[]')) 
+  for find in "${replacements[@]}"; do
+    find=$(echo $find | tr -d '\n')
+    replace=$(cat $releaseParams | yq -e ".release-tags.${find}")
```
cbandy, line 18: ♻️ In general, `cat one-file | ...` can be `< one-file ...`.

```suggestion
    replace=$(< "$releaseParams" yq -e ".release-tags.${find}")
```

hack/update-release-parameters.sh
```diff
@@ -0,0 +1,31 @@
...
+    draft=$(echo "$draft" | sed -e "s|{{$find}}|$replace|")
```
cbandy, line 19: 📝 Could be alright, but noting that replacing in a loop can... unnest placeholders. e.g. `{{sock}} -> {{shoe}} -> boot`.

🔧 Should use a `g` modifier here to handle lines with multiple placeholders.

```
$ echo '{{sock}} {{sock}}' | sed -e 's|{{sock}}|shoe|'
shoe {{sock}}

$ echo '{{sock}} {{sock}}' | sed -e 's|{{sock}}|shoe|g'
shoe shoe
```

```suggestion
    draft=$(echo "$draft" | sed -e "s|{{$find}}|$replace|g")
```

cbandy:
🌱 Not everything [`shellcheck`](https://www.shellcheck.net/) recommends is _necessary_, but following its advice is usually good. I recommend seeing if it has any recommendations when authoring shell scripts.

hack/release-prep.yaml
```diff
@@ -0,0 +1,139 @@
...
+
+updates:
+- targetFile: .github/ISSUE_TEMPLATE/bug_report.md
+  template: hack/release-templates/github-ISSUE_TEMPLATE-bug-report.md
+  replacements:
+  - ubi8-pgo-latest-image-tag
+
+- targetFile: .github/ISSUE_TEMPLATE/feature_request.md
+  template: hack/release-templates/github-ISSUE_TEMPLATE-feature-request.md
+  replacements:
```
tjmoore4, line 53: :+1: I like that you made the values being replaced explicit for each file rather than a search across every file.

:question: That said, could we add more guidance on how to add new values? If I understand correctly, it would amount to adding the new variable (with the `{{value}}` format) in the file(s), then adding the new `release-tags` value along with updating an existing (or adding a new) entry under `updates`. I think either more information in the comment or possible a short README would suffice.

hack/update-release-parameters.sh
```diff
@@ -0,0 +1,28 @@
...
+  readarray replacements <<< "$replacements"
+  for find in "${replacements[@]//\"/}"; do
+    find=$(tr -d '\n' <<< "$find")
+    replace=$(< "$releaseParams" yq -e ".release-tags.${find}")
+    draft=$(sed -e "s|{{$find}}|$replace|g" <<< "$draft")
+  done
+  echo "$draft" > "$targetFile"
+done
+
+make generate-crd
```
tony-landreth, line 28: @cbandy @tjmoore4 This works, but is this a bad idea?

tjmoore4 (commented):
Looks good to me. One suggestion for some usage guidance.

hack/update-release-parameters.sh
tjmoore4: Not that I can think of. I like the idea of the CRD being updated at the same time as everything else.

tjmoore4 (approved):

PR #3575, opened 2023-02-14
Tablespace Implementation (alpha)

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Tablespaces were not implemented by the operator

Implements and documents an alpha version of tablespaces, behind a feature flag

**Other Information**:
Issue: [sc-17759]


internal/controller/postgrescluster/postgres.go
```diff
@@ -561,6 +561,80 @@ func (r *Reconciler) reconcilePostgresDataVolume(
 	return pvc, err
 }
 
+// TODO think about multiple tablespace volumes
```
benjaminjb, line 564: I've done that, so removing this in the next commit

internal/postgres/config.go
```diff
@@ -201,6 +205,30 @@ func startupCommand(
 	version := fmt.Sprint(cluster.Spec.PostgresVersion)
 	walDir := WALDirectory(cluster, instance)
 
+	// If the user requests tablespaces, we want to create and chmod the directories
```
benjaminjb, line 208: That's not quite right, updating

config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
```diff
@@ -9456,6 +9456,201 @@ spec:
                               type: object
                           type: object
                       type: object
+                    tablespaceVolumes:
+                      description: The list of tablespaces, e.g., volumees to mount
```
tony-landreth, line 9460:
```suggestion
                      description: The list of tablespaces, e.g., volumes to mount
```

benjaminjb: Done

docs/content/guides/tablespaces.md
```diff
@@ -0,0 +1,294 @@
...
+performance on a particular Postgres object (a table, index, etc.). Some examples
+of use cases for tablespaces include:
+
+- Partitioning larger data sets across different volumes
+- Putting data onto archival systems
+- Utilizing faster/more performant hardware (or a storage class) for a particular database
+- Storing sensitive data on a volume that supports transparent data-encryption
+(TDE)
+
+and others.
```
tony-landreth, line 26:
```suggestion
- Storing sensitive data on a volume that supports transparent data-encryption (TDE)

and others.
```

docs/content/guides/tablespaces.md
```diff
@@ -0,0 +1,294 @@
...
+```
+
+So to create a tablespace called `books` in the new `books` volume, your command might look like
+
+```
+CREATE TABLESPACE books LOCATION '/tablespaces/books/data';
+```
+
+To break that path down: `tablespaces` is the mount point for all tablespace volumes; `books`
+is the name of the volume in the spec; and `data` is directory created with the appropriate
```
tony-landreth, line 161:
```suggestion
is the name of the volume in the spec; and `data` is a directory created with the appropriate
```

internal/naming/names.go
```diff
@@ -325,6 +325,19 @@ func InstancePostgresDataVolume(instance *appsv1.StatefulSet) metav1.ObjectMeta
 	}
 }
 
+// TODO(benjaminjb): We need to consider shortening the PVC name
+// or at least need to document so that people can avoid too-long names
+// InstanceTablesapceDataVolume returns the ObjectMeta for the tablespace data
+// volume for instance.
+func InstanceTablesapceDataVolume(instance *appsv1.StatefulSet, tablespaceName string) metav1.ObjectMeta {
```
tony-landreth, line 332:
```suggestion
// InstanceTablespaceDataVolume returns the ObjectMeta for the tablespace data
// volume for instance.
func InstanceTablespaceDataVolume(instance *appsv1.StatefulSet, tablespaceName string) metav1.ObjectMeta {
```

internal/controller/postgrescluster/instance.go
```diff
@@ -1086,11 +1087,14 @@ func (r *Reconciler) reconcileInstance(
 	if err == nil {
 		postgresWALVolume, err = r.reconcilePostgresWALVolume(ctx, cluster, spec, instance, observed, clusterVolumes)
 	}
+	if err == nil {
+		tablespaceVolumes, err = r.reconcileTablespaceVolumes(ctx, cluster, spec, instance, clusterVolumes)
+	}
 	if err == nil {
 		postgres.InstancePod(
```
tony-landreth, line 1094: :question: Should this function start taking a parameter struct to bring the number of arguments down?

internal/postgres/config.go
```diff
@@ -201,6 +205,46 @@ func startupCommand(
...
+		// - https://www.postgresql.org/docs/current/manage-ag-tablespaces.html
+		// However, unlike the PGDATA directory, Postgres will _not_ error out
+		// if the permissions are wrong on the tablespace directory.
+		// Instead, when a tablespace is created in Postgres, Postgres will
+		// chmod the tablespace directory to match permissions on the PGDATA directory,
+		// i.e., 700 or 750
+		// Postgres setting the directory permissions:
+		// - https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/backend/commands/tablespace.c;h=5411638696b4d1436a56f62e9cde7ec61797377b;hb=REL_14_0#l604
+		// Postgres setting directory permissions:
+		// - https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/common/file_perm.c;h=ab85765a0bfb5108265819a4c7e9ff87dc6add10;hb=REL_14_0#l34
```
cbandy, line 227: ⛏️ GitWeb URLs are long, but parts are redundant and can be omitted for brevity.

🤔 The two links have the same English above them but do two different things. Perhaps the 2nd explanation is unintentional?

```suggestion
		// Postgres setting the tablespace directory permissions:
		// - https://git.postgresql.org/gitweb/?p=postgresql.git;f=src/backend/commands/tablespace.c;hb=REL_14_0#l600
		// Postgres choosing directory permissions:
		// - https://git.postgresql.org/gitweb/?p=postgresql.git;f=src/common/file_perm.c;hb=REL_14_0#l27
```

benjaminjb: Heh, the first "setting" means "this is where PG sets the dir permissions"; the second "setting" is "this is where PG sets the internal variable it uses" -- I could've found another way to say that...

docs/content/guides/tablespaces.md
```diff
@@ -0,0 +1,307 @@
...
+- Storing sensitive data on a volume that supports transparent data-encryption (TDE)
+
+and others.
+
+In order to use Postgres tablespaces properly in a highly-available,
+distributed system, there are several considerations to ensure proper operations:
+
+- Each tablespace must have its own volume; this means that every tablespace for
+every replica in a system must have its own volume;
+- The filesystem map must be consistent across the cluster;
```
tony-landreth, line 32: :question: How does one check this?

benjaminjb: Reading that now, it seems unclear and I dislike `cluster` there; what I mean here is that every postgres pod that is part of the postgres cluster needs to have the same paths, e.g., if you start a postgres (leader) and create a tablespace at `/tablespaces/user/data` that command is going to be replayed on any replicas you have, so that path has to be valid for every pod.

.github/workflows/test.yaml
```diff
@@ -130,6 +130,7 @@ jobs:
             --env 'RELATED_IMAGE_POSTGRES_13_GIS_3.1=registry.developers.crunchydata.com/crunchydata/crunchy-postgres-gis:ubi8-13.9-3.1-2' \
             --env 'RELATED_IMAGE_POSTGRES_14=registry.developers.crunchydata.com/crunchydata/crunchy-postgres:ubi8-14.6-2' \
             --env 'RELATED_IMAGE_POSTGRES_14_GIS_3.1=registry.developers.crunchydata.com/crunchydata/crunchy-postgres-gis:ubi8-14.6-3.1-2' \
+            --env 'PGO_FEATURE_GATES=TablespaceVolumes=true' \
```
tony-landreth, line 133: :question: Should this be enabled in `deploy-dev`?

tony-landreth (approved):
Looks good! :+1:

internal/naming/names.go
```diff
@@ -325,6 +325,19 @@ func InstancePostgresDataVolume(instance *appsv1.StatefulSet) metav1.ObjectMeta
 	}
 }
 
+// TODO(benjaminjb): We need to consider shortening the PVC name
+// or at least need to document so that people can avoid too-long names
```
cbandy, line 329: 🤔 What is too long? I think this is saying 253 characters? https://github.com/CrunchyData/postgres-operator/blob/8b14582aaa9df139b8eb38fe48a47f546db61f02/internal/naming/limitations.md?plain=1#L89

internal/pgbackrest/reconcile.go
```diff
@@ -302,6 +303,14 @@ func addServerContainerAndVolume(
 		postgres.DataVolumeMount().Name: postgres.DataVolumeMount(),
 		postgres.WALVolumeMount().Name:  postgres.WALVolumeMount(),
 	}
+	if util.DefaultMutableFeatureGate.Enabled(util.TablespaceVolumes) {
+		for _, instance := range cluster.Spec.InstanceSets {
+			for _, vol := range instance.TablespaceVolumes {
+				tablespaceVolumeMount := postgres.TablespaceVolumeMount(vol.Name)
+				postgresMounts[postgres.TablespaceVolumeMount(vol.Name).Name] = tablespaceVolumeMount
```
cbandy, line 310: 🤔 Am I reading this correctly? Can one call be reduced?
```suggestion
				tablespaceVolumeMount := postgres.TablespaceVolumeMount(vol.Name)
				postgresMounts[tablespaceVolumeMount.Name] = tablespaceVolumeMount
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -478,6 +478,23 @@ type PostgresInstanceSetSpec struct {
...
+
+	// The list of tablespaces volumes to mount for this postgrescluster
+	// This function is currently in alpha as of PGO v5.4
+	// +optional
+	TablespaceVolumes []TablespaceVolume `json:"tablespaceVolumes,omitempty"`
+}
+
+type TablespaceVolume struct {
+	// The name for the tablespace, used as the path name for the volume.
+	// Must be unique in the instance set since they become the directory names.
```
cbandy, line 490: 🔧 Kubernetes can enforce uniqueness for us. Add `listType=map` and `listMapKey=name` on the `TablespaceVolumes` field.

https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy

https://github.com/CrunchyData/postgres-operator/blob/8b14582aaa9df139b8eb38fe48a47f546db61f02/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go#L97-L98

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -478,6 +478,23 @@ type PostgresInstanceSetSpec struct {
...
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
```
cbandy, line 492: 🤔 This value goes multiple places and should probably have validation to ensure those succeed.

- This value goes into a PVC name. I recall [PostgresUserSpec](https://github.com/CrunchyData/postgres-operator/blob/8b14582aaa9df139b8eb38fe48a47f546db61f02/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgres_types.go#L41) did similar considering Secret names.

- This value goes into a Volume name. [Docs for that](https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/volume/#Volume) say "DNS label," but K8s defines more than one of those. If you track down the exact validation, I would be grateful to see it explained in `internal/naming/limitations.md`. 🙇🏻 

- This value goes into a directory name, so no slashes? I don't know what else to consider here.
  🤔 `fsck` `lost+found`...?

(Other places?)

internal/postgres/reconcile.go
```diff
@@ -38,6 +38,11 @@ func DataVolumeMount() corev1.VolumeMount {
 	return corev1.VolumeMount{Name: "postgres-data", MountPath: dataMountPath}
 }
 
+// TablespaceVolumeMount returns the name and mount path of the PostgreSQL tablespace data volume.
+func TablespaceVolumeMount(tablespaceName string) corev1.VolumeMount {
+	return corev1.VolumeMount{Name: tablespaceName, MountPath: tablespaceMountPath + "/" + tablespaceName}
```
cbandy, line 43: ❓ Do we need something here to ensure a name like `postgres-data` doesn't confuse everything? A prefix, maybe?

.github/workflows/test.yaml
benjaminjb: Hmmm, I guess so because if you run `deploy-dev` for KUTTL testing (as I do), you'd want that on...

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -1116,6 +1118,23 @@ func (r *Reconciler) reconcileRestoreJob(ctx context.Context,
 		volumeMounts = append(volumeMounts, walVolumeMount)
 	}
 
+	if util.DefaultMutableFeatureGate.Enabled(util.TablespaceVolumes) {
+		for _, pgtablespaceVolume := range pgtablespaceVolumes {
+			tablespaceVolumeMount := postgres.TablespaceVolumeMount(
+				pgtablespaceVolume.Labels["postgres-operator.crunchydata.com/data"])
```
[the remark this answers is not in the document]
benjaminjb: Found and changed a few instances.

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -1509,8 +1528,16 @@ func (r *Reconciler) reconcilePostgresClusterDataSource(ctx context.Context,
 		return errors.WithStack(err)
 	}
 
+	pgtablespaces := []*corev1.PersistentVolumeClaim{}
+	if util.DefaultMutableFeatureGate.Enabled(util.TablespaceVolumes) {
```
benjaminjb, line 1532: @cbandy, what do you feel about removing these gates and leaving `reconcileTablespaceVolumes` as the... gatekeeper?

internal/controller/postgrescluster/postgres.go
```diff
@@ -561,6 +561,78 @@ func (r *Reconciler) reconcilePostgresDataVolume(
 	return pvc, err
 }
 
+// +kubebuilder:rbac:groups="",resources=persistentvolumeclaims,verbs=create;patch
```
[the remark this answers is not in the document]
benjaminjb: Worth changing these in the few places it exists?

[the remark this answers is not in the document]
benjaminjb: Actually we have a bunch of annotations that follow the first syntax, maybe it's worth a whole separate PR to bring it into line?

internal/naming/names.go
benjaminjb: Oh, yeah, I documented this (that it wouldn't likely be a problem because of the 253 character length), removing that TODO now

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
benjaminjb: Oof, good thinking. How do you feel about `^[a-zA-Z0-9]+$`? Like, what's the upshot to being permissive?

internal/postgres/reconcile.go
benjaminjb: Hmmm, [expletive]. If you try to create a postgrescluster with tablespaces named the same as our default mounts, the cluster gets created, and only in the pgo logs do you see the error. But we know users don't look there.

OK, how do you feel about the validation that makes sure that no `-` is used since all of our set names have that form? Ugh, until they don't...

internal/controller/postgrescluster/postgres.go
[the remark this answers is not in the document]
cbandy: Agreed. No need to touch lots of stuff in this PR.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
cbandy: > How do you feel about `^[a-zA-Z0-9]+$`?

["DNS labels"](https://github.com/CrunchyData/postgres-operator/blob/master/internal/naming/limitations.md#dns-label) need to be lowercase, and some need to start with a letter. Could `^[a-z][a-z0-9]*$` to cover both?

I also had yet to consider empty string. Good catch. 🤔 I don't know if "min length" validation has a better message or is overkill or what.

> Like, what's the upshot to being permissive?

Dunno. I do know we can always go from less-permissive now to more-permissive later. (We should not do the reverse.) That is, objects stored in the API (that are/were valid) should/must continue to be valid in the future.

internal/postgres/reconcile.go
benjaminjb: OK, I've had lunch and now I'm adding a prefix, which is the easy fix you suggested.

internal/controller/postgrescluster/instance.go
benjaminjb: I'm not feeling that as a necessity at this moment, but would definitely consider it in the future

.github/workflows/test.yaml
benjaminjb: Added that

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -1599,6 +1626,14 @@ func (r *Reconciler) reconcileCloudBasedDataSource(ctx context.Context,
 		return errors.WithStack(err)
 	}
 
+	// TODO(benjaminjb): do we really need this?
```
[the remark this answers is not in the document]
benjaminjb: I updated the code comment to be clear that I want to test cloudbased datasources more.

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -1599,6 +1626,14 @@ func (r *Reconciler) reconcileCloudBasedDataSource(ctx context.Context,
...
+	pgtablespaces := []*corev1.PersistentVolumeClaim{}
+	if util.DefaultMutableFeatureGate.Enabled(util.TablespaceVolumes) {
```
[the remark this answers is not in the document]
benjaminjb: Changed all instances where we're using `reconcileTablespaceVolumes` to not double-check this gate.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
benjaminjb: Added

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
benjaminjb: OK, updated: alphanumeric, lowercase only, alphabet to start. And added minLength for that clearer error message.

PR #3605, opened 2023-03-13
Add Trivy to REL_4_7

tony-landreth:
Issue: [sc-17407]

### DIFF
--- a/.github/workflows/trivy-scan.yaml
+++ b/.github/workflows/trivy-scan.yaml
@@ -1,0 +1,55 @@
+# Uses Trivy to scan every pull request, rejecting those with severe, fixable vulnerabilities.
+name: Trivy
+
+on:
+  pull_request:
+    branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/trivy-scan.yaml
```diff
@@ -0,0 +1,53 @@
...
+      # This report should match the uploaded code scan report below
+      # and is a convenience/redundant effort for those who prefer to
+      # read logs and/or if anything goes wrong with the upload.
+      - name: Log all detected vulnerabilities
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: fs
+          hide-progress: true
+          ignore-unfixed: true
+          skip-dirs: 'testing'
```
benjaminjb, line 35: ⛏️ Might want to add a comment here since this diverges from some of the other uses.

benjaminjb (approved):

Issue #3623, opened 2023-04-04
PGUpgrade does not remove wal files from old cluster


stemid:
## Overview

The pgupgrade process does not delete the old wal directory, meaning it "wastes" space on your volume unless you delete it yourself.

## Environment

Please provide the following details:

- Platform: Kubernetes
- Platform Version: 1.21.14
- PGO Image Tag: ubi8-5.3.0-0
- Postgres Version 14
- Storage: rook-ceph-block

### REPRO

Provide steps to get to the error condition:

1. Follow the docs for a postgres major version upgrade
2. run the commands and scripts at the end of the pg_upgrade output.

### EXPECTED

1. All old dirs should be deleted, pg13 and pg13_wal.

### ACTUAL

1. Only pg13 is deleted, pg13_wal is left.

dsessler7:
Hello @stemid,

This is actually by design. Our major upgrade procedure will delete the `/pgdata/pg##` and `/pgdata/pg##_wal` directories for the old version from the replicas; however, we leave them in place on the primary. We don't want to automatically delete them in case something goes wrong with the upgrade.

The `delete_old_cluster.sh` script referenced in our documentation is actually created by `pg_upgrade` and yes, it only deletes the old `pg##` directory, not the wal directory. All this script is doing is running a `rm` command on the directory. If you want to delete the old `pg##_wal` directory as well, you can manually run the command on that directory.

PR #3630, opened 2023-04-13
document that wal files are not deleted

stemid:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [x] Documentation
 - [ ] Testing enhancement
 - [ ] Other

As I mentioned in issue #3623 the old wal files are not cleaned up on the primary instance after a major version pgupgrade. This is not mentioned in docs and can be a surprise if you're wondering what is taking up more space than normal.

Just mention it in the docs so people can delete it manually when they're happy with the upgrade.

**Other Information**:

### DIFF
--- a/docs/content/guides/major-postgres-version-upgrade.md
+++ b/docs/content/guides/major-postgres-version-upgrade.md
@@ -152,7 +152,7 @@ rm -rf '/pgdata/pg{{< param fromPostgresVersion >}}'
 
 When you are satisfied with the upgrade, you can execute this command to remove the old data directory. Do so at your discretion.
 
-The script does not delete old WAL-files so these must be deleted manually.
+Note that the `delete_old_cluster.sh` script does not delete the old WAL files. These are typically found in `/pgdata/pg{{< param fromPostgresVersion >}}_wal`, although they can be stored elsewhere. If you would like to delete these files, this must be done manually.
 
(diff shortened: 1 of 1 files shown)


docs/content/guides/major-postgres-version-upgrade.md
```diff
@@ -152,6 +152,8 @@ rm -rf '/pgdata/{{< param fromPostgresVersion >}}'
 
 When you are satisfied with the upgrade, you can execute this command to remove the old data directory. Do so at your discretion.
 
+The script does not delete old WAL-files so these must be deleted manually.
```
dsessler7, line 155: Happy to add this to the documentation, although we would like to word it a little differently. How about the following:
```suggestion
Note that the `delete_old_cluster.sh` script does not delete the old WAL files. These are typically found in `/pgdata/pg{{< param fromPostgresVersion >}}_wal`, although they can be stored elsewhere. If you would like to delete these files, this must be done manually.
```

dsessler7: Writing the above comment also makes me realize that line 150 isn't quite right. Could you add an edit to that line to your PR? It should read:

`rm -rf '/pgdata/pg{{< param fromPostgresVersion >}}'`

dsessler7 (changes requested):

dsessler7 (approved):

PR #3667, opened 2023-06-05
Update depguard configuration for golangci-lint v1.53

cbandy:
The depguard v2 linter allows different rules to be applied to different sets of files. We can remove the path-based exclusion from the golangci-lint configuration.

See: https://github.com/OpenPeeDeeP/depguard#config
See: https://golangci-lint.run/usage/linters/#depguard

**Checklist:**

 
**Type of Changes:**

 
 - [x] Other

When no rulesets are configured, a default "Main" ruleset rejects everything except stdlib. Lint checks are failing with:

```
  Error: import 'k8s.io/apimachinery/pkg/util/intstr' is not allowed from list 'Main' (depguard)
```

Lint checks are passing.

**Other Information**:

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -21,17 +21,25 @@ linters:
 
 linters-settings:
   depguard:
-    include-go-root: true
-    packages-with-error-message:
-      - io/ioutil: >
-          Use the "io" and "os" packages instead.
-          See https://go.dev/doc/go1.16#ioutil
+    rules:
+      everything:
+        deny:
+          - pkg: io/ioutil
+            desc: >
+              Use the "io" and "os" packages instead.
+              See https://go.dev/doc/go1.16#ioutil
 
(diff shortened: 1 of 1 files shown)


tony-landreth (approved):

.golangci.yaml
```diff
@@ -21,17 +21,25 @@ linters:
...
+      everything:
+        deny:
+          - pkg: io/ioutil
+            desc: >
+              Use the "io" and "os" packages instead.
+              See https://go.dev/doc/go1.16#ioutil
 
-      - net/http/httptest: Should be used only in tests.
-      - testing/*: The "testing" packages should be used only in tests.
+      not-tests:
```
cbandy, line 32: 🌱 I wasn't excited by this name, but we'll not see it usually. It can change to any better name at any time.

PR #3678, opened 2023-06-28
Adding fix for hugepages/restore issue.

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Currently, a cluster restore will fail if the pods are scheduled to a node that has huge pages enabled and available but the postgrescluster is not requesting huge pages. This is due to our restore process creating a new `postgresql.conf` file that did not have the fix in place to workaround the previously addressed hugepages/kube issue.

Restores will now succeed regardless of hugepages settings on nodes/postgrescluster.

**Other Information**:

### DIFF
--- a/internal/controller/postgrescluster/pgbackrest.go
+++ b/internal/controller/postgrescluster/pgbackrest.go
@@ -1086,9 +1086,16 @@ func (r *Reconciler) reconcileRestoreJob(ctx context.Context,
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


benjaminjb (approved):
LGTM. 

I kinda wish our tests checked that the command output was what we expected, but that would be a new thing for this test. (I mean, we're not doing it yet at all, it looks like.)

PR #3694, opened 2023-07-24
Update test workflow

tony-landreth:
Issue: [sc-20728]

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -54,7 +54,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kubernetes: [v1.25, v1.21]
+        kubernetes: [v1.27, v1.24]
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


tjmoore4 (approved):

.github/workflows/test.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kubernetes: [v1.25, v1.21]
+        kubernetes: [1.27, 1.24]
```
tjmoore4, line 57: :thinking: I'm not sure, but do we need a `v` here? i.e.
```suggestion
        kubernetes: [v1.27, v1.24]
```

tjmoore4 (approved):

.github/workflows/test.yaml
tony-landreth: Thanks for catching that! The v is needed.

Issue #3690, opened 2023-07-13
PGO v5.4.0 issue with PostgreSQL 13 image


paulbadcock:
## Overview

PGO v5.4.0 has broken all instances of PostgreSQL 13 clusters. It's not generating the image to use in the stateful set but it works for version 14 and 15 clusters.

## Environment

Please provide the following details:

- Platform: `OpenShift`
- Platform Version:  `4.12.22`
- PGO Image Tag: `ubi8-5.4.0-0`
- Postgres Version  `13`
- Storage: `thin-vmdk`

### REPRO

Provide steps to get to the error condition:

1. Install the community operator version of PGO 5.4.0
2. Create a new instance of a PostgreSQL 13 database like so:

```yaml
kind: PostgresCluster
apiVersion: postgres-operator.crunchydata.com/v1beta1
metadata:
  name: example
  namespace: example
spec:
  backups:
    pgbackrest:
      repos:
        - name: repo1
          volume:
            volumeClaimSpec:
              accessModes:
                - ReadWriteOnce
              resources:
                requests:
                  storage: 1Gi
  instances:
    - dataVolumeClaimSpec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 1Gi
      replicas: 1
  postgresVersion: 13
```

4. Observe failure to start as the stateful set is missing the image line for the container

### EXPECTED

1. Database to start

### ACTUAL

1. Stateful set and the database is offline due to the missing image definition

## Logs

Stateful set will have the following events:

```text
Create Pod example-instance1-4stl-0 in StatefulSet example-instance1-4stl failed error: Pod "example-instance1-4stl-0" is invalid: [spec.containers[0].image: Required value, spec.containers[1].image: Required value, spec.initContainers[0].image: Required value, spec.initContainers[1].image: Required value]
```

## Additional Information

Using a PostgreSQL version 14 or 15 works as expected, this only breaks with a v13.

If you update the PostgresCluster spec file of the database with an image tag from the v5.3.0 operator the database will startup fine on v5.4.0 but this isn't correct but is a break fix work around.

```yaml
kind: PostgresCluster
apiVersion: postgres-operator.crunchydata.com/v1beta1
metadata:
  name: example
  namespace: example
spec:
  backups:
    pgbackrest:
      repos:
        - name: repo1
          volume:
            volumeClaimSpec:
              accessModes:
                - ReadWriteOnce
              resources:
                requests:
                  storage: 1Gi
[...]

PR #3697, opened 2023-07-25
Stop PostgresCluster reconciliation when required image not set and update PGUpgrade logic

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

In certain scenarios, an empty `image:` field can result in a failure to properly reconcile a PostgresCluster as described in #3690.

This update prevents empty image values from impacting the reconciliation of a PostgresCluster. With this change, the impacted cluster will not be updated until the necessary images are defined and a corresponding warning event will be created. PostgresClusters with images properly defined will reconcile normally.

Also adjusts the PGUpgrade logic to allow for easier recovery from a
missing image scenario. Specific Conditions are more clearly defined
and checking is added for the 'crunchy-upgrade' image.

Finally, a Kuttl test scenario is added.

**Other Information**:
Issue: [sc-21130]

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -98,3 +98,43 @@ func PostgresContainerImage(cluster *v1beta1.PostgresCluster) string {
 func PGONamespace() string {
 	return os.Getenv("PGO_NAMESPACE")
 }
+
+// VerifyImageValues checks that all container images required by the
+// spec are defined. If any are undefined, a list is returned in an error.
+func VerifyImageValues(cluster *v1beta1.PostgresCluster) error {
+
+	var images []string
+
+	if PGBackRestContainerImage(cluster) == "" {
+		images = append(images, "crunchy-pgbackrest")
+	}
+	if PGAdminContainerImage(cluster) == "" &&
+		cluster.Spec.UserInterface != nil &&
+		cluster.Spec.UserInterface.PGAdmin != nil {
+		images = append(images, "crunchy-pgadmin4")
+	}
+	if PGBouncerContainerImage(cluster) == "" &&
+		cluster.Spec.Proxy != nil &&
+		cluster.Spec.Proxy.PGBouncer != nil {
+		images = append(images, "crunchy-pgbouncer")
+	}
+	if PGExporterContainerImage(cluster) == "" &&
+		cluster.Spec.Monitoring != nil &&
+		cluster.Spec.Monitoring.PGMonitor != nil &&
+		cluster.Spec.Monitoring.PGMonitor.Exporter != nil {
+		images = append(images, "crunchy-postgres-exporter")
+	}
+	if PostgresContainerImage(cluster) == "" {
+		if cluster.Spec.PostGISVersion != "" {
+			images = append(images, "crunchy-postgres-gis")
+		} else {
+			images = append(images, "crunchy-postgres")
+		}
+	}
+
+	if len(images) > 0 {
+		return fmt.Errorf("Missing image(s): %s", images)
+	}
+
+	return nil
+}
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -145,3 +145,56 @@ func TestPostgresContainerImage(t *testing.T) {
 	cluster.Spec.Image = "spec-image"
 	assert.Equal(t, PostgresContainerImage(cluster), "spec-image")
 }
+
+func TestVerifyImageValues(t *testing.T) {
+	cluster := &v1beta1.PostgresCluster{}
+
+	verifyImageCheck := func(t *testing.T, envVar, errString string, cluster *v1beta1.PostgresCluster) {
+		unsetEnv(t, envVar)
+		err := VerifyImageValues(cluster)
+		assert.ErrorContains(t, err, errString)
+	}
+
+	t.Run("crunchy-postgres", func(t *testing.T) {
+		cluster.Spec.PostgresVersion = 14
+		verifyImageCheck(t, "RELATED_IMAGE_POSTGRES_14", "crunchy-postgres", cluster)
+	})
+
+	t.Run("crunchy-postgres-gis", func(t *testing.T) {
+		cluster.Spec.PostGISVersion = "3.3"
+		verifyImageCheck(t, "RELATED_IMAGE_POSTGRES_14_GIS_3.3", "crunchy-postgres-gis", cluster)
+	})
+
+	t.Run("crunchy-pgbackrest", func(t *testing.T) {
+		verifyImageCheck(t, "RELATED_IMAGE_PGBACKREST", "crunchy-pgbackrest", cluster)
+	})
+
+	t.Run("crunchy-pgbouncer", func(t *testing.T) {
+		cluster.Spec.Proxy = new(v1beta1.PostgresProxySpec)
+		cluster.Spec.Proxy.PGBouncer = new(v1beta1.PGBouncerPodSpec)
+		verifyImageCheck(t, "RELATED_IMAGE_PGBOUNCER", "crunchy-pgbouncer", cluster)
+	})
+
+	t.Run("crunchy-pgadmin4", func(t *testing.T) {
+		cluster.Spec.UserInterface = new(v1beta1.UserInterfaceSpec)
+		cluster.Spec.UserInterface.PGAdmin = new(v1beta1.PGAdminPodSpec)
+		verifyImageCheck(t, "RELATED_IMAGE_PGADMIN", "crunchy-pgadmin4", cluster)
+	})
+
+	t.Run("crunchy-postgres-exporter", func(t *testing.T) {
+		cluster.Spec.Monitoring = new(v1beta1.MonitoringSpec)
+		cluster.Spec.Monitoring.PGMonitor = new(v1beta1.PGMonitorSpec)
+		cluster.Spec.Monitoring.PGMonitor.Exporter = new(v1beta1.ExporterSpec)
+		verifyImageCheck(t, "RELATED_IMAGE_PGEXPORTER", "crunchy-postgres-exporter", cluster)
+	})
+
+	t.Run("multiple images", func(t *testing.T) {
+		err := VerifyImageValues(cluster)
+		assert.ErrorContains(t, err, "crunchy-postgres-gis")
+		assert.ErrorContains(t, err, "crunchy-pgbackrest")
+		assert.ErrorContains(t, err, "crunchy-pgbouncer")
+		assert.ErrorContains(t, err, "crunchy-pgadmin4")
+		assert.ErrorContains(t, err, "crunchy-postgres-exporter")
+	})
+
+}
--- a/internal/controller/pgupgrade/jobs_test.go
+++ b/internal/controller/pgupgrade/jobs_test.go
@@ -16,6 +16,7 @@ package pgupgrade
 
 import (
 	"context"
+	"os"
 	"strings"
 	"testing"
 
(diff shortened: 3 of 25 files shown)


internal/config/config.go
```diff
@@ -98,3 +98,43 @@ func PostgresContainerImage(cluster *v1beta1.PostgresCluster) string {
 func PGONamespace() string {
 	return os.Getenv("PGO_NAMESPACE")
 }
+
+// VerifyImageValues checks that all container images required by the
+// spec are defined. If any are undefined, a list is returned in an error.
```
tony-landreth, line 103: How can you tell which images are required by the spec, given what the [CRD](https://access.crunchydata.com/documentation/postgres-operator/latest/references/crd/postgrescluster#postgresclusterspecbackupspgbackrest) says?

internal/controller/pgupgrade/jobs_test.go
```diff
@@ -264,3 +265,55 @@ spec:
 status: {}
 	`))
 }
+
+func saveEnv(t testing.TB, key string) {
```
tony-landreth, line 269: :pick: Took me a minute to figure out why this function is needed. A comment might help here.

internal/controller/pgupgrade/pgupgrade_controller.go
```diff
@@ -331,6 +344,22 @@ func (r *PGUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	setStatusToProgressingIfReasonWas("PGClusterNotShutdown", upgrade)
 
+	// A separate check for primary identification allows for cases where the
```
tony-landreth, line 347: What should a user do in this scenario?

internal/config/config.go
tjmoore4: Since the `RELATED_IMAGES` implementation requires those fields to be optional, it's effectively inferred. The pgBackRest image is always required, either a Postgres or Postgres-GIS image is required and the remainder are determined by your configuration.

testing/kuttl/e2e/major-upgrade-missing-image/11-assert.yaml
```diff
@@ -0,0 +1,11 @@
...
+# Since the cluster is missing the annotation, we get this condition
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PGUpgrade
+metadata:
+  name: empty-image-upgrade
+status:
+  conditions:
+  - type:   "Progressing"
+    status: "False"
+    reason: "PGClusterPrimaryNotIdentified"
```
tony-landreth, line 11: :thinking: Wonder if we should put these `reasons`s in our docs.

testing/kuttl/e2e/major-upgrade-missing-image/12--start-and-update-version.yaml
```diff
@@ -0,0 +1,17 @@
+---
+# Shutdown the cluster -- but without the annotation.
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PostgresCluster
+metadata:
+  name: major-upgrade-empty-image
+spec:
+  shutdown: false
```
tony-landreth, line 8: :question: How does this field setting fit with the comment? `# Shutdown the cluster -- but without the annotation.`

testing/kuttl/e2e/major-upgrade-missing-image/12--start-and-update-version.yaml
```diff
@@ -0,0 +1,17 @@
+---
+# Shutdown the cluster -- but without the annotation.
```
tony-landreth, line 2: :wrench: This comment doesn't seem right.

testing/kuttl/e2e/major-upgrade-missing-image/14-assert.yaml
```diff
@@ -0,0 +1,22 @@
+---
+# Now that the postgres cluster is shut down and annotated, the pgupgrade
+# can finish reconciling. We know the reconciling is complete when
```
tony-landreth, line 3:
```suggestion
# can finish reconciling. We know the reconciliation is complete when
```

internal/controller/pgupgrade/pgupgrade_controller.go
tjmoore4: I'd like to investigate additional options, but, as of right now, I think the upgrade has to be abandoned.
This can result from when a PostgresCluster never initialized in the first place. In this case a new cluster should be created. The more complex scenario is when the cluster has previously initialized, but for some reason the correct StatefulSet cannot be identified.

testing/kuttl/e2e/major-upgrade-missing-image/10-assert.yaml
```diff
@@ -0,0 +1,12 @@
...
+# shutdown status.
+apiVersion: postgres-operator.crunchydata.com/v1beta1
+kind: PGUpgrade
+metadata:
+  name: empty-image-upgrade
+status:
+  conditions:
+  - type:   "Progressing"
+    status: "False"
+    reason: "PGClusterNotShutdown"
```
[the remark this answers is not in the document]
tjmoore4: No, the shutdown process identifies the startup instance (which is where we get `world.ClusterPrimary`, so this should be the correct order to check. Said another way, shutdown precedes the next check.

testing/kuttl/e2e/major-upgrade-missing-image/11-assert.yaml
tjmoore4: Agreed, once this is finalized.

dsessler7 (commented):
Other than what Tony has pointed out, looks good to me. I'll let him give the official Approval though.

tony-landreth (approved):

internal/config/config.go
```diff
@@ -98,3 +98,43 @@ func PostgresContainerImage(cluster *v1beta1.PostgresCluster) string {
...
+	if PGBouncerContainerImage(cluster) == "" &&
+		cluster.Spec.Proxy != nil &&
+		cluster.Spec.Proxy.PGBouncer != nil {
+		images = append(images, "crunchy-pgbouncer")
+	}
+	if PGExporterContainerImage(cluster) == "" &&
+		cluster.Spec.Monitoring != nil &&
+		cluster.Spec.Monitoring.PGMonitor != nil &&
+		cluster.Spec.Monitoring.PGMonitor.Exporter != nil {
+		images = append(images, "crunchy-postgres-exporter")
```
benjaminjb, line 125: ⛏️ Not a blocker, but I wonder if the order of checks here should have the `*Image` func last? Like: is this cluster asking for a proxy? Do we have an image for that requested proxy?

internal/config/config_test.go
```diff
@@ -145,3 +145,47 @@ func TestPostgresContainerImage(t *testing.T) {
...
+		verifyImageCheck(t, "RELATED_IMAGE_PGADMIN", "crunchy-pgadmin4", cluster)
+	})
+
+	t.Run("crunchy-postgres-exporter", func(t *testing.T) {
+		cluster.Spec.Monitoring = new(v1beta1.MonitoringSpec)
+		cluster.Spec.Monitoring.PGMonitor = new(v1beta1.PGMonitorSpec)
+		cluster.Spec.Monitoring.PGMonitor.Exporter = new(v1beta1.ExporterSpec)
+		verifyImageCheck(t, "RELATED_IMAGE_PGEXPORTER", "crunchy-postgres-exporter", cluster)
+	})
+
```
benjaminjb, line 190: 🤔 I might like a test that checked for multiple missing images

testing/kuttl/e2e/major-upgrade-missing-image/12--start-and-update-version.yaml
```diff
@@ -0,0 +1,17 @@
+---
+# Update the postgres version and start restart the cluster.
```
benjaminjb, line 2:
```suggestion
# Update the postgres version and restart the cluster.
```

testing/kuttl/e2e/major-upgrade-missing-image/17--check-data-and-version.yaml
```diff
@@ -0,0 +1,39 @@
+---
```
benjaminjb, line 1: 🔧  rename file -- not checking data, just the version, right?

testing/kuttl/e2e/major-upgrade-missing-image/01--valid-upgrade.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
benjaminjb, line 1: Feel like this test could use a README.md to describe the steps. I'm actually a little confused -- do we ever hit the "missing image" error?

(Alternately: what happens if we have a 13 cluster and try to upgrade but have no 14 related image?

benjaminjb (commented):
Some questions, mostly non-blockers.

internal/config/config.go
```diff
@@ -98,3 +98,43 @@ func PostgresContainerImage(cluster *v1beta1.PostgresCluster) string {
...
+	}
+	if PostgresContainerImage(cluster) == "" {
+		if version := cluster.Spec.PostGISVersion; version != "" {
```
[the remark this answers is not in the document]
tjmoore4: Not here, but the code I 'borrowed' from above does. :smile: I'll make this update.

internal/config/config.go
tjmoore4: Most of the time, I believe evaluation will stop at the first check the way it's written, so I figure that's worth something. I could see it either way I think.

testing/kuttl/e2e/major-upgrade-missing-image/17--check-data-and-version.yaml
tjmoore4: Yep, that was a copy/paste thing. Updating.

testing/kuttl/e2e/major-upgrade-missing-image/01--valid-upgrade.yaml
tjmoore4: Yes, because the initial steps use a non-available Postgres version. For the second case, assuming an appropriate upgrade image is used, the cluster would be unable to start until the required image was defined, but upgrade shouldn't be impacted. I've added a README as well.

testing/kuttl/e2e/major-upgrade-missing-image/README.md
```diff
@@ -0,0 +1,36 @@
+## Major upgrade missing image tests
+
+This is a variation of derived from our major upgrade KUTTL tests designed to
```
benjaminjb, line 3:
```suggestion
This is a variation derived from our major upgrade KUTTL tests designed to
```

testing/kuttl/e2e/major-upgrade-missing-image/10--cluster.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: major-upgrade-empty-image
+spec:
+  # postgres version that is no longer available
+  postgresVersion: 10
+  patroni:
+    dynamicConfiguration:
+      postgresql:
+        parameters:
+          shared_preload_libraries: pgaudit, set_user, pg_stat_statements, pgnodemx, pg_cron
```
jmckulk, line 15: why do we care about these shared libraries?

tjmoore4: This test was designed as a supplement to the `major-upgrade` test, so I tried to keep things similar where possible. I'm not sure the libraries are strictly required here, but there shouldn't be any harm in leaving them configured.

jmckulk: Looking back a little - there were issues where an upgrade would fail due to extensions not being installed in the upgrade container. The upgrade test was updated to include the libraries and test this case (sc-16369).

PR #3714, opened 2023-08-29
Initial changes/additions for exporter refactor

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

Currently, the `crunchy-postgres-exporter` image contains a number of setup.sql and queries.yaml files, as well as a start script that generates the final queries.yml file used by the `postgres_exporter` binary and starts the binary.

The `postgres-operator` image will now hold the setup.sql and queries.yaml files as well as the logic needed to build the master queries.yml file used by the `postgres_exporter` binary. PGO will create a ConfigMap that will hold the queries.yml file used by the `postgres_exporter` binary. The `crunchy-postgres-exporter` image will now be a component container that lives in its own repo. 

**Other Information**:


internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -399,11 +428,7 @@ func configureExporterTLS(cluster *v1beta1.PostgresCluster, template *corev1.Pod
 		}}
 
 		exporterContainer.VolumeMounts = append(exporterContainer.VolumeMounts, mounts...)
-		exporterContainer.Env = append(exporterContainer.Env, corev1.EnvVar{
-			// TODO (jmckulk): define path not dir
-			Name:  "WEB_CONFIG_DIR",
-			Value: "web-config/",
-		})
+		exporterContainer.Command = append(exporterContainer.Command, "--web.config.file=/web-config/web-config.yml")
```
jmckulk, line 431: It doesn't look like this flag is added to the expected command in the tests. I'm also not seeing this flag added when I deploy locally.

Is it being used? Why not define it above with the other flags?

build/postgres-operator/Dockerfile
```diff
@@ -4,6 +4,13 @@ COPY licenses /licenses
 
 COPY bin/postgres-operator /usr/local/bin
 
+RUN mkdir -p /opt/cpm/conf
+
+COPY hack/tools/pgmonitor/postgres_exporter/common /opt/cpm/conf
+COPY hack/tools/pgmonitor/postgres_exporter/linux /opt/cpm/conf
```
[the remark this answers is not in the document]
cbandy: 📝 `pgbackrest-info.sh` is inside the Postgres image.

internal/controller/postgrescluster/pgmonitor.go
dsessler7: It only gets added when custom tls is used.

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -320,6 +336,10 @@ func addPGMonitorExporterToInstancePodSpec(
 			Name: "exporter-config",
 			// this is the path for custom config as defined in the start.sh script for the exporter container
 			MountPath: "/conf",
+		}, {
```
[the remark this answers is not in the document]
dsessler7: It will be when we add support for custom queries.

internal/controller/postgrescluster/pgmonitor.go
jmckulk: It would be nice to have the command defined in the same place :pick:

build/postgres-operator/Dockerfile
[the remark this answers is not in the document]
jmckulk: Oh yeah, we have our own `pgbackrest-info.sh` :thinking:

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -51,6 +53,14 @@ const (
...
 )
 
+// Defaults for certain values used in queries.yml
+// TODO(dsessler7): make these values configurable via spec
+var defaultValuesForQueries = map[string]string{
+	"PGBACKREST_INFO_THROTTLE_MINUTES":    "10",
+	"PG_STAT_STATEMENTS_LIMIT":            "20",
+	"PG_STAT_STATEMENTS_THROTTLE_MINUTES": "-1",
+}
+
```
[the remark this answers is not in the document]
dsessler7: Yeah I wasn't sure of the best place to define these. Given that they are "default" values and that they might eventually be configurable by the user, I went with "package scope" but I can move them into the function.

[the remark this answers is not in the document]
jmckulk: Honestly, I'm on the fence. The thing that feels weird to me is that it is a var map instead of individual string constants

internal/controller/postgrescluster/pgmonitor.go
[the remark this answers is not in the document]
dsessler7: Agreed. I was going to make the changes so that all the queries files are in the same directory when working the custom queries issue.

internal/controller/postgrescluster/pgmonitor.go
dsessler7: I'm not sure how that would work... :thinking:

jmckulk: Oh sorry, I didn't realize this was a different function :sweat_smile: 

Where it is now makes sense.

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -269,32 +286,31 @@ func addPGMonitorToInstancePodSpec(
...
-			{Name: "PG_STAT_STATEMENTS_LIMIT", Value: "20"},
-			{Name: "PG_STAT_STATEMENTS_THROTTLE_MINUTES", Value: "-1"},
-			{Name: "EXPORTER_PG_HOST", Value: exporterHost},
-			{Name: "EXPORTER_PG_PORT", Value: fmt.Sprint(*cluster.Spec.Port)},
-			{Name: "EXPORTER_PG_DATABASE", Value: exporterDB},
-			{Name: "EXPORTER_PG_USER", Value: pgmonitor.MonitoringUser},
-			{Name: "EXPORTER_PG_PASSWORD", ValueFrom: &corev1.EnvVarSource{
+			{Name: "DATA_SOURCE_URI", Value: fmt.Sprintf("%s:%d/%s?%s", exporterHost, *cluster.Spec.Port, exporterDB, exporterPgParams)},
+			{Name: "DATA_SOURCE_USER", Value: pgmonitor.MonitoringUser},
+			{Name: "DATA_SOURCE_PASS", ValueFrom: &corev1.EnvVarSource{
```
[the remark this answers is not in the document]
jmckulk: We have an issue to change to a pass file. If you wanted to work it as part of this story, I think that would be fine, but we'll get to it soon either way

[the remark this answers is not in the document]
dsessler7: Yeah I think I'd rather keep that in a separate issue.

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -305,10 +291,10 @@ func addPGMonitorExporterToInstancePodSpec(
 		ImagePullPolicy: cluster.Spec.ImagePullPolicy,
 		Resources:       cluster.Spec.Monitoring.PGMonitor.Exporter.Resources,
 		Command: []string{
-			"/opt/cpm/bin/postgres_exporter", exporterExtendQueryPathFlag, exporterWebListenAddressFlag,
+			"postgres_exporter", exporterExtendQueryPathFlag, exporterWebListenAddressFlag,
 		},
 		Env: []corev1.EnvVar{
-			{Name: "DATA_SOURCE_URI", Value: fmt.Sprintf("%s:%d/%s?%s", exporterHost, *cluster.Spec.Port, exporterDB, exporterPgParams)},
+			{Name: "DATA_SOURCE_URI", Value: fmt.Sprintf("%s:%d/%s?", exporterHost, *cluster.Spec.Port, exporterDB)},
```
jmckulk, line 297:
```suggestion
			{Name: "DATA_SOURCE_URI", Value: fmt.Sprintf("%s:%d/%s", exporterHost, *cluster.Spec.Port, exporterDB)},
```
without parameters, I think we can remove the `?`

Link: https://www.postgresql.org/docs/current/libpq-connect.html section: 34.1.1.2. Connection URIs

internal/pgmonitor/util.go
```diff
@@ -19,6 +19,11 @@ import (
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
+const (
+	// Where we mount the queries.yaml file
+	QueriesConfigDir = "/opt/crunchy/conf"
```
jmckulk, line 24: I think it would be good to make this configurable. Currently, deploy dev won't work unless you make a code change. Pulling it from an env var then setting that for deploy-dev in the Makefiles would be helpful

Makefile
```diff
@@ -70,11 +70,15 @@ get-pgmonitor:
 	git -C '$(dir $(PGMONITOR_DIR))' clone https://github.com/CrunchyData/pgmonitor.git || git -C '$(PGMONITOR_DIR)' fetch origin
 	@git -C '$(PGMONITOR_DIR)' checkout '$(PGMONITOR_VERSION)'
 	@git -C '$(PGMONITOR_DIR)' config pull.ff only
+	[ -d /opt/crunchy/conf ] || sudo mkdir -p /opt/crunchy/conf
```
jmckulk, line 73: I don't think we should reach outside the `postgres-operator` repo - can we update this to point somewhere in the repo, maybe `hack/` something?

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -325,7 +325,7 @@ func addPGMonitorExporterToInstancePodSpec(
 		}, {
 			Name: "exporter-queries",
 			// this is the path for the default config, which we generate and then mount via ConfigMap
-			MountPath: pgmonitor.QueriesConfigDir,
+			MountPath: "/tmp",
```
jmckulk, line 328: Is there any overlap in the `tmp` directory with other containers? Is that an issue?

internal/controller/postgrescluster/pgmonitor.go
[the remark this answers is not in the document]
jmckulk: planning on leaving this as is, iiuc

jmckulk (approved):
lgtm

hack/bin/postgres_exporter/postgres_exporter
jmckulk, line 1: Need a gitignore for the binary and these files

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -104,7 +106,7 @@ func (r *Reconciler) reconcilePGMonitorExporter(ctx context.Context,
 	// the `EnableExporterInPostgreSQL` funcs; that way we are always running
 	// that function against an updated and running pod.
 	if pgmonitor.ExporterEnabled(cluster) {
-		sql, err := os.ReadFile(fmt.Sprintf("%s/pg%d/setup.sql", pgmonitor.GetQueriesConfigDir(), cluster.Spec.PostgresVersion))
+		sql, err := os.ReadFile(fmt.Sprintf("%s/pg%d/setup.sql", pgmonitor.GetQueriesConfigDir(log), cluster.Spec.PostgresVersion))
```
jmckulk, line 109: pass the context down then get the logger from inside the `GetQueriesConfigDir` function
```suggestion
		sql, err := os.ReadFile(fmt.Sprintf("%s/pg%d/setup.sql", pgmonitor.GetQueriesConfigDir(ctx), cluster.Spec.PostgresVersion))
```

It is more normal for us to pass around a context and sets us up to do more interesting things in the future with contexts

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -532,7 +534,7 @@ func generateQueries(ctx context.Context, cluster *v1beta1.PostgresCluster) stri
 	// }
 
 	for _, queryType := range baseQueries {
-		queriesContents, err := os.ReadFile(fmt.Sprintf("%s/queries_%s.yml", pgmonitor.GetQueriesConfigDir(), queryType))
+		queriesContents, err := os.ReadFile(fmt.Sprintf("%s/queries_%s.yml", pgmonitor.GetQueriesConfigDir(log), queryType))
```
jmckulk, line 537: :thought_balloon: you could grab this path once then just use a variable - instead of having to change each line every time something changes

jmckulk (approved):
lgtm

jmckulk (approved):

PR #3715, opened 2023-08-30
Update apply_test to handle changes for Kubernetes 1.28+

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

Prior to 1.28.0, certain no-op server-side apply updates bumped the resourceVersion value. For new Kubernetes versions this behavior has been adjusted so that resourceVersion is not bumped.

This change adds an additional check for the server version to allow the correct test to be executed.

**Other Information**:

### DIFF
--- a/internal/controller/postgrescluster/apply_test.go
+++ b/internal/controller/postgrescluster/apply_test.go
@@ -79,8 +79,19 @@ func TestServerSideApply(t *testing.T) {
 		after := constructor()
 		assert.NilError(t, cc.Patch(ctx, after, client.Apply, reconciler.Owner))
 		assert.Assert(t, after.GetResourceVersion() != "")
-		assert.Assert(t, after.GetResourceVersion() != before.GetResourceVersion(),
-			"expected https://github.com/kubernetes-sigs/controller-runtime/issues/1356")
+
+		switch {
+		// TODO(tjmoore4): The update currently impacts 1.28+ only, but may be
+		// backpatched in the future.
+		// - https://github.com/kubernetes/kubernetes/pull/116865
+		case serverVersion.LessThan(version.MustParseGeneric("1.28")):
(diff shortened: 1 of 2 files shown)


internal/controller/postgrescluster/apply_test.go
```diff
@@ -79,8 +79,19 @@ func TestServerSideApply(t *testing.T) {
 		after := constructor()
 		assert.NilError(t, cc.Patch(ctx, after, client.Apply, reconciler.Owner))
 		assert.Assert(t, after.GetResourceVersion() != "")
-		assert.Assert(t, after.GetResourceVersion() != before.GetResourceVersion(),
-			"expected https://github.com/kubernetes-sigs/controller-runtime/issues/1356")
+
+		switch {
```
tony-landreth, line 83: :question: Would an if-else work here?

tjmoore4: Yes, but there is a possibility we'll need additional cases in the future. Also, we use `switch` elsewhere for this type of behavior, so it's more consistent with existing coding patterns.

internal/controller/postgrescluster/apply_test.go
```diff
@@ -79,8 +79,19 @@ func TestServerSideApply(t *testing.T) {
...
+		// TODO(tjmoore4): The update currently impacts 1.28+ only, but may be
+		// backpatched in the future.
+		// - https://github.com/kubernetes/kubernetes/pull/116865
```
cbandy, line 86: 👍🏻 The backpatches look haphazard and stalled.

cbandy (approved):

PR #3719, opened 2023-09-13
Ability to set custom ccp_monitoring pass

jmckulk:
With this change, users can update the <cluster>-monitoring secret with a password, in either the `stringData` or `data` secret fields, and remove the `verifier` to update the ccp_monitoring password in Postgres. After changing the secret, the `watcher` logic will notice the change and restart the postgres_exporter process with the updated password. This change is to support monitoring for standby clusters. Before this, a standby cluster would be created without having access to the ccp_monitoring user password that was replicated from the primary cluster.

In addition to allowing a custom ccp_monitoring password, this change updates the exporter to use queries from pgMonitor 4.9 and refactors our e2e tests.


testing/kuttl/e2e/exporter-password-change/00-assert.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+# Check that all containers in the instance pod are ready
```
benjaminjb, line 4: To double-check my understanding: we have 00--create-cluster.yaml (maybe rename since it applies _and_ asserts?), which asserts that the pod is running; but this script verifies that all the containers are in the right state?

Or put another way: is there a chance the containers won't be / what happens if they aren't?

testing/kuttl/e2e/exporter-password-change/04-assert.yaml
```diff
@@ -0,0 +1,39 @@
...
+    retry() { bash -ceu 'printf "$1\nSleeping...\n" && sleep 5' - "$@"; }
+    contains() { bash -ceu '[[ "$1" == *"$2"* ]]' - "$@"; }
+    check_containers_ready() { bash -ceu ' \
+      echo "$1" | jq -e ".[] | select(.type==\"ContainersReady\") | .status==\"True\""' - "$@";
+    }
+    pod=$(kubectl get pods -o name -n $NAMESPACE \
+      -l postgres-operator.crunchydata.com/cluster=exporter-password-change \
+      -l postgres-operator.crunchydata.com/crunchy-postgres-exporter=true)
+    [ "$pod" = "" ] && retry "Pod not found" && exit 1
+    password=$(kubectl exec -n ${NAMESPACE} ${pod} -c exporter -- bash -c 'echo $DATA_SOURCE_PASS')
```
benjaminjb, line 20:
```suggestion
    password=$(kubectl exec -n ${NAMESPACE} ${pod} -c exporter -- bash -c 'printenv DATA_SOURCE_PASS')
```

I really have no reason for this suggestion. Maybe it's clearer to me that `printenv` is printing an env, but what else could `echo` be doing in this scenario?

testing/kuttl/e2e/exporter-password-change/04-assert.yaml
```diff
@@ -0,0 +1,39 @@
...
+    condition_json=$(kubectl get ${pod} -n ${NAMESPACE} -o jsonpath="{.status.conditions}")
+    [ "$condition_json" = "" ] && retry "conditions not found" && exit 1
+    { check_containers_ready "$condition_json"; } || {
+      retry "containers not ready"
+      exit 1
+    }
+    scrape_metrics=$(kubectl exec ${pod} -c exporter -n ${NAMESPACE} -- \
+      curl --silent http://localhost:9187/metrics | grep "pg_exporter_last_scrape_error")
+    { contains "${scrape_metrics}" 'pg_exporter_last_scrape_error 0'; } || {
+      retry "${scrape_metrics}" && exit 1
```
benjaminjb, line 34:
```suggestion
      retry "${scrape_metrics}"
      exit 1
```

testing/kuttl/e2e/exporter-password-change/README.md
```diff
@@ -0,0 +1,47 @@
+# Exporter Password Change
```
benjaminjb, line 1: 💯

benjaminjb (commented):
I know you wanted to get the passfile work in, so this is just an initial look at this step.

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -221,25 +221,38 @@ func (r *Reconciler) reconcileMonitoringSecret(
 
 	intent.Data = make(map[string][]byte)
 
-	if len(existing.Data["password"]) == 0 || len(existing.Data["verifier"]) == 0 {
+	// Copy existing password and verifier into the intent
+	password, set := existing.Data["password"]
+	if set {
```
cbandy, line 226: 🤔 Why distinguish between unset and zero/empty here when the blocks below don't?

(Similar code exists [over here](https://github.com/CrunchyData/postgres-operator/blob/eedf513582b7d90c7763f64eb55c5bb266d729f5/internal/controller/postgrescluster/postgres.go#L69-L73).)

internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -221,25 +221,38 @@ func (r *Reconciler) reconcileMonitoringSecret(
...
+		intent.Data["password"] = password
+		verifier, set := existing.Data["verifier"]
+		if set {
+			intent.Data["verifier"] = verifier
+		}
+	}
+
+	// When password is unset, generate a new one
+	// TODO (jmckulk): Consider password types (i.e., AlphaNumeric)
```
cbandy, line 235: 📝 As I recall, AlphaNumeric was added to accommodate tools that are difficult to configure with `!` and `$` and similar.

Passwords with symbols are stronger/better (kinda) than those without, so I prefer we use them when possible.

testing/kuttl/e2e/exporter-password-change/01-assert.yaml
```diff
@@ -0,0 +1,19 @@
...
+      -l postgres-operator.crunchydata.com/crunchy-postgres-exporter=true)
+    scrape_metrics=$(kubectl exec ${pod} -c exporter -n ${NAMESPACE} -- \
+      curl --silent http://localhost:9187/metrics | grep "pg_exporter_last_scrape_error")
+    { contains "${scrape_metrics}" 'pg_exporter_last_scrape_error 0'; } || {
+      retry "${scrape_metrics}" && exit 1
+    }
+collectors:
+- type: pod
+  selector: "postgres-operator.crunchydata.com/cluster=exporter-password-change,postgres-operator.crunchydata.com/crunchy-postgres-exporter=true"
+  container: exporter
```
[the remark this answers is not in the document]
jmckulk: Looking at the Kuttl docs - yes it looks like it will

testing/kuttl/e2e/exporter-password-change/04-assert.yaml
jmckulk: Yeah, printenv probably would be clearer. With pass file changes it will turn into a `cat /file` so most likely won't change it

testing/kuttl/e2e/exporter-password-change/04-assert.yaml
jmckulk: :+1: I'll push this up in the pass file work

testing/kuttl/e2e/exporter-password-change/00-assert.yaml
```diff
@@ -0,0 +1,20 @@
...
+- script: |
+    retry() { bash -ceu 'printf "$1\nSleeping...\n" && sleep 5' - "$@"; }
```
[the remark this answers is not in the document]
jmckulk: "retry" here is basically sleep then exit with a message (maybe there is a better name for the function :thinking: ). If the pod isn't found, the script fails and Kuttl runs it again.

So the whole script will be run every time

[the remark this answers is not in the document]
benjaminjb: Ah, right, because what we discovered in testing was that the `commands` script in a TestAssert gets rerun from the beginning if any element fails.

testing/kuttl/e2e/exporter-password-change/00-assert.yaml
jmckulk: > maybe rename since it applies and asserts?

I'd rather not rename the file - I think the TestStep is to create the cluster and that step has multiple parts 1) create and 2) assert that the expected resources are created.
I do see how something like `create-and-assert` would be more clear, but would we name every TestStep `<do thing>-and-<assert thing>`? Maybe the fact that it is a TestStep means it does a thing and asserts

> Or put another way: is there a chance the containers won't be / what happens if they aren't?

yes, the containers in the pod might not all be ready, so the `ContainersReady` condition wouldn't be found. At this point, the test would fail

internal/controller/postgrescluster/pgmonitor.go
jmckulk: So we would never want to generate an Alphanumeric password? Meaning we can remove this TODO

I know user passwords have a user requirement for Alphanumeric, but this password isn't something that any apps or users should interact with. Only the exporter should be connecting through ccp_monitoring

internal/controller/postgrescluster/pgmonitor.go
jmckulk: Oh yeah, we could just set the value from existing (assuming it isn't nil) every time and operator based on the intent value going forward. I'll look at changing this

internal/controller/postgrescluster/pgmonitor.go
jmckulk: I removed this comment - users shouldn't be using this password, so I think we can use whatever type we think is appropriate

testing/kuttl/e2e/exporter-password-change/04-assert.yaml
jmckulk: its using `cat` now

internal/pgmonitor/exporter.go
```diff
@@ -149,17 +146,19 @@ func ExporterStartCommand(commandFlags []string) []string {
 		// which uses up a lot of memory
 		`while read -r -t 3 -u "${fd}" || true; do`,
 
-		// If the directory's modify time is newer than our file descriptor's,
+		// If the either directory's modify time is newer than our file descriptor's,
```
benjaminjb, line 149:
```suggestion
		// If either directories' modify time is newer than our file descriptor's,
```

testing/kuttl/e2e/exporter-custom-queries/00-assert.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+# First, check that all containers in the instance pod are ready
+# Then, list the query files mounted to the exporter and check for expected files
+# Then, check the contents of the queries to ensure queries.yml was generated correctly
+# Finally, store the current exporter pid as an annotation
+- script: |
+    retry() { bash -ceu 'printf "$1\nSleeping...\n" && sleep 5' - "$@"; }
```
benjaminjb, line 9: I almost want to move the `exit 1` into `retry` so then we can say something like, "when this func is called, we sleep, then fail, to signal to KUTTL to restart this test step"

testing/kuttl/e2e/exporter-password-change/02-assert.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+# Ensure that the password has been updated in the exporter and it can still access
+# Postgres. 
+# - Check that the exporter pid has changed
+# - Check that the DATA_SOURCE_PASS_FILE contains the expected password (`password`)
+# - Check that all containers in the instance pod are ready after restart (TODO: review comment after watcher)
```
benjaminjb, line 8: Can we do that TODO?

testing/kuttl/e2e/exporter-replica/00-assert.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+# First, check that all containers in the instance(s) pod are ready
+# Then, grab the exporter metrics output and check that there were no scrape errors
```
benjaminjb, line 5: Are we doing this scrape errors check?

benjaminjb: (Also re that scrape error check: if we have the wrong password, does that metric go up?)

testing/kuttl/e2e/exporter-password-change/files/update-monitoring-password-checks.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - apiVersion: postgres-operator.crunchydata.com/v1beta1
+    blockOwnerDeletion: true
+    controller: true
+    kind: PostgresCluster
+    name: exporter-password-change
+data:
+  # ensure the password is encoded to 'password'
+  password: cGFzc3dvcmQ=
+---
+# TODO: Check that password is set as a file
```
benjaminjb, line 18: Is this TODO still valid?

benjaminjb (commented):
Just a few comments/questions, mostly re comments/naming things

testing/kuttl/e2e/exporter-custom-queries/00-assert.yaml
jmckulk: i like that

testing/kuttl/e2e-other/exporter-append-custom-queries/README.md
dsessler7, line 1: It looks like you moved this file from the normal exporter test (feature gate off) to the e2e-other exporter test (feature gate on), so the contents of the readme is no longer accurate...

testing/kuttl/e2e/exporter-password-change/00-assert.yaml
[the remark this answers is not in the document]
jmckulk: `exit_with_message()`
`wait_and_exit()` 

:thinking:

testing/kuttl/e2e/exporter-password-change/02-assert.yaml
jmckulk: With the watcher changes, the containers should ever be "not ready". I'm thinking it might be possible to remove that section of the test. I don't see what it tells us

testing/kuttl/e2e/exporter-replica/00-assert.yaml
jmckulk: nope, we should be :+1: good catch

yes, the metric will be `1` if there is an issue with the scrape, this includes incorrect password

testing/kuttl/e2e/exporter-password-change/00-assert.yaml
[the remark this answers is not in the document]
dsessler7: How many times would it retry?

testing/kuttl/e2e/exporter-password-change/files/update-monitoring-password-checks.yaml
jmckulk: I don't think we need that. We could check that the volume is mounted as expected but that would require checking a list :thinking: 

In the assert step we confirm that the contents of the file are what we expect - i think this accomplished the same thing

testing/kuttl/e2e/exporter-password-change/README.md
```diff
@@ -0,0 +1,47 @@
...
+If this step fails logs are collected from the exporter container.
+
+## 02-change-password:
+
+This TestStep will:
+1) Apply the `files/update-monitoring-password.yaml` file to set the monitoring password to `password`
+2) Assert that conditions outlined in `files/update-monitoring-password-checks.yaml` are met
+    - A `<cluster>-monitoring` secret exists with `data.password` set to the encoded value for `password`
+
+## 03-restart-exporter:
```
dsessler7, line 34: Steps 03 and 04 in the README don't actually exist...

testing/kuttl/e2e/exporter-custom-queries/00-assert.yaml
jmckulk: I'm not sure if this would work :thinking: would that just exit the `retry` function?

testing/kuttl/e2e/exporter-replica/00--create-cluster.yaml
dsessler7, line 1: What is the purpose of the `exporter-replica` test?

internal/pgmonitor/exporter.go
```diff
@@ -60,7 +60,7 @@ var DefaultValuesForQueries = map[string]string{
 func GenerateDefaultExporterQueries(ctx context.Context, cluster *v1beta1.PostgresCluster) string {
 	log := logging.FromContext(ctx)
 	var queries string
-	baseQueries := []string{"backrest", "global", "per_db", "nodemx"}
+	baseQueries := []string{"backrest", "global", "global_dbsize", "per_db", "nodemx"}
```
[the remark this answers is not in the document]
jmckulk: its required to get the db size metrics. That query isn't included in the global file anymore

testing/kuttl/e2e/exporter-replica/00--create-cluster.yaml
jmckulk: ensure that an exporter on a replica can gather metrics - it should work if a regular cluster is working but i was thinking a test case would be good

testing/kuttl/e2e-other/exporter-append-custom-queries/README.md
jmckulk: yup i copied and pasted that wrong. I'll fix it

internal/pgmonitor/exporter.go
```diff
@@ -149,17 +146,19 @@ func ExporterStartCommand(commandFlags []string) []string {
...
+		`  then`,
 
-		// We then want to get rid of the old file descriptor, get a fresh one
+		// When something changes we want to get rid of the old file descriptor, get a fresh one
 		// and restart the loop
-		`  then`,
-		`    exec {fd}>&- && exec {fd}<> <(:)`,
-		`    stat --format='Latest queries file dated %%y' "${directory}"`,
+		`    echo "Something changed..."`,
+		`    exec {fd}>&- & exec {fd}<> <(:)`,
```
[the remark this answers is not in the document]
jmckulk: What change are you talking about?

testing/kuttl/e2e/exporter-password-change/00-assert.yaml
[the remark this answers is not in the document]
jmckulk: the `retry` doesn't do anything special - it just waits and expects an exits 1. Kuttl will loop through the script until it exits 0 or hits a timeout

internal/pgmonitor/exporter.go
```diff
@@ -149,17 +146,19 @@ func ExporterStartCommand(commandFlags []string) []string {
 		// which uses up a lot of memory
 		`while read -r -t 3 -u "${fd}" || true; do`,
 
-		// If the directory's modify time is newer than our file descriptor's,
+		// If the either directories' modify time is newer than our file descriptor's,
```
benjaminjb, line 149:
```suggestion
		// If either directories' modify time is newer than our file descriptor's,
```

testing/kuttl/e2e/exporter-custom-queries/00-assert.yaml
benjaminjb: hmmm, what if we `set -e` at the top of the script? Not sure abt that, just a thought to maybe try. 

I'm curious: if you take `retry` out and just put in a sleep, does anything get printed if the script fails?

internal/pgmonitor/exporter.go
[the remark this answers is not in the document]
benjaminjb: `&& exec` in line 161 became `& exec` in line 159?

testing/kuttl/e2e/exporter-standby/00--create-certs.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+apply:
+- files/cluster-certs.yaml
```
benjaminjb, line 4:
```suggestion
- files/cluster-certs.yaml

```
Can one add a newline in a suggestion on GH?

testing/kuttl/e2e/exporter-standby/02--set-primary-password.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+apply:
+- files/update-primary-password.yaml
+assert:
+- files/update-primary-password-checks.yaml
```
benjaminjb, line 6:
```suggestion
- files/update-primary-password-checks.yaml

```

testing/kuttl/e2e/exporter-standby/03--create-standby.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+apply:
+- files/standby-cluster.yaml
+# - files/update-standby-password.yaml
```
benjaminjb, line 5: Are the commented out lines remnants of trying to create + update in one step?

testing/kuttl/e2e/exporter-standby/04--set-standby-password.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+apply:
+- files/update-standby-password.yaml
+assert:
+- files/update-standby-password-checks.yaml
```
benjaminjb, line 6:
```suggestion
- files/update-standby-password-checks.yaml

```

testing/kuttl/e2e/exporter-standby/04-assert.yaml
```diff
@@ -0,0 +1,38 @@
...
+      retry "containers not ready"
+      exit 1
+    }
+
+    scrape_metrics=$(kubectl exec ${pod} -c exporter -n ${NAMESPACE} -- \
+      curl --silent http://localhost:9187/metrics | grep "pg_exporter_last_scrape_error")
+    { contains "${scrape_metrics}" 'pg_exporter_last_scrape_error 0'; } || {
+      retry "${scrape_metrics}"
+      exit 1
+    }
```
benjaminjb, line 38:
```suggestion
    }

```
Wonder if we can add a linter or a git pre-commit hook? (though to be fair, I think I might be the person who most often loses newlines at the end of files.)

testing/kuttl/e2e/exporter-standby/README.md
```diff
@@ -0,0 +1,9 @@
+# Exporter connection on standby cluster
+
+The exporter standby test will deploy two clusters, one primary and one standby.
+Both clusters have monitoring enabled and are created in the same namespace to
+allow for easy connections over the network.
+
+The `ccp_monitoring` password for both clusters are updated to match allowing
+the exporter on the standby cluster to query postgres using the proper `ccp_monitoring`
+password.
```
benjaminjb, line 9: 🤔 Would it be easier or harder to pull the password from the primary and set that for the standby instead of resetting the password for both? I'm just thinking it might save a step. But this actually tests what we want it to, so this isn't a thing to change at this point.

testing/kuttl/e2e/exporter-standby/03--create-standby.yaml
jmckulk: I was doing both of these (create and update) in one step at one point. I can remove it

testing/kuttl/e2e/exporter-standby/README.md
jmckulk: It wouldn't be that difficult to grab the contents of the primary secret. But yeah, this is the actual workflow I would expect someone to use, and I don't think it adds too much complexity to the test

testing/kuttl/e2e/exporter-standby/00--create-certs.yaml
jmckulk: yep! i just have to apply the suggestions and rebase locally.

testing/kuttl/e2e/exporter-custom-queries/00-assert.yaml
jmckulk: I did a quick test and `set -e` does return as expected.

> I'm curious: if you take retry out and just put in a sleep, does anything get printed if the script fails?

I'm not tracking... like this?

```
{ contains x y } || { sleep 5 && exit 1}
```

jmckulk: with that, there is no error message - i'm not sure what would happen if the error continued until the timeout 

the first command just sleeps and exits but the second uses retry
```
    logger.go:42: 09:58:34 | exporter-no-tls/0-00--create-cluster | running command: [sh -c set -e
        retry() { bash -ceu 'printf "$1\nSleeping...\n" && sleep 5 && exit 1' - "$@"; }
...
        pod=$(kubectl get pods -o name -n "${NAMESPACE}" \
          -l postgres-operator.crunchydata.com/cluster=exporter-no-tls \
          -l postgres-operator.crunchydata.com/crunchy-postgres-exporter=true)
        [ "$pod" = "" ] && sleep 5 && exit 1
...
    logger.go:42: 09:58:40 | exporter-no-tls/0-00--create-cluster | running command: [sh -c set -e
...
        condition_json=$(kubectl get "${pod}" -n "${NAMESPACE}" -o jsonpath="{.status.conditions}")
        [ "$condition_json" = "" ] && retry "conditions not found"
        { check_containers_ready "$condition_json"; } || {
          retry "containers not ready"
        }
...
    logger.go:42: 09:58:40 | exporter-no-tls/0-00--create-cluster | false
    logger.go:42: 09:58:40 | exporter-no-tls/0-00--create-cluster | containers not ready
    logger.go:42: 09:58:40 | exporter-no-tls/0-00--create-cluster | Sleeping...
    logger.go:42: 09:58:46 | exporter-no-tls/0-00--create-cluster | running command: [sh -c set -e

File: testing/kuttl/e2e/exporter-password-change/00-assert.yaml (hunk 8 above)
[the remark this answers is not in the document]
Comment (benjaminjb) [reply testing/kuttl/e2e/exporter-password-change/00-assert.yaml:6]: Yeah, I guess this is a feature of TestAssert steps that has an unexpected side-effect when using `commands`:

* TestAsserts wait until their conditions are true -- so if you have a TestAssert looking for a pod with particular labels, KUTTL will keep checking until it finds that pod or until it times out.
* so if a `commands` section in a TestAssert fails, it just gets restarted.

If I put it that way, I get it, but I'm still surprised by that behavior the first few times I've seen it. Maybe worth a comment in the README?

File: testing/kuttl/e2e/exporter-custom-queries/00-assert.yaml (hunk 10 above)
Comment (benjaminjb) [reply testing/kuttl/e2e/exporter-custom-queries/00-assert.yaml:9]: Yeah, this 
```
{ contains x y } || { sleep 5 && exit 1}
```
is what I was thinking. I was wondering if KUTTL would report something when it gets to that `exit`, but no, thinking about it now, I don't see why it would.

That said, do you know where the `false` is coming from in 
```
logger.go:42: 09:58:40 | exporter-no-tls/0-00--create-cluster | false
```

Comment (dsessler7) [reply testing/kuttl/e2e/exporter-custom-queries/00-assert.yaml:9]: `check_containers_ready`... it takes the argument, the `.status.conditions` of the pod, uses `jq` and whatnot to get the `ContainersReady` status, and compares it to `True` and returns the result which in this case is `false`

File: testing/kuttl/e2e/exporter-password-change/00-assert.yaml (hunk 8 above)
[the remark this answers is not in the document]
Comment (jmckulk) [reply testing/kuttl/e2e/exporter-password-change/00-assert.yaml:6]: Yeah, i imagine a kuttl how to doc somewhere?

One thing of note - the script will loop until 1) it passes AND 2) the assert files are met. So even if a script exits 0, it will continue to run until the assert files (e.g., yaml asserts on a postgres cluster) also pass

File: testing/kuttl/e2e/exporter-password-change/README.md [hunk 24]
```diff
@@ -0,0 +1,47 @@
...
+2) Assert that conditions outlined in `files/initial-postgrescluster-checks.yaml` are met
+    - PostgresCluster exists with a single ready replica
+    - A pod with `cluster` and `crunchy-postgres-exporter` labels has the status `{phase: Running}`
+    - A `<cluster>-monitoring` secret exists with correct labels and ownerReferences
+
+## 00-assert:
+
+This TestAssert will loop through a script until:
+1) the instance pod has the `ContainersReady` condition with status `true`
+2) the asserts from `00--create-cluster` are met.
```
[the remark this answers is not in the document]
jmckulk: ha, i assume you didn't mean to link to pr # 2

um this is related to what I said in the other comment. The script will loop until both of those things are true. Script passes and asserts are met

testing/kuttl/e2e/exporter-custom-queries/00-assert.yaml
jmckulk: locally i have sent the output of `jq > /dev/null` so it doesn't have that line. I'll push that up shortly

dsessler7 (approved):
LGTM

testing/kuttl/e2e/exporter-no-tls/00-assert.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+# First, check that all containers in the instance pod are ready
+# Then, grab the exporter metrics output and check that there were no scrape errors
+# Then, check the exporter logs for the 'TLS is disabled' line
```
benjaminjb, line 6: ⛏️ 
```suggestion
# Then, check the exporter logs for the 'TLS is disabled' line
# Then, grab the exporter metrics output and check that there were no scrape errors
```
?

benjaminjb (approved):
LGTM

testing/kuttl/e2e/exporter-custom-queries/00-assert.yaml
jmckulk: I didn't end up pushing this change. Will add to backlog

testing/kuttl/e2e/exporter-password-change/00-assert.yaml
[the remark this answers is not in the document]
jmckulk: I didn't end up changing the name of the `retry` function - Will add to backlog

PR #3728, opened 2023-09-21
Make PGO backwards compatible with older crunchy-postgres-exporter im…

dsessler7:
…ages.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

Currently, the latest changes to PGO are only compatible with `crunchy-postgres-exporter` images that will be built with the latest changes in that repo. This is because the new exporter images will have the `postgres_exporter` binary on the PATH and latest PGO now expect this. Older exporter images do not have `postgres_exporter` on the PATH and will therefore fail to run properly.

This PR makes it so PGO checks to see if the `postgres_exporter` command is available. If it is, it must be a newer image. If it isn't, it must be an older exporter image and PGO will find the directory that holds the binary and add that directory to the PATH.

**Other Information**:

### DIFF
--- a/internal/pgmonitor/exporter.go
+++ b/internal/pgmonitor/exporter.go
@@ -122,16 +122,8 @@ func GenerateDefaultExporterQueries(ctx context.Context, cluster *v1beta1.Postgr
 // the source queries files.
 func ExporterStartCommand(commandFlags []string) []string {
 	script := strings.Join([]string{
-		// Check to see if the postgres_exporter command is successful. If it is, we are
-		// using an image that has the command on the PATH. If it isn't, we have an older
-		// image that does not have it on the PATH and we should find the directory that
-		// holds the postgres_exporter binary and add it to the PATH.
-		`rc=0`,
-		`postgres_exporter --version || rc=$?`,
-		`if [ $rc -ne 0 ]; then`,
-		`	dir=$(find /opt/cpm/bin/postgres_exporter* -type d)`,
-		`	export PATH=$PATH:$dir`,
-		`fi`,
+		// Older images do not have the command on the PATH.
+		`PATH="$PATH:$(echo /opt/cpm/bin/postgres_exporter-*)"`,
 
 		// Set up temporary file to hold postgres_exporter process id
 		`POSTGRES_EXPORTER_PIDFILE=/tmp/postgres_exporter.pid`,
--- a/internal/pgmonitor/exporter.go
+++ b/internal/pgmonitor/exporter.go
@@ -122,6 +122,17 @@ func GenerateDefaultExporterQueries(ctx context.Context, cluster *v1beta1.Postgr
 // the source queries files.
 func ExporterStartCommand(commandFlags []string) []string {
 	script := strings.Join([]string{
+		// Check to see if the postgres_exporter command is successful. If it is, we are
+		// using an image that has the command on the PATH. If it isn't, we have an older
+		// image that does not have it on the PATH and we should find the directory that
+		// holds the postgres_exporter binary and add it to the PATH.
+		`rc=0`,
+		`postgres_exporter --version || rc=$?`,
+		`if [ $rc -ne 0 ]; then`,
+		`	dir=$(find /opt/cpm/bin/postgres_exporter* -type d)`,
+		`	export PATH=$PATH:$dir`,
+		`fi`,
+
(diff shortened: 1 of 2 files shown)


internal/pgmonitor/exporter.go
```diff
@@ -122,6 +122,17 @@ func GenerateDefaultExporterQueries(ctx context.Context, cluster *v1beta1.Postgr
...
+		// Check to see if the postgres_exporter command is successful. If it is, we are
+		// using an image that has the command on the PATH. If it isn't, we have an older
+		// image that does not have it on the PATH and we should find the directory that
+		// holds the postgres_exporter binary and add it to the PATH.
+		`rc=0`,
+		`postgres_exporter --version || rc=$?`,
+		`if [ $rc -ne 0 ]; then`,
+		`	dir=$(find /opt/cpm/bin/postgres_exporter* -type d)`,
+		`	export PATH=$PATH:$dir`,
+		`fi`,
```
cbandy, line 134: 🤔 Can we let the shell do all the looking around?

```suggestion
		// Older images do not have the command on the PATH.
		`PATH="$PATH:$(echo /opt/cpm/bin/postgres_exporter-*)"`,
```

When the directory does not exist, `PATH` will include `/opt/cpm/bin/postgres_exporter-*` and be ignored (because it doesn't exist).

cbandy (approved):

internal/pgmonitor/exporter.go
dsessler7: I didn't think `PATH` could use wildcards like that... Also, doesn't it need to be exported?

benjaminjb: Old image:

```
➜  ~ docker run --rm -it registry.developers.crunchydata.com/crunchydata/crunchy-postgres-exporter:latest sh
sh-4.4$ export PATH="$PATH:$(echo /opt/cpm/bin/postgres_exporter-*)"
sh-4.4$ postgres_exporter --version
postgres_exporter, version 0.12.1 (branch: HEAD, revision: 1c063b1b1913db029d449818e9cd1750c2282198)
  build user:       root@a5fc99238ef0
  build date:       20230613-16:18:22
  go version:       go1.20.5
  platform:         linux/amd64
  tags:             netgo static_build
```

New image, just to check:

```
➜  crunchydata docker run --rm -it gcr.io/crunchy-dev-test/crunchy-postgres-exporter:latest sh
sh-4.4$ postgres_exporter --version
postgres_exporter, version 0.12.1 (branch: HEAD, revision: 1c063b1b1913db029d449818e9cd1750c2282198)
  build user:       root@a5fc99238ef0
  build date:       20230613-16:18:22
  go version:       go1.20.5
  platform:         linux/amd64
  tags:             netgo static_build
sh-4.4$ export PATH="$PATH:$(echo /opt/cpm/bin/postgres_exporter-*)"
sh-4.4$ postgres_exporter --version
postgres_exporter, version 0.12.1 (branch: HEAD, revision: 1c063b1b1913db029d449818e9cd1750c2282198)
  build user:       root@a5fc99238ef0
  build date:       20230613-16:18:22
  go version:       go1.20.5
  platform:         linux/amd64
  tags:             netgo static_build
```

cbandy: > I didn't think `PATH` could use wildcards like that...

The wildcard is expanded by the [shell](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/idx/shell.html) (pathname expansion) before running `echo`.

(( When nothing matches, the result is the pattern itself. Bash shopt `nullglob` changes this behavior, and/but both results turn out okay in this case. ))

> Also, doesn't it need to be exported?

Hmm. Only if we wanted to specifically affect child processes. IIUC, _this_ shell is the one we want to affect, so no need.

Also, it seems to [already] be exported everywhere I could think to check. I don't know what might govern that.

```
● docker run --rm alpine sh -c 'export -p'
export HOME='/root'
export HOSTNAME='b7e9d03ca262'
export PATH='/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'
export PWD='/'
export SHLVL='1'
```

benjaminjb (approved):

PR #3747, opened 2023-10-13
add pgadmin config

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

* kerberos env vars aren't used
* no config ingested by pgadmin

* remove unused kerberos env vars for now
* add init container to write script to read from mounted config file

**Other Information**:
Issue: PGO-547

### DIFF
--- a/internal/controller/standalone_pgadmin/pod.go
+++ b/internal/controller/standalone_pgadmin/pod.go
@@ -16,13 +16,25 @@ package standalone_pgadmin
 
 import (
 	"fmt"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/crunchydata/postgres-operator/internal/config"
 	"github.com/crunchydata/postgres-operator/internal/initialize"
 	"github.com/crunchydata/postgres-operator/internal/naming"
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
+)
+
+const (
+	configMountPath = "/etc/pgadmin/conf.d"
+	configFilePath  = "~postgres-operator/" + settingsConfigMapKey
+	clusterFilePath = "~postgres-operator/" + settingsClusterMapKey
+	ldapFilePath    = "~postgres-operator/ldap-bind-password"
+
+	// Nothing should be mounted to this location except the script our initContainer writes
+	scriptMountPath = "/etc/pgadmin"
 )
 
 // pod populates a PodSpec with the container and volumes needed to run pgAdmin.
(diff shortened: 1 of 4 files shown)


internal/controller/standalone_pgadmin/pod.go
```diff
@@ -245,5 +299,48 @@ done
 
 	wrapper := `monitor() {` + startScript + reloadScript + `}; export cluster_file="$1"; export -f monitor; exec -a "$0" bash -ceu monitor`
 
-	return []string{"bash", "-ceu", "--", wrapper, "pgadmin", fmt.Sprintf("%s/~postgres-operator/%s", configMountPath, settingsClusterMapKey)}
+	return []string{"bash", "-ceu", "--", wrapper, "pgadmin", fmt.Sprintf("%s/conf.d/~postgres-operator/%s", configMountPath, settingsClusterMapKey)}
```
cbandy, line 302: 🏕️ This isn't something you introduced, but this `configMountPath + conf.d/~postgres-operator + clusterMap` occurs multiple times as well. It should be a constant.

⛏️ `settingsClusterMapKey` is a Secret/ConfigMap _key_, but here it is being used as a file _path_. The names should be clear about what they mean or are for. Here's an example:

https://github.com/CrunchyData/postgres-operator/blob/b05eb5433b6aa80da380926ef99935de38eb945b/internal/pgadmin/config.go#L60-L64

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -245,5 +299,48 @@ done
...
+	// configurations when pgAdmin starts.
+	//
+	// Note: All pgAdmin settings are uppercase with underscores, so ignore any keys/names
+	// that are not.
+	//
+	// Note: set pgAdmin's LDAP_BIND_PASSWORD setting from the via Secret last
+	// in order to overwrite configuration of LDAP_BIND_PASSWORD via ConfigMap JSON.
+	const (
+		// ldapPasswordPath is the path for mounting the LDAP Bind Password
+		ldapPasswordPath         = "/conf.d/~postgres-operator/ldap-bind-password" /* #nosec */
```
cbandy, line 323: 🔧 This path is used on L238 as well. It should be a constant.

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -127,19 +148,51 @@ func pod(
...
+				ReadOnly:  false,
+			},
+			// Volume mounted here because of related mount path.
+			// If volume is not mounted here, K8s errors out trying to create
+			// initContainer and container with different read-only values at related mount path
+			{
+				Name:      mountedConfigsVolumeName,
+				MountPath: configMountPath + "/conf.d",
+				ReadOnly:  true,
+			},
```
cbandy, line 182: ❓ I'm not grokking this comment. Can you explain it to me?

I don't see `startupCommand()` reading from this directory. Perhaps we can omit the mount completely?

cbandy (changes requested):

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -25,6 +26,10 @@ import (
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
+const (
+	configMountPath = "/etc/pgadmin"
```
cbandy, line 30: 💭 IMV, this path is dangerous because pgAdmin executes Python from here. I sense that I want more commentary about how this path/constant is different from similarly named ones in other packages. Like, it isn't _the_ one directory where we plop files. Perhaps as more constants appear, the purpose of this path will be more clear and I'll feel better.

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -34,52 +39,81 @@ func pod(
...
-	configVolume.Projected = &corev1.ProjectedVolumeSource{
-		Sources: podConfigFiles(inConfigMap, *inPGAdmin),
+	// Volume used to write a custom config_system.py file in the initContainer
+	// which then loads the configs found in the `mountedConfigsVolume`
+	systemConfigsVolume := corev1.Volume{
+		Name: configSystemVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{
+				Medium: corev1.StorageMediumMemory,
+			},
```
cbandy, line 68: ❓ In the other package, we seem to be limiting the size of this writable volume. I have a vague memory that it was to try to mitigate some risk that a writable executable `config_system.py` could be exploited.

https://github.com/CrunchyData/postgres-operator/blob/b05eb5433b6aa80da380926ef99935de38eb945b/internal/pgadmin/reconcile.go#L210-L218

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -34,52 +39,81 @@ func pod(
 ) {
 	const (
 		// config and data volume names
-		configVolumeName = "standalone-pgadmin-config"
-		dataVolumeName   = "standalone-pgadmin-data"
+		mountedConfigsVolumeName = "standalone-pgadmin-config"
+		configSystemVolumeName   = "config-system"
+		dataVolumeName           = "standalone-pgadmin-data"
```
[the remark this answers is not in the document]
benjaminjb: I'm not crazy about this var names: `mountedConfigsVolumeName` vs `configSystemVolumeName`?

internal/controller/standalone_pgadmin/pod.go
benjaminjb: So here's what I was thinking and wanted to bring up:

1. "we're not using that `etc/pgadmin/conf.d` in the initContainer, so let's drop it" -- that led to an error with `read-only file system: unknown`
2. "hmmm, ok, well we've got one vol at `etc/pgadmin` and one _inside_ it at `etc/pgadmin/conf.d` -- what happens if I drop the `etc/pgadmin/conf.d` from the initContainer and instead mount it at the container at `var/pgadmin`?" -- and that seemed to work.

So it seemed like defining the parent volume (`etc/pgadmin`) as `rw` in the initcontainer and then the parent + child vols as `ro` in the container led to some confusion.

Not sure if that description makes sense.

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -123,28 +147,61 @@ func pod(
...
+		VolumeMounts: []corev1.VolumeMount{
+			// Volume mounted here because of related mount path.
+			// If volume is not mounted here, K8s errors out trying to create
+			// initContainer and container with different read-only values at related mount path
+			{
+				Name:      configVolumeName,
+				MountPath: configMountPath,
+				ReadOnly:  true,
+			},
+			// Volume to write a custom `config_system.py`` file to.
```
jmckulk, line 180:
```suggestion
			// Volume to write a custom `config_system.py` file to.
```

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -245,5 +303,47 @@ done
...
+	//
+	// This command writes a script in `/etc/pgadmin/config_system.py` that reads from
+	// the `pgadmin-settings.json` file and the `ldap-bind-password` file (if it exists)
+	// and sets those variables globally. That way those values are available as pgAdmin
+	// configurations when pgAdmin starts.
+	//
+	// Note: All pgAdmin settings are uppercase with underscores, so ignore any keys/names
+	// that are not.
+	//
+	// Note: set pgAdmin's LDAP_BIND_PASSWORD setting from the via Secret last
```
tjmoore4, line 323: :thinking: Should this be
```suggestion
	// Note: set pgAdmin's LDAP_BIND_PASSWORD setting from the Secret last
```
or perhaps
```suggestion
	// Note: set pgAdmin's LDAP_BIND_PASSWORD setting via Secret last
```
?

tjmoore4 (approved):
With the latest changes, this looks good to me :+1: 
(Friendly reminder to squash commits before merging.)

internal/controller/standalone_pgadmin/pod.go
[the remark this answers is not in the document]
benjaminjb: OK, I changed the var names to something I could live with.

internal/controller/standalone_pgadmin/pod.go
benjaminjb: We've got a few ways to deal with this, which I'd like to present from worst to best:

1. leave `/etc/pgadmin` rw in the main container -- yikes, we don't want things there because they can run
2. have the initContainer `mkdir etc/pgadmin/conf.d` and not mount the volume there -- it works, but it runs code
3. leave the unused `etc/pgadmin/conf.d` volume mounted to the initcontainer -- it's unused, but mounting an unused vol to an init (therefore temporary) container doesn't seem _wasteful_

Probably some other way to handle this, but with some experimentation, that's what I've got.

cbandy: I've looked into it, and here's what's happening:

1. Kubernetes (kubelet) preps a bunch of directories to act as the emptyDir, ConfigMap, and Secret volumes.
2. It passes a list of bind mounts (directories outside the container paired with paths inside the container) to runc during container start.
3. runc then mounts them in sequence, calling mkdir on the path within the container chroot, "rootfs."

It's here that runc finds the higher directory is readonly (via mount), and it cannot mkdir the subdirectory.

This does _not_ happen when the higher directory is a ConfigMap or Secret because Kubernetes does those mkdir during step 1: https://pr.k8s.io/60342

I've opened https://issue.k8s.io/121294 to see if Kubernetes may ever do this for us. In the meantime, we can explain what we do is a workaround for a known issue. 😁 

<br/>

> 2\. have the initContainer `mkdir etc/pgadmin/conf.d` and not mount the volume there -- it works, but it runs code

Now that I've looked into it, I like this option. This is exactly how Kubernetes does it, and "prepping the file system" in an init container seems fitting.

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -34,52 +46,77 @@ func pod(
 ) {
 	const (
 		// config and data volume names
-		configVolumeName = "standalone-pgadmin-config"
-		dataVolumeName   = "standalone-pgadmin-data"
+		configVolumeName = "pgadmin-config"
+		dataVolumeName   = "pgadmin-data"
+		logVolumeName    = "pgadmin-log"
+		scriptVolumeName = "config-system"
```
cbandy, line 52: 👍🏻 I like `script` and `config-system`.

⛏️ I have some mild anxiety that something else might try to inject a (generic sounding) `config-system` volume. I would feel more comfortable if this had `pgadmin` on the front like the other volume names.

```suggestion
		scriptVolumeName = "pgadmin-config-system"
```

internal/controller/standalone_pgadmin/pod.go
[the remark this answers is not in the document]
cbandy: 👍🏻 I like the new names, too.

internal/controller/standalone_pgadmin/pod_test.go
```diff
@@ -93,19 +89,51 @@ containers:
...
     name: tmp
+initContainers:
+- command:
+  - bash
+  - -ceu
+  - --
+  - (umask a-w && echo "$1" > /etc/pgadmin/config_system.py)
+  - startup
+  - |
+    import json, re, os
```
cbandy, line 111: 👍🏻 It's Python in Go, and/but you made it look good in YAML!

cbandy (approved):

internal/controller/standalone_pgadmin/pod.go
benjaminjb: Updated.

PR #3761, opened 2023-10-26
When requested, load the "citus" library first

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature

User-defined libraries are loaded last. CrunchyData/postgres-operator#3194

When included in user-defined libraries, `citus` is loaded first.

**Other Information**:

Fixes: CrunchyData/postgres-operator#3194
Replaces: CrunchyData/postgres-operator#3530

### DIFF
--- a/internal/patroni/config.go
+++ b/internal/patroni/config.go
@@ -227,17 +227,24 @@ func DynamicConfiguration(
 	// Override the above with mandatory parameters.
 	if pgParameters.Mandatory != nil {
 		for k, v := range pgParameters.Mandatory.AsMap() {
-			// Unlike other PostgreSQL parameters that have mandatory values,
-			// shared_preload_libraries is a comma separated list that can have
-			// other values appended in addition to the mandatory values. Below,
-			// any values provided in the CRD are appended after the mandatory
-			// values.
-			s, ok := parameters[k].(string)
-			if k == "shared_preload_libraries" && ok {
-				parameters[k] = v + "," + s
-			} else {
-				parameters[k] = v
+
+			// This parameter is a comma-separated list. Rather than overwrite the
+			// user-defined value, we want to combine it with the mandatory one.
+			// Some libraries belong at specific positions in the list, so figure
+			// that out as well.
+			if k == "shared_preload_libraries" {
+				// Load mandatory libraries ahead of user-defined libraries.
+				if s, ok := parameters[k].(string); ok && len(s) > 0 {
+					v = v + "," + s
+				}
+				// Load "citus" ahead of any other libraries.
+				// - https://github.com/citusdata/citus/blob/v12.0.0/src/backend/distributed/shared_library_init.c#L417-L419
+				if strings.Contains(v, "citus") {
+					v = "citus," + v
+				}
 			}
+
+			parameters[k] = v
 		}
 	}
 	postgresql["parameters"] = parameters
--- a/internal/pgmonitor/postgres.go
+++ b/internal/pgmonitor/postgres.go
@@ -50,14 +50,7 @@ func PostgreSQLParameters(inCluster *v1beta1.PostgresCluster, outParameters *pos
 		// Exporter expects that shared_preload_libraries are installed
 		// pg_stat_statements: https://access.crunchydata.com/documentation/pgmonitor/latest/exporter/
 		// pgnodemx: https://github.com/CrunchyData/pgnodemx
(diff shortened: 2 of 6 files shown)


internal/postgres/parameters_test.go
```diff
@@ -68,3 +68,23 @@ func TestParameterSet(t *testing.T) {
...
+
+	ps.AppendToList("full", "a")
+	assert.Equal(t, ps.Value("full"), "a")
+
+	ps.AppendToList("full", "b")
+	assert.Equal(t, ps.Value("full"), "a,b")
+
+	ps.AppendToList("full")
+	assert.Equal(t, ps.Value("full"), "a,b", "expected no change")
+}
```
benjaminjb, line 90: Might be unnecessary but if I'm reading the code right I wouldn't mind a test like:

```suggestion
	ps.AppendToList("multi", "a", "b")
	assert.Equal(t, ps.Value("multi"), "a,b")
	
	ps.AppendToList("multi", "c", "d")
	assert.Equal(t, ps.Value("multi"), "a,b,c,d")
}
```

internal/pgmonitor/postgres.go
```diff
@@ -50,14 +50,7 @@ func PostgreSQLParameters(inCluster *v1beta1.PostgresCluster, outParameters *pos
 		// Exporter expects that shared_preload_libraries are installed
 		// pg_stat_statements: https://access.crunchydata.com/documentation/pgmonitor/latest/exporter/
 		// pgnodemx: https://github.com/CrunchyData/pgnodemx
-		libraries := []string{"pg_stat_statements", "pgnodemx"}
-
-		defined, found := outParameters.Mandatory.Get("shared_preload_libraries")
-		if found {
-			libraries = append(libraries, defined)
-		}
```
benjaminjb, line 58: Making sure I'm following: this code here would've prepended "pg_stat_statements", "pgnodemx" to the list -- and we don't want them prepended, we want them appended, yeah?

internal/patroni/config.go
```diff
@@ -227,17 +227,24 @@ func DynamicConfiguration(
...
+			// that out as well.
+			if k == "shared_preload_libraries" {
+				// Load mandatory libraries ahead of user-defined libraries.
+				if s, ok := parameters[k].(string); ok && len(s) > 0 {
+					v = v + "," + s
+				}
+				// Load "citus" ahead of any other libraries.
+				// - https://github.com/citusdata/citus/blob/v12.0.0/src/backend/distributed/shared_library_init.c#L417-L419
+				if strings.Contains(v, "citus") {
+					v = "citus," + v
```
benjaminjb, line 243: Making sure I've got this (though the test helps): 

1. we want citus to be first; 
2. citus can be present several times without side effect;
3. removing citus from that list (from the non-first position) would be brittle

Is that a fair summation?

cbandy: > Is that a fair summation?

Yep.

> 2\. citus can be present several times without side effect;

There is a set of related parameters for loading libraries that amount to `LOAD` statements at certain times.

- https://www.postgresql.org/docs/current/runtime-config-client.html#RUNTIME-CONFIG-CLIENT-PRELOAD
- https://www.postgresql.org/docs/current/sql-load.html
  > If the file has been loaded already, the command does nothing.

Postgres [compares inodes](https://git.postgresql.org/gitweb/?p=postgresql.git;f=src/backend/utils/fmgr/dfmgr.c;hb=REL_11_0#l169) to detect symlink chains as well.

> 3\. removing citus from that list (from the non-first position) would be brittle

A variety of names can refer to the same library. As I recall, `citus`, `citus.so`, and a full path are acceptable. This does not handle the full path case, but in my experience that is quite unusual.

Being more deliberate with these names requires parsing, which #3530 did partially. Ultimately, matching on names is still a step removed from the issue: the Citus code that complains could be in a file with any name. :neckbeard:

internal/pgmonitor/postgres.go
cbandy: Yeah, this is a change.

internal/postgres/parameters_test.go
cbandy: I added something similar and chucked in some quotes:

```go
	ps.AppendToList("full", "a", "cd", `"e"`)
	assert.Equal(t, ps.Value("full"), `a,b,a,cd,"e"`)
```

benjaminjb (approved):

PR #3789, opened 2023-11-28
Add Parallel Pod Management Policy to standalone pgadmin StatefulSet.

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

**Other Information**:

### DIFF
--- a/internal/controller/postgrescluster/pgbackrest.go
+++ b/internal/controller/postgrescluster/pgbackrest.go
@@ -134,10 +134,32 @@ func (r *Reconciler) applyRepoHostIntent(ctx context.Context, postgresCluster *v
 	repoHostName string, repoResources *RepoResources,
 	observedInstances *observedInstances) (*appsv1.StatefulSet, error) {
 
-	repo, err := r.generateRepoHostIntent(postgresCluster, repoHostName, repoResources,
-		observedInstances)
+	repo, err := r.generateRepoHostIntent(postgresCluster, repoHostName, repoResources, observedInstances)
 	if err != nil {
 		return nil, err
+	}
+
+	// Previous versions of PGO used a StatefulSet Pod Management Policy that could leave the Pod
+	// in a failed state. When we see that it has the wrong policy, we will delete the StatefulSet
+	// and then recreate it with the correct policy, as this is not a property that can be patched.
+	// When we delete the StatefulSet, we will leave its Pods in place. They will be claimed by
+	// the StatefulSet that gets created in the next reconcile.
+	existing := &appsv1.StatefulSet{}
+	if err := errors.WithStack(r.Client.Get(ctx, client.ObjectKeyFromObject(repo), existing)); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return nil, err
+		}
+	} else {
+		if existing.Spec.PodManagementPolicy != repo.Spec.PodManagementPolicy {
+			// We want to delete the STS without affecting the Pods, so we set the PropagationPolicy to Orphan.
+			// The orphaned Pods will be claimed by the new StatefulSet that gets created in the next reconcile.
+			uid := existing.GetUID()
+			version := existing.GetResourceVersion()
+			exactly := client.Preconditions{UID: &uid, ResourceVersion: &version}
+			propagate := client.PropagationPolicy(metav1.DeletePropagationOrphan)
+
+			return repo, errors.WithStack(r.Client.Delete(ctx, existing, exactly, propagate))
+		}
 	}
 
 	if err := r.apply(ctx, repo); err != nil {
@@ -560,6 +582,14 @@ func (r *Reconciler) generateRepoHostIntent(postgresCluster *v1beta1.PostgresClu
 		// the cluster should not be shutdown, set this value to 1
 		repo.Spec.Replicas = initialize.Int32(1)
 	}
+
+	// Use StatefulSet's "RollingUpdate" strategy and "Parallel" policy to roll
+	// out changes to pods even when not Running or not Ready.
+	// - https://docs.k8s.io/concepts/workloads/controllers/statefulset/#rolling-updates
+	// - https://docs.k8s.io/concepts/workloads/controllers/statefulset/#forced-rollback
+	// - https://kep.k8s.io/3541
+	repo.Spec.PodManagementPolicy = appsv1.ParallelPodManagement
+	repo.Spec.UpdateStrategy.Type = appsv1.RollingUpdateStatefulSetStrategyType
 
 	// Restart containers any time they stop, die, are killed, etc.
 	// - https://docs.k8s.io/concepts/workloads/pods/pod-lifecycle/#restart-policy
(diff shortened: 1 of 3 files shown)


internal/controller/standalone_pgadmin/statefulset.go
```diff
@@ -33,6 +35,27 @@ func (r *PGAdminReconciler) reconcilePGAdminStatefulSet(
...
+	// to be manually deleted. We now assign the Parallel Pod Management Policy, however, this is not a property
+	// that can be patched, so if we run into a StatefulSet from the previous PGO version that does not have the
+	// parallel policy in place, we must delete and recreate the StatefulSet.
+	existing := &appsv1.StatefulSet{ObjectMeta: naming.StandalonePGAdmin(pgadmin)}
+	if err := errors.WithStack(r.Client.Get(ctx, client.ObjectKeyFromObject(existing), existing)); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+	} else {
+		if existing.Spec.PodManagementPolicy != appsv1.ParallelPodManagement {
```
cbandy, line 49: 🤔 The other function is the one that chooses the parallel policy. Perhaps it would be tidier to generate the STS and compare the existing object to that one?

```go
	sts := statefulset(r, pgadmin, configmap, dataVolume)

	// Previous versions of PGO used a StatefulSet update strategy and policy that
	// could leave the pod in a failed state. When we see it has the wrong policy,
	// delete the StatefulSet while leaving its pods in place. They will be
	// claimed by the StatefulSet created during the next reconciliation.
	// - https://docs.k8s.io/concepts/workloads/controllers/statefulset/#forced-rollback
	var existing appsv1.StatefulSet
	err := r.Client.Get(ctx, client.ObjectKeyFromObject(sts), &existing)

	if err == nil && existing.Spec.PodManagementPolicy != sts.Spec.PodManagementPolicy {
		// delete, 🤔 always return early?
	}

	if err := errors.WithStack(r.setControllerReference(pgadmin, sts)); err != nil {
		return err
	}
	return errors.WithStack(r.apply(ctx, sts))
```

internal/controller/standalone_pgadmin/statefulset.go
```diff
@@ -33,6 +35,27 @@ func (r *PGAdminReconciler) reconcilePGAdminStatefulSet(
...
+			// We want to delete the STS without affecting the Pods, so we set the PropagationPolicy to Orphan.
+			// The orphaned Pods will be claimed by the new StatefulSet that gets created below.
+			propagate := client.PropagationPolicy(metav1.DeletePropagationOrphan)
+			if err = errors.WithStack(r.Client.Delete(ctx, existing, propagate)); err != nil {
```
cbandy, line 53: 🔧 There is a chance that the object existed when we did the Get (read) and has since gone away when we Delete (write). When we want the object gone, Not Found is okay, so we can use `client.IgnoreNotFound()`.

🔧 There is a chance that the object changed after we did the Get (read). When deleting an object because of something you saw on it, it is good to use preconditions to delete the object you saw rather than the one that might be there now. In the PostgresCluster reconciler, it looks like we used `deleteControlled` most of the time for this.

```go
		uid := existing.GetUID()
		version := existing.GetResourceVersion()
		exactly := client.Preconditions{UID: &uid, ResourceVersion: &version}
		propagate := client.PropagationPolicy(metav1.DeletePropagationOrphan)

		return errors.WithStack(
			client.IgnoreNotFound(r.Client.Delete(ctx, existing, exactly, propagate))
```

internal/controller/standalone_pgadmin/statefulset.go
dsessler7: Do we not care about the error if the return from `r.Client.Get` isn't `nil`? Obviously if it is a "NotFound" error then we don't care, but if it's a different error?

I didn't make it return early when deleting because I figured that we wanted to do `r.apply(ctx, sts)` regardless... Is it a bad idea to do both during one reconciliation cycle?

internal/controller/standalone_pgadmin/statefulset.go
```diff
@@ -80,6 +103,10 @@ func statefulset(
 		Partition: initialize.Int32(0),
 	}
 
+	// Set the StatefulSet Pod Management Policy to "Parallel" so that the StatefulSet can
+	// recreate the pod, if it initially has an issue (like an bad image).
+	sts.Spec.PodManagementPolicy = appsv1.ParallelPodManagement
```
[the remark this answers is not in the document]
dsessler7: I'm not sure why we set the `UpdateStrategy` to "RollingUpdate" since that is the default... Maybe in case the default ever changes?

internal/controller/standalone_pgadmin/statefulset.go
cbandy: > Do we not care about the error if the return from `r.Client.Get` isn't `nil`? Obviously if it is a "NotFound" error then we don't care, but if it's a different error?

I couldn't decide. I think I was being too clever. We don't want to have to think about it, so returning it is best.

> I didn't make it return early when deleting because I figured that we wanted to do `r.apply(ctx, sts)` regardless... Is it a bad idea to do both during one reconciliation cycle?

It might not be gone yet. 🤔 We don't have a policy of trying to "do just one thing," but maybe that would be good. I trust that controller-runtime will wake us back up when the DELETE event comes through.

internal/controller/standalone_pgadmin/statefulset.go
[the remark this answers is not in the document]
cbandy: I like having it here so we don't have to think about the default. We want the StatefulSet controller to deploy the pods a certain way, so we set those fields.

🤔 I feel a little differently about the `rollingUpdate` field, though I'm unsure why. The docs linked there say:

> In most cases you will not need to use a partition, but they are useful if you want to stage an update, roll out a canary, or perform a phased roll out.

We're not trying to do those.

cbandy (approved):

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -561,6 +583,10 @@ func (r *Reconciler) generateRepoHostIntent(postgresCluster *v1beta1.PostgresClu
 		repo.Spec.Replicas = initialize.Int32(1)
 	}
 
+	// Set the StatefulSet Pod Management Policy to "Parallel" so that the StatefulSet can
+	// recreate the pod, if it initially has an issue (like an bad image).
+	repo.Spec.PodManagementPolicy = appsv1.ParallelPodManagement
```
cbandy, line 588: ⛏️ I like the links to `#rolling-updates` and `#forced-rollback` in the other files. I'd also like to see them here, but it's not a blocker.

⛏️ Similarly, I'd like to see the `UpdateStrategy.Type` set here to clarify that the STS controller will roll out pod changes.

cbandy (approved):

PR #3817, opened 2024-01-05
Tweak restart logic for exporter

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Flakes in testing in one CI/CD environment.

This separates out the kill/restart logic for the exporter; previously, if a file changed, we would kill/restart. This led to some test flakes where the restart would happen too quickly (hypothesis) before the port was free. This PR separates the logic:

If the watched files change, kill the exporter;
If no exporter is running, start the exporter.

This also adds a check to the start_postgres_exporter func: save the PID to a file only if that proc is running.

**Other Information**:
Issue: [PGO-420]

### DIFF
--- a/internal/pgmonitor/exporter.go
+++ b/internal/pgmonitor/exporter.go
@@ -152,11 +152,8 @@ func ExporterStartCommand(builtinCollectors bool, commandFlags ...string) []stri
 		`  echo "Starting postgres_exporter with the following flags..."`,
 		`  echo "${postgres_exporter_flags[@]}"`,
 		`  postgres_exporter "${postgres_exporter_flags[@]}" &`,
-		`  echo $! > $POSTGRES_EXPORTER_PIDFILE`,
+		`  [ -e /proc/$! ] && echo $! > $POSTGRES_EXPORTER_PIDFILE`,
 		`}`,
-
-		// run function to combine queries files and start postgres_exporter
-		`start_postgres_exporter`,
 
 		// Create a file descriptor with a no-op process that will not get
 		// cleaned up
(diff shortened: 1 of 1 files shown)


jmckulk (approved):

internal/pgmonitor/exporter.go
```diff
@@ -152,12 +152,9 @@ func ExporterStartCommand(builtinCollectors bool, commandFlags ...string) []stri
 		`  echo "Starting postgres_exporter with the following flags..."`,
 		`  echo "${postgres_exporter_flags[@]}"`,
 		`  postgres_exporter "${postgres_exporter_flags[@]}" &`,
-		`  echo $! > $POSTGRES_EXPORTER_PIDFILE`,
+		`  [ -e /proc/$! ] && echo $! > $POSTGRES_EXPORTER_PIDFILE`,
```
cbandy, line 155: 🤔 What effect does this change have?

internal/pgmonitor/exporter.go
```diff
@@ -152,12 +152,9 @@ func ExporterStartCommand(builtinCollectors bool, commandFlags ...string) []stri
...
 		`}`,
 
-		// run function to combine queries files and start postgres_exporter
-		`start_postgres_exporter`,
```
cbandy, line 159: 🤔 It seems reasonable to me to start it up when the entrypoint first runs. Why remove it?

internal/pgmonitor/exporter.go
```diff
@@ -167,19 +164,23 @@ func ExporterStartCommand(builtinCollectors bool, commandFlags ...string) []stri
...
 		// and restart the loop
 		`    echo "Something changed..."`,
 		`    exec {fd}>&- && exec {fd}<> <(:)`,
 		`    stat --format='Latest queries file dated %y' "/conf"`,
 		`    stat --format='Latest password file dated %y' "/opt/crunchy/password"`,
 		`  fi`,
+
+		// If postgres_exporter is not running, restart it
+		// Use the recorded pid as a proxy for checking if postgres_exporter is running
+		`  if [ ! -e $POSTGRES_EXPORTER_PIDFILE ] || [ ! -e /proc/$(head -1 ${POSTGRES_EXPORTER_PIDFILE?}) ] ; then`,
```
cbandy, line 181: 🤔 If we leave the initial `start_postgres_exporter` and remove the conditional file write, then the PID file will always exist and we can drop the first condition here.

internal/pgmonitor/exporter.go
benjaminjb: Only writing the pidfile if the proc exists? It was something that came up as leading to the flakes we saw. Not sure if we need it with the below changes.

internal/pgmonitor/exporter.go
benjaminjb: That's the way I originally had it. I have no qualms about changing it back

benjaminjb: What do you think of it now?

internal/pgmonitor/exporter.go
cbandy: 🔧 The command above this succeeded, so the process in `$!` very likely exists. I think it's best to always write the PID file so that later commands can always read it.

cbandy (approved):

PR #3826, opened 2024-01-18
Bridge via PGO

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Create a PG cluster in Bridge via a K8s yaml

**Other Information**:
Issue: [PGO-814]


internal/bridge/managedpostgrescluster/managedpostgrescluster_controller.go
```diff
@@ -0,0 +1,552 @@
...
+		UpdateFunc: func(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
+			handle(e.ObjectNew, q)
+		},
+		// TODO(managedpostgrescluster): Do we want to reconcile ManagedPostgresClusters when the secret is deleted?
+		DeleteFunc: func(e event.DeleteEvent, q workqueue.RateLimitingInterface) {
+			handle(e.Object, q)
+		},
+	}
+}
+
```
benjaminjb, line 114: Needs kubebuilder annotation for Listing ManagedPostgresCluster (as an example)

internal/bridge/managedpostgrescluster/managedpostgrescluster_controller.go
```diff
@@ -0,0 +1,552 @@
...
+			Message: fmt.Sprintf(
+				"Cannot create with bad secret: %v", "TODO(managedpostgrescluster)"),
+		})
+
+		// Don't automatically requeue Secret issues
+		// We are watching for related secrets,
+		// so will requeue when a related secret is touched
+		return ctrl.Result{}, nil
+	}
+
```
benjaminjb, line 222: Should we check that the key is valid? Are there different levels of access in bridge? (Looks like API key doesn't have granular controls.)

internal/bridge/managedpostgrescluster/managedpostgrescluster_controller.go
```diff
@@ -0,0 +1,552 @@
...
+	// Wonder if there's a better way to handle adding/checking/removing statuses
+	// We did something in the upgrade controller
+	// Exit early if we can't create from this K8s object
+	// unless this K8s object has been changed (compare ObservedGeneration)
+	invalid = meta.FindStatusCondition(managedpostgrescluster.Status.Conditions,
+		v1beta1.ConditionCreating)
+	if invalid != nil &&
+		invalid.Status == metav1.ConditionFalse &&
+		invalid.Message == "ClusterInvalid" &&
```
benjaminjb, line 280: How do we know for sure that the cluster is invalid?

internal/bridge/managedpostgrescluster/managedpostgrescluster_controller.go
```diff
@@ -0,0 +1,552 @@
...
+		invalid.Status == metav1.ConditionFalse &&
+		invalid.Reason == "ClusterInvalid" {
+		meta.RemoveStatusCondition(&managedpostgrescluster.Status.Conditions,
+			v1beta1.ConditionCreating)
+	}
+
+	// We should only be missing the ID if no create has been issued
+	// or the create was interrupted and we haven't received the ID.
+	if managedpostgrescluster.Status.ID == "" {
+		// START FIND
```
benjaminjb, line 296: Consider annotation gate or some other block to adoption, so that we only allow it if intentional.

internal/bridge/managedpostgrescluster/managedpostgrescluster_controller.go
```diff
@@ -0,0 +1,552 @@
...
+		return ctrl.Result{}, err
+	}
+	managedpostgrescluster.Status.Cluster = clusterDetails
+
+	clusterUpgradeDetails, err := r.NewClient().GetClusterUpgrade(ctx, key, managedpostgrescluster.Status.ID)
+	if err != nil {
+		log.Error(err, "whoops, cluster upgrade getting issue")
+		return ctrl.Result{}, err
+	}
+	managedpostgrescluster.Status.ClusterUpgrade = clusterUpgradeDetails
```
benjaminjb, line 378: Do we want to keep the history of upgrades somewhere? On the status in a different field, or as events?

benjaminjb: Bridge Audit log:

Screenshot 2024-01-30 at 2 41 43 PM

internal/bridge/managedpostgrescluster/managedpostgrescluster_controller.go
```diff
@@ -0,0 +1,552 @@
...
+	if haDiff {
+		return r.handleUpgradeHA(ctx, key, managedpostgrescluster)
+	}
+
+	// Doy, use the result to get an accurate log
+	log.Info("Reconciled", "requeue", err != nil ||
+		result.Requeue ||
+		result.RequeueAfter > 0)
+	// Always requeue?
+	return ctrl.Result{RequeueAfter: 3 * time.Minute}, nil
```
benjaminjb, line 419: Maybe doing this means we don't need the Watch?

internal/bridge/managedpostgrescluster/managedpostgrescluster_controller.go
```diff
@@ -0,0 +1,524 @@
...
+// Returns an error if the int is outside the range for Bridge min (10) or max (65535).
+func handleStorage(storageSpec resource.Quantity) (int64, error) {
+	scaledValue := storageSpec.ScaledValue(resource.Giga)
+
+	if scaledValue < 10 || scaledValue > 65535 {
+		return 0, fmt.Errorf("storage value must be between 10 and 65535")
+	}
+
+	return scaledValue, nil
+}
```
benjaminjb, line 442: @cbandy PGO-884

internal/bridge/managedpostgrescluster/managedpostgrescluster_controller.go
```diff
@@ -0,0 +1,524 @@
...
+				return ctrl.Result{Requeue: true}, nil
+			}
+		}
+
+		// END FIND
+
+		// if we've gotten here then no cluster exists with that name and we're missing the ID, ergo, create cluster
+
+		// TODO(managedpostgrescluster) Can almost just use the managed.Spec... except for the team, which we don't want
+		// users to set on the spec. Do we?
```
dsessler7, line 342: I was wondering about this myself... Not sure the team ID really needs to be kept secret. And now that we know that API Keys can have access to multiple teams, if we move the Team ID to the spec, then users don't need a separate Secret for each team.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/managed_postgrescluster_types.go
```diff
@@ -0,0 +1,204 @@
...
+	Plan            string             `json:"plan_id,omitempty"`
+	MajorVersion    int                `json:"major_version,omitempty"`
+	PostgresVersion intstr.IntOrString `json:"postgres_version_id,omitempty"`
+	Provider        string             `json:"provider_id,omitempty"`
+	Region          string             `json:"region_id,omitempty"`
+	Storage         int64              `json:"storage,omitempty"`
+	Team            string             `json:"team_id,omitempty"`
+	State           string             `json:"state,omitempty"`
+	// TODO(managedpostgrescluster): add other fields, DiskUsage, Host, IsProtected, IsSuspended, CPU, Memory, etc.
+}
```
dsessler7, line 130: Does this belong here? :thinking: Feels more like an intermediary struct we use for interacting with the Bridge API rather than part of the `managedpostgrescluster` API...

dsessler7 (approved):

.github/workflows/test.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
-          go-version: 1.x
+          go-version: '1.19'
```
benjaminjb, line 18: I'm not happy with this.

internal/bridge/managedpostgrescluster/managedpostgrescluster_controller.go
benjaminjb: Ticket made for that, so resolving this

internal/bridge/managedpostgrescluster/managedpostgrescluster_controller.go
benjaminjb: Created a ticket

pkg/apis/postgres-operator.crunchydata.com/v1beta1/managed_postgrescluster_types.go
benjaminjb: We've got a ticket, so resolving

internal/bridge/managedpostgrescluster/managedpostgrescluster_controller.go
dsessler7: Maybe add a TODO questioning whether we want this or the Watch or both?

.github/workflows/test.yaml
dsessler7: What are you not happy with?

benjaminjb: Just to have a record here: the `go-test` action was failing, I think in the `check-generate` action, and it seemed to be an issue with the Go version, so I pinned it to what we have in our go.mod. But I didn't like that this was the only PR that was failing that test, and I don't think we want to pin our Go here (since we don't pin it elsewhere).

PR #3840, opened 2024-02-07
Add Crunchy Bridge Cluster adoption annotation logic.

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Currently, a CrunchyBridgeCluster CR will automatically adopt an existing Bridge Cluster if they have the same name.

Now, the CrunchyBridgeCluster CR will not adopt an existing Bridge Cluster with the same name unless a specific annotation is applied to the CR with the cluster ID as the value. Detailed conditions and log messages will be emitted if the proper annotation is not in place.

**Other Information**:

### DIFF
--- a/internal/bridge/crunchybridgecluster/crunchybridgecluster_controller.go
+++ b/internal/bridge/crunchybridgecluster/crunchybridgecluster_controller.go
@@ -17,6 +17,7 @@ package crunchybridgecluster
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -38,6 +39,7 @@ import (
 
 	"github.com/crunchydata/postgres-operator/internal/bridge"
 	pgoRuntime "github.com/crunchydata/postgres-operator/internal/controller/runtime"
+	"github.com/crunchydata/postgres-operator/internal/naming"
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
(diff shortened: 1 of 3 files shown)


ValClarkson (approved):

internal/bridge/crunchybridgecluster/crunchybridgecluster_controller.go
```diff
@@ -328,9 +330,53 @@ func (r *CrunchyBridgeClusterReconciler) Reconcile(ctx context.Context, req ctrl
 
 		for _, cluster := range clusters {
 			if crunchybridgecluster.Name == cluster.Name {
```
cbandy, line 332: ❓ Should this be `crunchybridgecluster.Spec.ClusterName`?

internal/bridge/crunchybridgecluster/crunchybridgecluster_controller.go
```diff
@@ -328,9 +330,53 @@ func (r *CrunchyBridgeClusterReconciler) Reconcile(ctx context.Context, req ctrl
...
+				adoptionID, annotationExists := crunchybridgecluster.Annotations[naming.CrunchyBridgeClusterAdoptionAnnotation]
+				if strings.EqualFold(adoptionID, cluster.ID) {
+					// Annotation is present with correct ID value; adopt cluster by assigning ID to status.
+					crunchybridgecluster.Status.ID = cluster.ID
+					// Requeue now that we have a cluster ID assigned
+					return ctrl.Result{Requeue: true}, nil
+				} else {
+					// Adoption annotation value is incorrect. Check to see if annotation exists
+					// to determine proper log message and status condition
+					if annotationExists {
```
cbandy, line 343: 🤔 This branch doesn't seem necessary to me. The longer message on L360 has the information from the API call. If the user can see `status` they can also see `annotations` and the mismatch.

internal/naming/annotations_test.go
```diff
@@ -31,4 +31,5 @@ func TestAnnotationsValid(t *testing.T) {
 	assert.Assert(t, nil == validation.IsQualifiedName(PGBackRestRestore))
 	assert.Assert(t, nil == validation.IsQualifiedName(PGBackRestIPVersion))
 	assert.Assert(t, nil == validation.IsQualifiedName(PostgresExporterCollectorsAnnotation))
+	assert.Assert(t, nil == validation.IsQualifiedName(CrunchyBridgeClusterAdoptionAnnotation))
```
cbandy, line 34: 👍🏻

internal/bridge/crunchybridgecluster/crunchybridgecluster_controller.go
benjaminjb:
```suggestion
			if crunchybridgecluster.Spec.ClusterName == cluster.Name {
```
Yeah, this code was in place before we separated out the name of the CR from the name of the cluster.

I think we separated them out in order to make the distinction between the CR requirements and the bridge requirements for names. (Also there's differing constraints around uniqueness in k8s / Bridge.)

dsessler7: I'm partial to the idea of just having one name and enforcing constraints for both bridge and k8s when creating the CR... Only scenario that I can come up with where this is problematic is if the customer is adopting a pre-existing cluster in bridge that has a name that conflicts with certain kubernetes constraints, but if that's the case I guess we just make the customer change the name of the cluster in bridge?

cbandy: The K8s name and Bridge name are different ideas with different constraints, so we need separate fields.

We might be able to enforce some similarities or default-ish behavior using cross-field CEL validation in the future. In that case, the `spec.clusterName` field could become optional when the K8s name is also a valid Bridge name.

cbandy: 🤔 Notably, K8s names are immutable and Bridge names are mutable.

cbandy (approved):

PR #3848, opened 2024-02-14
Cleanup EnvTest binaries more aggressively

cbandy:
I had trouble running "make clean" while switching between Linux and macOS in the same working directory.


dsessler7 (approved):
LGTM

PR #3854, opened 2024-02-16
Allow customers to specify roles that they want credentials for...

dsessler7:
in the CrunchyBridgeCluster spec. When specified, PGO will create/update a corresponding Secret and fill with role name, password, and connection URI during the reconcile loop. If role is deleted from spec or secret name is changed, PGO will delete unused secret.

**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

There is no user/role functionality.

The customer can now add roles (that correspond to available roles in Bridge) to their spec and specify a secret name and get a Secret in their k8s cluster that contains that role's password and connection URI.

**Other Information**:


internal/bridge/crunchybridgecluster/postgres.go
```diff
@@ -0,0 +1,149 @@
...
+	// if err == nil {
+	// 	err = r.reconcilePostgresUsersInPostgreSQL(ctx, cluster, instances, users, secrets)
+	// }
+	// if err == nil {
+	// 	// Copy PostgreSQL users and passwords into pgAdmin. This is here because
+	// 	// reconcilePostgresRoleSecrets is building a (default) PostgresUserSpec
+	// 	// that is not in the PostgresClusterSpec. The freshly generated Secrets
+	// 	// are available here, too.
+	// 	err = r.reconcilePGAdminUsers(ctx, cluster, users, secrets)
+	// }
```
ValClarkson, line 81: :question: looks like the above comments should be removed?

ValClarkson (approved):

pkg/apis/postgres-operator.crunchydata.com/v1beta1/crunchy_bridgecluster_types.go
```diff
@@ -79,6 +88,21 @@ type CrunchyBridgeClusterSpec struct {
...
 }
 
+type CrunchyBridgeClusterRoleSpec struct {
+	// The name of this PostgreSQL role. The value may contain only lowercase
+	// letters, numbers, and hyphen so that it fits into Kubernetes metadata.
+	// The above is problematic for us as Bridge has a role with an underscore.
+	// TODO: figure out underscore dilemma
+	// +kubebuilder:validation:Pattern=`^[A-Za-z][A-Za-z0-9\-_ ]*[A-Za-z0-9]$`
+	// +kubebuilder:validation:Type=string
+	Name string `json:"name"`
```
cbandy, line 98: ❓ Is this effectively the `name` field of [ClusterRole](https://docs.crunchybridge.com/api/cluster-role#cluster-role)? Can we drop regex validation and let the API tell us what's valid?

```suggestion
	// Name of the role within Crunchy Bridge.
	// More info: https://docs.crunchybridge.com/concepts/users
	Name string `json:"name"`
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/crunchy_bridgecluster_types.go
```diff
@@ -79,6 +88,21 @@ type CrunchyBridgeClusterSpec struct {
...
+
+	// The name of the Secret that will hold the role credentials.
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+	// +kubebuilder:validation:Type=string
+	SecretName string `json:"secretName"`
```
cbandy, line 103: ❓ Secret names are DNS subdomains. Is there a validation rule/shortcut for that?

🤔 We can start with a more limited rule (the one here) and expand it later.

dsessler7: According to [this](https://github.com/CrunchyData/postgres-operator/blob/master/config/crd/bases/postgres-operator.crunchydata.com_crunchybridgeclusters.yaml#L232-L234), the regex we want is `[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`

dsessler7: I would think that the `kubebuilder:validation:Format` marker would accept an `IsDNS1123Subdomain` option given that the kubebuilder codebase has functions written specifically for validating this, but it doesn't seem to work and I can't find any documentation for what options they actually accept for the "Format" validation...

ValClarkson (approved):

pkg/apis/postgres-operator.crunchydata.com/v1beta1/crunchy_bridgecluster_types.go
cbandy: AFAIK, `format` comes from OpenAPIv3. I see references to JSON Schema Validation elsewhere. I'm also not sure what Kubernetes has implemented (or not).

- https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation →
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#schema-object →
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#dataTypeFormat

- https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules →
https://swagger.io/specification/#data-types →
https://tools.ietf.org/html/draft-bhutton-json-schema-validation-00#section-7.3
  <br/>`hostname` looks close, but I'm not excited by following 3–5 RFCs down to the early internet. :notsure:

I see the same regex you pasted when I look [here](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Subdomain). So that plus max-length sounds good to me.

internal/bridge/crunchybridgecluster/crunchybridgecluster_controller.go
```diff
@@ -403,38 +420,36 @@ func (r *CrunchyBridgeClusterReconciler) Reconcile(ctx context.Context, req ctrl
...
-	// Consider cluster details as a separate func.
+	// If we reach this point, our CrunchyBridgeCluster object has an ID, so we want
+	// to fill in the details for the cluster, cluster status, and cluster upgrades
+	// from the Bridge API.
 
+	// Get Cluster
 	clusterDetails, err := r.NewClient().GetCluster(ctx, key, crunchybridgecluster.Status.ID)
 	if err != nil {
-		log.Error(err, "whoops, cluster getting issue")
+		log.Error(err, "whoops, issue getting cluster")
```
benjaminjb, line 430: Might be time to take out the "whoops" from the logging

internal/bridge/client.go
```diff
@@ -45,6 +46,172 @@ type Client struct {
...
+type ClusterDiskUsageApiResource struct {
+	DiskAvailableMB int64 `json:"disk_available_mb,omitempty"`
+	DiskTotalSizeMB int64 `json:"disk_total_size_mb,omitempty"`
+	DiskUsedMB      int64 `json:"disk_used_mb,omitempty"`
+}
+
+// ClusterGroupApiResource holds information on a ClusterGroup
+type ClusterGroupApiResource struct {
+	ID       string                `json:"id,omitempty"`
+	Clusters []*ClusterApiResource `json:"clusters,omitempty"`
```
benjaminjb, line 109: `ClusterApiResource.ClusterGroup.Clusters` == `ClusterApiResource`, yeah? Just without any clustergroup info of their own? (I never hit this part of the API)

internal/bridge/crunchybridgecluster/crunchybridgecluster_controller.go
```diff
@@ -85,6 +86,22 @@ func (r *CrunchyBridgeClusterReconciler) SetupWithManager(
 		Complete(r)
 }
 
+// The owner reference created by controllerutil.SetControllerReference blocks
+// deletion. The OwnerReferencesPermissionEnforcement plugin requires that the
+// creator of such a reference have either "delete" permission on the owner or
+// "update" permission on the owner's "finalizers" subresource.
+// - https://docs.k8s.io/reference/access-authn-authz/admission-controllers/
+// +kubebuilder:rbac:groups="postgres-operator.crunchydata.com",resources="pgupgrades/finalizers",verbs={update}
```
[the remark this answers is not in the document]
dsessler7: Yep, I actually have that fixed in a local commit that I haven't pushed yet.

internal/bridge/crunchybridgecluster/crunchybridgecluster_controller.go
```diff
@@ -445,23 +460,23 @@ func (r *CrunchyBridgeClusterReconciler) Reconcile(ctx context.Context, req ctrl
...
 		return ctrl.Result{RequeueAfter: 3 * time.Minute}, nil
 	}
 
 	// Check if there's an upgrade difference for the three upgradeable fields that hit the upgrade endpoint
 	// Why PostgresVersion and MajorVersion? Because MajorVersion in the Status is sure to be
-	// an int of the major version, whereas Status.Cluster.PostgresVersion might be the ID
-	if (storageVal != crunchybridgecluster.Status.Cluster.Storage) ||
-		crunchybridgecluster.Spec.Plan != crunchybridgecluster.Status.Cluster.Plan ||
-		crunchybridgecluster.Spec.PostgresVersion != crunchybridgecluster.Status.Cluster.MajorVersion {
+	// an int of the major version, whereas Status.Responses.Cluster.PostgresVersion might be the ID
```
benjaminjb, line 469:
```suggestion
	// an int of the major version, whereas Status.PostgresVersion might be the ID
```

internal/bridge/client.go
```diff
@@ -45,6 +46,172 @@ type Client struct {
...
+// changing its plan, upgrading its major version, or increasing its storage size.
+type PostClustersUpgradeRequestPayload struct {
+	Plan             string             `json:"plan_id,omitempty"`
+	PostgresVersion  intstr.IntOrString `json:"postgres_version_id,omitempty"`
+	UpgradeStartTime string             `json:"starting_from,omitempty"`
+	Storage          int64              `json:"storage,omitempty"`
+}
+
+// PutClustersUpgradeRequestPayload is used for updating an ongoing or scheduled upgrade.
+type PutClustersUpgradeRequestPayload struct {
```
benjaminjb, line 188: Are we going to be doing this?

internal/bridge/crunchybridgecluster/postgres.go
```diff
@@ -0,0 +1,143 @@
...
+		err = errors.WithStack(
+			r.Client.List(ctx, secrets,
+				client.InNamespace(cluster.Namespace),
+				client.MatchingLabelsSelector{Selector: selector},
+			))
+	}
+
+	// Index secrets by PostgreSQL role name and delete any that are not in the
+	// cluster spec.
+	roleSecrets := make(map[string]*corev1.Secret, len(secrets.Items))
```
benjaminjb, line 106: I'm curious about the use of this map of secrets. 

* You use it to store a secret if it exists as a K8s object and should exist (line 114)
* You use it to store a secret before it gets created (line 132)
* And then you use it to return all the secrets (line 142)

But when it gets returned, it might have secrets that were generated but not created, right? Like if there's an error on creation, it's still in the map.

Or, taking a step back, since we're not using that list of secrets that gets returned, maybe we don't need it? 

Like: 
1. we iterate over the list of secrets returned; 
2. if there's a secret in that list that doesn't match the spec.roles, we can delete it; 
3. if there's a secret that does match the spec.roles, we can skip it; 
4. and if there's spec.roles that don't exist yet as secrets, we create it.

internal/naming/selectors.go
```diff
@@ -145,3 +145,16 @@ func ClusterPrimary(cluster string) metav1.LabelSelector {
...
+
+// CrunchyBridgeClusterPostgresRoles selects things labeled for CrunchyBridgeCluster
+// PostgreSQL roles in cluster.
+func CrunchyBridgeClusterPostgresRoles(cluster string) metav1.LabelSelector {
+	return metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			LabelCluster: cluster,
+		},
+		MatchExpressions: []metav1.LabelSelectorRequirement{
+			{Key: LabelCrunchyBridgeClusterPostgresRole, Operator: metav1.LabelSelectorOpExists},
```
benjaminjb, line 157: This seems fine to me, though might it be simpler to add
```
naming.LabelRole: RoleCrunchyBridgeClusterPostgresRole
```
to the MatchLabels?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/crunchy_bridgecluster_types.go
```diff
@@ -79,70 +87,92 @@ type CrunchyBridgeClusterSpec struct {
...
 
+type CrunchyBridgeClusterRoleSpec struct {
+	// Name of the role within Crunchy Bridge.
+	// More info: https://docs.crunchybridge.com/concepts/users
+	Name string `json:"name"`
+
+	// The name of the Secret that will hold the role credentials.
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	// +kubebuilder:validation:Type=string
+	SecretName string `json:"secretName"`
```
benjaminjb, line 98: I think there was some discussion around why we want to let them name the secret -- maybe something to do with the role names from Bridge? If so, might be worth a code comment.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/crunchy_bridgecluster_types.go
```diff
@@ -79,70 +87,92 @@ type CrunchyBridgeClusterSpec struct {
...
 
-	// The cluster upgrade as represented by Bridge
+	// Most recent, raw responses from Bridge API
 	// +optional
-	ClusterUpgrade *ClusterUpgrade `json:"clusterUpgradeResponse,omitempty"`
-}
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:validation:Type=object
+	Responses APIResponses `json:"responses"`
```
benjaminjb, line 154: Ah, ok, so we use this to shove the raw response from cluster, status, upgrade?

internal/bridge/client.go
dsessler7: A slice of ClusterApiResource, yeah.

internal/bridge/crunchybridgecluster/crunchybridgecluster_controller.go
dsessler7: As of right now we only have MajorVersion on the status since that is what PGO cares about.

internal/bridge/client.go
dsessler7: Not sure. Maybe at some point. I guess this is just there as a placeholder.

internal/bridge/crunchybridgecluster/postgres.go
```diff
@@ -0,0 +1,143 @@
...
+			continue
+		}
+		if err == nil {
+			roleSecrets[roleName], err = r.generatePostgresRoleSecret(cluster, role, clusterRole)
+		}
+		if err == nil {
+			err = errors.WithStack(r.apply(ctx, roleSecrets[roleName]))
+		}
+		if err != nil {
+			log.Error(err, "Issue creating role secret.")
```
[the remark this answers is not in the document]
dsessler7: I figured that the error in `err` would contain information on exactly what failed, although I don't know for sure what it would contain...

pkg/apis/postgres-operator.crunchydata.com/v1beta1/crunchy_bridgecluster_types.go
dsessler7: Yep.

internal/bridge/crunchybridgecluster/postgres.go
dsessler7: So, a lot of this was modeled after the user/secret creation for `postgresclusters`. You make a good point that we add the generated secret to the map before ensuring that it was successfully created in k8s which seems wrong. I can make a change to only add it to the map after we confirm that it is created in k8s (this should be changed for `postgrescluster` too).

Also good point that as of right now, we don't really care what the map contains since we don't use it, although that could change in the future if we add pgAdmin support or something similar.

In your numbered outline, I don't think we want to do #3 (skip a secret if it's already there) because we want to refresh it in case the password/URI string has changed.

internal/bridge/crunchybridgecluster/postgres.go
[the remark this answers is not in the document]
benjaminjb: oh, doy on my part, we're logging the err, that's probably got the info I'm looking for, right?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/crunchy_bridgecluster_types.go
benjaminjb: OK, from chatting with Drew: we can't use role as the secret name because of disallowed characters/different constraints around role names and secret names, so we want to define the secret on the pgo side; and rather than define our secret pattern in PGO, we will let the users define their own roles. 

(I am commenting here to keep this comment _out_ of the CRD.)

(And as discussed with Drew, letting users define secret names means we need to have some guards against duplicate names within the namespace (whether those secret names come from one or multiple clusters).)

internal/bridge/crunchybridgecluster/postgres.go
dsessler7: Spoke to Chris about this... The convention is that if there is an error, we don't pay attention to the other return values. So it essentially doesn't matter that the map has a secret that wasn't successfully created in k8s because the failed k8s creation returns an error that negates everything else.

internal/bridge/quantity.go
```diff
@@ -0,0 +1,53 @@
...
+
+package bridge
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func FromCPU(n int64) *resource.Quantity {
```
benjaminjb, line 22: We're not using this right now, are we?

internal/bridge/crunchybridgecluster/postgres.go
```diff
@@ -0,0 +1,174 @@
...
+	specRoles := cluster.Spec.Roles
+
+	// Index role specifications by PostgreSQL role name and make sure that none of the
+	// secretNames are identical in the spec
+	secretNames := make(map[string]bool)
+	roleSpecs := make(map[string]*v1beta1.CrunchyBridgeClusterRoleSpec, len(specRoles))
+	for i := range specRoles {
+		if secretNames[specRoles[i].SecretName] {
+			// Duplicate secretName found, return early with error
+			err := errors.New("There are duplicate Role SecretNames in the spec. SecretNames must be unique.")
```
benjaminjb, line 95: I have a vague memory of us preferring `fmt.Errorf`, let me check on that...

Yeah, I'm seeing more of that.

benjaminjb (commented):
This all makes sense to me, but I do wish we had some other solution for the secret name only because it seems kind of cumbersome for the operator to

(a) check for secret name uniqueness on the spec,
(b) check for secret name uniqueness in the namespace (ignoring secrets with certain labels -- heck, you could include that in this K8s API list action: list secrets that don't have certain selectors),
(c) and then get the secrets for this cluster to compare against the roles requested.

But I can't really see a way around it. My only other thoughts are 

(a) "let's use a UUID" -- which make the secrets hard to use;
(b) "let's transform the disallowed characters from the PG role into allowed characters for K8s" -- which might just collide with another secret. (Unless we add the cluster as part of the secret, and even then, that's not 100% proof against collisions.)

So yeah, I have no other solutions, so won't block on this.

internal/bridge/crunchybridgecluster/postgres.go
dsessler7: Okay, I can switch to using that.

internal/bridge/quantity.go
dsessler7: No, we're not. It would be useful if we ever decide to include CPU on the status... Not sure if I should keep it or just delete it for now.

benjaminjb: Yeah, I'm on the fence...

internal/bridge/crunchybridgecluster/postgres.go
dsessler7: Although when I search the repo I see 74 results for `errors.New` and 55 results for `fmt.Errorf`...

benjaminjb (approved):

internal/bridge/crunchybridgecluster/postgres.go
dsessler7: Probably just the preference of whoever was writing the code...

PR #3873, opened 2024-03-18
feat: generate postgres user secret with r2dbc uris

bullshit:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

The operator generates user secrets. For example if you use a r2dbc based application you cannot reference the jdbc-uri key from the secret.

r2dbc-uri keys are added to the secrets starting with `r2dbc:postgresql://....`

**Other Information**:

### DIFF
--- a/internal/controller/postgrescluster/postgres.go
+++ b/internal/controller/postgrescluster/postgres.go
@@ -128,6 +128,14 @@ func (r *Reconciler) generatePostgresUserSecret(
 			Path:     database,
 			RawQuery: query.Encode(),
 		}).String())
+		// The R2DBC driver requires a different URI scheme than the JDBC driver
+		// - https://r2dbc.io/spec/1.0.0.RELEASE/spec/html/#overview.connection.url
+		intent.Data["r2dbc-uri"] = []byte((&url.URL{
+			Scheme:   "r2dbc:postgresql",
+			Host:     net.JoinHostPort(hostname, port),
+			Path:     database,
+			RawQuery: query.Encode(),
+		}).String())
 	}
 
 	// When PgBouncer is enabled, include values for connecting through it.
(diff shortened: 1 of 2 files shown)


bullshit:
Any feedback regarding this PR?  Maybe @dsessler7 could approve this change.

bullshit:
Is there any reason this contribution is stagnates?

andrewlecuyer:
Hi @bullshit, I was actually just reviewing your changes yesterday.  And as we continue to see requests for additional flavors/variants of the connection details within the user Secret (such as additional URI variants, e.g. for `r2dbc`, in addition  to different flavors of the various keys within the Secret, e.g. user vs. username, or db vs. database vs. dbname, etc.),
I am currently assessing options that might provide even more flexibility for the many connectivity use-cases/needs we see out in the wild.

So in other words, rather than endlessly polluting the Secret with access keys, URI's, etc. for that many different variants out there, if would be great to provide users with the ability to define (via the PostgresCluster spec) the exact keys, URI's, etc. they need.  And right now I have a couple ideas/designs that I would like to think through before adding another yet key to the Secret.  E.g. here is a POC for what something like this might looks like:

```yaml
spec:
  users:
    - name: rhino
      secret:
        credentialMappings:
          user:
          - username
          - uname
          password:
          - pass
          - pwd
          - passwd
          dbname:
          - database
          - databasename
          - db
          host:
          - hostname
          - hostName
          port:
          - pgPort
        customURIs:
        # user has access to "user", "password", "host" & "port" to build/template out
        # any custom URI needed
        - r2dbc: "r2dbc:postgres://$(user):$(password)@$(host):$(port)/$(dbname)
```

Which as you cam see, would be another way to get you the `r2dbc` URI you're looking for:

```yaml
$ kubectl get secret -n postgres-operator hippo-pguser-someone -o yaml
apiVersion: v1
data:
...
  r2dbc: cjJkYmM6Ly9zb21lb25lOkh2OSUzRitfX1d4PWlWSHhSWjY5eVlZM3dWQGhpcHBvLXByaW1hcnkucG9zdGdyZXMtb3BlcmF0b3Iuc3ZjOjU0MzIvc29tZW9uZQ==
...
```

I welcome any thoughts/feedback.

PR #3881, opened 2024-04-04
Consolidate registration logic and add more tests

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement
 - [x] Other

**Other Information**:

Issue: PGO-1105


internal/controller/postgrescluster/controller.go
```diff
@@ -213,35 +210,11 @@ func (r *Reconciler) Reconcile(
...
-
-		if registrationRequiredStatusFound(cluster) {
-			cluster.Status.RegistrationRequired = nil
-			r.Recorder.Event(cluster, corev1.EventTypeNormal, "Token Verified", "Thank you for registering your installation of Crunchy Postgres for Kubernetes.")
-		}
+	if r.Registration != nil && r.Registration.Required(r.Recorder, cluster, &cluster.Status.Conditions) {
+		registration.SetAdvanceWarning(r.Recorder, cluster, &cluster.Status.Conditions)
 	}
+	cluster.Status.RegistrationRequired = nil
+	cluster.Status.TokenRequired = ""
```
tony-landreth, line 217: Will these lines eventually go away?

internal/registration/runner.go
```diff
@@ -0,0 +1,195 @@
...
+		defer ticker.Stop()
+		ticks = ticker.C
+	}
+
+	log := logging.FromContext(ctx).WithValues("controller", "registration")
+
+	for {
+		select {
+		case <-ticks:
+			_, before := r.state()
```
tony-landreth, line 180: :question: How does the Runner's state persist across a restart? Or does it have to?

internal/controller/postgrescluster/controller.go
cbandy: Eventually. I noticed some odd behavior with `managedFields` when removing these from the CRD. We should zero them out (while in the CRD) before removing them entirely.

internal/registration/runner.go
cbandy: The state is loaded once during startup before the Manager calls this function:

https://github.com/CrunchyData/postgres-operator/blob/3315e0cb3bbb70b12bf635758c57b1b0ac1bdad3/cmd/postgres-operator/main.go#L105

tony-landreth (approved):

PR #3884, opened 2024-04-10
Add shared functions for Kubernetes in tests

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

### DIFF
--- a/internal/bridge/crunchybridgecluster/helpers_test.go
+++ b/internal/bridge/crunchybridgecluster/helpers_test.go
@@ -18,23 +18,17 @@ package crunchybridgecluster
 import (
 	"context"
 	"os"
-	"path/filepath"
 	"strconv"
-	"strings"
-	"sync"
 	"testing"
 	"time"
 
-	"gotest.tools/v3/assert"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	"sigs.k8s.io/yaml"
 
 	"github.com/crunchydata/postgres-operator/internal/bridge"
-	"github.com/crunchydata/postgres-operator/internal/controller/runtime"
 	"github.com/crunchydata/postgres-operator/internal/initialize"
 	"github.com/crunchydata/postgres-operator/internal/testing/require"
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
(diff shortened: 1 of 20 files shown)


tjmoore4 (approved):

internal/testing/require/kubernetes.go
```diff
@@ -0,0 +1,173 @@
...
+// function to ensure they share.
+func Kubernetes(t testing.TB) client.Client {
+	t.Helper()
+	_, cc := kubernetes3(t)
+	return cc
+}
+
+// Kubernetes2 is the same as [Kubernetes] but also returns a copy of the client
+// configuration.
+func Kubernetes2(t testing.TB) (*rest.Config, client.Client) {
```
dsessler7, line 81: `Kubernetes2` feels little strange to me... Maybe we could rename it `KubernetesAndClientConfig` or something?

internal/bridge/crunchybridgecluster/helpers_test.go
```diff
@@ -59,102 +54,41 @@ func init() {
...
-// "/config/crd/bases" directory. When any of these fail, it calls t.Fatal.
-// It deletes CRDs and stops the local API using t.Cleanup.
-//
-// This function was duplicated from the postgrescluster package.
-// TODO: Pull these duplicated functions out into a separate, shared package.
-//
-//nolint:unparam
-func setupKubernetes(t testing.TB) (*envtest.Environment, client.Client) {
+// that uses it. See [require.Kubernetes] for more details.
+func setupKubernetes(t testing.TB) client.Client {
```
dsessler7, line 59: Could we pull this function and `setupNamespace` out into a helper package rather than having them defined multiple times in different packages? :thinking:

internal/testing/require/kubernetes.go
```diff
@@ -0,0 +1,173 @@
...
+	t.Helper()
+	env, cc := kubernetes3(t)
+	return rest.CopyConfig(env.Config), cc
+}
+
+func kubernetes3(t testing.TB) (*envtest.Environment, client.Client) {
```
dsessler7, line 87: Same here... :shrug:

internal/bridge/crunchybridgecluster/helpers_test.go
cbandy: Yep, but not me in this PR.

The namespace dumping wasn't helpful for the tests I was writing, so I like having it be optional.

`setupNamespace` is effectively moved to `require.Namespace`, but I didn't want to edit all those calls.

internal/testing/require/kubernetes.go
cbandy: I started with this exported because a couple `setupKubernetes` had this shape. After seeing everything, it wasn't much to change those places to `Kubernetes2` and unexport this.

cbandy: I was glad to unexport it because I have no idea how to safely share an `envtest.Environment` value or what concurrency is okay on it. `Kubernetes2` returns a copy, which seems safe so far.

internal/testing/require/kubernetes.go
cbandy: This name isn't great, but I'm gonna leave it.

PR #3886, opened 2024-04-12
pgAdmin Gunicorn hosting

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

This update updates the namespace scoped pgAdmin implementation to use Gunicorn for hosting. Server configuration is available via the PGAdmin manifest under spec.config.gunicorn.

**Other Information**:
Issue: PGO-546

### DIFF
--- a/internal/controller/standalone_pgadmin/configmap.go
+++ b/internal/controller/standalone_pgadmin/configmap.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
+	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 
@@ -74,6 +75,11 @@ func configmap(pgadmin *v1beta1.PGAdmin,
 	clusterSettings, err := generateClusterConfig(clusters)
 	if err == nil {
 		configmap.Data[settingsClusterMapKey] = clusterSettings
+	}
+
+	gunicornSettings, err := generateGunicornConfig(pgadmin)
+	if err == nil {
+		configmap.Data[gunicornConfigKey] = gunicornSettings
 	}
 
 	return configmap, err
@@ -179,5 +185,38 @@ func generateClusterConfig(
 		"Servers": clusterServers,
 	}
 	err := encoder.Encode(servers)
+	return buffer.String(), err
+}
+
+// generateGunicornConfig generates the config settings for the gunicorn server
+// - https://docs.gunicorn.org/en/latest/settings.html
+func generateGunicornConfig(pgadmin *v1beta1.PGAdmin) (string, error) {
+	settings := map[string]any{
+		// Bind to all IPv4 addresses and set 25 threads by default.
+		// - https://docs.gunicorn.org/en/latest/settings.html#bind
+		// - https://docs.gunicorn.org/en/latest/settings.html#threads
+		"bind":    "0.0.0.0:" + strconv.Itoa(pgAdminPort),
+		"threads": 25,
+	}
+
+	// Copy any specified settings over the defaults.
+	for k, v := range pgadmin.Spec.Config.Gunicorn {
+		settings[k] = v
+	}
+
+	// Write mandatory settings over any specified ones.
+	// - https://docs.gunicorn.org/en/latest/settings.html#workers
+	settings["workers"] = 1
+
+	// To avoid spurious reconciles, the following value must not change when
+	// the spec does not change. [json.Encoder] and [json.Marshal] do this by
+	// emitting map keys in sorted order. Indent so the value is not rendered
+	// as one long line by `kubectl`.
+	buffer := new(bytes.Buffer)
+	encoder := json.NewEncoder(buffer)
+	encoder.SetEscapeHTML(false)
+	encoder.SetIndent("", "  ")
+	err := encoder.Encode(settings)
+
 	return buffer.String(), err
 }
--- a/internal/controller/standalone_pgadmin/pod.go
+++ b/internal/controller/standalone_pgadmin/pod.go
@@ -28,10 +28,11 @@ import (
 )
 
 const (
-	configMountPath = "/etc/pgadmin/conf.d"
-	configFilePath  = "~postgres-operator/" + settingsConfigMapKey
-	clusterFilePath = "~postgres-operator/" + settingsClusterMapKey
-	ldapFilePath    = "~postgres-operator/ldap-bind-password"
+	configMountPath        = "/etc/pgadmin/conf.d"
+	configFilePath         = "~postgres-operator/" + settingsConfigMapKey
+	clusterFilePath        = "~postgres-operator/" + settingsClusterMapKey
+	ldapFilePath           = "~postgres-operator/ldap-bind-password"
+	gunicornConfigFilePath = "~postgres-operator/" + gunicornConfigKey
 
 	// Nothing should be mounted to this location except the script our initContainer writes
 	scriptMountPath = "/etc/pgadmin"
@@ -210,6 +211,10 @@ func podConfigFiles(configmap *corev1.ConfigMap, pgadmin v1beta1.PGAdmin) []core
 							Key:  settingsClusterMapKey,
 							Path: clusterFilePath,
 						},
+						{
+							Key:  gunicornConfigKey,
+							Path: gunicornConfigFilePath,
+						},
 					},
 				},
 			},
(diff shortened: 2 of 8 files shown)


internal/controller/standalone_pgadmin/configmap.go
```diff
@@ -181,3 +187,38 @@ func generateClusterConfig(
...
+
+// generateGunicornConfig generates the config settings for the gunicorn server
+// - https://docs.gunicorn.org/en/latest/settings.html
+func generateGunicornConfig(pgadmin *v1beta1.PGAdmin) (string, error) {
+	settings := map[string]any{
+		// Bind to all IPv4 addresses and set 25 threads by default.
+		// - https://docs.gunicorn.org/en/latest/settings.html#bind
+		// - https://docs.gunicorn.org/en/latest/settings.html#threads
+		"bind":    "0.0.0.0:" + strconv.Itoa(pgAdminPort),
+		"threads": 25,
```
jmckulk, line 199: Why are we setting this to 25 threads?

What is the default? Could we just use that?

tjmoore4: This can certainly set this to whatever we'd like, but I thought this was a good place to start.

jmckulk: What does setting the value do for pgAdmin? I didn't think it was able to use multi threads or processes. IIUC, [this](https://www.pgadmin.org/docs/pgadmin4/development/server_deployment.html#requirements) seems to reflect that

tjmoore4: According to this [comment](https://github.com/pgadmin-org/pgadmin4/blob/REL-8_5/pkg/docker/entrypoint.sh#L101-L102), multiple threads are supported (multiple processes aren't). Also, when using the pgAdmin project's container implementation, the default is also set to 25, see the relevant [docs](https://github.com/pgadmin-org/pgadmin4/blob/REL-8_5/docs/en_US/container_deployment.rst?plain=1#L161-L167) and [script](https://github.com/pgadmin-org/pgadmin4/blob/REL-8_5/pkg/docker/entrypoint.sh#L114-L118). With that in mind, I think that setting a default of 25 is reasonable while allowing for user configuration if desired.

internal/controller/standalone_pgadmin/configmap.go
```diff
@@ -181,3 +187,38 @@ func generateClusterConfig(
...
+	// Copy any specified settings over the defaults.
+	for k, v := range pgadmin.Spec.Config.Gunicorn {
+		settings[k] = v
+	}
+
+	// Write mandatory settings over any specified ones.
+	// - https://docs.gunicorn.org/en/latest/settings.html#workers
+	// - https://docs.gunicorn.org/en/latest/settings.html#workers
+	settings["workers"] = 1
+	settings["chdir"] = "/usr/local/lib/python3.11/site-packages/pgadmin4"
```
cbandy, line 211: @dsessler7 you and I spoke about this path the other day.

cbandy: ❓ I see this path in a variable in the startup script. Is it possible to use that variable as a arg/flag to gunicorn?

internal/controller/standalone_pgadmin/configmap.go
jmckulk: IIUC, those comments and scripts are specific to the pgAdmin docker image and its entrypoint. I'm not seeing anywhere in pgAdmin docs where it says it supports threads outside that container or a/their container deployment docs. I'm not understanding what that setting that value does, but I definitely feel like I'm missing something. The value is a Gunicorn setting, so maybe it doesn't have anything to do with pgAdmin? Is that where I'm getting mixed up?

internal/controller/standalone_pgadmin/configmap.go
tjmoore4: Yeah, that's an option as well. The command works out to be
`gunicorn -c /etc/pgadmin/gunicorn_config.py --chdir $PGADMIN_DIR pgAdmin4:app &`
This way is also a [more authoritative option](https://docs.gunicorn.org/en/latest/configure.html), so a value given through the spec should be overridden.
I'll make this change.

internal/controller/standalone_pgadmin/configmap.go
tjmoore4: pgAdmin has a `THREADED_MODE` setting mentioned in the pgAdmin [config.py docs](https://www.pgadmin.org/docs/pgadmin4/development/config_py.html) that "Allows flask application to response to the each request asynchronously" (which is what the server side pgAdmin code is written with). I can't speak to exactly how pgAdmin interacts with these values without digging in further, but everything seems to point to pgAdmin supporting multi-threading, assuming there are multiple threads configured by the web server.

jmckulk (approved):
lgtm

internal/controller/standalone_pgadmin/pod_test.go
```diff
@@ -181,21 +216,47 @@ containers:
...
+            stat --format='Loaded shared servers dated %y' "${cluster_file}"
+        fi
+        if [ ! -d /proc/$(cat $PGADMIN4_PIDFILE) ]
+        then
+        gunicorn -c /etc/pgadmin/gunicorn_config.py pgAdmin4:app &
+            echo $! > $PGADMIN4_PIDFILE
+            echo "Restarting pgAdmin4"
+        fi
+    done
+    }; export cluster_file="$1"; export -f monitor; exec -a "$0" bash -ceu monitor
```
dsessler7, line 259: Nice.

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -303,17 +311,17 @@ loadServerCommand
...
-		echo "Restarting pgAdmin4"
-	fi
+    if [ "${cluster_file}" -nt "/proc/self/fd/${fd}" ] && loadServerCommand
+    then
+        exec {fd}>&- && exec {fd}<> <(:)
+        stat --format='Loaded shared servers dated %y' "${cluster_file}"
+    fi
+    if [ ! -d /proc/$(cat $PGADMIN4_PIDFILE) ]
+    then
+    gunicorn -c /etc/pgadmin/gunicorn_config.py pgAdmin4:app &
```
dsessler7, line 321: Why do we not need the `--chdir` flag here? :thinking:

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -352,17 +360,36 @@ with open('` + configMountPath + `/` + configFilePath + `') as _f:
...
+`
+		// gunicorn reads from the `/etc/pgadmin/gunicorn_config.py` file during startup
+		// after all other config files.
+		// - https://docs.gunicorn.org/en/latest/configure.html#configuration-file
+		//
+		// This command writes a script in `/etc/pgadmin/gunicorn_config.py` that reads
+		// from the `gunicorn-config.json` file and sets those variables globally.
+		// That way those values are available as settings when gunicorn starts.
+		//
+		// Note: All pgAdmin settings are lowercase with underscores, so ignore
```
dsessler7, line 372: Should this say "All gunicorn settings are lowercase..."?

internal/controller/standalone_pgadmin/pod.go
tjmoore4: We do. I missed that change during the refactor. Thanks for the catch; I'll update this.

internal/controller/standalone_pgadmin/pod.go
tjmoore4: Yep, I'll make that change.

internal/controller/standalone_pgadmin/pod.go
dsessler7: Could also consider pulling that out into a function like we did with `loadServerCommand`, but not a blocker

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -303,17 +319,21 @@ loadServerCommand
...
-	fi
+    if [ "${cluster_file}" -nt "/proc/self/fd/${fd}" ] && loadServerCommand
+    then
+        exec {fd}>&- && exec {fd}<> <(:)
+        stat --format='Loaded shared servers dated %y' "${cluster_file}"
+    fi
+    if [ ! -d /proc/$(cat $PGADMIN4_PIDFILE) ]
+    then
+        if [ $APP_RELEASE -eq 7 ]; then
+            ` + startCommandV7 + `
```
jmckulk, line 330: I like having the string added in inline over passing it in through a print :heart: I think it makes the bash more readable

jmckulk (approved):
lgtm!

PR #3899, opened 2024-04-22
Reject invalid Postgres user specs

cbandy:
The PASSWORD option was never effective. CEL validation rules, available in beta since Kubernetes 1.25, can detect and reject values that the RE2 pattern validation of [github.com/go-openapi could not](https://pkg.go.dev/github.com/go-openapi/validate@v0.24.0#hdr-Known_limitations).

**Checklist:**

 
**Type of Changes:**

 
 - [x] Bug fix
 - [x] Testing enhancement

The PASSWORD option and SQL comments are accepted in Postgres user specs on the PostgresCluster object.

These values are now rejected at the Kubernetes API and removed before being sent to Postgres.

A new package checks CRD validation using envtest.

**Other Information**:

Issue: PGO-1094

### DIFF
--- a/internal/postgres/users.go
+++ b/internal/postgres/users.go
@@ -19,10 +19,42 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"strings"
+
+	pg_query "github.com/pganalyze/pg_query_go/v5"
 
 	"github.com/crunchydata/postgres-operator/internal/logging"
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
+
+func sanitizeAlterRoleOptions(options string) string {
+	const AlterRolePrefix = `ALTER ROLE "any" WITH `
+
+	// Parse the options and discard them completely when incoherent.
+	parsed, err := pg_query.Parse(AlterRolePrefix + options)
+	if err != nil || len(parsed.GetStmts()) != 1 {
+		return ""
+	}
+
+	// Rebuild the options list without invalid options. TODO(go1.21) TODO(slices)
+	orig := parsed.GetStmts()[0].GetStmt().GetAlterRoleStmt().GetOptions()
+	next := make([]*pg_query.Node, 0, len(orig))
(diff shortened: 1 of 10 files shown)


internal/postgres/users.go
```diff
@@ -19,11 +19,42 @@ import (
...
 	"github.com/crunchydata/postgres-operator/internal/logging"
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
+func sanitizeAlterRoleOptions(options string) string {
+	const AlterRolePrefix = `ALTER ROLE "any" WITH `
+
+	// Parse the options and discard them completely when incoherent.
+	parsed, err := pg_query.Parse(AlterRolePrefix + options)
+	if err != nil || len(parsed.Stmts) != 1 || parsed.Stmts[0].Stmt.GetAlterRoleStmt() == nil {
```
cbandy, line 35: 🔧 The linter advice is good. I'll update these.

internal/postgres/users.go
```diff
@@ -19,11 +19,43 @@ import (
...
+	"strings"
+
+	pg_query "github.com/pganalyze/pg_query_go/v5"
 
 	"github.com/crunchydata/postgres-operator/internal/logging"
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
+func sanitizeAlterRoleOptions(options string) string {
+	const AlterRolePrefix = `ALTER ROLE "any" WITH `
```
tjmoore4, line 31: :thinking: Just to make sure I understand, since you're just passing in the options and not the entire statement, you have to append this prefix in order to use the `GetAlterRoleStmt().GetOptions()` functions of pg_query. Or is there more to it than that?

tjmoore4 (approved):
Looks good to me. One non-blocking question.

internal/postgres/users.go
cbandy: Exactly right. The parser works only on complete statements.

PR #3901, opened 2024-04-24
Remove pgAdmin admin user from pgadmin secret. Generate random passwo…

dsessler7:
…rd in setup script. Adjust tests accordingly.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

Currently, PGO adds the default pgAdmin user that is created and used during pgAdmin setup to a k8s Secret where it can be accessed by PGO users. 

In 5.6 and later, the default admin pgAdmin user will no longer be reconciled in the pgAdmin secret since we now have the user management feature that allows PGO users to create their own pgAdmin users. 

**Other Information**:


internal/controller/standalone_pgadmin/pod.go
```diff
@@ -276,6 +267,7 @@ func startupScript(pgadmin *v1beta1.PGAdmin) []string {
 	// - https://www.pgadmin.org/docs/pgadmin4/development/server_deployment.html#standalone-gunicorn-configuration
 	// - https://docs.gunicorn.org/en/latest/configure.html
 	var startScript = fmt.Sprintf(`
+export PGADMIN_SETUP_PASSWORD="${RANDOM}${RANDOM}${RANDOM}"
```
tjmoore4, line 270: :question: I know this password isn't exposed, but could it be used (or otherwise exploited) at all?

If so, I'm thinking we may want to generate the random password a different way here. Based on what I was able to find [documented](https://tldp.org/LDP/abs/html/randomvar.html), this method could result in an all numeric password that is as small as three characters in length. That seems less than ideal. There are quite a few other options for generating a more substantial password consistently. One I found that might work is something like
```
$ export foo=$(date +%s | sha256sum | base64 | head -c 32) # 32 characters in length; adjust accordingly
$ echo $foo
MDRiNjAxNDRjNGVlNWIxNDAyNWFmNjQ2
```

tjmoore4 (approved):

jmckulk (approved):

PR #3906, opened 2024-05-01
sort secrets with the same labels 

ValClarkson:
sort secrets with the same labels by "pguser" being in the name and then sort by creation timestamp, if "pguser' is not in the name sort by creation time stamp if the creation timestamps are  equal sort by name

**Checklist:**

 
 - [x ] Have you added an explanation of what your changes do and why you'd like them to be included?
 - [ x] Have you updated or added documentation for the change, as applicable?
 - [ x] Have you tested your changes on all related environments with successful results, as applicable?

**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

the creation of secrets with the same labels will be randomly selected for use by pgo

the secrets are now sorted and pgo will use a secret that has been sorted into a proper order

**Other Information**:

### DIFF
--- a/internal/controller/postgrescluster/postgres.go
+++ b/internal/controller/postgrescluster/postgres.go
@@ -23,6 +23,7 @@ import (
 	"net"
 	"net/url"
 	"regexp"
+	"sort"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


internal/controller/postgrescluster/postgres.go
```diff
@@ -375,6 +376,36 @@ func (r *Reconciler) reconcilePostgresUserSecrets(
...
+		// If one secret has "pguser" and the other does not,
+		// move the one without "pguser" to the front
+		if isIPgUser && !isJPgUser {
+			return false
+		} else if !isIPgUser && isJPgUser {
+			return true
+		}
+
+		// If both secrets have "pguser" or neither have "pguser",
+		// sort by creation timestamp
```
tony-landreth, line 400: Does this comment go with the line 406?

ValClarkson: it does thanks for catching that

internal/controller/postgrescluster/postgres.go
```diff
@@ -375,6 +376,36 @@ func (r *Reconciler) reconcilePostgresUserSecrets(
...
+	// If one secret has "pguser" in its name and the other does not, the
+	// one without "pguser" is moved to the front.
+	// If both secrets have "pguser" in their names or neither has "pguser", they
+	// are sorted by creation timestamp.
+	// If two secrets have the same creation timestamp, they are further sorted by name.
+	// The secret to be used by PGO is put at the end of the sorted slice.
+	sort.Slice(secrets.Items, func(i, j int) bool {
+		// Check if either secrets have "pguser" in their names
+		isIPgUser := strings.Contains(secrets.Items[i].Name, "pguser")
+		isJPgUser := strings.Contains(secrets.Items[j].Name, "pguser")
```
cbandy, line 389: 🌱 Seeing this reminded me that we only really care about picking the right secret _for each user_. That is, the order between different users doesn't matter. 🤔 That might simplify things here, but I'm not sure.

internal/controller/postgrescluster/postgres.go
```diff
@@ -375,6 +376,36 @@ func (r *Reconciler) reconcilePostgresUserSecrets(
...
 	}
 
+	// Sorts the slice of secrets.Items based on secrets with identical labels
+	// If one secret has "pguser" in its name and the other does not, the
+	// one without "pguser" is moved to the front.
+	// If both secrets have "pguser" in their names or neither has "pguser", they
+	// are sorted by creation timestamp.
+	// If two secrets have the same creation timestamp, they are further sorted by name.
+	// The secret to be used by PGO is put at the end of the sorted slice.
+	sort.Slice(secrets.Items, func(i, j int) bool {
```
cbandy, line 386: 🌱 This comparison function only uses `secrets.Items`, so it could be separated into another function if you want to test that. Something like:

```go
func sortPostgresUserSecrets(secrets []corev1.Secret) {
  sort.Slice(secrets, func(i, j int) bool {
    …
  })
}
```

cbandy (approved):

PR #3911, opened 2024-05-08
Set SeccompProfile to RuntimeDefault

tony-landreth:
Issue: PGO-845

### DIFF
--- a/internal/initialize/security_test.go
+++ b/internal/initialize/security_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	"gotest.tools/v3/assert"
+	corev1 "k8s.io/api/core/v1"
 
 	"github.com/crunchydata/postgres-operator/internal/initialize"
 )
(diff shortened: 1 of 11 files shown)


internal/initialize/security_test.go
```diff
@@ -59,9 +60,8 @@ func TestPodSecurityContext(t *testing.T) {
 		assert.Assert(t, psc.RunAsUser == nil,
 			`Containers must not set runAsUser to 0`)
 
-		// TODO(cbandy): delegate to v1.SecurityContext
 		assert.Assert(t, psc.SeccompProfile == nil,
```
cbandy, line 63: 🤔 I think I was suggesting we do an assertion similar to `psc.RunAsNonRoot` a few lines up.

tony-landreth: Like this?
```go
if assert.Check(t, psc.SeccompProfile == nil) {
  assert.Assert(t, initialize.RestrictedSecurityContext().SeccompProfile != nil,
    `SeccompProfile should be delegated to the container-level v1.SecurityContext`)
}
```

cbandy: Yes! Thanks!

PR #3918, opened 2024-05-23
Move standalone pgadmin tests from e2e-other to e2e.

dsessler7:
**Checklist:**

 
 - [x] Have you added an explanation of what your changes do and why you'd like them to be included?
 - [ ] Have you updated or added documentation for the change, as applicable?
 - [ ] Have you tested your changes on all related environments with successful results, as applicable?
   - [ ] Have you added automated tests?

**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

**What is the current behavior (link to any open issues here)?**

Tests for standalone pgadmin functionality are in e2e-other and therefore do not get run nightly.

**What is the new behavior (if this is a feature change)?**
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

The standalone pgadmin tests are now all in e2e and will get run nightly.

**Other Information**:

PR #3920, opened 2024-05-29
pgData Volume Autogrow

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Adds autogrow capability for the pgData volumes.

**Other Information**:
Issues:
- PGO-1214
- PGO-1215
- PGO-1270


internal/postgres/config.go
```diff
@@ -172,16 +172,39 @@ func reloadCommand(name string) []string {
...
+  size=$(df --human-readable --block-size=M /pgdata | awk 'FNR == 2 {print $2}')
+  use=$(df --human-readable /pgdata | awk 'FNR == 2 {print $5}')
+  sizeInt="${size//M/}"
+  # Use the sed punctuation class, because the shell will not accept the percent sign in an expansion.
+  useInt=$(echo $use | sed 's/[[:punct:]]//g')
+  triggerExpansion="$((useInt > 75))"
+  if [ $triggerExpansion -eq 1 ]; then
+    newSize="$(((sizeInt / 2)+sizeInt))"
+    newSizeMi="${newSize}Mi"
+    d='[{"op": "add", "path": "/metadata/annotations/diskstarved", "value": "'"$newSizeMi"'"}]'
```
tony-landreth, line 205: :question: What do we think about `suggested-pgdata-pvc-size` instead? Too verbose?

internal/controller/postgrescluster/instance.go
```diff
@@ -342,6 +381,65 @@ func (r *Reconciler) observeInstances(
...
+		}
+	}
+
+	if limitSet && current.Value() > previous.Value() {
+		r.Recorder.Eventf(cluster, corev1.EventTypeNormal, "VolumeAutoGrow",
+			"pgData volume expansion to %v requested for %s/%s.",
+			current.String(), cluster.Name, instanceSetName)
+	}
+
+	// If the desired size was not observed, update with previously stored value.
```
tony-landreth, line 435: :question: What causes desired size not to be observed, when `desiredRequestBackup` is set?

internal/controller/postgrescluster/postgres.go
```diff
@@ -628,6 +635,71 @@ func (r *Reconciler) reconcilePostgresDataVolume(
 	return pvc, err
 }
 
+// determineVolumeSize compares the potential sizes from the instance spec, status
+// and limit and sets the appropriate current value.
```
tony-landreth, line 639: :pick: This method seems more accurately described as `setVolumeSize`.

internal/controller/postgrescluster/postgres.go
```diff
@@ -628,6 +635,71 @@ func (r *Reconciler) reconcilePostgresDataVolume(
...
+		pvc.Spec.Resources.Requests.Storage().Value() > pvc.Spec.Resources.Limits.Storage().Value() {
+		r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "VolumeRequestOverLimit",
+			"pgData volume request (%v) for %s/%s is greater than set limit (%v). Limit value will be used.",
+			pvc.Spec.Resources.Requests.Storage(), cluster.Name, instanceSpecName, pvc.Spec.Resources.Limits.Storage())
+
+		pvc.Spec.Resources.Requests = corev1.ResourceList{
+			corev1.ResourceStorage: *resource.NewQuantity(pvc.Spec.Resources.Limits.Storage().Value(), resource.BinarySI),
+		}
+		// Otherwise, if the limit is not set or the feature gate is not enabled, do not autogrow.
+	} else if !pvc.Spec.Resources.Limits.Storage().IsZero() && util.DefaultMutableFeatureGate.Enabled(util.AutoGrowVolumes) {
```
tony-landreth, line 659: This seems to be the definition of when auto-grow applies:
```
 !pvc.Spec.Resources.Limits.Storage().IsZero() && util.DefaultMutableFeatureGate.Enabled(util.AutoGrowVolumes) {
```
A function called something like `autogrowPermitted` might be nice to keep tabs on that definition.

internal/controller/postgrescluster/postgres.go
```diff
@@ -620,6 +621,12 @@ func (r *Reconciler) reconcilePostgresDataVolume(
 
 	pvc.Spec = instanceSpec.DataVolumeClaimSpec
 
+	r.determineVolumeSize(ctx, cluster, pvc, instance.Name, instanceSpec.Name)
+
+	// Clear any set limit before applying PVC. This is needed to allow the limit
+	// value to change later.
```
tony-landreth, line 627: :question: What is it about setting to nil here that allows the limit to change later?

internal/controller/postgrescluster/instance.go
tjmoore4: In a scenario where the annotation on the Pod was missing. The easiest example is when the cluster is shutdown, but it could also happen if a Pod had been restarted and the annotation had not yet been updated.

internal/controller/postgrescluster/postgres.go
tjmoore4: The `DataVolumeClaimSpec`, with the exception of the request value, is immutable once the PVC is bound. Effectively we're managing the limit ourselves but never applying it.

internal/controller/postgrescluster/postgres.go
tjmoore4: That works for me, I'll make that change.

internal/controller/postgrescluster/postgres.go
tjmoore4: I thought about that, but in `observeInstances` I only check for the feature gate (because limits can vary per instance set). For now, I'm inclined to leave this as is, but might be worth talking through. Another thing worth noting is that the limits equal to zero (e.g. `0Gi`) will not trigger autogrow despite being acceptable in the manifest. I think this is a reasonable approach, but happy to hear your thoughts on it.

internal/controller/postgrescluster/instance.go
tony-landreth: Can that be added to the comment?

tjmoore4: Sure :+1:

tony-landreth: Thanks!

internal/controller/postgrescluster/instance.go
```diff
@@ -320,10 +324,25 @@ func (r *Reconciler) observeInstances(
 
 	observed := newObservedInstances(cluster, runners.Items, pods.Items)
 
+	// Save desired volume size values in case the status is removed.
+	// This may happen in cases where the Pod is restarted, the cluster
+	// is shutdown, etc. Only save values for instances defined in the spec.
+	desiredRequestsBackup := make(map[string]string)
```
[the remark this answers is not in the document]
tjmoore4: Sure, that makes sense to me. I'll make that change.

internal/postgres/config.go
tjmoore4: Per discussion, updated.

dsessler7:
The `standalone-pgadmin` kuttl test is still failing because[ this PR](https://github.com/CrunchyData/postgres-operator/pull/3918) still needs to be reviewed and merged (hint hint).

internal/controller/postgrescluster/postgres.go
```diff
@@ -628,6 +635,71 @@ func (r *Reconciler) reconcilePostgresDataVolume(
...
+	pvc *corev1.PersistentVolumeClaim, instanceSpecName string) {
+	log := logging.FromContext(ctx)
+
+	// Capture the largest pgData volume size currently defined for a given instance set.
+	volumeRequestSize := pvc.Spec.Resources.Requests.Storage() //.Value()
+
+	// If the request value is greater than the set limit, use the limit and issue
+	// a warning event. A limit of 0 is ignorned.
+	if !pvc.Spec.Resources.Limits.Storage().IsZero() &&
+		pvc.Spec.Resources.Requests.Storage().Value() > pvc.Spec.Resources.Limits.Storage().Value() {
```
dsessler7, line 650: :pick: 
```suggestion
		volumeRequestSize.Value() > pvc.Spec.Resources.Limits.Storage().Value() {
```

internal/controller/postgrescluster/postgres.go
```diff
@@ -628,6 +635,71 @@ func (r *Reconciler) reconcilePostgresDataVolume(
...
+					}
+				}
+			}
+		}
+
+		// If the volume request size is greater than or equal to the limit and the
+		// limit is not zero, update the request size to the limit value.
+		// If the user manually requests a lower limit that is smaller than the current
+		// or requested volume size, it will be ignored in favor of the limit value.
+		if volumeRequestSize.Value() >= pvc.Spec.Resources.Limits.Storage().Value() {
```
dsessler7, line 681: :pick: 
```suggestion
		if volumeRequestSize.Value() >= volumeLimitSize.Value() {
```

internal/controller/postgrescluster/postgres.go
```diff
@@ -628,6 +635,71 @@ func (r *Reconciler) reconcilePostgresDataVolume(
...
+
+			r.Recorder.Eventf(cluster, corev1.EventTypeNormal, "VolumeLimitReached",
+				"pgData volume(s) for %s/%s are at size limit (%v).", cluster.Name,
+				instanceSpecName, pvc.Spec.Resources.Limits.Storage())
+
+			// If the volume size request is greater than the limit, issue an
+			// additional event warning.
+			if volumeRequestSize.Value() > pvc.Spec.Resources.Limits.Storage().Value() {
```
dsessler7, line 689: :pick: 
```suggestion
			if volumeRequestSize.Value() > volumeLimitSize.Value() {
```

internal/controller/postgrescluster/postgres.go
```diff
@@ -628,6 +635,71 @@ func (r *Reconciler) reconcilePostgresDataVolume(
...
+				r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "DesiredVolumeAboveLimit",
+					"The desired size (%v) for the %s/%s pgData volume(s) is greater than the size limit (%v).",
+					volumeRequestSize, cluster.Name, instanceSpecName, pvc.Spec.Resources.Limits.Storage())
+			}
+
+			volumeRequestSize = pvc.Spec.Resources.Limits.Storage()
```
dsessler7, line 695: :pick
```suggestion
			volumeRequestSize = volumeLimitSize
```

dsessler7 (commented):
A few nitpicky suggestions

internal/controller/postgrescluster/postgres.go
```diff
@@ -628,6 +635,71 @@ func (r *Reconciler) reconcilePostgresDataVolume(
...
 }
 
+// setVolumeSize compares the potential sizes from the instance spec, status
+// and limit and sets the appropriate current value.
+func (r *Reconciler) setVolumeSize(ctx context.Context, cluster *v1beta1.PostgresCluster,
+	pvc *corev1.PersistentVolumeClaim, instanceSpecName string) {
+	log := logging.FromContext(ctx)
+
+	// Capture the largest pgData volume size currently defined for a given instance set.
+	volumeRequestSize := pvc.Spec.Resources.Requests.Storage() //.Value()
```
dsessler7, line 645: Maybe do the same for the limit for readability's sake? (I thought I had this in my review before, but it was lost somehow)
```suggestion
	volumeRequestSize := pvc.Spec.Resources.Requests.Storage()
	volumeLimitSize := pvc.Spec.Resources.Limits.Storage()
```

tjmoore4: I'll definitely get rid of the obsolete comment, but part of what I'm trying to do is distinguish the values from the spec vs the value that I'm determining (i.e. `volumeRequestSize`). I'm concerned that renaming here will make that distinction less clear. Also, when I'm ultimately setting `pvc.Spec.Resources.Requests`, it's done by reference, so I'm trying to avoid any extra copies if possible.

internal/controller/postgrescluster/postgres.go
tjmoore4: See the comment above. I think I'd like to keep these values as explicitly from the pvc spec.

internal/controller/postgrescluster/postgres.go
tjmoore4: See the comment above. I think I'd like to keep these values as explicitly from the pvc spec.

internal/controller/postgrescluster/postgres.go
tjmoore4: See the comment above. I think I'd like to keep these values as explicitly from the pvc spec.

internal/controller/postgrescluster/postgres.go
tjmoore4: See the comment above. I think I'd like to keep these values as explicitly from the pvc spec.

internal/controller/postgrescluster/postgres.go
tjmoore4: I thought about this a bit more and talked it over with @tony-landreth. I think we can maintain the distinction with carefully named variables. I'll push a change in a few. Please let me know what you think.

internal/controller/postgrescluster/postgres.go
tjmoore4: Updated, see comment.

internal/controller/postgrescluster/postgres.go
tjmoore4: Updated, see comment.

internal/controller/postgrescluster/postgres.go
tjmoore4: Updated, see comment.

internal/controller/postgrescluster/postgres.go
tjmoore4: Updated, see comment.

internal/postgres/config.go
```diff
@@ -172,16 +172,39 @@ func reloadCommand(name string) []string {
...
+  use=$(df --human-readable /pgdata | awk 'FNR == 2 {print $5}')
+  sizeInt="${size//M/}"
+  # Use the sed punctuation class, because the shell will not accept the percent sign in an expansion.
+  useInt=$(echo $use | sed 's/[[:punct:]]//g')
+  triggerExpansion="$((useInt > 75))"
+  if [ $triggerExpansion -eq 1 ]; then
+    newSize="$(((sizeInt / 2)+sizeInt))"
+    newSizeMi="${newSize}Mi"
+    d='[{"op": "add", "path": "/metadata/annotations/suggested-pgdata-pvc-size", "value": "'"$newSizeMi"'"}]'
+    curl --cacert ${CACERT} --header "Authorization: Bearer ${TOKEN}" -XPATCH "${APISERVER}/api/v1/namespaces/${NAMESPACE}/pods/${HOSTNAME}?fieldManager=kubectl-annotate" -H "Content-Type: application/json-patch+json" --data "$d"
```
andrewlecuyer, line 206: 🤔 Thinking we should force ownership here.  I believe this is done via another query param, e.g. `force=true`.

https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts
> **Overwrite value, become sole manager:** If overwriting the value was intentional (or if the applier is an automated process like a controller) the applier should set the force query parameter to true (for kubectl apply, you use the --force-conflicts command line parameter), and make the request again. This forces the operation to succeed, changes the value of the field, and removes the field from all other managers' entries in managedFields.

tony-landreth: Makes sense. Thanks, Andrew!

tony-landreth: Not permitted:

```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "PatchOptions.meta.k8s.io \"\" is invalid: force: Forbidden: may not be specified for non-apply patch",
  "reason": "Invalid",
  "details": {
    "group": "meta.k8s.io",
    "kind": "PatchOptions",
    "causes": [
      {
        "reason": "FieldValueForbidden",
        "message": "Forbidden: may not be specified for non-apply patch",
        "field": "force"
      }
    ]
  },
  "code": 422

```

tjmoore4 (commented):
@tony-landreth I can't approve this PR because I created it, but I reviewed your portion of the code and it looks good to me :+1:

internal/controller/postgrescluster/postgres.go
dsessler7: Looks good

tony-landreth (approved):

Issue #3568, opened 2023-02-09
Users don't get CREATE permissions with postgresVersion: 15 


dbackeus:
When creating a `PostgresCluster` using `postgresVersion: 14` it's possible to use either the default or explicitly created users + databases and have all the expected CRUD privileges enabled for that user.

However when using `postgresVersion: 15` we get `ERROR:  permission denied for schema public` when trying to eg. `CREATE TABLE ...` for the default user and database.

I'm guessing this is related to the following mentioned in the Postgres 15 [release notes](https://www.postgresql.org/about/news/postgresql-15-released-2526/):
> PostgreSQL 15 also [revokes the CREATE permission from all users](https://www.postgresql.org/docs/15/ddl-schemas.html#DDL-SCHEMAS-PATTERNS) except a database owner from the public (or default) schema.

But this begs the question how users / database management in PGO is supposed to work when using Postgres 15?

andrewlecuyer:
@dbackeus thanks for reaching out! 

As you mentioned, the behavior you're seeing is simply a result of the change you referenced in PG 15.

Is there any specific behavior you're looking here?  For instance, is there something else you'd like and/or expect PGO to be doing here?

As you're seeing, our current approach is to simply align with the new PG 15 behavior here, but we're open to thoughts, suggestions, etc. as to how to best approach this.

saosebastiao:
To clarify, this is an intended behavioral change introduced in PG15, nothing to do with the Postgres Operator. 

https://fluca1978.github.io/2022/07/15/PostgreSQL15PublicSchema.html

dbackeus:
Right. After realizing that perhaps it isn't in the scope of PGO to give a convenient solution for this we figured out that we could solve this via `databaseInitSQL`.

Eg:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: init-sql
data:
  init.sql: |
    \c <database>
    GRANT CREATE ON SCHEMA public TO "<user>";
---
apiVersion: postgres-operator.crunchydata.com/v1beta1
kind: PostgresCluster
metadata:
  name: my-cluster
spec:
  postgresVersion: 15
  databaseInitSQL:
    key: init.sql
    name: init-sql
  # ...
```

pcornelissen:
I don't agree that this has noting to do with the postgres operator. The PGO has the ability to create users and databases, but starting with pg15 they serve no purpose, as you can't do something with the generated users.
[...]

PR #3940, opened 2024-06-21
Create schemas for users in granted databases

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

To help developers set up and connect quickly, the operator can now create schemas for `spec.users` without using an init SQL script.

This is a gated feature: to turn on set the FeatureGate `AutoCreateUserSchema=true`.

If turned on, a cluster can be annotated with
`postgres-operator.crunchydata.com/autoCreateUserSchema=true`.

If the feature is turned on and the cluster is annotated, PGO will create a schema named after the user in every database where that user has permissions.

(PG note: creating a schema with the same name as the user means that the PG `search_path` should not need to be updated, since `search_path` defaults to `"$user", public`.)

As with our usual pattern, the operator does not remove/delete PG objects (users, databases) that are removed from the spec.

NOTE: There are several schema names that would be dangerous to the cluster's operation; for instance, if you had pgbouncer enabled (which would create a `pgbouncer` schema) it would be dangerous to create a user named `pgbouncer` and use this feature to create a schema for that user.

**Other Information**:

Issues: [PGO-1333]
Fixes https://github.com/CrunchyData/postgres-operator/issues/3568

### DIFF
--- a/internal/postgres/users.go
+++ b/internal/postgres/users.go
@@ -29,6 +29,12 @@ import (
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
+var RESERVED_SCHEMA_NAMES = map[string]bool{
+	"public":    true,
+	"pgbouncer": true,
+	"monitor":   true,
+}
+
 func sanitizeAlterRoleOptions(options string) string {
 	const AlterRolePrefix = `ALTER ROLE "any" WITH `
 
@@ -190,6 +196,13 @@ func WriteUsersSchemasInPostgreSQL(ctx context.Context, exec Executor,
 
 	for i := range users {
 		spec := users[i]
+
+		if RESERVED_SCHEMA_NAMES[string(spec.Name)] {
+			log.V(1).Info("Skipping schema creation for user with reserved name",
+				"name", string(spec.Name))
+			continue
+		}
+
 		databases := []string{}
 		for _, db := range spec.Databases {
 			databases = append(databases, string(db))
--- a/internal/postgres/users.go
+++ b/internal/postgres/users.go
@@ -184,7 +184,6 @@ SELECT pg_catalog.format('GRANT ALL PRIVILEGES ON DATABASE %I TO %I',
 }
 
 // WriteUsersSchemasInPostgreSQL will create a schema for each user in each database that user has access to
-// NOTE: Does this mean if no dbs are specified, no schemas are created?
 func WriteUsersSchemasInPostgreSQL(ctx context.Context, exec Executor,
 	users []v1beta1.PostgresUserSpec) error {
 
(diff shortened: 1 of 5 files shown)


internal/postgres/users.go
```diff
@@ -162,5 +170,76 @@ SELECT pg_catalog.format('GRANT ALL PRIVILEGES ON DATABASE %I TO %I',
...
+		_, _ = sql.WriteString(`SELECT UNNEST(ARRAY[:databases]);`)
+
+		stdout, stderr, err = exec.ExecInDatabasesFromQuery(ctx,
+			sql.String(),
+			strings.Join([]string{
+				// Quiet NOTICE messages from IF EXISTS statements.
+				// - https://www.postgresql.org/docs/current/runtime-config-client.html
+				`SET client_min_messages = WARNING;`,
+
+				"CREATE SCHEMA IF NOT EXISTS AUTHORIZATION :username;",
```
benjaminjb, line 231: This creates a schema named :username, owned by :username.

Our only other schema that we lock down that I can see is `pgbouncer`, and we lock that with 

```sql
REVOKE ALL PRIVILEGES
    ON SCHEMA :"namespace" FROM PUBLIC, :"username";
 GRANT USAGE
    ON SCHEMA :"namespace" TO :"username";
```
--[cite](https://github.com/CrunchyData/postgres-operator/blob/master/internal/pgbouncer/postgres.go#L169-L172)

The difference being something like (rhino1 created with the `authorization` code, rhino2 granted/revoked as per other code):

```
postgres=# \dn+
                                       List of schemas
  Name  |       Owner       |           Access privileges            |      Description       
--------+-------------------+----------------------------------------+------------------------
 public | pg_database_owner | pg_database_owner=UC/pg_database_owner+| standard public schema
        |                   | =U/pg_database_owner                   | 
 rhino1 | rhino1            |                                        | 
 rhino2 | postgres          | postgres=UC/postgres                  +| 
        |                   | rhino2=U/postgres                      | 
(3 rows)
```

cbandy (changes requested):

internal/postgres/users.go
cbandy: These permissions LGTM.

internal/postgres/users.go
```diff
@@ -162,5 +170,76 @@ SELECT pg_catalog.format('GRANT ALL PRIVILEGES ON DATABASE %I TO %I',
...
+			}, "\n"),
+			map[string]string{
+				"databases": strings.Join(databases, ", "),
```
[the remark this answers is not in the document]
benjaminjb: hmmm, and spec.users.databases has no requirements...

internal/postgres/users_test.go
```diff
@@ -192,3 +197,52 @@ COMMIT;`))
...
+		) error {
+			calls++
+
+			b, err := io.ReadAll(stdin)
+			assert.NilError(t, err)
+
+			// The command strings will contain either of two possibilities, depending on the user called.
+			commands := strings.Join(command, ",")
+			re := regexp.MustCompile("--set=databases=\\[\"db1\"\\],--set=username=user-single-db|--set=databases=\\[\"db1\",\"db2\"\\],--set=username=user-multi-db")
+			assert.Assert(t, cmp.Regexp(re, commands))
```
cbandy, line 217: 👍🏻 Interesting to see that the `postgres.Exec*` design makes it hard to test variables.

cbandy (approved):

PR #3946, opened 2024-06-29
Make feature gates available via Context

cbandy:
Using Context improves the isolation and parallelism of tests involving feature gates. Each now builds and injects its own gate.

Gates are still implemented using `k8s.io/component-base/featuregate`, but the new `feature` package exports smaller interfaces and produces gates containing PGO features.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement
 - [x] Other


internal/feature/features.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+const (
+	// Support appending custom queries to default PGMonitor queries
+	AppendCustomQueries = "AppendCustomQueries"
+
+	// Support automatically growing volumes
+	AutoGrowVolumes = "AutoGrowVolumes"
+
+	BridgeIdentifiers = "BridgeIdentifiers"
```
tony-landreth, line 84: Should this have a descriptive comment?

cmd/postgres-operator/main.go
```diff
@@ -122,8 +118,9 @@ func main() {
 	log := logging.FromContext(ctx)
 	log.V(1).Info("debug flag set to true")
 
-	log.Info("feature gates enabled",
-		"PGO_FEATURE_GATES", os.Getenv("PGO_FEATURE_GATES"))
+	features := feature.NewGate()
```
benjaminjb, line 121: Petition to rename `features` as `gates`? (Largely so that line 162 reads 

```
if gates.Enabled(feature.BridgeIdentifiers)
```

internal/feature/features.go
```diff
@@ -0,0 +1,126 @@
...
+// Enabled indicates if a Feature is enabled in the Gate contained in ctx. It
+// returns false when there is no Gate.
+func Enabled(ctx context.Context, f Feature) bool {
+	gate, ok := ctx.Value(contextKey{}).(Gate)
+	return ok && gate.Enabled(f)
+}
+
+// NewContext returns a copy of ctx containing gate. Check it using [Enabled].
+func NewContext(ctx context.Context, gate Gate) context.Context {
+	return context.WithValue(ctx, contextKey{}, gate)
```
benjaminjb, line 125: Why `contextKey{}` as the key?

cbandy: The key is passed and compared as an interface, so this unexported type gives a unique value that also requires fewer (no?) allocations.

- https://pkg.go.dev/context#WithValue
- https://go.dev/ref/spec#Comparison_operators<br><br>
  > Two interface values are equal if they have [identical](https://go.dev/ref/spec#Type_identity) dynamic types and equal dynamic values or if both have value `nil`.
- https://dave.cheney.net/2014/03/25/the-empty-struct

internal/feature/features.go
cbandy: It has the right comment for now.

dsessler7 (approved):
LGTM

PR #3947, opened 2024-06-29
Follow ShellCheck's style guide for Bash scripts

cbandy:
Recent versions of ShellCheck recommend using `[[` in Bash scripts and `|| true` to indicate when errors can be ignored. I like the built-in `:` for this purpose.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Other

Changes to container entrypoints cause rollouts.

**Other Information**:

See: https://github.com/koalaman/shellcheck/wiki/SC2292
See: https://github.com/koalaman/shellcheck/wiki/SC2312
See: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#colon

### DIFF
--- a/internal/controller/postgrescluster/instance_test.go
+++ b/internal/controller/postgrescluster/instance_test.go
@@ -708,21 +708,21 @@ func TestAddPGBackRestToInstancePodSpec(t *testing.T) {
   - --
   - |-
     monitor() {
-    exec {fd}<> <(:)
+    exec {fd}<> <(:||:)
     until read -r -t 5 -u "${fd}"; do
       if
-        [ "${filename}" -nt "/proc/self/fd/${fd}" ] &&
+        [[ "${filename}" -nt "/proc/self/fd/${fd}" ]] &&
         pkill -HUP --exact --parent=0 pgbackrest
       then
-        exec {fd}>&- && exec {fd}<> <(:)
+        exec {fd}>&- && exec {fd}<> <(:||:)
         stat --dereference --format='Loaded configuration dated %y' "${filename}"
       elif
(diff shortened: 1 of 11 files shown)


internal/controller/postgrescluster/instance_test.go
```diff
@@ -708,21 +708,21 @@ func TestAddPGBackRestToInstancePodSpec(t *testing.T) {
   - --
   - |-
     monitor() {
-    exec {fd}<> <(:)
+    exec {fd}<> <(:||:)
```
tony-landreth, line 711: What does this line do?

tony-landreth (approved):

internal/postgres/config.go
```diff
@@ -303,27 +303,27 @@ chmod +x /tmp/pg_rewind_tde.sh
...
 		// Log the effective user ID and all the group IDs.
 		`echo Initializing ...`,
-		`results 'uid' "$(id -u)" 'gid' "$(id -G)"`,
+		`results 'uid' "$(id -u ||:)" 'gid' "$(id -G ||:)"`,
 
 		// Abort when the PostgreSQL version installed in the image does not
 		// match the cluster spec.
-		`results 'postgres path' "$(command -v postgres)"`,
-		`results 'postgres version' "${postgres_version:=$(postgres --version)}"`,
+		`results 'postgres path' "$(command -v postgres ||:)"`,
```
benjaminjb, line 310: Should we be checking earlier in this script that we have postgres on this container?

benjaminjb (approved):

internal/postgres/config.go
cbandy: Most things above here are function declarations. The only calls ahead of this are `printf` and `id`, so seems early enough to me.

PR #3962, opened 2024-07-29
archive-async by default with spool-path

tony-landreth:
Issue: PGO-1371 PGO-1142


internal/postgres/config.go
```diff
@@ -307,6 +312,12 @@ chmod +x /tmp/pg_rewind_tde.sh
 		`echo Initializing ...`,
 		`results 'uid' "$(id -u ||:)" 'gid' "$(id -G ||:)"`,
 
+		// The pgbackrest spool path should be co-located with wal.
+		// If a wal volume exists, link the spool-path to it.
+		`if [[ "${pgwal_directory}" == *"pgwal/"* ]] && [[ ! -d "/pgwal/pgbackrest-spool" ]];then rm -rf "/pgdata/pgbackrest-spool" && mkdir -p "/pgwal/pgbackrest-spool" && ln --force --symbolic "/pgwal/pgbackrest-spool" "/pgdata/pgbackrest-spool";fi`,
+		// When a pgwal volume is removed, force pgbackrest to recreate spool-path.
+		`if [[ ! "${pgwal_directory}" == *"pgwal/"* ]];then rm -rf /pgdata/pgbackrest-spool;fi`,
```
tjmoore4, line 319: :thinking: Would this result in the `pgbackrest-spool` directory being deleted on startup every time, even when a WAL volume has not previously been defined?

tony-landreth: @tjmoore4 Ready for another look.

tjmoore4 (approved):

PR #3965, opened 2024-08-02
Create stanza after repohost is added

tony-landreth:
Given a cloud host is already in place,
the user should still be able to add a
repohost.

### DIFF
--- a/internal/pgbackrest/pgbackrest.go
+++ b/internal/pgbackrest/pgbackrest.go
@@ -32,6 +32,10 @@ const (
 	// is detected while attempting stanza creation
 	errMsgConfigHashMismatch = "postgres operator error: pgBackRest config hash mismatch"
 
+	// errMsgStaleReposWithVolumesConfig is the error message displayed when a volume-backed repo has been
+	// configured, but the configuration has not yet propagated into the container.
+	errMsgStaleReposWithVolumesConfig = "postgres operator error: pgBackRest stale volume-backed repo configuration"
+
 	// errMsgBackupDbMismatch is the error message returned from pgBackRest when PG versions
 	// or PG system identifiers do not match between the PG instance and the existing stanza
 	errMsgBackupDbMismatch = "backup and archive info files exist but do not match the database"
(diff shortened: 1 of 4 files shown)


internal/pgbackrest/pgbackrest_test.go
```diff
@@ -34,15 +39,19 @@ func TestStanzaCreateOrUpgrade(t *testing.T) {
 	ctx := context.Background()
 	configHash := "7f5d4d5bdc"
 	expectedCommand := []string{"bash", "-ceu", "--", `
-declare -r hash="$1" stanza="$2" message="$3" cmd="$4"
+declare -r hash="$1" stanza="$2" message="$3" cmd="$4" check_repo_cmd="$5"
 if [[ "$(< /etc/pgbackrest/conf.d/config-hash)" != "${hash}" ]]; then
     printf >&2 "%s" "${message}"; exit 1;
+elif ! bash -c "${check_repo_cmd}"; then
```
tjmoore4, line 45: :pick: Could we add a comment to explain why we are checking all the repo paths are in place?

tjmoore4 (approved):

internal/pgbackrest/pgbackrest.go
```diff
@@ -60,22 +62,44 @@ func (exec Executor) StanzaCreateOrUpgrade(ctx context.Context, configHash strin
...
 	// propagated to the container, and if not, it prints an error and returns with exit code 1).
 	// Otherwise, it runs the pgbackrest command, which will either be "stanza-create" or
 	// "stanza-upgrade", depending on the value of the boolean "upgrade" parameter.
 	const script = `
-declare -r hash="$1" stanza="$2" message="$3" cmd="$4"
+declare -r hash="$1" stanza="$2" message="$3" cmd="$4" check_repo_cmd="$5"
 if [[ "$(< /etc/pgbackrest/conf.d/config-hash)" != "${hash}" ]]; then
     printf >&2 "%s" "${message}"; exit 1;
+elif ! bash -c "${check_repo_cmd}"; then
+		printf >&2 "%s" "${message}"; exit 1;
```
andrewlecuyer, line 95: ⛏️ Looks like an extra tab in here.

andrewlecuyer, line 95: 🤔 Should we consider another/different error message here?  The current message specifically specifically references a "config hash mismatch", which really isn't applicable this new check.

At the very least, prob worth adding adding a note/comment in there around the reasoning for leveraging that same error message, e.g. to sure we're not confused with other hashing logic in the future.

tony-landreth: Good idea 👍

internal/pgbackrest/pgbackrest_test.go
tony-landreth: Will do.

ValClarkson:
I've tested with multiple repo set ups on creation and it works as expected.  I've hit an error when i remove a repo thats an s3 bucket then you run backrest info this error occurs "error (missing stanza path)" for that repo that was removed.

internal/pgbackrest/pgbackrest.go
```diff
@@ -60,22 +62,44 @@ func (exec Executor) StanzaCreateOrUpgrade(ctx context.Context, configHash strin
...
 	}
 
+	var reposWithVolumes []v1beta1.PGBackRestRepo
+	for _, repo := range postgresCluster.Spec.Backups.PGBackRest.Repos {
+		if repo.Volume != nil {
+			reposWithVolumes = append(reposWithVolumes, repo)
+		}
+	}
+
+	grep := "grep %s-path /etc/pgbackrest/conf.d/pgbackrest_instance.conf"
```
tony-landreth, line 72: To the point @ValClarkson made, this will handle repos added, but not repos removed.

tony-landreth, line 72: @andrewlecuyer If we checked that the number of repos in the container config doesn't equal what's in the spec, wouldn't we get the added benefit of handling cases where a repo is removed (regardless of the type of repo)? The hash would still cover the cases where number of repos stays the same, but endpoint changes, right?

internal/pgbackrest/pgbackrest_test.go
tony-landreth: I put the comment on the source, rather than the test.

PR #3966, opened 2024-08-02
Updates for an always-on pgBackRest repo host

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

To support the 'backup-standby' pgBackRest configuration setting as well as to simplify the overall pgBackRest integration, this commit changes the pgBackRest repo host behavior to always be created. This allows backups to be consistently run on the repo host Pod instead of being run at times on the primary instance Pod.

Note that in cases where a repo host volume is not defined in the PostgresCluster spec, no volume will be created and pgBackRest log files will not be available in the Pod.

**Other Information**:
Issue: PGO-562

### DIFF
--- a/internal/controller/postgrescluster/pgbackrest.go
+++ b/internal/controller/postgrescluster/pgbackrest.go
@@ -33,7 +33,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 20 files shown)


internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -299,10 +298,8 @@ func (r *Reconciler) cleanupRepoResources(ctx context.Context,
 				owned.GetName() != naming.PGBackRestSSHSecret(postgresCluster).Name {
```
tony-landreth, line 298: There's a TODO on line 296/7 that might be worth handling.

tony-landreth (approved):

internal/controller/postgrescluster/pgbackrest.go
tjmoore4: I think that's true, but it's a little beyond the scope of this story. I think it's good to get this PR merged now, so I'll leave that in for the time being.

PR #3973, opened 2024-08-20
Api discovery

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

**Other Information**:

### DIFF
--- a/internal/controller/runtime/api_discovery.go
+++ b/internal/controller/runtime/api_discovery.go
@@ -1,0 +1,90 @@
+// Copyright 2023 - 2024 Crunchy Data Solutions, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package runtime
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// API is a combination of Group, Version, and Kind that can be used to check
+// what is available in the Kubernetes API. There are four ways to populate it:
+//  1. Group without Version nor Kind means any resource in that Group.
+//  2. Group with Version but no Kind means any resource in that GV.
+//  3. Group with Kind but no Version means that Kind in any Version of the Group.
+//  4. Group with Version and Kind means that exact GVK.
+type API = schema.GroupVersionKind
+
+type APIs interface {
+	Has(API) bool
+	HasAll(...API) bool
+	HasOne(...API) bool
+}
+
+// APISet implements [APIs] using empty struct for minimal memory consumption.
+type APISet map[API]struct{}
+
+func NewAPISet(api ...API) APISet {
+	s := make(APISet)
+
+	for i := range api {
+		s[api[i]] = struct{}{}
+		s[API{Group: api[i].Group}] = struct{}{}
+		s[API{Group: api[i].Group, Version: api[i].Version}] = struct{}{}
+		s[API{Group: api[i].Group, Kind: api[i].Kind}] = struct{}{}
+	}
+
+	return s
+}
+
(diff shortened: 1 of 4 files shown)


internal/controller/runtime/api_discovery.go
```diff
@@ -0,0 +1,246 @@
...
+	Has(API) bool
+	HasAll(...API) bool
+	HasOne(...API) bool
+}
+
+// APISet implements APIs using empty struct for minimal memory consumption.
+type APISet map[API]struct{}
+
+// APISet implmenents [APIs].
+var _ APIs = (*APISet)(nil)
```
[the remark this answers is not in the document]
dsessler7: Okay, after some reading I think I'm tracking, but let me write my understanding for confirmation...

Generally, the line is ensuring that `APISet` type satisfies the `APIs` interface. A blank identifier is used since the variable is only declared for the aforementioned purpose. The code on the right side of the `=` is a type conversion where you are converting a `nil` value to the "pointer to `APISet`" type. I presume the reason for doing this rather than just using `&APISet{}` is to save memory?

internal/controller/runtime/api_discovery.go
```diff
@@ -0,0 +1,246 @@
...
+	defer r.have.RUnlock()
+	return r.have.HasOne(api...)
+}
+
+type apiContextKey struct{}
+
+// Kubernetes returns the APIs previously stored by [NewAPIContext].
+// When nothing was stored, it returns an empty [APISet].
+func Kubernetes(ctx context.Context) APIs {
+	if apis, ok := ctx.Value(apiContextKey{}).(APIs); ok {
```
dsessler7, line 237: What is the `.(APIs)` doing?

cbandy: This is a [type assertion](https://go.dev/ref/spec#Type_assertions). `context.Value` returns `any`, an empty interface type, and this checks to see that what is inside/behind that interface implements `APIs`.

internal/controller/runtime/api_discovery.go
```diff
@@ -0,0 +1,246 @@
...
+			{Group: "security.openshift.io", Kind: "SecurityContextConstraints"},
+			{Group: "snapshot.storage.k8s.io", Kind: "VolumeSnapshot"},
+			{Group: "trust.cert-manager.io", Kind: "Bundle"},
+		},
+	}
+
+	return runner, err
+}
+
+// NeedLeaderElection returns false so that r runs any [manager.Manager],
```
cbandy, line 128: ⛏️ 

```suggestion
// NeedLeaderElection returns false so that r runs on any [manager.Manager],
```

internal/controller/runtime/api_discovery.go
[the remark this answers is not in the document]
cbandy: I don't know why I use `(*T)(nil)` instead of `&T{}` or `new(T)`. This article uses `nil`: https://go.dev/doc/effective_go#blank

🤔 Maybe this belongs inside a test. It's already covered by `TestAPIContext`, so maybe we drop it here.

cbandy, line 51: ⛏️ 
```suggestion
// APISet implements [APIs] using empty struct for minimal memory consumption.
type APISet map[API]struct{}
```

internal/controller/runtime/api_discovery.go
```diff
@@ -0,0 +1,246 @@
...
+	r.have.RLock()
+	defer r.have.RUnlock()
+	return r.have.HasOne(api...)
+}
+
+type apiContextKey struct{}
+
+// Kubernetes returns the APIs previously stored by [NewAPIContext].
+// When nothing was stored, it returns an empty [APISet].
+func Kubernetes(ctx context.Context) APIs {
```
cbandy, line 236: ❓ What do you think of this interface/experience? IIUC, callers will look like this:

```go
if runtime.Kubernetes(ctx).Has(runtime.API{
  Group: "snapshot.storage.k8s.io", Kind: "VolumeSnapshot",
}) {
  // ...
}
```

dsessler7: Yeah this feels good.

internal/controller/runtime/api_discovery.go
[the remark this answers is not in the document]
cbandy: I've pushed two commits that move these into a test. Increases coverage a bit, too.

PR #3974, opened 2024-08-21
Use snapshot during clone

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

Currently, the VolumeSnapshots that are created are not used by the operator. A user would have to manually target them.

Now, when a postgrescluster clone is being created from another postgrescluster and the source cluster has VolumeSnapshots turned on and has a ReadyToUse snapshot available, the VolumeSnapshot will be used to create the pgdata pvc.

**Other Information**:


internal/controller/postgrescluster/postgres.go
```diff
@@ -621,6 +621,24 @@ func (r *Reconciler) reconcilePostgresDataVolume(
 
 	pvc.Spec = instanceSpec.DataVolumeClaimSpec
 
+	// If a source cluster was provided and VolumeSnapshots are turned on in the source cluster
+	// and there is a VolumeSnapshot available for the source cluster that is ReadyToUse, use it
+	// as the source for the PVC. If there is an error when retrieving VolumeSnapshots, continue
+	// creating PVC in the usual fashion.
```
tony-landreth, line 627: We don't want to surface the error?

dsessler7: Maybe we log it, but continue creating the PVC?

tony-landreth: Works for me.

tony-landreth (approved):

internal/controller/postgrescluster/postgres.go
```diff
@@ -621,6 +622,26 @@ func (r *Reconciler) reconcilePostgresDataVolume(
...
+			snapshot := getLatestReadySnapshot(snapshots)
+			if snapshot != nil {
+				pvc.Spec.DataSource = &corev1.TypedLocalObjectReference{
+					APIGroup: initialize.String("snapshot.storage.k8s.io"),
+					Kind:     "VolumeSnapshot",
+					Name:     snapshot.Name,
+				}
+			}
+		} else {
+			log.Error(err, fmt.Sprintf("Could not get snapshots for %v, proceeding with typical restore process.", sourceCluster.Name))
```
cbandy, line 641: 🔧 If we want people to see this, it should be a warning event on `cluster` instead.

internal/controller/postgrescluster/postgres.go
```diff
@@ -621,6 +622,26 @@ func (r *Reconciler) reconcilePostgresDataVolume(
...
 
+	// If a source cluster was provided and VolumeSnapshots are turned on in the source cluster
+	// and there is a VolumeSnapshot available for the source cluster that is ReadyToUse, use it
+	// as the source for the PVC. If there is an error when retrieving VolumeSnapshots, log it,
+	// but continue creating PVC in the usual fashion.
+	if sourceCluster != nil && sourceCluster.Spec.Backups.Snapshots != nil && feature.Enabled(ctx, feature.VolumeSnapshots) {
+		snapshots, err := r.getSnapshotsForCluster(ctx, sourceCluster)
+		if err == nil {
+			snapshot := getLatestReadySnapshot(snapshots)
+			if snapshot != nil {
```
cbandy, line 633: ❓ I recognize that this can happen. Is it something we should warn (event) about? 🤔 or maybe inform? (info event)

🤔 Perhaps we want the opposite: inform people when we use a snapshot? (info event)

dsessler7: Maybe all of the above? If we find a good snapshot, create an Info event saying "Ready snapshot found; using snapshot to bootstrap pgdata pvc for _ cluster." If snapshots are enabled in the sourceCluster, but we don't find any snapshots, or we find one, but it isn't ReadyToUse, then we issue a Warning event.

internal/controller/postgrescluster/postgres.go
```diff
@@ -621,6 +621,32 @@ func (r *Reconciler) reconcilePostgresDataVolume(
...
+				r.Recorder.Eventf(cluster, corev1.EventTypeNormal, "BootstrappingWithSnapshot",
+					"Snapshot found for %v; bootstrapping cluster with snapshot.", sourceCluster.Name)
+				pvc.Spec.DataSource = &corev1.TypedLocalObjectReference{
+					APIGroup: initialize.String("snapshot.storage.k8s.io"),
+					Kind:     "VolumeSnapshot",
+					Name:     snapshot.Name,
```
cbandy, line 638: 🌱 I wanted to suggest that we read all these fields from `snapshot`, but getting the Group is wonky.

internal/controller/postgrescluster/postgres_test.go
```diff
@@ -251,26 +252,187 @@ func TestReconcilePostgresVolumes(t *testing.T) {
...
+		assert.NilError(t, err)
+
+		assert.Assert(t, metav1.IsControlledBy(pvc, cluster))
+
+		assert.Equal(t, pvc.Labels[naming.LabelCluster], cluster.Name)
+		assert.Equal(t, pvc.Labels[naming.LabelInstance], instance.Name)
+		assert.Equal(t, pvc.Labels[naming.LabelInstanceSet], spec.Name)
+		assert.Equal(t, pvc.Labels[naming.LabelRole], "pgdata")
+
+		assert.Assert(t, marshalMatches(pvc.Spec, `
```
cbandy, line 284: 🏕️ Do you mind changing the ones in this test to [cmp.MarshalMatches](https://github.com/CrunchyData/postgres-operator/blob/b4eb42aa3af6ef917755183bb78c4442bd33245f/internal/testing/cmp/cmp.go#L65) while you're here?

```suggestion
		assert.Assert(t, cmp.MarshalMatches(pvc.Spec, `
```

cbandy (approved):

cbandy (approved):

PR #3994, opened 2024-09-16
Always try "stanza-upgrade" after a failed "stanza-create"

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Error messages could be on either stderr or stdout depending on logging options. Error messages and exit codes could change unexpectedly, so use a shell list to run "stanza-upgrade" any time "stanza-create" exits non-zero.

**Other Information**:
Issue: PGO-1558

### DIFF
--- a/internal/controller/postgrescluster/pgbackrest.go
+++ b/internal/controller/postgrescluster/pgbackrest.go
@@ -2678,8 +2678,7 @@ func (r *Reconciler) reconcileStanzaCreate(ctx context.Context,
 	}
 
 	// Always attempt to create pgBackRest stanza first
-	configHashMismatch, err := pgbackrest.Executor(exec).StanzaCreateOrUpgrade(ctx, configHash,
-		false, postgresCluster)
+	configHashMismatch, err := pgbackrest.Executor(exec).StanzaCreateOrUpgrade(ctx, configHash, postgresCluster)
 	if err != nil {
 		// record and log any errors resulting from running the stanza-create command
 		r.Recorder.Event(postgresCluster, corev1.EventTypeWarning, EventUnableToCreateStanzas,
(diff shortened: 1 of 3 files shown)


benjaminjb:
Makes sense, but I wonder if it would be worthwhile to capture the errors and return that to the user (which would take some flexibility since, as you point out, an error might be on stderr or stdout)

tjmoore4:
> Makes sense, but I wonder if it would be worthwhile to capture the errors and return that to the user (which would take some flexibility since, as you point out, an error might be on stderr or stdout)

This should keep our error behavior the same overall. The main impact is that we try an stanza upgrade immediately after a stanza create fails rather than depending on a particular error to trigger the upgrade. The existing code would similarly not print the initial create error without first trying an upgrade.

benjaminjb:
> This should keep our error behavior the same overall. The main impact is that we try an stanza upgrade immediately after a stanza create fails rather than depending on a particular error to trigger the upgrade. The existing code would similarly not print the initial create error without first trying an upgrade.

Yeah, I guess I'm wondering if we get an error on create and then get an error on upgrade, we will only ever see the upgrade error, right? Not a blocker for me, I was just wondering.

tjmoore4:
> > This should keep our error behavior the same overall. The main impact is that we try an stanza upgrade immediately after a stanza create fails rather than depending on a particular error to trigger the upgrade. The existing code would similarly not print the initial create error without first trying an upgrade.
> 
> Yeah, I guess I'm wondering if we get an error on create and then get an error on upgrade, we will only ever see the upgrade error, right? Not a blocker for me, I was just wondering.

Yeah, that's my understanding. The approach is effectively, if the `create` error is solved by calling `upgrade` instead, don't bother the user. Definitely an argument to be made either way, but I think this approach is reasonable.

PR #3996, opened 2024-09-17
Reject pull requests that change imported licenses

cbandy:
We import dependencies that use a handful of open-source licenses. We want to be intentional about any change to these licenses, so this automation flags pull requests that do so.

Go modules are immutable, so checking during pull requests and pushes should suffice.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement
 - [x] Other

**Other Information**:

Issue: PGO-1556

### DIFF
--- a/trivy.yaml
+++ b/trivy.yaml
@@ -1,0 +1,14 @@
+# https://aquasecurity.github.io/trivy/latest/docs/references/configuration/config-file/
+---
+# Specify an exact list of recognized and acceptable licenses.
+# [A GitHub workflow](/.github/workflows/trivy.yaml) rejects pull requests that
+# import licenses not in this list.
+#
+# https://aquasecurity.github.io/trivy/latest/docs/scanner/license/
+license:
+  ignored:
+    - Apache-2.0
+    - BSD-2-Clause
+    - BSD-3-Clause
+    - ISC
+    - MIT
--- a/.github/workflows/trivy.yaml
+++ b/.github/workflows/trivy.yaml
@@ -1,5 +1,3 @@
-# Uses Trivy to scan every pull request, rejecting those with severe, fixable vulnerabilities.
-# Scans on PR to master and weekly with same behavior.
 name: Trivy
 
 on:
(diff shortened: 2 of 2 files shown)


trivy.yaml
```diff
@@ -0,0 +1,13 @@
+---
+# Specify an exact list of recognized and acceptable licenses.
+# [A GitHub workflow](/.github/workflows/trivy.yaml) rejects pull requests that
+# import licenses not in this list.
+#
+# https://aquasecurity.github.io/trivy/latest/docs/scanner/license/
+license:
+  notice:
+    #- Apache-2.0
```
cbandy, line 9: 🔧 Omitting Apache for now to confirm that the PR check will fail.

cbandy:
Here's what it looks like when a license is not in the configured list: https://github.com/CrunchyData/postgres-operator/actions/runs/10914660621/job/30293145553?pr=3996

trivy.yaml
```diff
@@ -0,0 +1,14 @@
+# https://aquasecurity.github.io/trivy/latest/docs/references/configuration/config-file/
+---
+# Specify an exact list of recognized and acceptable licenses.
+# [A GitHub workflow](/.github/workflows/trivy.yaml) rejects pull requests that
+# import licenses not in this list.
+#
+# https://aquasecurity.github.io/trivy/latest/docs/scanner/license/
+license:
+  ignored:
```
benjaminjb, line 9: So right now we're telling Trivy to ignore these licenses; if another license shows up that's not in this list, we'll get an alert because the job is configured with `--exit-code=1`: any license at any level -- `Forbidden`, `Restricted`, `Unencumbered`, et al. -- will alert if it is not ignored, right?

benjaminjb (approved):

trivy.yaml
cbandy: Right. Every license that Trivy recognizes has a severity, and this job alerts on all severities.

We'll also be alerted if Trivy can't detect the license. Those get a severity of `UNKNOWN`.

PR #3997, opened 2024-09-18
Snapshot dedicated pvc

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Currently, when the operator creates a VolumeSnapshot, it does it on the cluster's primary pgdata volume with the database running.

When snapshots are enabled a dedicated snapshot volume (DSV) will be created and every time a backup completes, a delta restore will be done in the DSV. When the restore finishes successfully, a snapshot will be created of the DSV. 

**Other Information**:

### DIFF
--- a/internal/controller/postgrescluster/pgbackrest.go
+++ b/internal/controller/postgrescluster/pgbackrest.go
@@ -32,6 +32,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/crunchydata/postgres-operator/internal/config"
+	"github.com/crunchydata/postgres-operator/internal/feature"
 	"github.com/crunchydata/postgres-operator/internal/initialize"
 	"github.com/crunchydata/postgres-operator/internal/logging"
 	"github.com/crunchydata/postgres-operator/internal/naming"
@@ -197,7 +198,7 @@ func (r *Reconciler) applyRepoVolumeIntent(ctx context.Context,
 // getPGBackRestResources returns the existing pgBackRest resources that should utilized by the
 // PostgresCluster controller during reconciliation.  Any items returned are verified to be owned
 // by the PostgresCluster controller and still applicable per the current PostgresCluster spec.
-// Additionally, and resources identified that no longer correspond to any current configuration
+// Additionally, any resources identified that no longer correspond to any current configuration
 // are deleted.
 func (r *Reconciler) getPGBackRestResources(ctx context.Context,
 	postgresCluster *v1beta1.PostgresCluster,
@@ -374,6 +375,15 @@ func (r *Reconciler) cleanupRepoResources(ctx context.Context,
 			if !backupsSpecFound {
 				break
 			}
+
+			// If the restore job has the PGBackRestBackupJobCompletion annotation, it is
+			// used for volume snapshots and should not be deleted (volume snapshots code
+			// will clean it up when appropriate).
+			if _, ok := owned.GetAnnotations()[naming.PGBackRestBackupJobCompletion]; ok {
+				ownedNoDelete = append(ownedNoDelete, owned)
+				delete = false
+			}
+
 			// When a cluster is prepared for restore, the system identifier is removed from status
 			// and the cluster is therefore no longer bootstrapped.  Only once the restore Job is
 			// complete will the cluster then be bootstrapped again, which means by the time we
@@ -762,7 +772,7 @@ func (r *Reconciler) generateRepoVolumeIntent(postgresCluster *v1beta1.PostgresC
 }
 
 // generateBackupJobSpecIntent generates a JobSpec for a pgBackRest backup job
-func generateBackupJobSpecIntent(postgresCluster *v1beta1.PostgresCluster,
+func generateBackupJobSpecIntent(ctx context.Context, postgresCluster *v1beta1.PostgresCluster,
 	repo v1beta1.PGBackRestRepo, serviceAccountName string,
 	labels, annotations map[string]string, opts ...string) *batchv1.JobSpec {
 
@@ -771,6 +781,11 @@ func generateBackupJobSpecIntent(postgresCluster *v1beta1.PostgresCluster,
 		"--stanza=" + pgbackrest.DefaultStanzaName,
 		"--repo=" + repoIndex,
 	}
+	// If VolumeSnapshots are enabled, use archive-copy and archive-check options
+	if postgresCluster.Spec.Backups.Snapshots != nil && feature.Enabled(ctx, feature.VolumeSnapshots) {
+		cmdOpts = append(cmdOpts, "--archive-copy=y", "--archive-check=y")
+	}
+
 	cmdOpts = append(cmdOpts, opts...)
 
 	container := corev1.Container{
@@ -1634,6 +1649,9 @@ func (r *Reconciler) reconcilePostgresClusterDataSource(ctx context.Context,
 		return errors.WithStack(err)
 	}
 
+	// TODO(snapshots): If pgdata is being sourced by a VolumeSnapshot then don't perform a typical restore job;
+	// we only want to replay the WAL.
+
 	// reconcile the pgBackRest restore Job to populate the cluster's data directory
 	if err := r.reconcileRestoreJob(ctx, cluster, sourceCluster, pgdata, pgwal, pgtablespaces,
 		dataSource, instanceName, instanceSetName, configHash, pgbackrest.DefaultStanzaName); err != nil {
@@ -2362,7 +2380,7 @@ func (r *Reconciler) reconcileManualBackup(ctx context.Context,
 	backupJob.ObjectMeta.Labels = labels
 	backupJob.ObjectMeta.Annotations = annotations
 
-	spec := generateBackupJobSpecIntent(postgresCluster, repo,
+	spec := generateBackupJobSpecIntent(ctx, postgresCluster, repo,
 		serviceAccount.GetName(), labels, annotations, backupOpts...)
 
 	backupJob.Spec = *spec
@@ -2523,7 +2541,7 @@ func (r *Reconciler) reconcileReplicaCreateBackup(ctx context.Context,
 	backupJob.ObjectMeta.Labels = labels
 	backupJob.ObjectMeta.Annotations = annotations
 
-	spec := generateBackupJobSpecIntent(postgresCluster, replicaCreateRepo,
+	spec := generateBackupJobSpecIntent(ctx, postgresCluster, replicaCreateRepo,
 		serviceAccount.GetName(), labels, annotations)
 
 	backupJob.Spec = *spec
@@ -2886,8 +2904,7 @@ func (r *Reconciler) reconcilePGBackRestCronJob(
 	labels := naming.Merge(
 		cluster.Spec.Metadata.GetLabelsOrNil(),
 		cluster.Spec.Backups.PGBackRest.Metadata.GetLabelsOrNil(),
-		naming.PGBackRestCronJobLabels(cluster.Name, repo.Name, backupType),
-	)
+		naming.PGBackRestCronJobLabels(cluster.Name, repo.Name, backupType))
 	objectmeta := naming.PGBackRestCronJob(cluster, backupType, repo.Name)
 
 	// Look for an existing CronJob by the associated Labels. If one exists,
@@ -2951,7 +2968,7 @@ func (r *Reconciler) reconcilePGBackRestCronJob(
 	// set backup type (i.e. "full", "diff", "incr")
 	backupOpts := []string{"--type=" + backupType}
 
-	jobSpec := generateBackupJobSpecIntent(cluster, repo,
+	jobSpec := generateBackupJobSpecIntent(ctx, cluster, repo,
 		serviceAccount.GetName(), labels, annotations, backupOpts...)
 
 	// Suspend cronjobs when shutdown or read-only. Any jobs that have already
--- a/internal/controller/postgrescluster/snapshots.go
+++ b/internal/controller/postgrescluster/snapshots.go
@@ -6,6 +6,8 @@ package postgrescluster
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -16,8 +18,12 @@ import (
 
 	volumesnapshotv1 "github.com/kubernetes-csi/external-snapshotter/client/v8/apis/volumesnapshot/v1"
 
+	"github.com/crunchydata/postgres-operator/internal/config"
 	"github.com/crunchydata/postgres-operator/internal/feature"
+	"github.com/crunchydata/postgres-operator/internal/initialize"
 	"github.com/crunchydata/postgres-operator/internal/naming"
+	"github.com/crunchydata/postgres-operator/internal/pgbackrest"
+	"github.com/crunchydata/postgres-operator/internal/postgres"
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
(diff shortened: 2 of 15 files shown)


internal/controller/postgrescluster/snapshots.go
```diff
@@ -16,53 +18,82 @@ import (
...
+//  7. When the snapshot job completes, we delete the restore job.
 func (r *Reconciler) reconcileVolumeSnapshots(ctx context.Context,
-	postgrescluster *v1beta1.PostgresCluster, instances *observedInstances,
-	clusterVolumes []corev1.PersistentVolumeClaim) error {
+	postgrescluster *v1beta1.PostgresCluster, pvc *corev1.PersistentVolumeClaim) error {
 
 	// Get feature gate state
 	volumeSnapshotsFeatureEnabled := feature.Enabled(ctx, feature.VolumeSnapshots)
 
+	// TODO(dsessler7): If user has a snapshots section in their cluster spec, indicating
```
tony-landreth, line 55: Do we warn like this for other feature gates?

internal/controller/postgrescluster/snapshots.go
```diff
@@ -133,49 +148,286 @@ func (r *Reconciler) reconcileVolumeSnapshots(ctx context.Context,
...
+	addTMPEmptyDir(&restoreJob.Spec.Template)
+
+	restoreJob.Annotations[naming.PGBackRestBackupJobCompletion] = backupJob.Status.CompletionTime.String()
+	return errors.WithStack(r.apply(ctx, restoreJob))
+}
+
+// generateSnapshotOfDedicatedSnapshotVolume will generate a VolumeSnapshot of
+// the dedicated snapshot PersistentVolumeClaim and annotate it with the
+// provided backup job's UID.
+func (r *Reconciler) generateSnapshotOfDedicatedSnapshotVolume(
```
tony-landreth, line 419: :pick: Feels like the code in `generateVolumeSnapshot` should be in here.

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -1634,6 +1649,9 @@ func (r *Reconciler) reconcilePostgresClusterDataSource(ctx context.Context,
 		return errors.WithStack(err)
 	}
 
+	// TODO(dsessler): If pgdata is being sourced by a VolumeSnapshot then don't perform a typical restore job;
```
tony-landreth, line 1652: Can this TODO be deleted?

internal/controller/postgrescluster/snapshots.go
```diff
@@ -133,49 +148,286 @@ func (r *Reconciler) reconcileVolumeSnapshots(ctx context.Context,
...
+		return pvc, nil
 	}
-	// Return error if primary instance not found
-	if primaryInstance.Name == "" {
-		return nil, errors.New("Could not find primary instance. Cannot create volume snapshot.")
+
+	// If pvc has been annotated with latest backup job's completion time, then restore was
+	// successful and there's nothing left to do.
+	backupJobCompletionTime := backupJob.Status.CompletionTime.String()
+	// TODO: change this to check whether backupJobCompletionTime is more recent than pvc annotation.
```
tony-landreth, line 225: Is the `==` good enough here?

internal/controller/postgrescluster/snapshots.go
```diff
@@ -133,49 +148,286 @@ func (r *Reconciler) reconcileVolumeSnapshots(ctx context.Context,
...
+	if pvc.GetAnnotations()[naming.PGBackRestBackupJobCompletion] == backupJobCompletionTime {
+		return pvc, nil
+	}
+
+	// If we've made it here, the pvc has not been restored with latest backup.
+	// Find the dedicated snapshot volume restore job if it exists. Since we delete
+	// successful restores after we annotate the PVC and stop making restore jobs
+	// if a failed DSV restore job exists, there should only ever be one DSV restore
+	// job in existence at a time.
+	// TODO(dsessler7): Should this function throw an error or something if multiple
```
tony-landreth, line 235: I think our current setup prevents multiple restore jobs from being created, due to name collisions, right? But if there were multiple restore jobs, you'd want to make sure you retrieve the right one, because the `PGBackRestBackupJobCompletion` annotation will be propagated from whichever one is chosen.

internal/controller/postgrescluster/snapshots.go
```diff
@@ -133,49 +148,286 @@ func (r *Reconciler) reconcileVolumeSnapshots(ctx context.Context,
...
+		return pvc, nil
+	}
+
+	// If we made it here, the restore job is still running and we should do nothing.
+	return pvc, err
+}
+
+// createDedicatedSnapshotVolume creates/updates/gets the dedicated snapshot volume.
+// It expects that the volume name and GVK has already been set on the pvc that is passed in.
+// TODO(dsessler7): The above fact makes this method feel a little awkward to me.
```
tony-landreth, line 283: If you feel it won't compromise the readability of `reconcileDedicatedSnapshotVolume`, go for it.

internal/controller/postgrescluster/snapshots.go
dsessler7: I kept it in its separate function so that it can be reused if we ever snapshot other volumes.

internal/controller/postgrescluster/pgbackrest.go
dsessler7: It hasn't been implemented, so I figured it would stay...

internal/controller/postgrescluster/snapshots.go
dsessler7: I believe so... Unless someone manually changes the annotation value on the PVC to an incorrect timestamp it will always either be equal to the `backupJobCompletionTime`, indicating that the PVC is up to date, or have an earlier timestamp (!= `backupJobCompletionTime`), indicating that it needs to be restored.

internal/controller/postgrescluster/snapshots.go
dsessler7: I can add some logic to `getDedicatedSnapshotVolumeRestoreJob` to always get the latest restore job in case there ever is more than one, but there really shouldn't ever be more than one because we only ever create a restore job if none are found (lines 243-246).

internal/controller/postgrescluster/snapshots.go
dsessler7: On second thought, I think `==` isn't good enough. Example scenario: A user is triggering backups with the manual process and has already had a successful backup that was then restored and snapshotted. When they trigger a new manual backup, the old one is deleted. Before the new backup completes, the only existing completed backup is the initial `replica-create` backup we always do on cluster creation. The timestamp for that initial backup is grabbed and is different than the timestamp in the pvc annotation so a restore job is created before the new manual backup completes.

internal/pgbackrest/config.go
```diff
@@ -263,6 +263,67 @@ mv "${pgdata}" "${pgdata}_bootstrap"`
...
+BACKUP_LABEL="xxx"
+if [[ -e "` + pgdata + `/backup_label" ]]
+then
+  if BACKUP_LABEL=$(md5sum ` + pgdata + `/backup_label); then
+    BACKUP_LABEL=$(echo "${BACKUP_LABEL}" | awk '{print $1}')
+  else
+    echo "Error getting checksum of backup_label."
+    exit 1
+  fi
+fi
```
cbandy, line 298: ⛏️ It looks like `pgdata` (Go) is passed into the script and assigned to `pgdata` (Bash). If so, these string concat can be replaced with shell strings:

```diff
-if [[ -e "` + pgdata + `/backup_label" ]]
+if [[ -e "${pgdata}/backup_label" ]]
```

:pick: The label command looks the same both before and after. If so, we can drop the call to `awk`.

:memo: `bash -e` halts the script whenever a command list exits non-zero.

🤔 If the error message from `md5sum` is sufficient, we can combine all that and let it halt the script:

```suggestion
	restoreScript := `declare -r pgdata="$1" opts="$2"
BACKUP_LABEL=$([[ ! -e "${pgdata}/backup_label" ]] || md5sum "${pgdata}/backup_label")
```

internal/controller/postgrescluster/pgbackrest.go
cbandy: 🌱 Rather than attach my name, I've started using topics in the parentheses. I'm not sure what word makes sense.

```suggestion
	// TODO(snapshots): If pgdata is being sourced by a VolumeSnapshot then don't perform a typical restore job;
```

tony-landreth (approved):

internal/controller/postgrescluster/snapshots.go
```diff
@@ -133,49 +148,295 @@ func (r *Reconciler) reconcileVolumeSnapshots(ctx context.Context,
...
+	// successful, set/update the annotation on the PVC and delete the restore job.
+	if restoreJob.Status.Succeeded == 1 {
+		if pvc.GetAnnotations() == nil {
+			pvc.Annotations = map[string]string{}
+		}
+		pvc.Annotations[naming.PGBackRestBackupJobCompletion] = restoreJob.GetAnnotations()[naming.PGBackRestBackupJobCompletion]
+		annotations := fmt.Sprintf(`{"metadata":{"annotations":{"%s": "%s"}}}`,
+			naming.PGBackRestBackupJobCompletion, pvc.Annotations[naming.PGBackRestBackupJobCompletion])
+
+		patch := client.RawPatch(client.Merge.Type(), []byte(annotations))
```
cbandy, line 267: 🤔 Interesting. @cbandy

internal/controller/postgrescluster/snapshots.go
```diff
@@ -133,49 +148,295 @@ func (r *Reconciler) reconcileVolumeSnapshots(ctx context.Context,
...
+
+	pgdata := postgres.DataDirectory(cluster)
+	repoName := backupJob.GetLabels()[naming.LabelPGBackRestRepo]
+
+	opts := []string{
+		"--stanza=" + pgbackrest.DefaultStanzaName,
+		"--pg1-path=" + pgdata,
+		"--repo=" + regexRepoIndex.FindString(repoName),
+		"--delta",
+		// "--link-all",
```
cbandy, line 354: 🤔 IIUC, we don't want a symlink to a WAL volume that doesn't exist. Perhaps instead of saying nothing about `--link-all` we should say "no" symlinks, however that's spelled. `x=false` ? `--no-x` ?

❓ What happens with tablespaces?

internal/controller/postgrescluster/snapshots.go
```diff
@@ -133,49 +148,292 @@ func (r *Reconciler) reconcileVolumeSnapshots(ctx context.Context,
...
+	// and consistently use the same one as the user *could* have multiple instanceSets with different
+	// storage settings. Do we want to simply always choose the first instanceSet and instance in their
+	// respective arrays? Or do we want to add something to the API for the user to specify an instanceSet?
+	// For now, I am choosing the first instanceSet/instance in their arrays.
+	instanceSet := &cluster.Spec.InstanceSets[0]
+	instance := instances.bySet[instanceSet.Name][0].Runner
+
+	// How are we going to handle tablespace volumes? Feels like if they exist we need to provide them to
+	// the restore command otherwise they will break in any clones made from a snapshot.
+	pgtablespaces, err := r.reconcileTablespaceVolumes(ctx, cluster, instanceSet, instance, clusterVolumes)
```
cbandy, line 367: ❓ What is returned here? Is this the regular set of tablespace volumes? Do we need dedicated copies to quiesce and snapshot?

internal/controller/postgrescluster/snapshots.go
```diff
@@ -16,53 +18,82 @@ import (
...
+	// TODO(dsessler7): If user has a snapshots section in their cluster spec, indicating
+	// that they would like to use snapshots, but the VolumeSnapshots feature gate is not
+	// turned on, should we emit an event or log something?
+
 	// Check if the Kube cluster has VolumeSnapshots installed. If VolumeSnapshots
-	// are not installed we need to return early. If user is attempting to use
+	// are not installed, we need to return early. If user is attempting to use
 	// VolumeSnapshots, return an error, otherwise return nil.
-	volumeSnapshotsExist, err := r.GroupVersionKindExists("snapshot.storage.k8s.io/v1", "VolumeSnapshot")
+	volumeSnapshotKindExists, err := r.GroupVersionKindExists("snapshot.storage.k8s.io/v1", "VolumeSnapshot")
```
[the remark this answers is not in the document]
dsessler7: Yeah... kind of want to get this PR merged, but maybe after it is merged I can add to the api discovery PR to make snapshots use it and then we can merged that one.

internal/controller/postgrescluster/snapshots.go
```diff
@@ -16,53 +18,82 @@ import (
...
+	// cluster has a Spec.Backups.Snapshots section defined).
 
+	// The specified VolumeSnapshotClass needs to exist.
+	volumeSnapshotClass := &volumesnapshotv1.VolumeSnapshotClass{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: postgrescluster.Spec.Backups.Snapshots.VolumeSnapshotClassName,
+		},
+	}
+	err = errors.WithStack(r.Client.Get(ctx, client.ObjectKeyFromObject(volumeSnapshotClass),
+		volumeSnapshotClass))
```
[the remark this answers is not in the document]
dsessler7: Yes, this is why we need those permissions on `volumesnapshotclass`.

I believe it will fail/error if you attempt to use a `volumesnapshotclass` that doesn't exist (I will double check this).

Are you thinking that we don't bother checking that the class exists but just pass the error/create an event when the snapshot creation fails due to a non-existent `volumesnapshotclass` being specified?

internal/controller/postgrescluster/snapshots.go
```diff
@@ -133,49 +148,295 @@ func (r *Reconciler) reconcileVolumeSnapshots(ctx context.Context,
...
+	pvcAnnotationTimestampString, pvcAnnotationExists := pvc.GetAnnotations()[naming.PGBackRestBackupJobCompletion]
+	if pvcAnnotationExists {
+		// If pvc has been annotated, we want to compare its timestamp to the latest backup job's
+		// timestamp. If the pvc's timestamp is older than the latest backup job's
+		// completion time, then we want to do a restore.
+		backupJobCompletionTime := backupJob.Status.CompletionTime.Time
+		pvcAnnotationTime, err := time.Parse(time.RFC3339, pvcAnnotationTimestampString)
+		if err != nil {
+			return pvc, err
 		}
```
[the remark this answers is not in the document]
dsessler7: Hmmm... Yes we manage the annotation so I wouldn't expect it to ever fail to parse unless the user messed with it... What would repairing it look like?

internal/controller/postgrescluster/snapshots.go
```diff
@@ -16,53 +18,82 @@ import (
...
+	// are not installed, we need to return early. If user is attempting to use
 	// VolumeSnapshots, return an error, otherwise return nil.
-	volumeSnapshotsExist, err := r.GroupVersionKindExists("snapshot.storage.k8s.io/v1", "VolumeSnapshot")
+	volumeSnapshotKindExists, err := r.GroupVersionKindExists("snapshot.storage.k8s.io/v1", "VolumeSnapshot")
 	if err != nil {
 		return err
 	}
-	if !*volumeSnapshotsExist {
+	if !*volumeSnapshotKindExists {
 		if postgrescluster.Spec.Backups.Snapshots != nil && volumeSnapshotsFeatureEnabled {
```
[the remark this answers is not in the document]
dsessler7: Not sure I follow. If the GVK check errors we return the error, so if we get to this `if`, the GVK check didn't error. If the GVK doesn't exist, but the user has snapshots turned on we want to error, but if they don't have snapshots turned on then we just return `nil`...

internal/controller/postgrescluster/snapshots.go
dsessler7: Thanks for bringing this up. Brian was using `--link-all` in his script. I wasn't sure if we wanted to use it which is why it is there, but commented out.

With tablespaces, we do the same thing that we do in our other `RestoreCommand` function...

internal/controller/postgrescluster/snapshots.go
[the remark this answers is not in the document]
dsessler7: @dsessler7 What happens when user leaves `volumeSnapshotClass` blank/empty string?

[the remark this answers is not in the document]
dsessler7: It errors and does not attempt to create a snapshot if you give an empty string for `volumeSnapshotClassName`: 
```
ERRO[2024-10-01T12:48:49-07:00] Reconciler error                              PostgresCluster=postgres-operator/hippo controller=postgrescluster controllerGroup=postgres-operator.crunchydata.com controllerKind=PostgresCluster error="admission webhook \"validation-webhook.snapshot.storage.k8s.io\" denied the request: Spec.VolumeSnapshotClassName must not be the empty string" file="internal/controller/postgrescluster/snapshots.go:145" func="postgrescluster.(*Reconciler).reconcileVolumeSnapshots" name=hippo namespace=postgres-operator reconcileID=642232f6-ca1f-40df-8e63-8d51ddc4bc40 version=
```
However, if you give a name for a non-existent `VolumeSnapshotClass` then it will create a VolumeSnapshot object, but the snapshot will have an error in its `Status`. We currently create a warning event when we detect an error in any of the existing VolumeSnapshots, but we don't log an error and we don't stop the operator creating more snapshots... Should this behavior change in some way? :thinking:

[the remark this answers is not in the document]
cbandy: > Should this behavior change in some way? 🤔

We can prevent the empty string with API validation; a `+kubebuilder:validation:MinLength=1` in here:

https://github.com/CrunchyData/postgres-operator/blob/66174ec98044d395bef1165a69f2fc11139c0a5f/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go#L693-L698

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -1634,6 +1649,9 @@ func (r *Reconciler) reconcilePostgresClusterDataSource(ctx context.Context,
 		return errors.WithStack(err)
 	}
 
+	// TODO(snapshots): If pgdata is being sourced by a VolumeSnapshot then don't perform a typical restore job;
+	// we only want to replay the WAL.
```
cbandy, line 1653: ❓ This sounds like the behavior in `pgbackrest.DedicatedSnapshotVolumeRestoreCommand`. If so, should this `TODO` go away?

internal/controller/postgrescluster/snapshots.go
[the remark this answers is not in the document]
cbandy: 🤔 🌱 Disabling the gate should stop all activity for the feature. That is, I don't think we should clean up when the gate is disabled.

- gate disabled, cluster snapshots configured → do nothing
- gate disabled, cluster snapshots omitted → do nothing
- gate enabled, cluster snapshots configured → keep/do snapshots
- gate enabled, cluster snapshots omitted → clean up snapshots, volume, jobs, etc.

🤔 Is there any reason not to do it that way?

cbandy (approved):
Please squash into fewer commits before merging.

internal/controller/postgrescluster/pgbackrest.go
dsessler7: The `DedicatedSnapshotVolumeRestoreCommand` does the delta restore into the dedicated volume and then the volume is snapshotted. Theoretically, because of the way we backup/restore/snapshot, when a clone is created using a snapshot, we should be able to just replay WAL to get the clone up to speed with the source cluster. However, as of this moment, we still do a delta restore after the clone pgdata volume has been sourced with a snapshot. This TODO is a reminder to implement the WAL replay for clones made from snapshots in the future.

internal/controller/postgrescluster/snapshots.go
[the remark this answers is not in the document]
dsessler7: Okay, I'll change the behavior.

PR #4012, opened 2024-10-09
Use the upstream Trivy action to scan licenses

cbandy:
The upstream action no longer runs in a container, so it can access the job environment and Go modules. The license job no longer needs any `ghcr.io` package, so it is no longer affected by the recent rate limiting issues.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

### DIFF
--- a/.github/workflows/trivy.yaml
+++ b/.github/workflows/trivy.yaml
@@ -19,31 +19,15 @@ jobs:
         with: { go-version: stable }
       - run: go mod download
 
-      # Login to the GitHub Packages registry to avoid rate limiting.
-      # - https://aquasecurity.github.io/trivy/v0.55/docs/references/troubleshooting/#github-rate-limiting
-      # - https://github.com/aquasecurity/trivy/issues/7580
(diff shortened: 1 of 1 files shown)


benjaminjb (approved):
LGTM. (Though ❓ : you prefer pinning to master rather than the 0.26.0 tag?)

What do you think about their current suggestion re: caching the vuln db in a default branch daily and using that cached db in other branches?

cbandy:
> LGTM. (Though ❓ : you prefer pinning to master rather than the 0.26.0 tag?)

👍🏻 I'll bump these actions to the tag. [Dependabot](https://github.com/CrunchyData/postgres-operator/actions/workflows/dependabot/dependabot-updates) should keep us informed/updated now.

> What do you think about their current suggestion re: caching the vuln db in a default branch daily and using that cached db in other branches?

I'm going to stop worrying about this in GitHub. They know the problem now, so I'm OK waiting for them to make the necessary changes to their action.

PR #4015, opened 2024-10-19
Update controller-gen and remove our code generation workarounds

cbandy:
Recent [controller-gen](https://github.com/kubernetes-sigs/controller-tools/releases) does everything we had to do with Ruby and Kustomize.

This removes the `build/crd`, `config/singlenamespace`, and `installers` directories. In their place, I've added two validation rules to _most_ of our PVC specs. The pgAdmin APIs don't have these rules, and adding them causes a test to fail.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Other

**Other Information**:

```console
$ make install
kubectl apply --server-side -k ./config/crd
customresourcedefinition.apiextensions.k8s.io/crunchybridgeclusters.postgres-operator.crunchydata.com serverside-applied
customresourcedefinition.apiextensions.k8s.io/pgadmins.postgres-operator.crunchydata.com serverside-applied
customresourcedefinition.apiextensions.k8s.io/pgupgrades.postgres-operator.crunchydata.com serverside-applied
customresourcedefinition.apiextensions.k8s.io/postgresclusters.postgres-operator.crunchydata.com serverside-applied

# or make deploy-dev

$ kubectl get crd -l app.kubernetes.io/name=pgo -L app.kubernetes.io/version
NAME                                                      CREATED AT             VERSION
crunchybridgeclusters.postgres-operator.crunchydata.com   2024-10-19T02:36:33Z   latest
pgadmins.postgres-operator.crunchydata.com                2024-10-19T02:36:33Z   latest
pgupgrades.postgres-operator.crunchydata.com              2024-10-19T02:36:33Z   latest
postgresclusters.postgres-operator.crunchydata.com        2024-10-19T02:36:33Z   latest
```


internal/patroni/rbac.go
```diff
@@ -12,25 +12,25 @@ import (
 )
 
 // "list", "patch", and "watch" are required. Include "get" for good measure.
-// +kubebuilder:rbac:namespace=patroni,groups="",resources="pods",verbs={get}
-// +kubebuilder:rbac:namespace=patroni,groups="",resources="pods",verbs={list,watch}
-// +kubebuilder:rbac:namespace=patroni,groups="",resources="pods",verbs={patch}
```
tjmoore4, line 17: :thinking: What was `namespace=patroni` doing/signifying here?

config/crd/kustomization.yaml
```diff
@@ -1,8 +1,17 @@
...
 - bases/postgres-operator.crunchydata.com_pgadmins.yaml
+
+patches:
+- target:
+    kind: CustomResourceDefinition
+  patch: |-
+    - op: add
+      path: /metadata/labels
+      value:
+        app.kubernetes.io/name: pgo
```
tjmoore4, line 16: :question: I like moving these labels to one location, but I am curious if this changes "what" the CRD is.

To hopefully explain that better, it seems the CRD was a single file (`config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml`) and now it's the base file + this patch. Does that sound right?

internal/patroni/rbac.go
benjaminjb: IIRC, the `namespace` part of the kubebuilder RBAC marker is a way to tell kubebuilder whether we want a ClusterRole or a Role specifically for a namespace:
```
If not set, the Rule belongs to the generated ClusterRole. If set, the Rule belongs to a Role, whose namespace is specified by this field.
```
[Citation](https://book.kubebuilder.io/reference/markers/rbac)

We were setting the namespace here (from the beginning of this file's existence, afaict) as a way to remind ourselves that these permissions were required by patroni; and then in our RBAC generation, we were (essentially) stripping that namespace out so that these permissions would just be part of the ClusterRole/Role that you need to run PGO. 

Now with some updates to controller-gen, we don't need to do that bit of processing; and since this rbac is specifically in the patroni directory, that's enough to tell us what these permissions are for, so we can drop the `namespace` field.

tjmoore4: :+1: That makes sense, thanks for the explanation.

tjmoore4 (approved):

config/crd/kustomization.yaml
cbandy: Yes, it does a bit. These files in `config/crd/bases` differ, but this `config/crd` [kustomization](https://kubectl.docs.kubernetes.io/references/kustomize/glossary/#kustomization) produces the same things as before. The remaining Make targets and scripts use the kustomization.

jmckulk (approved):

benjaminjb (approved):
Seems a little funny not to mention singlenamespace, but I suppose we do that elsewhere, e.g., pgo-examples

config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
```diff
@@ -3101,10 +3103,14 @@ spec:
...
-                                  - accessModes
-                                  - resources
                                   type: object
+                                  x-kubernetes-validations:
+                                  - message: missing accessModes
+                                    rule: has(self.accessModes) && size(self.accessModes)
+                                      > 0
+                                  - message: missing storage request
+                                    rule: has(self.resources) && has(self.resources.requests)
+                                      && has(self.resources.requests.storage)
```
cbandy, line 3113: 🤔 This affects the API reference documentation. The **required** column changes from yes to no.

PR #4021, opened 2024-10-28
Check that snapshot.Status is not nil when checking Status properties.

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

We currently don't check that a snapshot's Status isn't nil when checking Status properties. There is an edge case where a snapshot does not yet have a `Status` which can lead to a nil pointer dereference.

We now check for the existence of `snapshot.Status` when checking a snapshot's Status properties.

**Other Information**:


tjmoore4 (commented):
:question: These changes look good, but is there some way we can update our tests for this scenario?

tjmoore4 (approved):
:+1: Looks good

PR #4028, opened 2024-11-05
Create warning event when postgres version is EOL.

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

When a user has a postgrescluster that is using an End-Of-Life version of postgres, we don't warn them.

We add a warning event to the postgrescluster if the postgres version is End-Of-Life.

**Other Information**:

### DIFF
--- a/internal/controller/postgrescluster/controller.go
+++ b/internal/controller/postgrescluster/controller.go
@@ -136,6 +136,16 @@ func (r *Reconciler) Reconcile(
 			return runtime.ErrorWithBackoff(err)
 		}
 	}
+	// Issue Warning Event if postgres version is EOL according to PostgreSQL:
+	// https://www.postgresql.org/support/versioning/
+	currentTime := time.Now()
+	if postgres.ReleaseIsFinal(cluster.Spec.PostgresVersion, currentTime) {
+		r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "EndOfLifePostgresVersion",
+			"The last minor version of Postgres %[1]v has been released."+
+				" PG %[1]v will no longer receive updates. We recommend upgrading."+
+				" See https://www.postgresql.org/support/versioning",
+			cluster.Spec.PostgresVersion)
+	}
 
 	if cluster.Spec.Standby != nil &&
 		cluster.Spec.Standby.Enabled &&
--- a/internal/postgres/versions.go
+++ b/internal/postgres/versions.go
@@ -1,0 +1,26 @@
+// Copyright 2021 - 2024 Crunchy Data Solutions, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package postgres
+
+import "time"
+
+// https://www.postgresql.org/support/versioning
+var finalReleaseDates = map[int]time.Time{
+	10: time.Date(2022, time.November+1, 10, 0, 0, 0, 0, time.UTC),
+	11: time.Date(2023, time.November+1, +9, 0, 0, 0, 0, time.UTC),
+	12: time.Date(2024, time.November+1, 14, 0, 0, 0, 0, time.UTC),
+	13: time.Date(2025, time.November+1, 13, 0, 0, 0, 0, time.UTC),
+	14: time.Date(2026, time.November+1, 12, 0, 0, 0, 0, time.UTC),
+	15: time.Date(2027, time.November+1, 11, 0, 0, 0, 0, time.UTC),
+	16: time.Date(2028, time.November+1, +9, 0, 0, 0, 0, time.UTC),
+	17: time.Date(2029, time.November+1, +8, 0, 0, 0, 0, time.UTC),
+}
+
+// ReleaseIsFinal returns whether or not t is definitively past the final
+// scheduled release of a Postgres version.
+func ReleaseIsFinal(majorVersion int, t time.Time) bool {
+	known, ok := finalReleaseDates[majorVersion]
+	return ok && t.After(known)
+}
--- a/internal/controller/postgrescluster/controller_test.go
+++ b/internal/controller/postgrescluster/controller_test.go
@@ -556,4 +556,66 @@ spec:
 			Expect(instance.Spec.Replicas).To(PointTo(BeEquivalentTo(1)))
 		})
 	})
+
+	Context("Postgres version EOL", func() {
+		var cluster *v1beta1.PostgresCluster
+
+		BeforeEach(func() {
+			cluster = create(`
+metadata:
+  name: old-postgres
+spec:
+  postgresVersion: 11
+  image: postgres
+  instances:
+  - name: instance1
+    dataVolumeClaimSpec:
+      accessModes:
+      - "ReadWriteMany"
+      resources:
+        requests:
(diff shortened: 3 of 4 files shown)


internal/postgres/config.go
```diff
@@ -425,3 +426,29 @@ chmod +x /tmp/pg_rewind_tde.sh
 
 	return append([]string{"bash", "-ceu", "--", script, "startup"}, args...)
 }
+
+// ReleaseIsFinal returns whether or not t is definitively past the final
+// scheduled release of a Postgres version.
+// See: https://www.postgresql.org/support/versioning/
+func ReleaseIsFinal(majorVersion int, t time.Time) bool {
```
cbandy, line 433: ⛏️ This doesn't strike me as configuration. How about putting it in a new file, `internal/postgres/versions.go`?

internal/postgres/config_test.go
```diff
@@ -506,3 +507,25 @@ EOF
...
+	testDate, err := time.Parse("2006-Jan-02", "2024-Nov-04")
+	assert.NilError(t, err)
+	assert.Check(t, ReleaseIsFinal(10, testDate))
+	assert.Check(t, ReleaseIsFinal(11, testDate))
+	assert.Check(t, !ReleaseIsFinal(12, testDate))
+	assert.Check(t, !ReleaseIsFinal(13, testDate))
+	assert.Check(t, !ReleaseIsFinal(14, testDate))
+	assert.Check(t, !ReleaseIsFinal(15, testDate))
+	assert.Check(t, !ReleaseIsFinal(16, testDate))
+	assert.Check(t, !ReleaseIsFinal(17, testDate))
```
cbandy, line 522: 🤔 We have to test every version to get coverage. Is it better 🤷🏻 for the implementation to be more "data" than "behavior?"

(untested)
```go
// https://www.postgresql.org/support/versioning
var finalReleaseDates = map[int]time.Time{
	10: time.Date(2022, time.November+1, 10, 0, 0, 0, 0, time.UTC),
	11: time.Date(2023, time.November+1, +9, 0, 0, 0, 0, time.UTC),
…
}

// ReleaseIsFinal returns whether or not t is definitively past the final
// scheduled release of a Postgres version.
func ReleaseIsFinal(majorVersion int, t time.Time) bool {
	known, ok := finalReleaseDates[majorVersion]
	return ok && t.After(known)
}
```

cbandy (approved):
The implementation looks good. I'll let others comment on the warning reason or text.

internal/postgres/config_test.go
dsessler7: Yeah I think I like this better

internal/postgres/versions.go
```diff
@@ -0,0 +1,26 @@
...
+// https://www.postgresql.org/support/versioning
+var finalReleaseDates = map[int]time.Time{
+	10: time.Date(2022, time.November+1, 10, 0, 0, 0, 0, time.UTC),
+	11: time.Date(2023, time.November+1, +9, 0, 0, 0, 0, time.UTC),
+	12: time.Date(2024, time.November+1, 14, 0, 0, 0, 0, time.UTC),
+	13: time.Date(2025, time.November+1, 13, 0, 0, 0, 0, time.UTC),
+	14: time.Date(2026, time.November+1, 12, 0, 0, 0, 0, time.UTC),
+	15: time.Date(2027, time.November+1, 11, 0, 0, 0, 0, time.UTC),
+	16: time.Date(2028, time.November+1, +9, 0, 0, 0, 0, time.UTC),
+	17: time.Date(2029, time.November+1, +8, 0, 0, 0, 0, time.UTC),
```
cbandy, line 18: 🤩 symmetry!

internal/controller/postgrescluster/controller.go
```diff
@@ -136,6 +136,14 @@ func (r *Reconciler) Reconcile(
 			return runtime.ErrorWithBackoff(err)
 		}
 	}
+	// Issue Warning Event if postgres version is EOL according to PostgreSQL:
+	// https://www.postgresql.org/support/versioning/
+	currentTime := time.Now()
+	if postgres.ReleaseIsFinal(cluster.Spec.PostgresVersion, currentTime) {
+		r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "EndOfLifePostgresVersion",
+			"Postgres Version %v is End-Of-Life. We recommend upgrading to the latest supported version.",
```
tsykoduk, line 144: Perhaps something like

"Postgres version %v is the last release of major version %m. We recommend upgrading... blah blah"

tsykoduk (changes requested):

internal/controller/postgrescluster/controller.go
dsessler7: Hmm... I think we would have to exec into the postgres container to get the minor version. The `cluster.Spec.PostgresVersion` is just the major version.

tsykoduk: Hmmm...

Is %v the major and minor or just the major?

"Postgres version %v is the last release of this version. We recommend upgrading... blah blah" may work.

dsessler7: Also, we won't necessary know that they are using the last release of a particular major version. We just know that the Major Version they are using is End-Of-Life.

tsykoduk: Ok good context thanks. 


"Postgres version %v is nearing the end of community support. We recommend upgrading. See https://www.postgresql.org/support/versioning/."

tsykoduk (approved):

internal/controller/postgrescluster/controller.go
```diff
@@ -136,6 +136,15 @@ func (r *Reconciler) Reconcile(
...
 		}
 	}
+	// Issue Warning Event if postgres version is EOL according to PostgreSQL:
+	// https://www.postgresql.org/support/versioning/
+	currentTime := time.Now()
+	if postgres.ReleaseIsFinal(cluster.Spec.PostgresVersion, currentTime) {
+		r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "EndOfLifePostgresVersion",
+			"The last minor version of Postgres %v has been released. PG %v will no longer receive "+
+				"updates. We recommend upgrading. See https://www.postgresql.org/support/versioning/.",
+			cluster.Spec.PostgresVersion, cluster.Spec.PostgresVersion)
```
cbandy, line 146: ⛏️ I recommend passing the version only once to `fmt`.
⛏️ I find these messages easier to read in the code when broken up between sentences.

⛏️ I worry that a program that turns URLs into links will slurp up the trailing dot. For some reason, I find trailing slashes in URLs unappealing, too.

```suggestion
			"The last minor version of Postgres %[1]v has been released." +
				" PG %[1]v will no longer receive updates. We recommend upgrading."+
				" See https://www.postgresql.org/support/versioning",
			cluster.Spec.PostgresVersion)
```

cbandy (approved):
One last pile of nitpicks about formatting messages. LGTM

PR #4029, opened 2024-11-05
Update feature gate logging to include default on

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

After the 5.7 release, when AutoUserSchemaCreate was graduated to default on/true, we discovered that our current system (and the underlying featuregate implementation) treats features explicitly turned on by the user differently than features turned on by default.

This PR updates that logging to make clear that the features are those specifically requested by the user (kept as a string to help debugging) and revises the 'ShowGates' function to include those set through defaults.

**Other Information**:

Issues: [PGO-1824]


internal/feature/features.go
```diff
@@ -122,11 +126,23 @@ func NewContext(ctx context.Context, gate Gate) context.Context {
 	return context.WithValue(ctx, contextKey{}, gate)
 }
 
+// ShowGates returns all the gates that are on by default
+// (if not turned off by the user) or were explicitly set
+// on by the user.
 func ShowGates(ctx context.Context) string {
```
cbandy, line 132: ❓ Can we name this `ShowFeatures` instead?

```suggestion
// ShowFeatures returns all the features enabled in the Gate contained in ctx.
func ShowFeatures(ctx context.Context) string {
```

internal/feature/features.go
```diff
@@ -52,6 +55,7 @@ type Feature = featuregate.Feature
 type Gate interface {
 	Enabled(Feature) bool
 	String() string
```
cbandy, line 57: 🔧 I misunderstood what upstream's `String()` does. Let's explain it here.

```suggestion
	// String returns the features enabled or disabled by Set and SetFromMap.
	String() string
```

<p><br>
🤔 Maybe we drop String from the interface and pivot fully to context functions?
</p>

```go
// ShowAssigned returns the features enabled or disabled by Set and SetFromMap
// in the Gate contained in ctx.
func ShowAssigned(ctx context.Context) string

// ShowEnabled returns the features enabled in the Gate contained in ctx.
func ShowEnabled(ctx context.Context) string
```

cbandy (approved):
The implementation is good, but I've soured on some of the naming. I misunderstood what "String" did, and now want better method names.

I can do the work of renaming a bunch of things if you're ready to be done with this PR.

internal/feature/features.go
```diff
@@ -122,11 +126,23 @@ func NewContext(ctx context.Context, gate Gate) context.Context {
...
+			// `gate.Enabled` first checks if the feature is enabled;
+			// then (if not explicitly set by the user),
+			// it checks if the feature is on/true by default
+			if gate.Enabled(feature) {
+				featuresEnabled = append(featuresEnabled, fmt.Sprintf("%s=true", feature))
+			}
+		}
 	}
-	return featuresEnabled
+	sort.Strings(featuresEnabled)
```
cbandy, line 146: ⛏️ 
```suggestion
	slices.Sort(featuresEnabled)
```

internal/feature/features_test.go
```diff
@@ -53,13 +53,21 @@ func TestContext(t *testing.T) {
 	t.Parallel()
 	gate := NewGate()
 	ctx := NewContext(context.Background(), gate)
```
cbandy, line 55: 🤔 This seems right to me in this moment. Time to step away!

```go
	assert.Equal(t, ShowAssigned(ctx), "")
	assert.Assert(t, ShowEnabled(ctx) != "") // This assumes some feature is enabled by default.

	assert.NilError(t, gate.Set("TablespaceVolumes=true"))
	assert.Assert(t, Enabled(ctx, TablespaceVolumes))
	assert.Equal(t, ShowAssigned(ctx), "TablespaceVolumes=true")
	assert.Assert(t,
		strings.Contains(ShowEnabled(ctx), "TablespaceVolumes=true"),
		"got: %v", ShowEnabled(ctx))

	assert.NilError(t, gate.SetFromMap(map[string]bool{TablespaceVolumes: false}))
	assert.Assert(t, !Enabled(ctx, TablespaceVolumes))
	assert.Equal(t, ShowAssigned(ctx), "TablespaceVolumes=false")
	assert.Assert(t,
		!strings.Contains(ShowEnabled(ctx), "TablespaceVolumes"),
		"got: %v", ShowEnabled(ctx))
```

internal/feature/features.go
```diff
@@ -122,11 +126,23 @@ func NewContext(ctx context.Context, gate Gate) context.Context {
...
 }
 
+// ShowGates returns all the gates that are on by default
+// (if not turned off by the user) or were explicitly set
+// on by the user.
 func ShowGates(ctx context.Context) string {
-	featuresEnabled := ""
+	featuresEnabled := []string{}
 	gate, ok := ctx.Value(contextKey{}).(Gate)
 	if ok {
```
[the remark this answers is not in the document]
benjaminjb: This is the only place we need GetAll, so makes sense to me

internal/feature/features.go
benjaminjb: The only place we use `gate.String()` is in main.go when we report what the user set -- but if we're just telling the user what they set, why not just print the env var?

internal/feature/features_test.go
benjaminjb: Yeah, as we graduate feature gates, we'd have to keep updating the test.

internal/feature/features_test.go
```diff
@@ -24,16 +26,17 @@ func TestDefaults(t *testing.T) {
 	assert.Assert(t, false == gate.Enabled(TablespaceVolumes))
 	assert.Assert(t, false == gate.Enabled(VolumeSnapshots))
 
-	assert.Equal(t, gate.String(), "")
+	assert.Equal(t, ShowAssigned(ctx), "")
```
cbandy, line 29: ⛏️ I think we can drop this assertion from here. ShowAssigned is tested later.

internal/feature/features_test.go
```diff
@@ -24,16 +26,17 @@ func TestDefaults(t *testing.T) {
...
 
 func TestStringFormat(t *testing.T) {
 	t.Parallel()
 	gate := NewGate()
+	ctx := NewContext(context.Background(), gate)
 
 	assert.NilError(t, gate.Set(""))
 	assert.NilError(t, gate.Set("TablespaceVolumes=true"))
-	assert.Equal(t, gate.String(), "TablespaceVolumes=true")
+	assert.Equal(t, ShowAssigned(ctx), "TablespaceVolumes=true")
```
cbandy, line 39: ⛏️ and drop this assertion here.

internal/upgradecheck/http_test.go
```diff
@@ -69,7 +69,7 @@ func TestCheckForUpgrades(t *testing.T) {
 		assert.Equal(t, data.RegistrationToken, "speakFriend")
 		assert.Equal(t, data.BridgeClustersTotal, 2)
 		assert.Equal(t, data.PGOClustersTotal, 2)
-		assert.Equal(t, data.FeatureGatesEnabled, "TablespaceVolumes=true")
+		assert.Equal(t, data.FeatureGatesEnabled, "AutoCreateUserSchema=true,TablespaceVolumes=true")
```
cbandy, line 72: 👍🏻 This isn't `Contains`, but I don't mind seeing the format. 🌱 We can change the test if it is ever a problem.

internal/feature/features.go
cbandy: Whatever the shape on the way in, we have a canonical representation in the logs.

cbandy (approved):
LGTM. Please squash.

cmd/postgres-operator/main.go
```diff
@@ -134,7 +134,12 @@ func main() {
 
 	features := feature.NewGate()
 	assertNoError(features.Set(os.Getenv("PGO_FEATURE_GATES")))
-	log.Info("feature gates enabled", "PGO_FEATURE_GATES", features.String())
+
+	ctx = feature.NewContext(ctx, features)
+	// This logs just the feature gates as set by the user
+	log.Info("feature gates set during deployment", "PGO_FEATURE_GATES", feature.ShowAssigned(ctx))
+	// This logs the feature gates that are enabled, including gates that are on by default
+	log.Info(fmt.Sprintf("feature gates enabled: %s", feature.ShowEnabled(ctx)))
```
cbandy, line 142: 🤔 Usually, a `fmt` indicates we should have key/value pairing. Maybe a single log message? 🤷🏻 

```suggestion
	log.Info("feature gates",
		// These are set by the user
		"PGO_FEATURE_GATES", feature.ShowAssigned(ctx),
		// These are enabled, including features that are on by default
		"enabled", feature.ShowEnabled(ctx))
```

🤔 🤔 are the letters `PGO_FEATURE_GATES` useful at this moment? `assigned` and `enabled` attributes? 🤔

cbandy (approved):

cmd/postgres-operator/main.go
benjaminjb: Single log message makes sense to me; I like `PGO_FEATURE_GATES` since that's the env var that the user sets, and I would almost drop it (since the user could just check the deployment) but I like the convenience of getting all the info at one point.

I do want to update the documentation we have [here](https://access.crunchydata.com/documentation/postgres-operator/latest/installation/feature-gates) to add "look at your logs to see what's enabled."

PR #4032, opened 2024-11-15
Refactor: Tidy some imports and tests

cbandy:
This adds two functions to `internal/controller/runtime` to reduce the imports in the pgBackRest controller and adds a linter to keep `github.com/pkg/errors` out of tests.

This also uses [`golang.org/x/tools/go/packages`](https://pkg.go.dev/golang.org/x/tools/go/packages) to load VolumeSnapshot CRDs from the (already downloaded) client Go module.

### DIFF
--- a/internal/controller/runtime/conversion.go
+++ b/internal/controller/runtime/conversion.go
@@ -1,0 +1,73 @@
+// Copyright 2021 - 2024 Crunchy Data Solutions, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package runtime
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type (
+	GR  = schema.GroupResource
+	GV  = schema.GroupVersion
+	GVK = schema.GroupVersionKind
+	GVR = schema.GroupVersionResource
+)
+
+// These functions call the [runtime.DefaultUnstructuredConverter] with some additional type safety.
+// An [unstructured.Unstructured] should always be paired with a [client.Object], and
+// an [unstructured.UnstructuredList] should always be paired with a [client.ObjectList].
+
+// FromUnstructuredList returns a copy of list by marshaling through JSON.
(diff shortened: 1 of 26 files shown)


internal/controller/runtime/conversion.go
```diff
@@ -0,0 +1,55 @@
...
+
+type (
+	GR  = schema.GroupResource
+	GV  = schema.GroupVersion
+	GVK = schema.GroupVersionKind
+	GVR = schema.GroupVersionResource
+)
+
+// ConvertList copies source into target by marshaling through JSON.
+func ConvertList(source *unstructured.UnstructuredList, target client.ObjectList) error {
```
benjaminjb, line 22: 🤔 Is `ConvertObject|List` too broad? Should this be `FromUnstructuredList`?

benjaminjb: Also, much more ⛏️ : I have a symmetrical urge to have `___List` and `___Object` for both to and from funcs.

benjaminjb (approved):

internal/controller/runtime/conversion.go
cbandy: 👍🏻 Agreed. How about:

```go
// FromUnstructuredList copies source into target by marshaling through JSON.
func FromUnstructuredList(source *u.UL, target c.OL) error
func FromUnstructuredObject(source *u.U, target c.O) error

func ToUnstructuredList(source c.OL) (*u.UL, error)
func ToUnstructuredObject(source c.O) (*u.U, error)
```

I am almost certain I can use generics as well, though the caller will need to pass a type parameter:

```go
// FromUnstructuredList returns a copy of source my marshaling through JSON.
func FromUnstructuredList[T …](source *u.UL) (*T, error) {}

list, err := runtime.FromUnstructuredList[appsv1.StatefulSetList](uList)
```

I'm leaning toward the latter because "copy into" seems harder to reason about. (What happens when the target is not empty?)

benjaminjb: Had a meeting to discuss this -- sounds good to me.

cbandy: I've amended the original commit.

- All these functions take one argument and return a value and an error.
- The "source" parameters are now named "object" or "list."
- The "target" parameters are gone. Each function builds up a "result" to return.

benjaminjb (approved):

PR #4036, opened 2024-11-19
Add a Runner that watches for Kubernetes APIs

cbandy:
This adds an `internal/kubernetes` package with types and functions for interrogating the Kubernetes API version and the APIs it has installed.

- `kubernetes.APIs` is an interface for some read-only methods of [k8s.io/…/sets.Set](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set).
- `kubernetes.APISet` implementes the `APIs` interface using a `map`.
- `kubernetes.DiscoveryRunner` implements the `APIs` interface by periodically calling the Kubernetes API.

`main()` configures a DiscoveryRunner and adds it to the Manager and its Context so that the rest of our code can use it with these new functions:

- `kubernetes.Has(Context, API) bool`
- `kubernetes.IsOpenShift(Context) bool`
- `kubernetes.VersionString(Context) string`

**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [x] Testing enhancement

- When VolumeSnapshots are enabled, we query for VolumeSnapshot APIs every PostgresCluster reconciliation.
  - When this query fails, reconciliation fails.
- We query for OpenShift APIs once at startup.
- We have "is OpenShift" logic in multiple controllers.

- We query for a handful of APIs more than once and less than every reconciliation.
  - When this query fails, reconciliation continues unabated with old data.
- Tests use Context and map to emulate different API configurations.
- We have a canonical definition of "is OpenShift" with guidance for its use.
- We have replaced one use of "is OpenShift" with a query for a specific API.

**Other Information**:

This is similar to #3973, but has grown to replace all our runtime use of the upstream discovery client.

### DIFF
--- a/internal/kubernetes/apis.go
+++ b/internal/kubernetes/apis.go
@@ -1,0 +1,60 @@
+// Copyright 2024 Crunchy Data Solutions, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package kubernetes
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// API is a combination of Group, Version, and Kind that can be used to check
+// what is available in the Kubernetes API. There are four ways to populate it:
+//  1. Group without Version nor Kind means any resource in that Group.
+//  2. Group with Version but no Kind means any resource in that GV.
+//  3. Group with Kind but no Version means that Kind in any Version of the Group.
+//  4. Group with Version and Kind means that exact GVK.
+type API = schema.GroupVersionKind
+
+type APIs interface {
+	Has(API) bool
+	HasAll(...API) bool
+	HasAny(...API) bool
+}
+
+// APISet implements [APIs] using empty struct for minimal memory consumption.
+type APISet = sets.Set[API]
+
+func NewAPISet(api ...API) APISet {
+	// Start with everything that's passed in; full GVKs are here.
+	s := sets.New(api...)
+
+	// Add the other combinations; Group, GV, and GK.
+	for i := range api {
+		s.Insert(
+			API{Group: api[i].Group},
+			API{Group: api[i].Group, Version: api[i].Version},
+			API{Group: api[i].Group, Kind: api[i].Kind},
+		)
+	}
+
+	return s
+}
+
+type apiContextKey struct{}
+
+// Has returns true when api was previously stored by [NewAPIContext].
+func Has(ctx context.Context, api API) bool {
+	if i, ok := ctx.Value(apiContextKey{}).(interface{ Has(API) bool }); ok {
+		return i.Has(api)
+	}
+	return false
+}
+
+// NewAPIContext returns a copy of ctx containing apis. Interrogate it using [Has].
+func NewAPIContext(ctx context.Context, apis APIs) context.Context {
+	return context.WithValue(ctx, apiContextKey{}, apis)
+}
--- a/internal/kubernetes/discovery.go
+++ b/internal/kubernetes/discovery.go
@@ -1,0 +1,208 @@
+// Copyright 2024 Crunchy Data Solutions, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package kubernetes
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/version"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+
+	"github.com/crunchydata/postgres-operator/internal/logging"
+)
+
+type Version = version.Info
+
+// DiscoveryRunner implements [APIs] by reading from a Kubernetes API client.
+// Its methods are safe to call concurrently.
+type DiscoveryRunner struct {
+	// NOTE(tracing): The methods of [discovery.DiscoveryClient] do not take
+	// a Context so their API calls won't have a parent span.
+	Client interface {
+		ServerGroups() (*metav1.APIGroupList, error)
+		ServerResourcesForGroupVersion(string) (*metav1.APIResourceList, error)
+		ServerVersion() (*version.Info, error)
+	}
+
+	refresh time.Duration
+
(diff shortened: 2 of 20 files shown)


internal/kubernetes/apis.go
```diff
@@ -0,0 +1,58 @@
...
+type APISet = sets.Set[API]
+
+func NewAPISet(api ...API) APISet {
+	s := sets.New[API](api...)
+
+	for i := range api {
+		s.Insert(
+			API{Group: api[i].Group},
+			API{Group: api[i].Group, Version: api[i].Version},
+			API{Group: api[i].Group, Kind: api[i].Kind},
```
benjaminjb, line 38: We don't want to insert the full GVK?

cbandy: That's happening on L32. The full extent of what they passed us is added there, and these add the... lesser combos. Worth a code comment, but I'm not sure how to describe it atm.

cbandy: 😆 That wasn't so bad.

internal/postgres/reconcile.go
```diff
@@ -293,9 +293,9 @@ func PodSecurityContext(cluster *v1beta1.PostgresCluster) *corev1.PodSecurityCon
 	// - https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids
 	// - https://docs.k8s.io/tasks/configure-pod-container/security-context/
 	// - https://docs.openshift.com/container-platform/4.8/authentication/managing-security-context-constraints.html
-	if cluster.Spec.OpenShift == nil || !*cluster.Spec.OpenShift {
-		podSecurityContext.FSGroup = initialize.Int64(26)
+	if !initialize.FromPointer(cluster.Spec.OpenShift) {
```
benjaminjb, line 296: ah, OK, had to read the func / re-read the word "initialize" to get it: we give the value of the pointer or the null value of the type if the pointer is null.
 
So: Spec.OpenShift: true => true ; and false/nil => false.

internal/kubernetes/discovery_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+func TestVersionString(t *testing.T) {
+	ctx := context.Background()
+	assert.Equal(t, "", VersionString(ctx))
+
+	runner := new(DiscoveryRunner)
+	runner.have.Version = version.Info{
+		Major: "1", Minor: "2", GitVersion: "asdf",
+	}
+	assert.Equal(t, "asdf", VersionString(NewAPIContext(ctx, runner)))
```
benjaminjb, line 76: TIL `String()` only returns the GitVersion, cf. https://github.com/kubernetes/apimachinery/blob/a8f449e276fe566efddb149992049c78f0088492/pkg/version/types.go#L35-L37

internal/kubernetes/discovery.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// NewDiscoveryRunner creates a [DiscoveryRunner] that periodically reads from
+// the Kubernetes at config.
+func NewDiscoveryRunner(config *rest.Config) (*DiscoveryRunner, error) {
+	dc, err := discovery.NewDiscoveryClientForConfig(config)
+
+	runner := &DiscoveryRunner{
+		Client:  dc,
+		refresh: 10 * time.Minute,
```
benjaminjb, line 50: Have you done a manual test (more for fun, really) where you log what the runner has, add some CRD, and see it log the new APISet after the refresh time?

internal/kubernetes/discovery.go
```diff
@@ -0,0 +1,221 @@
...
+type DiscoveryRunner struct {
+	Client interface {
+		ServerGroups() (*metav1.APIGroupList, error)
+		ServerResourcesForGroupVersion(string) (*metav1.APIResourceList, error)
+		ServerVersion() (*version.Info, error)
+	}
+
+	refresh time.Duration
+
+	want []API
```
benjaminjb, line 35: 🤔 I can't come up with a really better term for this, but `want` had me thinking we want these object types to be present, when really, these are the GV/GK/GVK that we want details on if present.

internal/kubernetes/apis.go
```diff
@@ -0,0 +1,60 @@
...
+type APIs interface {
+	Has(API) bool
+	HasAll(...API) bool
+	HasAny(...API) bool
+}
+
+// APISet implements [APIs] using empty struct for minimal memory consumption.
+type APISet = sets.Set[API]
+
+func NewAPISet(api ...API) APISet {
```
benjaminjb, line 31: Do we ever use this outside of tests?

internal/postgres/reconcile.go
cbandy: Yeah, this call doesn't read well. A function that returns zero when a pointer is nil _sounds_ intuitive, but I dunno in practice. Other functions in this package act on the pointers their passed, so that combo is confusing, too.

K8s has a not-versioned pile of `util` with a [ptr](https://pkg.go.dev/k8s.io/utils/ptr) `Deref(p *T, default T) T`. That doesn't read much better to me.

internal/kubernetes/discovery.go
cbandy: I'll see how `relevant` goes.

internal/kubernetes/discovery.go
cbandy: 👍🏻 I hadn't, and now I have. I added the VolumeSnapshots CRDs; they appeared. I removed the CRDs; they disappeared.

⛏️ This should log when it's successful. My first attempt to dump the set using [UnsortedList](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set.UnsortedList) looked terrible.
🌱 I'll do something small and leave a TODO.

internal/kubernetes/apis.go
cbandy: Nope. I considered using it in the runner, but accumulating a slice in those nested loops to pass to this loop didn't feel right.

dsessler7 (approved):

internal/kubernetes/apis.go
benjaminjb: Worth moving to testing?

cbandy: I think the constructor for this type should live here.

internal/kubernetes/discovery.go
cbandy: I've added a debug log message.

PR #4040, opened 2024-11-27
Refactor: Pass client.Object types by reference

cbandy:
I'm confident about the first commit. The others switch from List types, which seem like a Kubernetes API concern, to slices of "plain old objects." That seems better positioned for generics and [`slices`](https://pkg.go.dev/slices), but I didn't look for those opportunities here. 🤔 Perhaps those commits should wait until then.


benjaminjb (approved):
LGTM
Per your comment on the PR (should commits 2-4 wait): I think getting them in early doesn't increase our complexity, so seems fine to merge in now, yeah?

internal/controller/runtime/reconcile.go
```diff
@@ -7,9 +7,21 @@ package runtime
 import (
 	"time"
 
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+// Requests converts objects to a slice of [reconcile.Request].
+func Requests[T client.Object](objects ...T) []reconcile.Request {
```
cbandy, line 15: This has convinced me that "slice of plain objects" pays dividends, even now.

benjaminjb (approved):
Less code ++

PR #4041, opened 2024-12-02
Update OpenTelemetry and use an internal package

cbandy:
This updates OpenTelemetry modules and imports them into an `internal/tracing` package.

- `main()` now uses the [autoexport](https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport) and [autoprop](https://pkg.go.dev/go.opentelemetry.io/contrib/propagators/autoprop) packages to configure standard exporters according to the current [Specification](https://opentelemetry.io/docs/specs/otel/configuration).
- When a trace exporter is enabled, the `internal/logging` package [correlates log messages](https://opentelemetry.io/docs/specs/otel/compatibility/logging_trace_context) with spans.
- The [spancheck linter](https://golangci-lint.run/usage/linters/#spancheck) is configured to remind us to call `Span.End` and `Span.RecordError`.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Other

**Other Information**:

Issue: PGO-1954


internal/controller/standalone_pgadmin/controller.go
```diff
@@ -145,7 +148,7 @@ func (r *PGAdminReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		log.V(1).Info("Reconciled pgAdmin")
 	}
 
-	return ctrl.Result{}, err
+	return ctrl.Result{}, tracing.Escape(span, err)
```
tjmoore4, line 151: :thinking: Why is this using `Escape` instead of `Check`? Is it because we're leaving the current scope?

internal/controller/pgupgrade/pgupgrade_controller.go
```diff
@@ -93,7 +95,10 @@ func (r *PGUpgradeReconciler) findUpgradesForPostgresCluster(
 // Reconcile does the work to move the current state of the world toward the
 // desired state described in a [v1beta1.PGUpgrade] identified by req.
 func (r *PGUpgradeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
-	log := ctrl.LoggerFrom(ctx)
+	ctx, span := tracing.Start(ctx, "reconcile-pgupgrade")
+	log := logging.FromContext(ctx)
+	defer span.End()
+	defer func(s tracing.Span) { _ = tracing.Escape(s, err) }(span)
```
tjmoore4, line 101: :thinking: This is another place where I am unsure why we're using `Escape` instead of `Check` (there are others, but I think it's generally the same question, so I won't mention any more).

tjmoore4 (approved):
Looks good to me, a couple of comments but no blockers.

internal/controller/standalone_pgadmin/controller.go
cbandy: > Is it because we're leaving the current scope?

Precisely.

You can add error information to a span separately from indicating the span failed. OTel has [this page](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-spans) describing how to represent error information. Footnote **[3] `exception.escaped`** explains what "escape" means to them.

For us, yeah, escaping is likely that we're returning an error while ending a span.

`Check` adds error information without the "escaped" indicator.

Both functions do the thing we always have to do, `err != nil`, so we can use `Escape` kinda blindly. The code in this discussion/thread is doing `defer span.End()`, so all the early returns do an "add 'error escaped' when err != nil."

tjmoore4: :+1: Makes sense, thanks for the explanation.

PR #4045, opened 2024-12-06
bump controller-runtime to v0.19.3

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

**Other Information**:


go.mod
```diff
@@ -8,27 +8,27 @@ require (
...
-	k8s.io/api v0.30.2
-	k8s.io/apimachinery v0.30.2
-	k8s.io/client-go v0.30.2
-	k8s.io/component-base v0.30.2
-	sigs.k8s.io/controller-runtime v0.18.4
+	k8s.io/api v0.31.0
+	k8s.io/apimachinery v0.31.0
+	k8s.io/client-go v0.31.0
+	k8s.io/component-base v0.31.0
+	sigs.k8s.io/controller-runtime v0.19.3
```
cbandy, line 31: 👍🏻

cbandy (approved):
There are a couple `TODO(controller-runtime)` that I think can be addressed now.

I believe we can stop turning a Runner's context.Canceled into nil.

```diff
-			if errors.Is(ctx.Err(), context.Canceled) {
-				return nil
-			}
 			return ctx.Err()
```

benjaminjb (approved):
Whoa, never saw that lint check before. So the problem is that there are 6 types of `PersistentVolumeClaimConditionType` and our switch only covers 4?

I lean toward adding a `nolint:exhaustive` to that to ignore the lint check there. (I'm not even sure it's worth making a ticket to cover the two missing cases, but we could for now and revisit it.)

cbandy:
> I lean toward adding a `nolint:exhaustive` to that to ignore the lint check there. (I'm not even sure it's worth making a ticket to cover the two missing cases, but we could for now and revisit it.)

👍🏻 The issue raised by the linter shouldn't block this PR. I'm fine with ignoring or `nolint` or anything that unblocks here.

🔧 I would like an issue for the missing cases. They sound exactly like the resizing that code is supposed to understand.

cbandy (approved):

PR #4047, opened 2024-12-09
Configure Patroni Logs to be stored in a file

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Patroni is configured to send logs to stdout

This change allows for the configuration of the Postgres instance
Patroni logs to go to a file on the 'pgdata' volume instead of to
stdout. This file is located at '/pgdata/patroni/log/patroni.log'.

Both the log size limit and log level are configurable; only the
size limit setting is required. If not set, the default behavior
of sending all logs to stdout is maintained.

Changes to this configuration require a reload to take effect.

- https://patroni.readthedocs.io/en/latest/yaml_configuration.html#log

  

**Other Information**:
Issue: PGO-1701


tjmoore4:
:memo: We have a story in our backlog to address the linter issue separately, so I'll leave that only for now.

internal/patroni/config.go
```diff
@@ -50,6 +50,17 @@ func clusterYAML(
...
 
+		// Configure the Patroni log path directory
+		// - https://patroni.readthedocs.io/en/latest/yaml_configuration.html#log
+		//
+		// Setting file_size to 0 means that there will be no rollover.
+		// - https://github.com/patroni/patroni/blob/v3.3.4/patroni/log.py#L431-L432
+		// - https://docs.python.org/3.11/library/logging.handlers.html#rotatingfilehandler
+		"log": map[string]any{
+			"dir":       naming.PatroniPGDataLogPath,
+			"file_size": 0,
```
tjmoore4, line 61: :thought_balloon: I'm conflicted about this. On one hand, this makes referencing the correct file simpler and removes the possibility of losing logs over time based on file size limits. However, this means the file will grow larger and larger over time if not managed. I think we should discuss this further before merging.

tjmoore4: Per discussion with team, this is now configurable by the user.

internal/patroni/config.go
```diff
@@ -152,6 +152,20 @@ func clusterYAML(
...
 
+	// if a Patroni log file size is configured, configure volume file storage
+	if patroniLogFileSize != 0 {
+
+		root["log"] = map[string]any{
+			// Configure the Patroni log settings
+			// - https://patroni.readthedocs.io/en/latest/yaml_configuration.html#log
+			"dir":       naming.PatroniPGDataLogPath,
+			"file_num":  1,
+			"file_size": patroniLogFileSize,
```
cbandy, line 163: ❓ Does `file_num` go into [RotatingFileHandler](https://docs.python.org/3/library/logging.handlers.html#logging.handlers.RotatingFileHandler).backupCount? If so, my understanding is that there will be two files of `file_size` bytes, `patroni.log` and `patroni.log.1`.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/patroni_types.go
```diff
@@ -48,6 +52,21 @@ type PatroniSpec struct {
 	// - https://patroni.readthedocs.io/en/latest/kubernetes.html
 }
 
+type PatroniLogConfig struct {
+
+	// Patroni Log file size
+	// https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity
+	// +required
+	FileSize *string `json:"fileSize"`
```
cbandy, line 60: I want the field we expose to restrict the _total_ space occupied by all log files combined. I think I prefer `limit` or `storageLimit` for this. WDYT?

----

I went looking for some examples in the Kubernetes API:

CSIStorageCapacity: `maximumVolumeSize`
PodDisruptionBudget: `maxUnavailable`, `minAvailable` (Pods)
LimitRange: `max: { cpu, memory }`, `maxLimitRequestRatio: { cpu, memory }`
QueuingConfiguration: `queueLengthLimit`
TopologySpreadConstraint: `maxSkew`, `minDomains`
HorizontalPodAutoscaler: `maxReplicas`, `minReplicas`

Typing things out:

```yaml
history: 10Mi
maximumHistory: 10Mi
limit: 10Mi
storageLimit: 10Mi

# Knowing the Patroni YAML, this is ambiguous to me.
#  - Is this a single file, or is this all files combined?
maximumSize: 10Mi
maximumTotalSize: 10Mi
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/patroni_types.go
```diff
@@ -23,6 +23,10 @@ type PatroniSpec struct {
 	// +kubebuilder:validation:Minimum=3
 	LeaderLeaseDurationSeconds *int32 `json:"leaderLeaseDurationSeconds,omitempty"`
 
+	// Patroni log configuration settings.
+	// +optional
+	Log *PatroniLogConfig `json:"log,omitempty"`
```
cbandy, line 28: 🤔 This matches the Patroni naming, but I want to deviate from Patroni exposes. What do you think of naming this `logging`?

```yaml
spec:
  patroni:
    logging:
      level: INFO
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/patroni_types.go
```diff
@@ -48,6 +52,21 @@ type PatroniSpec struct {
...
+	// https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity
+	// +required
+	FileSize *string `json:"fileSize"`
+
+	// Patroni log level
+	// https://docs.python.org/3.6/library/logging.html#levels
+	// +kubebuilder:validation:Enum={CRITICAL,ERROR,WARNING,INFO,DEBUG,NOTSET}
+	// +kubebuilder:default:=INFO
+	// +optional
+	LogLevel *string `json:"logLevel,omitempty"`
```
cbandy, line 67: ⛏️ `log` prefix here seems redundant, we're in the `log` section already.

internal/controller/postgrescluster/cluster.go
```diff
@@ -53,6 +54,32 @@ func (r *Reconciler) reconcileClusterConfigMap(
...
+
+				fileSize, err := resource.ParseQuantity(*cluster.Spec.Patroni.Log.FileSize)
+				sizeInBytes := fileSize.Value()
+
+				if err != nil {
+					r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "InvalidPatroniLogFileSize",
+						"Error when parsing spec.patroni.log.fileSize. File size will default to 25M.")
+
+					sizeInBytes = 25000000
+				}
```
cbandy, line 75: ⛏️ I'd prefer that we prevent this with validation somehow.

❓ What happens to Patroni if someone picks something tiny, like `.1Ki` ?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/patroni_types.go
```diff
@@ -48,6 +52,21 @@ type PatroniSpec struct {
...
+type PatroniLogConfig struct {
+
+	// Patroni Log file size
+	// https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity
+	// +required
+	FileSize *string `json:"fileSize"`
+
+	// Patroni log level
+	// https://docs.python.org/3.6/library/logging.html#levels
+	// +kubebuilder:validation:Enum={CRITICAL,ERROR,WARNING,INFO,DEBUG,NOTSET}
```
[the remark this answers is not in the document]
cbandy: `NOTSET` is a terrible name, but it's official... WDYT of calling it `ALL` here and converting that to `NOTSET` in our code?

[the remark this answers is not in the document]
tjmoore4: I was conflicted about this as well but landed on just including all possible options. That said, `NOTSET` is somewhat confusing. Per the [docs I've read](https://docs.python.org/3/library/logging.html#logging-levels), this is the expected behavior:
> When set on a logger, indicates that ancestor loggers are to be consulted to determine the effective level. If that still resolves to NOTSET, then all events are logged. When set on a handler, all events are handled.

If I understand correctly, that means that since we don't have an ancestor loggers, it will always be equivalent to `ALL` logs. If that's the case, I'm good with changing this to `ALL`, I would just want to make sure I'm not missing any other possibilities.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/patroni_types.go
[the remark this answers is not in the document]
tjmoore4: For the first question, ultimately, the value will be converted to the number of bytes, so I think that's probably ok. Maybe just add a note in the description that integers = number of bytes?

For the second question, yes more or less. That's probably the better way to go since the current implementation's defaulting behavior would be avoided altogether this way.

internal/controller/postgrescluster/cluster.go
tjmoore4: Per the comment above, it looks like exposing `resource.Quantity` is the better option. I think reworking the current function to force a log file size minimum may make sense. Any thoughts on what size? The Patroni minimum of 25M doesn't feel terrible.

internal/patroni/config.go
tjmoore4: Good catch, that looks right, which is not what I intended. I'm thinking that means I'll want to set that value to `0` instead. I'll test to make sure that works as expected.

tjmoore4: It doesn't.... Apparently to get the file rotation behavior, `file_num` needs to be >= 1
https://github.com/python/cpython/blob/3.11/Lib/logging/handlers.py#L134
If set to zero, the file just grows continually. An easy solution would be to divide the size given by two, leave the `file_size` setting at 1 and update the name (as you mention below) so that it's clear what we're doing. Thoughts on that approach?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/patroni_types.go
tjmoore4: `storageLimit` makes sense to me, especially in light of the [discussion above](https://github.com/CrunchyData/postgres-operator/pull/4047#discussion_r1888951453). I'll make that change.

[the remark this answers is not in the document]
tjmoore4: Just to follow up on this, with the latest changes, the minimum value will equate to `25000000`, so hopefully that will lead to fewer configuration issues.

internal/patroni/config.go
cbandy: div 2 sounds good

internal/controller/postgrescluster/cluster.go
cbandy: Yeah, 25M works for me. We can go lower in the future should the need arise.

internal/controller/postgrescluster/cluster.go
```diff
@@ -53,6 +53,31 @@ func (r *Reconciler) reconcileClusterConfigMap(
...
+
+	if cluster.Spec.Patroni != nil {
+		if cluster.Spec.Patroni.Logging != nil {
+			if cluster.Spec.Patroni.Logging.StorageLimit != nil {
+
+				sizeInBytes := cluster.Spec.Patroni.Logging.StorageLimit.Value()
+
+				if sizeInBytes < 25000000 {
+					r.Recorder.Eventf(cluster, corev1.EventTypeNormal, "PatroniLogStorageLimitTooSmall",
+						"Configured Patroni log storage limit is too small. File size will default to 25M.")
```
cbandy, line 70: ⛏️ IIUC, we can use CEL validation for this [eventually](https://kubernetes.io/docs/reference/using-api/cel/) (quantity library). Please add a `TODO(validation)` or so.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/patroni_types.go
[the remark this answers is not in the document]
cbandy: 👍🏻 That's my understanding as well. This Patroni setting goes into the root logger, so it sounds like "all."

But! This doesn't make sense (to me) in the hierarchy. Does a parent logger set to `DEBUG` have anything to do with debug messages from a child logger set to `INFO`? 🤔 cc: @benjaminjb

I've lost interest in the usability of this field! I think what you have now behaves exactly as Patroni intends. I don't have a better UX for logging than upstream Python. 🤷🏻 👍🏻 🌴

cbandy (approved):

PR #4057, opened 2024-12-19
Consolidate arguments to a Patroni function

cbandy:
When called at runtime, the second argument is always derived from the first. This simplifies those call sites and clarifies the behavior in tests.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

### DIFF
--- a/internal/patroni/config_test.go
+++ b/internal/patroni/config_test.go
@@ -229,8 +229,7 @@ func TestDynamicConfiguration(t *testing.T) {
 
 	for _, tt := range []struct {
 		name     string
-		cluster  *v1beta1.PostgresCluster
-		input    map[string]any
+		spec     v1beta1.PostgresClusterSpec
 		hbas     postgres.HBAs
 		params   postgres.Parameters
 		expected map[string]any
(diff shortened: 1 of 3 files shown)


internal/patroni/config_test.go
```diff
@@ -797,8 +856,10 @@ func TestDynamicConfiguration(t *testing.T) {
 				"loop_wait": int32(10),
 				"ttl":       int32(30),
 				"postgresql": map[string]any{
-					"bin_name":      map[string]any{"pg_rewind": string("/tmp/pg_rewind_tde.sh")},
-					"parameters":    map[string]any{},
+					"bin_name": map[string]any{"pg_rewind": string("/tmp/pg_rewind_tde.sh")},
+					"parameters": map[string]any{
+						"encryption_key_command": "echo test",
+					},
```
cbandy, line 862: 📝 Here's where the expection changed.

internal/patroni/config_test.go
```diff
@@ -250,13 +249,17 @@ func TestDynamicConfiguration(t *testing.T) {
 		},
 		{
 			name: "top-level passes through",
-			input: map[string]any{
-				"retry_timeout": 5,
+			spec: v1beta1.PostgresClusterSpec{
+				Patroni: &v1beta1.PatroniSpec{
+					DynamicConfiguration: map[string]any{
+						"retry_timeout": int64(5),
```
tjmoore4, line 255: :thinking: Why do we need `int64` here and below?

tjmoore4 (approved):

internal/patroni/config_test.go
cbandy: The `DeepCopy()` function is only happy with types produced by the `json` package plus `int64`. When parsed, these numbers are `float64`. 🤔 Maybe I should use that type. Another approach could be YAML instead of structs here.

```go
	var object map[string]any
	fmt.Println(yaml.Unmarshal([]byte(`{
		asdf: 1,
	}`), &object))

	fmt.Printf("%#v\n", object)
	fmt.Printf("%T\n", object["asdf"])
```

```
<nil>
map[string]interface {}{"asdf":1}
float64
```

cbandy: I pushed another commit that unmarshals yaml. WDYT?

tjmoore4: Maybe `float64` would make sense, but I'm also fine with leaving this as is. I mainly wanted to understand the reason for the type change and now I do :+1:

tjmoore4: The new changes work for me, perhaps it's even a little more decipherable?

PR #4084, opened 2025-01-30
pgbouncer user password change should not require verifier

philrhurst:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Currently, it is not possible to clear the verifier for the _crunchypgbouncer Secret and only change the password.

The new behavior mirrors what is possible for the ccp_monitoring user. The verifier can be cleared and the password change. CPK will provide the SCRAM verifier.

**Other Information**:


andrewlecuyer:
@philrhurst any thoughts on an integration test to verify this change?

For instance, I do see some existing tests out there to verify the proper contents of the the PgBouncer Secret:

https://github.com/CrunchyData/postgres-operator/blob/740400d7f70b42385528a23cc55790004e46114b/internal/pgbouncer/reconcile_test.go#L52

Maybe this test can be expanded (or a similar test created, etc.) to cover the functionality you've implemented?

internal/pgbouncer/reconcile.go
```diff
@@ -52,14 +54,28 @@ func Secret(ctx context.Context,
...
 	// NOTE(cbandy): We don't have a function to compare a plaintext password
 	// to a SCRAM verifier.
 	password := string(inSecret.Data[passwordSecretKey])
 	verifier := string(inSecret.Data[verifierSecretKey])
 
-	if err == nil && (len(password) == 0 || len(verifier) == 0) {
-		password, verifier, err = generatePassword()
+	// If the password is empty, generate a new one.
+	// Ignore the verifier for now.
+	if err == nil && len(password) == 0 {
```
andrewlecuyer, line 71: It looks like this `err == nil` check is unnecessary?  I.e., there is no chance `err` will be anything but `nil` at this point, correct?

internal/pgbouncer/reconcile_test.go
```diff
@@ -90,6 +90,34 @@ func TestSecret(t *testing.T) {
...
+	// Simulate the generation of SCRAM verifier
+	existing.Data = map[string][]byte{
+		"pgbouncer-verifier": []byte("SCRAM-SHA-256$4096:randomsalt:storedkey:serverkey"),
+	}
+
+	assert.NilError(t, Secret(ctx, cluster, root, existing, service, intent))
+
+	// Verify that the SCRAM verifier is correctly set in the intent secret
+	assert.Assert(t, len(intent.Data["pgbouncer-verifier"]) != 0)
+	assert.Equal(t, string(intent.Data["pgbouncer-verifier"]), "SCRAM-SHA-256$4096:randomsalt:storedkey:serverkey")
```
andrewlecuyer, line 118: I'm not clear on the behavior being verified here.  We expect the user to either provide both the verifier and the password, or just a password, correct?  With the latter being the main intent of this change, i.e. allowing the user to simply provide a password without having to provide the verifier as well.

Therefore, should the test reflect that, e.g. setting a password and verifying that the verifier is generated?

```go
intent = new(corev1.Secret)
existing.Data = map[string][]byte{
    "pgbouncer-password": []byte("password"),
}
assert.NilError(t, Secret(ctx, cluster, root, existing, service, intent))
assert.Assert(t, len(intent.Data["pgbouncer-verifier"]) != 0)
```

And then also verifying both are simply left as is if the user provides both?

```go
intent = new(corev1.Secret)
existing.Data = map[string][]byte{
    "pgbouncer-verifier": []byte("SCRAM-SHA-256$4096:randomsalt:storedkey:serverkey"),
    "pgbouncer-password": []byte("password"),
}
assert.NilError(t, Secret(ctx, cluster, root, existing, service, intent))
assert.Equal(t, string(intent.Data["pgbouncer-verifier"]), "SCRAM-SHA-256$4096:randomsalt:storedkey:serverkey")
assert.Equal(t, string(intent.Data["pgbouncer-password"]), "password")
```

internal/pgbouncer/reconcile.go
philrhurst: yep, good point. I removed the extra check.

internal/pgbouncer/reconcile_test.go
philrhurst: you're right - this makes more sense now.

internal/pgbouncer/reconcile.go
```diff
@@ -52,14 +54,28 @@ func Secret(ctx context.Context,
...
+		password, err = util.GenerateASCIIPassword(32)
```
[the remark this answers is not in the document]
philrhurst: Hmm...this is an interesting edge case, and I see where I missed a crucial step.

Comparing this section to two other sections where we handle passwords and verifiers, I see a principle where we don't  _trust_ that a user-provided verifier is correct. When a password is unset, the verifier is always re-generated. That's the design I was trying to follow (but I missed the step of clearing out the verifier). I'll refactor to make it clear.

- https://github.com/CrunchyData/postgres-operator/blob/main/internal/controller/postgrescluster/postgres.go#L69-L98
- https://github.com/CrunchyData/postgres-operator/blob/main/internal/controller/postgrescluster/pgmonitor.go#L192-L213

internal/pgbouncer/reconcile_test.go
andrewlecuyer: Looks good now, and definitely covers the behavior covered by this change (i.e., allowing the user to either provide both the verifier and the password, or just a password).

🌱 I could see adding one more test to verify the behavior of the user providing no password, and a verifier only (e.g. to confirm that both the password and the verifier will be generated).  This would also be the same as a user simply deleting the password and not the verifier, effectively triggering a password rotation.

philrhurst: I added a test for this. Thanks for the feedback.

andrewlecuyer (approved):

internal/pgbouncer/reconcile_test.go
andrewlecuyer: Awesome! Thanks @philrhurst.

PR #4089, opened 2025-02-09
Otel api and export logs

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Our observability solution doesn't currently have an API and does not allow for users to specify exporters for the OTel collector. Our patroni and pgadmin logs are missing certain desirable attributes.

There is now an instrumentation API for configuring our OTel observability implementation. Users can use this API to configure exporters and add them to the logs pipelines. The processed patroni logs now have expected kubernetes attributes. The processed pgAdmin logs now have certain expected attributes they were missing before.

**Other Information**:


internal/collector/pgadmin.go
```diff
@@ -56,10 +60,32 @@ func EnablePgAdminLogging(ctx context.Context, configmap *corev1.ConfigMap) erro
...
+					// the log record body with the message field.
+					//
+					// https://github.com/open-telemetry/semantic-conventions/blob/v1.29.0/docs/general/logs.md
+					`set(attributes["log.record.original"], body)`,
 					`set(cache, ParseJSON(body))`,
 					`merge_maps(attributes, ExtractPatterns(cache["message"], "(?P<webrequest>[A-Z]{3}.*?[\\d]{3})"), "insert")`,
+					`set(body, cache["message"])`,
+
+					// Set process.title to the "name" from each log record.
+					// TODO: confirm that the "name" entry in the pgAdmin logs is in fact referring
```
tony-landreth, line 76: Looks like this always going to say either "pgadmin" or "gunicorn.access". The [docs](https://docs.python.org/3/library/logging.html#logrecord-attributes) define name as "Name of the logger used to log the call."

dsessler7: Okay, in that case this seems to me like it should be mapped to `instrumentation_scope.name`... @cbandy do you concur?

internal/collector/patroni.go
```diff
@@ -86,14 +100,27 @@ func EnablePatroniLogging(ctx context.Context,
 			}},
 		}
 
+		// If there are exporters to be added to the logs pipelines defined in
+		// the spec, add them to the pipeline. Otherwise, add the DebugExporter.
```
cbandy, line 104: 👍🏻

internal/collector/pgadmin.go
cbandy: Agreed. If the component has an idea of named loggers, then that name goes into the scope. [This](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-scope) and the term after it talk about it.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/instrumentation_types.go
```diff
@@ -0,0 +1,55 @@
...
+	Files []corev1.VolumeProjection `json:"files,omitempty"`
+}
+
+// InstrumentationLogsSpec defines the configuration for collecting logs via
+// OpenTelemetry.
+type InstrumentationLogsSpec struct {
+	// Exporters allows users to specify which exporters they want to use in
+	// the logs pipeline.
+	// +optional
+	Exporters []string `json:"exporters,omitempty"`
```
cbandy, line 54: 🌱 It may be possible to validate that values in this list exist as keys in `config.exporters`.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -95,6 +95,11 @@ type PostgresClusterSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,order=2
 	InstanceSets []PostgresInstanceSetSpec `json:"instances"`
 
+	// Configuration for the OpenTelemetry collector container used to collect
+	// logs and metrics.
+	// +optional
+	Instrumentation *InstrumentationSpec `json:"instrumentation,omitempty"`
```
cbandy, line 101: 👍🏻 pointer, optional

cbandy (approved):

internal/collector/pgadmin.go
dsessler7: Thank you for the link! I was having a hard time finding documentation on Instrumentation Scope.

PR #4090, opened 2025-02-10
Ensure OTel directories are writable

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Bug fix

The pgAdmin log offsets are stored in `/tmp` to avoid conflicting permissions.

The pgAdmin log directories are created and updated to be writable during startup. Log offsets are now stored alongside logs.


internal/shell/paths.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2024 - 2025 Crunchy Data Solutions, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// We want the [filepath] to package to behave correctly for Linux containers.
```
benjaminjb, line 5:
```suggestion
// We want the [filepath] package to behave correctly for Linux containers.
```

internal/collector/pgadmin.go
```diff
@@ -21,24 +21,20 @@ func EnablePgAdminLogging(ctx context.Context, spec *v1beta1.InstrumentationSpec
...
-		"create_directory": true,
-		"fsync":            true,
-	}
-	otelConfig.Extensions["file_storage/gunicorn"] = map[string]any{
-		"directory":        "/var/log/gunicorn" + "/receiver",
-		"create_directory": true,
+
+	otelConfig.Extensions["file_storage/pgadmin_data_logs"] = map[string]any{
+		"directory":        "/var/lib/pgadmin/logs/receiver",
+		"create_directory": false,
```
benjaminjb, line 27: OK, so we're creating the `/receiver` directory in pod.go

```
shell.MakeDirectories(0o775, dataMountPath, path.Join(LogDirectoryAbsolutePath, "receiver")),
```

hmmm, not sure I have a question here, just interested in the different way we're doing this.

benjaminjb (approved):

internal/collector/pgadmin.go
cbandy: I've put TODOs elsewhere to create the directory using our function. This `create_directory` option does not do what we need in Kubernetes; [opentelemetry-collector-contrib#37774](https://www.github.com/open-telemetry/opentelemetry-collector-contrib/issues/37774)

IMV, this directory is a collector concern and belongs in its (the sidecar's) shell entrypoint or other init.

PR #4095, opened 2025-02-14
Rotate pgbouncer logs

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

No log rotation.

Allows the rotation of pgbouncer logs.

**Other Information**:

### DIFF
--- a/internal/collector/config.go
+++ b/internal/collector/config.go
@@ -5,11 +5,22 @@
 package collector
 
 import (
+	_ "embed"
+	"errors"
+	"fmt"
+	"regexp"
+
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/yaml"
 
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
+
+// The contents of "logrotate.conf" as a string.
+// See: https://pkg.go.dev/embed
+//
+//go:embed "logrotate.conf"
+var logrotateConfigFormatString string
 
 // ComponentID represents a component identifier within an OpenTelemetry
 // Collector YAML configuration. Each value is a "type" followed by an optional
@@ -101,4 +112,64 @@ func NewConfig(spec *v1beta1.InstrumentationSpec) *Config {
 	}
 
 	return config
+}
+
+func generateLogrotateConfig(logFilePath string, retentionPeriod string, postrotateScript string) (string, error) {
+	retentionPeriodMap, err := parseRetentionPeriodForLogrotate(retentionPeriod)
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf(
+		logrotateConfigFormatString,
+		logFilePath,
+		retentionPeriodMap["number"],
+		retentionPeriodMap["interval"],
+		postrotateScript,
+	), err
+}
+
+func parseRetentionPeriodForLogrotate(retentionPeriod string) (map[string]string, error) {
+	// logrotate does not have an "hourly" interval, but if an interval is not
+	// set it will rotate whenever logrotate is called, so set an empty string
+	// in the config file for hourly
+	unitIntervalMap := map[string]string{
+		"h":    "",
+		"hr":   "",
+		"hour": "",
+		"d":    "daily",
+		"day":  "daily",
+		"w":    "weekly",
+		"wk":   "weekly",
+		"week": "weekly",
+	}
+
+	// Define duration regex and capture matches
+	durationMatcher := regexp.MustCompile(`(?P<number>\d+)\s*(?P<interval>[A-Za-zµ]+)`)
+	matches := durationMatcher.FindStringSubmatch(retentionPeriod)
+
+	// If three matches were not found (whole match and two submatch captures),
+	// the retentionPeriod format must be invalid. Return an error.
+	if len(matches) < 3 {
+		return nil, errors.New("invalid retentionPeriod; must be number of hours, days, or weeks")
+	}
+
+	// Create a map with "number" and "interval" keys
+	retentionPeriodMap := make(map[string]string)
+	for i, name := range durationMatcher.SubexpNames() {
+		if i > 0 && i <= len(matches) {
+			retentionPeriodMap[name] = matches[i]
+		}
+	}
+
+	// If the duration unit provided is found in the unitIntervalMap, set the
+	// "interval" value to the "interval" string expected by logrotate; otherwise,
+	// return an error
+	if _, exists := unitIntervalMap[retentionPeriodMap["interval"]]; exists {
+		retentionPeriodMap["interval"] = unitIntervalMap[retentionPeriodMap["interval"]]
+	} else {
+		return nil, fmt.Errorf("invalid retentionPeriod; %s is not a valid unit", retentionPeriodMap["interval"])
+	}
+
+	return retentionPeriodMap, nil
 }
--- a/internal/collector/config.go
+++ b/internal/collector/config.go
@@ -166,7 +166,7 @@ func generateLogrotateConfig(logFilePath string, retentionPeriod *v1beta1.Durati
 // number of hours and "hourly"; otherwise, we will round up to the nearest day
 // and return the day count and "daily"
 func parseDurationForLogrotate(retentionPeriod *v1beta1.Duration) (int, string) {
-	hours := math.Round(retentionPeriod.AsDuration().Hours())
+	hours := math.Ceil(retentionPeriod.AsDuration().Hours())
 	if hours < 24 {
 		return int(hours), "hourly"
 	}
--- a/internal/collector/config.go
+++ b/internal/collector/config.go
@@ -149,16 +149,10 @@ func generateLogrotateConfig(logFilePath string, retentionPeriod *v1beta1.Durati
 	)
 }
 
-// FIXME: If the rotate count is 0, logrotate will not keep any archives, it will
-// essentially just empty out the working file during rotation, but it will still
-// rotate on the interval specified (hourly, daily, weekly, etc). If the user
-// sets the retentionPeriod to 0, we currently default to hourly regardless of
-// what unit the user provides. E.g. If the user sets the retentionPeriod to "0w"
-// they might expect that it will rotate once a week, but not keep any archives,
-// but we will actually rotate every hour... Is there a way to get the unit
-// provided in the Duration??
+// parseDurationForLogrotate takes a retention period and returns the rotate
+// number and interval string that should be used in the logrotate config
 func parseDurationForLogrotate(retentionPeriod *v1beta1.Duration) (int, string) {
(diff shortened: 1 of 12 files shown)


internal/collector/config_test.go
```diff
@@ -61,3 +63,134 @@ service:
...
+		{
+			retentionPeriod: "35hour",
+			number:          1,
+			interval:        "daily",
+		},
+		{
+			retentionPeriod: "36 hours",
+			number:          2,
+			interval:        "daily",
+		},
```
cbandy, line 161: 👍🏻 These tests!

🤔 Does logrotate "1 daily" mean (1) current file + (1) one day saved? If so, the contents of the two files combined is between 24 and 48 hours, right?
- just before rotation = 24 hours + 24 hours
- just after rotation = 0 hours + 24 hours

🔧 If that's how it works, I think I want both these cases to be `2, "daily"` so that right after rotation, we retain (more than) 35 hours. (Swap `math.Round` for `math.Ceil`? I dunno!)

internal/collector/instance.go
```diff
@@ -41,11 +41,13 @@ func AddToPod(
 	outPod *corev1.PodSpec,
 	volumeMounts []corev1.VolumeMount,
 	sqlQueryPassword string,
+	includeLogrotate bool,
```
cbandy, line 44: ❓ Do you expect we'll stop passing in `false` here eventually?

internal/controller/postgrescluster/pgbouncer.go
```diff
@@ -103,6 +103,13 @@ func (r *Reconciler) reconcilePGBouncerConfigMap(
 		(feature.Enabled(ctx, feature.OpenTelemetryLogs) || feature.Enabled(ctx, feature.OpenTelemetryMetrics)) {
 		err = collector.AddToConfigMap(ctx, otelConfig, configmap)
 	}
+	// If OTel logging is enabled and retentionPeriod is set, add logrotate config
+	if err == nil && otelConfig != nil && feature.Enabled(ctx, feature.OpenTelemetryLogs) &&
+		cluster.Spec.Instrumentation != nil && cluster.Spec.Instrumentation.Logs != nil &&
+		cluster.Spec.Instrumentation.Logs.RetentionPeriod != nil {
```
cbandy, line 109: 🔧 We should always rotate so that logs do not fill the disk. Without a user-specified retention, use some "small" amount that allows PgBouncer log collection to work. 🤔 "small" is less than a day? 🤷🏻

internal/naming/names.go
```diff
@@ -158,6 +158,15 @@ const (
...
 	PGBouncerLogPath = "/tmp"
 
+	// PGbouncerFullLogPath is the full path to the pgbouncer log file
+	PGBouncerFullLogPath = PGBouncerLogPath + "/pgbouncer.log"
+
+	// PGBouncerPostRotateScript is the script that is run after pgBouncer's log
+	// files have been rotated. The pgbouncer process is sent a sighup signal.
+	// TODO: Should this go here or in controller/postgrescluster/pgbouncer.go?
+	// Or somewhere in the internal/pgbouncer folder, for that matter?
+	PGBouncerPostRotateScript = "pkill -HUP --exact pgbouncer"
```
cbandy, line 168: 🌱 I don't like that `naming` is becoming a dumping ground for constants. This is fine for now; I know there are some circular imports to wrangle.

cbandy (changes requested):

internal/naming/names.go
benjaminjb, line 168: Kinda feel like this should be in the pgbouncer.go

internal/collector/config.go
```diff
@@ -102,3 +114,47 @@ func NewConfig(spec *v1beta1.InstrumentationSpec) *Config {
...
+}
+
+// parseDurationForLogrotate takes a retention period and returns the rotate
+// number and interval string that should be used in the logrotate config
+func parseDurationForLogrotate(retentionPeriod *v1beta1.Duration) (int, string) {
+	hours := math.Round(retentionPeriod.AsDuration().Hours())
+	if hours < 24 {
+		return int(hours), "hourly"
+	}
+	return int(math.Round(hours / 24)), "daily"
```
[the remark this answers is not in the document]
benjaminjb: Similarly, if I put in 36 hours, I might expect 36 hours, not 2 days.

internal/collector/config_test.go
dsessler7: Ahh yes, this makes sense. I'll change it.

internal/collector/instance.go
dsessler7: I think we will still pass `false` for pgAdmin because it has its own log rotation mechanism.

internal/naming/names.go
dsessler7: Agreed. Feels weird. I think I'll move it to `collector/pgbouncer.go`

dsessler7: Agreed. Moving it.

internal/collector/config.go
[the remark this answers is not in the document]
dsessler7: Well, 3 weeks of logs *is* 21 days of logs, we will just be keeping 21 log files where each is a day's worth of logs rather than 3 log files where each is a week's worth of logs... But yeah, we will definitely need plenty of emphasis on the fact that we will be rounding up to the nearest day for hour amounts that aren't multiples of 24.

internal/controller/postgrescluster/pgbouncer.go
dsessler7: Maybe just 1 day?

internal/collector/instance.go
cbandy: Ah! right.

internal/collector/config.go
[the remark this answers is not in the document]
cbandy: I don't think that needs to be documented. The user specifies _how much_ to retain, **not** the method for retaining it.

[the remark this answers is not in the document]
dsessler7: Good point. Should probably still make a note that we round up to the nearest day so users aren't confused at least.

internal/collector/instance.go
```diff
@@ -99,6 +105,32 @@ func AddToPod(
 		VolumeMounts:    append(volumeMounts, configVolumeMount),
 	}
 
+	// If a retentionPeriod is set and this is a pod that uses logrotate for
+	// log rotation, add config volume and mount for logrotate config
+	if includeLogrotate && spec != nil && spec.Logs != nil && spec.Logs.RetentionPeriod != nil {
```
benjaminjb, line 110: Do we require `RetentionPeriod` or are we allowing that to be blank and we'll default to a day?

internal/collector/instance.go
```diff
@@ -108,5 +140,26 @@ func AddToPod(
...
-	outPod.Volumes = append(outPod.Volumes, configVolume)
+}
+
+// startCommand generates the command script used by the collector container
+func startCommand(includeLogrotate bool) []string {
+	var startScript = `
+/otelcol-contrib --config /etc/otel-collector/config.yaml
+`
+
+	if includeLogrotate {
```
benjaminjb, line 151: We need a way to make sure that the logrotate.conf is valid and present before calling logrotate -- I mean, currently, above we only mount the config is it passes

```
includeLogrotate && spec != nil && spec.Logs != nil && spec.Logs.RetentionPeriod != nil
```

But we alter the startCommand just based on the `includeLogrotate` boolean.

internal/collector/config.go
```diff
@@ -102,3 +114,61 @@ func NewConfig(spec *v1beta1.InstrumentationSpec) *Config {
...
+	)
+}
+
+// parseDurationForLogrotate takes a retention period and returns the rotate
+// number and interval string that should be used in the logrotate config.
+// If the retentionPeriod is less than 24 hours, the function will return the
+// number of hours and "hourly"; otherwise, we will round up to the nearest day
+// and return the day count and "daily"
+func parseDurationForLogrotate(retentionPeriod *v1beta1.Duration) (int, string) {
+	hours := math.Round(retentionPeriod.AsDuration().Hours())
```
cbandy, line 169: 🔧 We reject fractions of hours in the CRD right now, but this logic should still use ceiling here; e.g. `1.4h` → `2, "hourly"`

NB: 🤦🏻 There's also a bug in `strfmt` regarding fractions. If you add a test for this, use multiple units, like: `1 hour 20 min`.

```suggestion
	hours := math.Ceil(retentionPeriod.AsDuration().Hours())
```

internal/controller/postgrescluster/pgbouncer.go
```diff
@@ -103,6 +103,11 @@ func (r *Reconciler) reconcilePGBouncerConfigMap(
 		(feature.Enabled(ctx, feature.OpenTelemetryLogs) || feature.Enabled(ctx, feature.OpenTelemetryMetrics)) {
 		err = collector.AddToConfigMap(ctx, otelConfig, configmap)
 	}
+	// If OTel logging is enabled, add logrotate config
+	if err == nil && otelConfig != nil && feature.Enabled(ctx, feature.OpenTelemetryLogs) {
+		err = collector.AddLogrotateConfig(ctx, cluster.Spec.Instrumentation, configmap,
+			naming.PGBouncerFullLogPath, collector.PGBouncerPostRotateScript)
```
cbandy, line 109: 🤔 Seeing a `collector` constant passed to a `collector` function has me wondering: Should this happen in `collector.NewConfigForPgBouncerPod` ?

internal/collector/instance.go
dsessler7: We allow it to be blank and default to a day.

internal/collector/instance.go
dsessler7: Yeah I need to remove the checking for `spec` and whatnot on line 110.

internal/controller/postgrescluster/pgbouncer.go
dsessler7: Hmm... Well right now `collector.NewConfigForPgBouncerPod` just creates the otel config, so I don't think this would go there (unless we changed it to return a separate `logrotate` config that then gets passed into `reconcilePGBouncerConfigMap` :thinking: )... Another option would be to have this happen in `collector.AddToConfigMap`, although then that function would have a bunch of `AddLogrotateConfig` calls for all the various components, but I guess that's not problematic? :thinking:

internal/collector/instance.go
dsessler7: Not sure how we would check that `logrotate.conf` is present and valid before calling `logrotate` :thinking:

internal/collector/instance.go
dsessler7: I'll remove the check for `spec` etc here

benjaminjb (approved):

cbandy (approved):

PR #4100, opened 2025-02-19
Add an API struct representing a single Secret value

cbandy:
This adds validation to the recurring pattern of selecting a single value from a Secret. Note that the "name" field is now required.

Secrets are best mounted as files, and the logic for translating these references into volume projections is now consolidated in two exported methods.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Bug fix
 - [x] Other

Secret name is not validated as required, and forgetting to add it causes reconciliation errors.

The field is now required, but clusters that lack it are broken anyway.


config/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml
```diff
@@ -1327,24 +1327,24 @@ spec:
                       More info: https://www.pgadmin.org/docs/pgadmin4/latest/ldap.html
                     properties:
                       key:
-                        description: The key of the secret to select from.  Must be
-                          a valid secret key.
+                        description: Name of the data field within the Secret.
+                        maxLength: 253
+                        minLength: 1
+                        pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?([.][a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
```
dsessler7, line 1333: What exactly is this regex pattern checking for?

dsessler7 (approved):
LGTM

config/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml
```diff
@@ -973,24 +973,24 @@ spec:
                       More info: https://www.pgadmin.org/docs/pgadmin4/latest/external_database.html
                     properties:
                       key:
-                        description: The key of the secret to select from.  Must be
-                          a valid secret key.
+                        description: Name of the data field within the Secret.
+                        maxLength: 253
+                        minLength: 1
+                        pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?([.][a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
```
cbandy, line 979: 🔧 This isn't right! Subdomain length, yes, but the format isn't so limited.

https://github.com/kubernetes/apimachinery/blob/v0.27.0/pkg/util/validation/validation.go#L425-L441

cbandy: Amended.

config/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml
```diff
@@ -973,24 +973,27 @@ spec:
...
-                        description: The key of the secret to select from.  Must be
-                          a valid secret key.
+                        description: Name of the data field within the Secret.
+                        maxLength: 253
+                        minLength: 1
+                        pattern: ^[-._a-zA-Z0-9]+$
                         type: string
+                        x-kubernetes-validations:
+                        - message: cannot be "." or start with ".."
+                          rule: self != "." && !self.startsWith("..")
```
cbandy, line 983: 📝 Having a CEL rule felt like a bit much, but if we allow these values up-front, then we'll error trying to send them into volume source, secret ref, etc during reconcile.

PR #4101, opened 2025-02-22
Rotate pgAdmin/gunicorn logs according to retentionPeriod

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

pgAdmin/Gunicorn log rotation is hard-coded

pgAdmin has a built in logfile rotation system, as does gunicorn. We want to use those systems, but respect the retention setting that a user sets in the instrumentation section. This PR uses the RetentionPeriod to
- set the number of backups for gunicorn and pgAdmin;
- set the rotation cycle for pgAdmin.

**Other Information**:

Issues: [PGO-2168]

### DIFF
--- a/internal/controller/standalone_pgadmin/pod.go
+++ b/internal/controller/standalone_pgadmin/pod.go
@@ -7,14 +7,16 @@ package standalone_pgadmin
 import (
 	"context"
 	"fmt"
-	"path"
+	"strconv"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
+	"github.com/crunchydata/postgres-operator/internal/collector"
 	"github.com/crunchydata/postgres-operator/internal/config"
+	"github.com/crunchydata/postgres-operator/internal/feature"
 	"github.com/crunchydata/postgres-operator/internal/initialize"
 	"github.com/crunchydata/postgres-operator/internal/kubernetes"
 	"github.com/crunchydata/postgres-operator/internal/naming"
@@ -45,6 +47,7 @@ const (
 
 // pod populates a PodSpec with the container and volumes needed to run pgAdmin.
 func pod(
+	ctx context.Context,
 	inPGAdmin *v1beta1.PGAdmin,
 	inConfigMap *corev1.ConfigMap,
 	outPod *corev1.PodSpec,
@@ -81,15 +84,6 @@ func pod(
 			// - https://kubernetes.io/docs/concepts/storage/volumes/#emptydir
 			// NOTE: tmpfs blocks are PAGE_SIZE, usually 4KiB, and size rounds up.
 			SizeLimit: resource.NewQuantity(32<<10, resource.BinarySI),
-		},
-	}
-
-	// create a temp volume for restart pid/other/debugging use
-	// TODO: discuss tmp vol vs. persistent vol
-	tmpVolume := corev1.Volume{Name: "tmp"}
-	tmpVolume.VolumeSource = corev1.VolumeSource{
-		EmptyDir: &corev1.EmptyDirVolumeSource{
-			Medium: corev1.StorageMediumMemory,
 		},
 	}
 
@@ -140,10 +134,6 @@ func pod(
 				MountPath: scriptMountPath,
 				ReadOnly:  true,
 			},
-			{
-				Name:      tmpVolume.Name,
-				MountPath: "/tmp",
-			},
 		},
 	}
 
@@ -168,7 +158,7 @@ func pod(
 
 	startup := corev1.Container{
 		Name:            naming.ContainerPGAdminStartup,
-		Command:         startupCommand(),
+		Command:         startupCommand(ctx, inPGAdmin),
 		Image:           container.Image,
 		ImagePullPolicy: container.ImagePullPolicy,
 		Resources:       container.Resources,
@@ -192,7 +182,6 @@ func pod(
 		configVolume,
 		dataVolume,
 		scriptVolume,
-		tmpVolume,
 	}
 	outPod.Containers = []corev1.Container{container}
 	outPod.InitContainers = []corev1.Container{startup}
@@ -352,7 +341,7 @@ done
 }
 
 // startupCommand returns an entrypoint that prepares the filesystem for pgAdmin.
-func startupCommand() []string {
+func startupCommand(ctx context.Context, inPgadmin *v1beta1.PGAdmin) []string {
 	// pgAdmin reads from the `/etc/pgadmin/config_system.py` file during startup
 	// after all other config files.
 	// - https://github.com/pgadmin-org/pgadmin4/blob/REL-7_7/docs/en_US/config_py.rst
@@ -374,10 +363,11 @@ func startupCommand() []string {
 
 		// configDatabaseURIPath is the path for mounting the database URI connection string
 		configDatabaseURIPathAbsolutePath = configMountPath + "/" + configDatabaseURIPath
+	)
 
-		// The constants set in configSystem will not be overridden through
-		// spec.config.settings.
-		configSystem = `
+	// The values set in configSystem will not be overridden through
+	// spec.config.settings.
+	var configSystem = `
 import glob, json, re, os, logging
 DEFAULT_BINARY_PATHS = {'pg': sorted([''] + glob.glob('/usr/pgsql-*/bin')).pop()}
 with open('` + configMountPath + `/` + configFilePath + `') as _f:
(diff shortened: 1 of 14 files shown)


internal/controller/standalone_pgadmin/pod.go
```diff
@@ -430,7 +457,7 @@ logconfig_dict['loggers']['gunicorn.error']['handlers'] = ['file']
 logconfig_dict['handlers']['file'] = {
   'class': 'logging.handlers.RotatingFileHandler',
   'filename': '` + GunicornLogFileAbsolutePath + `',
-  'backupCount': 1, 'maxBytes': 2 << 20, # MiB
+  'backupCount': ` + maxBackupRetentionNumber + `, 'maxBytes': 2 << 20, # MiB
```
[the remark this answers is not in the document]
benjaminjb: Wait, hold on...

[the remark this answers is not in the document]
benjaminjb: OK, this was bugging me, so I looked into it and we can do timed rotating for gunicorn.

(I tested by manually setting the code to rotate on minutes and saw the rotation happen as expected.)

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -374,10 +378,39 @@ func startupCommand() []string {
...
 		// configDatabaseURIPath is the path for mounting the database URI connection string
 		configDatabaseURIPathAbsolutePath = configMountPath + "/" + configDatabaseURIPath
+	)
+	var (
+		maxBackupRetentionNumber = "1"
+		// One day in minutes for pgadmin rotation
+		pgAdminRetentionPeriod = "24 * 60"
+		// Daily rotation for gunicorn rotation
+		gunicornRetentionPeriod = "D"
+	)
```
dsessler7, line 388: Thinking out loud: So it seems like we are defaulting to retaining a day's worth of logs whether we have the OpenTelemetryLogs feature enabled or not? Which looks like it was already the case for pgAdmin, but for gunicorn the behavior is changing from retaining based on size to time?

dsessler7 (approved):
LGTM

internal/controller/standalone_pgadmin/pod.go
benjaminjb: 🤔 Yeah -- and actually if we go back far enough, we weren't actually making any changes to the default logging until we added OTel. Which makes me think we should maybe only add these logging configs if OTel is enabled...

dsessler7: mmhmm

config/crd/bases/postgres-operator.crunchydata.com_pgupgrades.yaml
```diff
@@ -1134,11 +1138,30 @@ spec:
                       type: string
                   type: object
                 type: array
+              transferMethod:
+                description: |-
+                  The method pg_upgrade should use to transfer files to the new cluster.
+                  More info: https://www.postgresql.org/docs/current/pgupgrade.html
+                enum:
```
cbandy, line 1145: 📝 Something about the rebase. These commits are on main already.

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -390,47 +395,86 @@ if os.path.isfile('` + ldapPasswordAbsolutePath + `'):
...
+	// This command writes a script in `/etc/pgadmin/gunicorn_config.py` that reads
+	// from the `gunicorn-config.json` file and sets those variables globally.
+	// That way those values are available as settings when Gunicorn starts.
+	//
+	// Note: All Gunicorn settings are lowercase with underscores, so ignore
+	// any keys/names that are not.
+	//
+	// Gunicorn uses the Python logging package, which sets the following attributes:
+	// https://docs.python.org/3/library/logging.html#logrecord-attributes.
+	// JsonFormatter is used to format the log: https://pypi.org/project/jsonformatter/
```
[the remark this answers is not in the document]
benjaminjb: Yeah, this is part of that issue I noticed where we're (currently) changing our default logging behavior regardless of the collector, so I'm moving things around. 

(Also reminded me that we've made changes re: volumes for logging that I don't want to undo, which means if you upgrade to this PGO version, your pgadmins will restart even if you don't have collector turned on.)

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -140,10 +134,6 @@ func pod(
 				MountPath: scriptMountPath,
 				ReadOnly:  true,
 			},
-			{
-				Name:      tmpVolume.Name,
-				MountPath: "/tmp",
```
cbandy, line 145: 📝 It's possible that *something* will want a writable `/tmp`. 🤔 I *think* I'd prefer to leave this. (We're the ones who set ReadOnlyRootFS.)

internal/controller/standalone_pgadmin/statefulset.go
```diff
@@ -132,8 +134,10 @@ func statefulset(
 		}
 
 		collector.AddToPod(ctx, pgadmin.Spec.Instrumentation, pgadmin.Spec.ImagePullPolicy,
-			configmap, &sts.Spec.Template.Spec, volumeMounts, "", []string{}, false)
+			configmap, &sts.Spec.Template.Spec, volumeMounts, "", []string{LogDirectoryAbsolutePath}, false)
 	}
 
+	postgrescluster.AddTMPEmptyDir(&sts.Spec.Template)
```
cbandy, line 140: ❓ Oh hmm. Does the STS still have a writable `/tmp` ?

benjaminjb: Yeah, I noticed that in most other cases, we don't add `/tmp` directly in our spec but through this func that adds it to all containers. 

I ran into this problem with pgadmin (where /tmp was only avail to the main container) when the collector tried to start up and write a PID file to /tmp. I thought about revising that to use a var rather than file (as we talked about when we merged that PR), but I also wanted the collector container to have access to /tmp anyway, just in case. (Could also do both.)

internal/controller/standalone_pgadmin/pod.go
benjaminjb: OK, we made some changes about where we're logging, so I couldn't put all the logging stuff in as optional, but made some changes.

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -393,44 +383,82 @@ if os.path.isfile('` + configDatabaseURIPathAbsolutePath + `'):
...
-    _conf, _data = re.compile(r'[a-z_]+'), json.load(_f)
-    if type(_data) is dict:
-        globals().update({k: v for k, v in _data.items() if _conf.fullmatch(k)})
-
+		gunicornConfig = gunicornConfig + `
+import collections, copy, gunicorn, gunicorn.glogging
 gunicorn.SERVER_SOFTWARE = 'Python'
 logconfig_dict = copy.deepcopy(gunicorn.glogging.CONFIG_DEFAULTS)
 logconfig_dict['loggers']['gunicorn.access']['handlers'] = ['file']
 logconfig_dict['loggers']['gunicorn.error']['handlers'] = ['file']
```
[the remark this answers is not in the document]
benjaminjb: We start with the default settings from [here](https://github.com/benoitc/gunicorn/blob/23.0.0/gunicorn/glogging.py#L47), so

```
    "loggers": {
        "gunicorn.error": {
            "level": "INFO",
            "handlers": ["error_console"],
            "propagate": True,
            "qualname": "gunicorn.error"
        },

        "gunicorn.access": {
            "level": "INFO",
            "handlers": ["console"],
            "propagate": True,
            "qualname": "gunicorn.access"
        }
    },
```

So we set all our `logger.*.handlers` to `file`, and then below we set the `handlers.file`, so we skip using the default handlers (right below in the default settings):

```
    "handlers": {
        "console": {
            "class": "logging.StreamHandler",
            "formatter": "generic",
            "stream": "ext://sys.stdout"
        },
        "error_console": {
            "class": "logging.StreamHandler",
            "formatter": "generic",
            "stream": "ext://sys.stderr"
        },
    },
```

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -393,44 +383,82 @@ if os.path.isfile('` + configDatabaseURIPathAbsolutePath + `'):
...
+				pgAdminRetentionPeriod = "60"
+				gunicornRetentionPeriod = "H"
+			}
+		}
+
+		configSystem = configSystem + `
+LOG_ROTATION_AGE = ` + pgAdminRetentionPeriod + ` # minutes
 LOG_ROTATION_SIZE = 5 # MiB
-LOG_ROTATION_MAX_LOG_FILES = 1
+LOG_ROTATION_MAX_LOG_FILES = ` + maxBackupRetentionNumber + `
```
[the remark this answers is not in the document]
benjaminjb: Happy to, though not sure why we originally had both -- for safety/because pgadmin makes it easy?

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -393,44 +383,82 @@ if os.path.isfile('` + configDatabaseURIPathAbsolutePath + `'):
...
 
 JSON_LOGGER = True
 CONSOLE_LOG_LEVEL = logging.WARNING
 FILE_LOG_LEVEL = logging.INFO
 FILE_LOG_FORMAT_JSON = {'time': 'created', 'name': 'name', 'level': 'levelname', 'message': 'message'}
```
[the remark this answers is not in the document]
benjaminjb: We could move some of this to the `generateConfig` (with some vars for the logging settings); gunicorn also lets you read from a json file for some configs (with `logconfig_json`), so I could imagine moving some that to the configmap.

In my mind, the more we move to the configmap, the less disruptive a change is (as long as the configmap is being remounted and the process checks to reload). Is there another benefit you're thinking of?

(That said, I don't off the top of my head remember the config order of precedence here if we move this to `generateConfig`.)

internal/controller/standalone_pgadmin/pod.go
dsessler7: Yeah, as of right now the OTel start script puts the PID in a file in `tmp` (although we did talk previously about moving the PID to a bash variable)

dsessler7: nevermind, that's specific to the collector container

dsessler7: nevermind, nevermind, I see it now

dsessler7 (approved):

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -277,7 +267,11 @@ func startupScript(pgadmin *v1beta1.PGAdmin) []string {
 
 	// startCommands (v8 image includes Gunicorn)
 	var startCommandV7 = "pgadmin4 &"
-	var startCommandV8 = "gunicorn -c /etc/pgadmin/gunicorn_config.py --chdir $PGADMIN_DIR pgAdmin4:app &"
+	// For Gunicorn, watch the logging config and reload if changes are detected.
+	var startCommandV8 = "gunicorn -c /etc/pgadmin/gunicorn_config.py" +
+		" --reload-extra-file " + configMountPath + `/` + gunicornLogConfFilePath +
+		" --log-config-json " + configMountPath + `/` + gunicornLogConfFilePath +
```
cbandy, line 273: 👍🏻 Nice! We can do a lot with config files.

❓ My read of [`reload_extra_files`](https://docs.gunicorn.org/en/latest/settings.html#reload-extra-files) is that it requires `reload` to be enabled. Do we use that?

❓ I see `reload_extra_files` and `logconfig_json` mentioned in the config file settings. Is it possible to set these through `settings` in `generateGunicornConfig()`? 🤔 If so, we might end up with all JSON and no change to the entrypoint?

cbandy (approved):
Thank you for making those changes. One question, but not a blocker.

internal/controller/standalone_pgadmin/pod.go
benjaminjb: See https://github.com/benoitc/gunicorn/issues/3353 for an issue I ran into.

Not in that GH issue, but as a precursor to it: all the `reload` stuff is expressly for development. So I was thinking of adding a SIGHUP signal into our mtime checking loop, only SIGHUP doesn't work -- but SIGKILL might if we want to do that every time a cluster changes or the config changes. (The mtime for all those files change together since they are in the same configmap.)

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -325,10 +311,15 @@ loadServerCommand
...
+	// In order to get gunicorn to reload the logging config
+	// we need to send a KILL rather than a HUP signal.
+	// - https://github.com/benoitc/gunicorn/issues/3353
+	// Right now the config file is on the same configMap as the cluster file
+	// so if the mtime changes for any of those files, it will change for all.
 	var reloadScript = `
 exec {fd}<> <(:||:)
 while read -r -t 5 -u "${fd}" ||:; do
-    if [[ "${cluster_file}" -nt "/proc/self/fd/${fd}" ]] && loadServerCommand
+    if [[ "${cluster_file}" -nt "/proc/self/fd/${fd}" ]] && loadServerCommand && kill -KILL $(head -1 ${PGADMIN4_PIDFILE?});
```
cbandy, line 322: 🌱 I expect this increases the restart count when changing settings. WFM!

cbandy (approved):

internal/controller/standalone_pgadmin/pod.go
benjaminjb: I haven't seen any restarts with KILL -- our gunicorn is wrapped by our loop script.

PR #4107, opened 2025-02-26
Calculate Postgres parameters in the controller

cbandy:
The controller assigned mandatory and default values but did nothing with values defined in the spec.

The `patroni.DynamicConfiguration` function was interpreting its schemaless fields while also combining Postgres parameters from elsewhere. Its tests were long and complicated.

 1. Postgres parameters are now extracted from the schemaless Patroni field into their own function with its own tests. Unusual types are handled more deliberately.
 2. The PostgresCluster controller now creates a single set of Postgres parameters based on all the fields of the PostgresCluster spec.
 3. The DynamicConfiguration function is simpler (44 lines, 30% smaller) and easier to test.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement


internal/patroni/postgres_test.go
```diff
@@ -0,0 +1,110 @@
...
+		spec := new(v1beta1.PatroniSpec)
+		require.UnmarshalInto(t, spec, `{
+			dynamicConfiguration: {},
+		}`)
+		result := PostgresParameters(spec)
+
+		assert.Assert(t, result != nil)
+		assert.DeepEqual(t, result.AsMap(), map[string]string{})
+
+		t.Run("WrongType", func(t *testing.T) {
```
benjaminjb, line 42: Putting the "WrongType" test inside the "No_" test kinda confuses me.

benjaminjb: I guess it fits because the wrong type means the result is empty still?

benjaminjb (approved):

PR #4109, opened 2025-02-27
Rotate postgres logs

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Currently, the postgres logs are hard-coded to rotate daily and to retain a weeks worth of logs.

With the OpenTelemetryLogs feature enabled, the retention of the postgres logs will be determined by the `retentionPeriod` in the instrumentation spec. The operator will retain at least the amount specified; it might retain more depending on the value given. 

**Other Information**:

### DIFF
--- a/internal/collector/postgres.go
+++ b/internal/collector/postgres.go
@@ -9,7 +9,11 @@ import (
 	_ "embed"
 	"encoding/json"
 	"fmt"
+	"math"
 	"slices"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/crunchydata/postgres-operator/internal/feature"
 	"github.com/crunchydata/postgres-operator/internal/naming"
@@ -84,14 +88,22 @@ func EnablePostgresLogging(
 		// PostgreSQL v8.3 adds support for CSV logging, and
 		// PostgreSQL v15 adds support for JSON logging. The latter is preferred
 		// because newlines are escaped as "\n", U+005C + U+006E.
-		if inCluster.Spec.PostgresVersion < 15 {
+		if inCluster != nil && inCluster.Spec.PostgresVersion < 15 {
 			outParameters.Add("log_destination", "csvlog")
 		} else {
 			outParameters.Add("log_destination", "jsonlog")
 		}
 
-		// Keep seven days of logs named for the day of the week;
-		// this has been the default produced by `initdb` for some time now.
+		// If retentionPeriod is set in the spec, use that value; otherwise, we want
+		// to use a reasonably short duration. Defaulting to 1 day.
+		retentionPeriod := metav1.Duration{Duration: 24 * time.Hour}
+		if inCluster != nil && inCluster.Spec.Instrumentation != nil &&
+			inCluster.Spec.Instrumentation.Logs != nil &&
+			inCluster.Spec.Instrumentation.Logs.RetentionPeriod != nil {
+			retentionPeriod = inCluster.Spec.Instrumentation.Logs.RetentionPeriod.AsDuration()
+		}
+		logFilename, logRotationAge := generateLogFilenameAndRotationAge(retentionPeriod)
+
 		// NOTE: The automated portions of log_filename are *entirely* based
 		// on time. There is no spelling that is guaranteed to be unique or
 		// monotonically increasing.
@@ -100,9 +112,9 @@ func EnablePostgresLogging(
 		// probably requires another process that deletes the oldest files.
 		//
 		// The ".log" suffix is replaced by ".json" for JSON log files.
-		outParameters.Add("log_filename", "postgresql-%a.log")
+		outParameters.Add("log_filename", logFilename)
 		outParameters.Add("log_file_mode", "0660")
-		outParameters.Add("log_rotation_age", "1d")
+		outParameters.Add("log_rotation_age", logRotationAge)
 		outParameters.Add("log_rotation_size", "0")
 		outParameters.Add("log_truncate_on_rotation", "on")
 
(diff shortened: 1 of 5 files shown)


internal/collector/postgres.go
```diff
@@ -84,14 +84,25 @@ func EnablePostgresLogging(
...
 
-		// Keep seven days of logs named for the day of the week;
-		// this has been the default produced by `initdb` for some time now.
+		var retentionPeriod *v1beta1.Duration
+		if inCluster != nil && inCluster.Spec.Instrumentation != nil &&
+			inCluster.Spec.Instrumentation.Logs != nil &&
+			inCluster.Spec.Instrumentation.Logs.RetentionPeriod != nil {
+			retentionPeriod = inCluster.Spec.Instrumentation.Logs.RetentionPeriod
+		} else {
+			retentionPeriod, _ = v1beta1.NewDuration("1d")
```
dsessler7, line 99: I wish I could create a `Duration` struct literal so that I wouldn't need to handle an error... @cbandy is this possible?

internal/collector/postgres.go
```diff
@@ -272,3 +283,49 @@ func EnablePostgresLogging(
...
 }
+
+// generateLogFilenameAndRotationAge takes a retentionPeriod and returns a
+// log_filename and log_rotation_age to be used to configure postgres logging
+func generateLogFilenameAndRotationAge(
+	retentionPeriod *v1beta1.Duration,
+) (logFilename, logRotationAge string) {
+	number, interval := parseDurationForLogrotate(retentionPeriod)
+
+	// If the retention is a single hour or as single day, the log filename
```
benjaminjb, line 294:
```suggestion
	// If the retention is a single hour or a single day, the log filename
```

internal/collector/postgres.go
```diff
@@ -272,3 +283,49 @@ func EnablePostgresLogging(
...
+	logFilename = "postgresql.log"
+
+	// If the duration is 1 hour, set log_rotation_age to "1h"
+	if interval == "hourly" {
+		if number == 1 {
+			logRotationAge = "1h"
+		} else {
+			// If interval is "hourly" but duration is not 1 hour, we will round
+			// retention up to a day
+			logRotationAge = "1d"
```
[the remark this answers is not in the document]
dsessler7: Yes

internal/collector/postgres.go
```diff
@@ -272,3 +283,49 @@ func EnablePostgresLogging(
...
+// generateLogFilenameAndRotationAge takes a retentionPeriod and returns a
+// log_filename and log_rotation_age to be used to configure postgres logging
+func generateLogFilenameAndRotationAge(
+	retentionPeriod *v1beta1.Duration,
+) (logFilename, logRotationAge string) {
+	number, interval := parseDurationForLogrotate(retentionPeriod)
+
+	// If the retention is a single hour or as single day, the log filename
+	// will not have any escape characters in it
+	logFilename = "postgresql.log"
```
benjaminjb, line 296: Have you tried a 1h rotation and seen this file be replaced with the new log?

internal/collector/postgres.go
```diff
@@ -272,3 +283,49 @@ func EnablePostgresLogging(
...
+
+		// If the duration is greater than a day and less than or equal to a week,
+		// we will round retention up to a week
+		if number > 1 && number <= 7 {
+			logFilename = "postgresql-%w.log"
+		}
+
+		// If the duration is greater than a week and less than or equal to 28 days,
+		// we will round retention up to a month
+		if number > 7 && number <= 28 {
```
benjaminjb, line 320: What would it look like if we went up to 31? I would think, if a month had less than 31 days, PG wouldn't write over last month's last few days until the next month that had 31 days. That is, if Jan writes 1-31, then Feb writes 1-28, then 29, 30, 31 are all from Jan. Is that right?

benjaminjb (commented):
LGTM, just some smaller questions I want to check on

internal/collector/postgres.go
dsessler7: That's how I understand it.

internal/collector/postgres.go
cbandy: 🤔 I recommend changing `collector.parseDurationForLogrotate` to accept `metav1.Duration` or `time.Duration`.

dsessler7: Done. Mind taking a look?

internal/collector/postgres.go
```diff
@@ -272,3 +285,56 @@ func EnablePostgresLogging(
 		}
 	}
 }
+
+// generateLogFilenameAndRotationAge takes a retentionPeriod and returns a
+// log_filename and log_rotation_age to be used to configure postgres logging
+func generateLogFilenameAndRotationAge(
+	retentionPeriod metav1.Duration,
+) (logFilename, logRotationAge string) {
+	number, interval := parseDurationForLogrotate(retentionPeriod)
```
cbandy, line 294: 🤔 I wonder if translating the retention period twice is worth it. Would the logic below be easier if we went straight off a duration?

```go
hours := math.Ceil(retentionPeriod.Hours())

switch true {
case hours <= 1:	… // minutes
case hours <= 24:	… // hours
case hours <= 24 * 7:	… // weekdays
case hours <= 24 * 28:	… // days of month
default:	… // days of year
}
```

cbandy (approved):

internal/collector/postgres.go
```diff
@@ -272,3 +285,56 @@ func EnablePostgresLogging(
...
+	// will always need to make up the total retention period with multiple log
+	// files that hold subunits of the total time (e.g. if the retentionPeriod
+	// is an hour, there will be 60 1-minute long files; if the retentionPeriod
+	// is a day, there will be 24 1-hour long files, etc)
+
+	// If the duration is 1 hour, set log_rotation_age to 1 minute
+	if interval == "hourly" {
+		if number == 1 {
+			logFilename = "postgresql-%M.log"
+			logRotationAge = "1"
```
cbandy, line 306: ⛏️ I can't remember the base unit a particular parameter uses. 📝 Unit formats described in [this section](https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-NAMES-VALUES).

```suggestion
			logRotationAge = "1min"
```

PR #4111, opened 2025-02-28
Calculate Postgres HBA rules in the controller

cbandy:
The controller assigned mandatory and default rules but did nothing with rules defined in the spec.

The `patroni.DynamicConfiguration` function was interpreting its schemaless fields while also combining Postgres HBA rules from elsewhere. Its tests were long and complicated.

1. Postgres HBA rules are now extracted from the schemaless Patroni field in their own function with its own tests.
2. The PostgresCluster controller now creates a single set of HBA rules based on all the fields of the PostgresCluster spec.
3. The DynamicConfiguration function is simpler (19 lines, 18% smaller) and easier to test.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement


internal/controller/postgrescluster/postgres.go
```diff
@@ -41,6 +42,35 @@ import (
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
+// generatePostgresHBAs produces the HBA rules for cluster that incorporates,
+// from highest to lowest precedence:
+//  1. mandatory rules determined by controllers
+//  2. rules in cluster.spec.patroni.dynamicConfiguration
```
benjaminjb, line 48: This makes sense (as the status quo), though would we ever want to loosen this to let user-defined rules take precedence?

internal/postgres/hba.go
```diff
@@ -157,3 +152,45 @@ func (hba *HostBasedAuthentication) String() string {
 	return strings.TrimSpace(fmt.Sprintf("%s %s %s %s %s %s",
 		hba.origin, hba.database, hba.user, hba.address, hba.method, hba.options))
 }
+
+// OrderedHBAs is an append-only sequence of pg_hba.conf lines.
+type OrderedHBAs struct {
+	records []string
+}
+
+// Append renders and adds pg_hba.conf lines to o. Nil pointers are ignored.
```
benjaminjb, line 161: Just checking: it wouldn't be worth filtering out the nils before the `grow`, would it?

internal/patroni/config_test.go
```diff
@@ -359,142 +354,28 @@ func TestDynamicConfiguration(t *testing.T) {
...
 		},
 		{
-			name: "postgresql.pg_hba: no default when input",
+			name: "HBA pass through",
 			spec: `{
 				patroni: {
 					dynamicConfiguration: {
 						postgresql: {
-							pg_hba: [custom],
+							pg_hba: [calculated, elsewhere],
```
benjaminjb, line 368: These don't come through?

internal/controller/postgrescluster/postgres.go
```diff
@@ -41,6 +42,35 @@ import (
...
+	builtin := postgres.NewHBAs()
+	pgmonitor.PostgreSQLHBAs(ctx, cluster, &builtin)
+	pgbouncer.PostgreSQL(cluster, &builtin)
+
+	// Postgres processes HBA rules in order. Start with mandatory rules
+	// so connections are matched against them first.
+	result := new(postgres.OrderedHBAs)
+	result.Append(builtin.Mandatory...)
+
+	// Append any rules specified in the the Patroni section.
```
dsessler7, line 62: :pick: 
```suggestion
	// Append any rules specified in the Patroni section.
```

internal/controller/postgrescluster/postgres.go
```diff
@@ -41,6 +42,35 @@ import (
...
+	before := result.Length()
+	result.AppendUnstructured(patroni.PostgresHBAs(cluster.Spec.Patroni)...)
+
+	// When there are no specified rules, include the recommended defaults.
+	if result.Length() == before {
+		result.Append(builtin.Default...)
+	}
```
dsessler7, line 69: So if the user provides anything, we don't set any defaults? Do we not want to provide defaults for rules they didn't touch?

internal/postgres/hba.go
cbandy: 🤔 I might have added the filtering after I had used Grow. I think I'll skip Grow for now.

internal/patroni/config_test.go
cbandy: This function doesn't do that work anymore. Now,

1. `patroni.PostgresHBAs` reads them out,
2. `postgrescluster.generatePostgresHBAs` stitches them together in the right order, then passes that result
3. here, to `patroni.DynamicConfiguration`.

internal/controller/postgrescluster/postgres.go
cbandy: Yep, same as before: https://github.com/CrunchyData/postgres-operator/blob/22f7989f1a1a83ed1dbc64ae8e9b0f6404c54079/internal/patroni/config.go#L261-L266

dsessler7 (approved):
LGTM

PR #4112, opened 2025-02-28
Clone embedded metrics variable to avoid continuous appending.

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Operator always appends to the same global variable, the contents of which are put in a ConfigMap which triggers a reconcile, which forces another appending, etc, etc, until the ConfigMap gets too large.

The metrics variable in question is cloned so that we no longer continually append to it. ConfigMap doesn't get overloaded.

**Other Information**:


benjaminjb (approved):
Nice catch!

cbandy (approved):

PR #4113, opened 2025-03-01
Allow tuning of logs batching behavior

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature

Batches of logs are exported every 200ms.

Users can configure the size and frequency of batches via the API. Documentation on each field explains what they do and why each is useful.

**Other Information**:

Issue: PGO-2176


internal/collector/config_test.go
```diff
@@ -55,13 +60,55 @@ processors:
...
+	t.Run("LogsBatches", func(t *testing.T) {
+		var spec *v1beta1.InstrumentationSpec
+		require.UnmarshalInto(t, &spec, `{
+			logs: {
+				batches: {
+					maxDelay: 5min 12sec,
+					maxRecords: 123,
+					minRecords: 45,
+				},
+			},
```
cbandy, line 83: 📝 I'm pretty good but not 💯 on the name `batches` for this section of the API. LMK

internal/collector/pgadmin.go
```diff
@@ -102,7 +101,7 @@ func EnablePgAdminLogging(ctx context.Context, spec *v1beta1.InstrumentationSpec
 		Processors: []ComponentID{
 			"resource/pgadmin",
 			"transform/pgadmin_log",
-			SubSecondBatchProcessor,
+			LogsBatchProcessor,
 			CompactingProcessor,
 		},
 		Exporters: exporters,
```
cbandy, line 107: ❓ This and the pipeline below it differ only in their receivers. Is that correct?

internal/collector/config_test.go
benjaminjb: "For my pgadmin, I have this specification: for the instrumentation, I want my logs to be retained for a period of X days and to be batched X at a time." Seems reasonable.

But we only allow a user to set one batch processor, so `batches` seems odd (as a plural noun, unless it's a verb...)

internal/collector/pgadmin.go
benjaminjb: Yeah, iirc, the only distinction in the `transform` step is a pattern match that will hit for pgadmin and not for gunicorn (or vice versa) -- so, yeah, we could combine them.

internal/collector/pgbackrest.go
```diff
@@ -107,7 +105,7 @@ func NewConfigForPgBackrestRepoHostPod(
 			Processors: []ComponentID{
 				"resource/pgbackrest",
 				"transform/pgbackrest_logs",
-				SubSecondBatchProcessor,
```
benjaminjb, line 110: Are we going to use this Processor? Seems like we could ditch it (I'm 10 files in)

internal/collector/patroni.go
```diff
@@ -16,6 +17,11 @@ func EnablePatroniLogging(ctx context.Context,
 	inCluster *v1beta1.PostgresCluster,
 	outConfig *Config,
 ) {
+	var spec *v1beta1.InstrumentationLogsSpec
+	if inCluster != nil && inCluster.Spec.Instrumentation != nil {
+		spec = inCluster.Spec.Instrumentation.Logs
+	}
+
```
benjaminjb, line 24: We don't use this var outside of the `if feature.Enabled` block, do we? And technically only use it once?

internal/testing/validation/pgadmin_test.go
```diff
@@ -32,6 +32,131 @@ func TestPGAdminInstrumentation(t *testing.T) {
...
+				valid   bool
+			}{
+				{valid: true, batches: ``},              // both null
+				{valid: true, batches: `minRecords: 1`}, // one null
+				{valid: true, batches: `maxDelay: 1s`},  // other null
+
+				{valid: false, batches: `minRecords: 0`}, // one zero
+				{valid: false, batches: `maxDelay: 0m`},  // other zero
+
+				{valid: true, batches: `minRecords: 0, maxDelay: 0m`}, // both zero
```
benjaminjb, line 48:
```suggestion
				{valid: true, batches: `minRecords: 0, maxDelay: 0m`}, // both zero
				{valid: true, batches: `minRecords: 1, maxDelay: 1s`}, // both non-zero
```
Just to complete the square/use test as documentation

pkg/apis/postgres-operator.crunchydata.com/v1beta1/instrumentation_types.go
```diff
@@ -71,3 +75,59 @@ type InstrumentationLogsSpec struct {
...
+	// Number of records to wait for before exporting a batch. Higher numbers
+	// allow more records to be deduplicated and compressed before export.
+	// ---
+	// +kubebuilder:validation:Minimum=0
+	// +default=8192
+	// +optional
+	MinRecords *int32 `json:"minRecords,omitempty"`
+}
+
+func (s *OpenTelemetryLogsBatchSpec) Default() {
```
benjaminjb, line 125: Do we need both kubebuilder defaults (when do those run, on admission?) and a default func?

internal/collector/config_test.go
cbandy: I was thinking: "logs go out in batches." Maybe "batching" ?

https://pkg.go.dev/go.opentelemetry.io/collector/processor/batchprocessor#section-readme

> The batch processor … places them into batches. Batching helps…

internal/collector/pgbackrest.go
cbandy: It's still used in some metrics pipelines. 🤔 Worth a review later.

dsessler7 (approved):

internal/collector/config_test.go
benjaminjb: "Set this field for custom batching rules." 👍

PR #4115, opened 2025-03-03
Provide a way to enable resource detectors

cbandy:
Resource detectors are a simple way to gather resource attributes from the environment automatically. They are especially useful in managed or cloud environments.

**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature

GCP identifies our logs as coming from its `generic_node` resource.

Setting `spec.instrumentation.config.detectors: [{ name: gcp }]` is enough to have GCP recognize our logs as its `k8s_container` [monitored resource](https://cloud.google.com/monitoring/api/resources).

**Other Information**:

Issue: PGO-2193


dsessler7 (approved):
LGTM

pkg/apis/postgres-operator.crunchydata.com/v1beta1/instrumentation_types.go
```diff
@@ -71,3 +85,23 @@ type InstrumentationLogsSpec struct {
...
+type OpenTelemetryResourceDetector struct {
+	// Name of the resource detector to enable: `aks`, `eks`, `gcp`, etc.
+	// ---
+	// +kubebuilder:validation:MaxLength=20
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Name string `json:"name"`
+
+	// Attributes to use from this detector. Detectors usually add every attribute
+	// they know automatically. Names omitted here behave according detector defaults.
```
benjaminjb, line 100:
```suggestion
	// they know automatically. Names omitted here behave according to detector defaults.
```

internal/collector/config_test.go
```diff
@@ -56,12 +62,43 @@ processors:
...
+  resourcedetection:
+    aks:
+      resource_attributes:
+        k8s.cluster.name:
+          enabled: true
+    detectors:
+    - gcp
+    - aks
+    override: false
+    timeout: 30s
```
benjaminjb, line 99: Huh, so this form lets us avoid specifying the resourcedetection sources like `resourcedetection/ec2`.

(hmmm, for a moment I had `/aws` there before double-checking -- not here, but maybe in the docs we want to say what the allowable options are or say that there are only a set list and send people to the otel docs.

cbandy: Yeah, this component handles lots of ... sources. 🤔 Now that I think about it, that seems unusual because it means you must include/compile support for multiple clouds, even if you don't use them.

PR #4120, opened 2025-03-04
Add a validated field for Postgres HBA rules

cbandy:
These structured fields are easier and safer to use than raw HBA records.

**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature

HBA rules must be crafted by hand in the `patroni.dynamicConfiguration` section of the spec.

HBA rules can be specified and validated in the `authentication.rules` section. 

**Other Information**:

Issue: PGO-2263


internal/postgres/hba.go
```diff
@@ -106,7 +156,7 @@ func (hba *HostBasedAuthentication) NoSSL() *HostBasedAuthentication {
 func (hba *HostBasedAuthentication) Options(opts map[string]string) *HostBasedAuthentication {
 	hba.options = ""
 	for k, v := range opts {
-		hba.options = fmt.Sprintf("%s %s=%s", hba.options, k, hba.quote(v))
+		hba.options = fmt.Sprintf("%s %s=%s", hba.options, hba.quote(k), hba.quote(v))
 	}
 	return hba
```
cbandy, line 161: 🔧 These need to be sorted.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgres_types.go
```diff
@@ -9,6 +9,15 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+type PostgresAuthenticationSpec struct {
+	// More info: https://www.postgresql.org/docs/current/auth-pg-hba-conf.html
+	// ---
+	// +kubebuilder:validation:MaxItems=10
```
tjmoore4, line 15: :thinking: Is this limit on max items and the other similar limits below Postgres limitations or are they something we're setting as reasonable maximums?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgres_types.go
```diff
@@ -68,6 +77,77 @@ type PostgresConfig struct {
...
+// +kubebuilder:validation:XValidation:rule=`has(self.hba) ? true : has(self.connection) && has(self.method)`,message=`"connection" and "method" are required`
+//
+// +structType=atomic
+type PostgresHBARuleSpec struct {
+	//// 🌱 Files containing HBA records that are included in the "pg_hba.conf" file.
+	//// ---
+	//// +kubebuilder:validation:MaxItems=20
+	//// +listType=atomic
+	//// +optional
+	//Files []OptionalFileRef `json:"files,omitempty"`
```
tjmoore4, line 137: :question: should this be cleaned up?

tjmoore4 (commented):
Looks good overall. A couple of questions/comments, mostly non-blockers with one potential area for cleanup.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgres_types.go
cbandy: These are reasonable maximums. Kubernetes estimates the complexity of all the CEL rules in your CRD and rejects CRDs with rules that might cost «too much.» There's no way to see the current cost or the threshold for rejection. 😢 

I've started putting these on new fields so we can add CEL rules later. If a user wants more than we've allowed, we can safely raise it. If we start unbounded and later wish to add CEL rules, we'd have to add a limit, which is a breaking change.

https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#resource-use-by-validation-functions

> Because of this, it is considered best practice to put a limit via `maxItems`, `maxProperties`, and `maxLength` for anything that will be processed in a validation rule in order to prevent validation errors during cost estimation.

internal/postgres/hba.go
```diff
@@ -50,10 +52,51 @@ func NewHBA() *HostBasedAuthentication {
...
 func (*HostBasedAuthentication) quote(value string) string {
 	return `"` + strings.ReplaceAll(value, `"`, `""`) + `"`
 }
 
+func (hba *HostBasedAuthentication) quoteDatabase(name string) string {
+	// Since PostgreSQL 16, a quoted string beginning with slash U+002F is
+	// interpreted as a regular expression. Express these names as a Postgres
+	// regex that exactly matches the entire name.
+	if len(name) > 0 && name[0] == '/' {
+		name = "/^" +
```
cbandy, line 73: 📝 @tony-landreth, this is what I _didn't_ need to patch. Slash means regex, and regex can express an exact match.

tjmoore4 (approved):

PR #4123, opened 2025-03-06
Pgadmin oauth secrets

philrhurst:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

The PGAdmin API only allows users to specify OAUTH configuration data via plaintext in the manifest.

Users can now specify their OAUTH configuration via a Secret and pass that to the PGAdmin with an entry containing the secret name in  `spec.config.oauthConfigurations`.

**Other Information**:


internal/controller/standalone_pgadmin/statefulset.go
```diff
@@ -135,5 +140,51 @@ func statefulset(
...
+	secretHash = hex.EncodeToString(hash.Sum(nil))
+	currentHash = sts.Spec.Template.Annotations["configMapHash"]
+
+	if currentHash != secretHash {
+		if sts.Spec.Template.Annotations == nil {
+			sts.Spec.Template.Annotations = map[string]string{}
+		}
+		sts.Spec.Template.Annotations["configMapHash"] = secretHash
+	}
+}
```
dsessler7, line 190: I'm probably just missing something, but I don't understand the purpose of this code here... It seems like it's just updating the hash in the annotation if it is different from the current hash, but I don't see any references to either of these hash annotations in the codebase, so I'm wondering what their purpose is in the first place...

pkg/apis/postgres-operator.crunchydata.com/v1beta1/standalone_pgadmin_types.go
```diff
@@ -42,6 +42,14 @@ type StandalonePGAdminConfiguration struct {
...
 	// +kubebuilder:validation:Type=object
 	Settings SchemalessObject `json:"settings,omitempty"`
+
+	// OauthConfigurations allows the user to reference one or more Secrets
+	// containing OAUTH2 configuration settings for pgAdmin.
+	// Each Secret shall contain a single data key called oauth-config.json
+	// whose value is a JSON object containing the OAUTH2 configuration settings.
+	// More info: https://www.pgadmin.org/docs/pgadmin4/latest/oauth2.html
+	// +optional
+	OauthConfigurations []corev1.LocalObjectReference `json:"oauthConfigurations,omitempty"`
```
dsessler7, line 52: I think it's probably fine to tell the user that they have to name the file a specific thing; however, if we wanted to allow the user to name it whatever they want, you could use `SecretKeyRef` found in `pkg/apis/postgres-operator.crunchydata.com/v1beta1/config_types.go`

internal/controller/standalone_pgadmin/statefulset.go
philrhurst: I ran into an issue when I updated one of the entries `spec.config.settings.OAUTH2_CONFIG`. The associated ConfigMap would update, but it would not trigger a restart of the PGAdmin Pod (so the change would not be applied). I found that updating the hash in the stateful set _would_ trigger the rollout. I'm open to other suggestions though.

dsessler7: Ahhh, I see... Yeah, this makes sense as the "reloading" part of our startup script for pgAdmin just watches the `pgadmin-shared-clusters.json` file and then runs the `load-servers` command when it sees changes. We could do something similar that watches the `oauthConfigDir` and then restarts the pgAdmin process if it notices a change; however, your solution is probably fine. The only disadvantage that I see with the current implementation is that the whole Pod gets rolled back out when the config/secret changes. This not only restarts the pgAdmin container but any sidecar containers (e.g. the OTel collector ) as well, and generally takes longer than just killing the process and starting it back up again in the shell... @cbandy do you have thoughts on this?

philrhurst: I agree with you on the disadvantage of the current approach. But since the OAuth config is loaded through the InitContainer, I'm not sure what could be done (without a lot of restructuring) but I'm open to ideas.

dsessler7: Ahh, another good point. Okay, let's stick with the current implementation, but maybe we just make the comment above where `checkOauthSecretsChange` and `checkConfigMapChange` are called a bit more descriptive... Something like: "If certain configuration changes, we need to not only restart the pgAdmin process, but also re-run the init container. We therefore store hashes of these configurations in annotations on the pgAdmin statefulset, which force a Pod restart when they change."

dsessler7 (approved):
Just need to beef up that one comment for clarity, otherwise, looks good.

andrewlecuyer:
@philrhurst just a general question here.  Could a user effectively put any pgAdmin configuration into the Secret?  E.g. even setting that have nothing to do with OAuth?  Or are the `OAUTH2` settings the only settings that are accepted?

internal/controller/standalone_pgadmin/statefulset.go
philrhurst: I've updated the comments - thanks for the assist.

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -212,6 +215,24 @@ func podConfigFiles(configmap *corev1.ConfigMap, pgadmin v1beta1.PGAdmin) []core
...
+		for _, secret := range oauthSecrets {
+			config = append(config, corev1.VolumeProjection{
+				Secret: &corev1.SecretProjection{
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: secret.Name,
+					},
+					Items: []corev1.KeyToPath{
+						{
+							Key:  "oauth-config",
+							Path: fmt.Sprintf("%s%s.json", oauthConfigDir, secret.Name),
```
cbandy, line 228: 📝 I think I see. Secret names are unique in this namespace, ensuring we have a unique filename in this directory.

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -185,7 +187,8 @@ func pod(
 
 // podConfigFiles returns projections of pgAdmin's configuration files to
 // include in the configuration volume.
-func podConfigFiles(configmap *corev1.ConfigMap, pgadmin v1beta1.PGAdmin) []corev1.VolumeProjection {
+func podConfigFiles(configmap *corev1.ConfigMap, pgadmin v1beta1.PGAdmin,
+	oauthSecrets []corev1.Secret) []corev1.VolumeProjection {
```
cbandy, line 191: 📝 We may be able to pass only projections or refs here rather than entire Secret objects.

config/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml
```diff
@@ -1353,6 +1353,30 @@ spec:
...
+                          description: |-
+                            Name of the referent.
+                            This field is effectively required, but due to backwards compatibility is
+                            allowed to be empty. Instances of this type with an empty value here are
+                            almost certainly wrong.
+                            More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                          type: string
+                      type: object
+                      x-kubernetes-map-type: atomic
+                    type: array
```
cbandy, line 1379: 🤔 The pgAdmin setting is an array, so the array here makes sense.

❓ Are there any other constraints on the OAuth settings in pgAdmin, like can't use the same `OAUTH2_NAME` twice or so?

cbandy (commented):
Some thoughts and questions. I'm trying out the API locally right now.

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -373,6 +394,20 @@ import glob, json, re, os
...
+        for filename in os.listdir(folder_path):
+            with open(os.path.join(folder_path, filename), "r", encoding="utf-8") as f:
+                try:
+                    oath = json.load(f)
+                    if oath.get("OAUTH2_NAME") not in [
+                        o.get("OAUTH2_NAME") for o in _data.get("OAUTH2_CONFIG")]:
+                        _data.get("OAUTH2_CONFIG").append(oath)
+                    for o in _data.get("OAUTH2_CONFIG"):
+                        if o.get("OAUTH2_NAME") == oath.get("OAUTH2_NAME"):
+                            o.update(oath)
```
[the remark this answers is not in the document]
philrhurst: If the Secret and Spec both contain data, then the Secret will win.

internal/controller/standalone_pgadmin/pod.go
philrhurst: yes - right now the OAUTH Secret would have to exist within the same namespace as PGAdmin, but that seemed reasonable.

internal/controller/standalone_pgadmin/pod.go
```diff
@@ -373,6 +394,20 @@ import glob, json, re, os
 DEFAULT_BINARY_PATHS = {'pg': sorted([''] + glob.glob('/usr/pgsql-*/bin')).pop()}
 with open('` + configMountPath + `/` + configFilePath + `') as _f:
     _conf, _data = re.compile(r'[A-Z_0-9]+'), json.load(_f)
+    folder_path = '` + configMountPath + `/` + oauthConfigDir + `'
+    if os.path.isdir(folder_path):
+        for filename in os.listdir(folder_path):
+            with open(os.path.join(folder_path, filename), "r", encoding="utf-8") as f:
+                try:
+                    oath = json.load(f)
```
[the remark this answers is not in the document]
philrhurst: yes they have!

pkg/apis/postgres-operator.crunchydata.com/v1beta1/standalone_pgadmin_types.go
```diff
@@ -42,6 +42,14 @@ type StandalonePGAdminConfiguration struct {
...
 	// +kubebuilder:validation:Type=object
 	Settings SchemalessObject `json:"settings,omitempty"`
+
+	// OauthConfigurations allows the user to reference one or more Secrets
+	// containing OAUTH2 configuration settings for pgAdmin.
+	// Each Secret shall contain a single data key called oauth-config
+	// whose value is a JSON object containing the OAUTH2 configuration settings.
+	// More info: https://www.pgadmin.org/docs/pgadmin4/latest/oauth2.html
+	// +optional
+	OauthConfigurations []corev1.LocalObjectReference `json:"oauthConfigurations,omitempty"`
```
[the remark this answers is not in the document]
philrhurst: I was trying to keep the API changes to a minimum, i.e. give the name of the Secret and nothing else is required. The documentation would describe the format the Secret should take.

config/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml
philrhurst: As far as I can tell - there is nothing stopping you in pgAdmin from using the same OAUTH2_NAME twice, but if I read their code correctly, they end up using `OAUTH2_NAME` as the key for a dictionary.

philrhurst: so that would mean duplicate usage could collapse in unexpected ways within pgAdmin.

cbandy: I see. This kinda thing, yeah? https://github.com/pgadmin-org/pgadmin4/blob/846d0eafe6026907b26ccabd2cedb219a1d38a5e/web/pgadmin/authenticate/oauth2.py#L109-L114

philrhurst: yes - that's the thing

PR #4126, opened 2025-03-06
Otel custom queries

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

No way to customize queries for metrics when using OTel feature.

You can now both provide custom queries/metrics and turn off existing ones when using the OpenTelemetryMetrics feature.

**Other Information**:


pkg/apis/postgres-operator.crunchydata.com/v1beta1/instrumentation_types.go
```diff
@@ -91,10 +101,70 @@ type InstrumentationLogsSpec struct {
...
+	// Set a max length to keep rule costs low.
+	// +kubebuilder:validation:MaxLength=20
+	//
+	// +required
+	Name string `json:"name"`
+
+	// A ConfigMap holding the yaml file that contains the queries.
+	// ---
+	// +required
+	Queries ConfigMapKeyRef `json:"queries"`
```
dsessler7, line 143: Kind of want to rename this, but not sure what it should be...

internal/naming/labels.go
```diff
@@ -40,6 +40,10 @@ const (
 	// LabelMovePGWalDir is used to identify the Job that moves an existing pg_wal directory.
 	LabelMovePGWalDir = labelPrefix + "move-pgwal-dir"
 
+	// LabelCollectorDiscovery is added to Pods running the OpenTelemetry "collector"
+	// container to support discovery by Prometheus
+	LabelCollectorDiscovery = labelPrefix + "crunchy-otel-collector"
```
benjaminjb, line 45: If we are planning on not changing the Prometheus setup, then I think the pods we want Prometheus to scrape should have the usual metrics labels? Ugh, but that label is `crunchy-postgres-exporter`, which I do not want on these pods, so we should adjust the prometheus config to look for either, I suppose. I'm looking at the monitoring stack, so I'll add that to my ticket.

internal/collector/instance.go
```diff
@@ -136,18 +150,22 @@ func AddToPod(
...
 			Protocol:      corev1.ProtocolTCP,
 		}}
 	}
 
-	outPod.Containers = append(outPod.Containers, container)
+	template.Spec.Containers = append(template.Spec.Containers, container)
+
+	// add the proper label to support Pod discovery by Prometheus
+	initialize.Labels(template)
+	template.Labels[naming.LabelCollectorDiscovery] = "true"
```
benjaminjb, line 168: Should all the labeling and the metric-specific file work be done inside the `if feature.Enabled(ctx, feature.OpenTelemetryMetrics)` block?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/instrumentation_types.go
benjaminjb: Hmmm, they are queries, though this field is more specifically the name of a configmap that has yaml-formatted queries, right? I think `metrics.add[].queries` makes sense, but I could see some wiggle room here. What aspect of this would you want to capture in a different name?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/instrumentation_types.go
```diff
@@ -91,10 +101,70 @@ type InstrumentationLogsSpec struct {
...
+	// +kubebuilder:validation:Pattern=`^((PT)?( *[0-9]+ *(?i:(ms|s|m)|(milli|sec|min)s?))+|0)$`
+	//
+	// `controller-gen` needs to know "Type=string" to allow a "Pattern".
+	// +kubebuilder:validation:Type=string
+	//
+	// Set a max length to keep rule costs low.
+	// +kubebuilder:validation:MaxLength=20
+	// +kubebuilder:validation:XValidation:rule=`duration("0") <= self && self <= duration("60m")`
+	//
+	// +default="5s"
```
benjaminjb, line 159: What happens if you leave this field blank?

internal/collector/postgres_metrics.go
```diff
@@ -38,44 +40,100 @@ var ltPG16 json.RawMessage
...
+			log.Error(err, "error compiling postgres metrics")
+		}
+
+		// Remove any queries that user has specified in the spec
+		if inCluster.Spec.Instrumentation != nil &&
+			inCluster.Spec.Instrumentation.Metrics != nil &&
+			inCluster.Spec.Instrumentation.Metrics.CustomQueries != nil &&
+			inCluster.Spec.Instrumentation.Metrics.CustomQueries.Remove != nil {
+
+			// Convert json to array of maps
```
benjaminjb, line 75: We start with a yaml for ease of reading; then we generate a json for ease of putting into required OTel; then we unmarshal into a golang map for ease of removing _if_ there's anything to remove.

I don't have any particular ask here, but this feels like a place where we might want to rethink our formats.

internal/collector/postgres_metrics.go
```diff
@@ -110,3 +196,53 @@ func appendToJSONArray(a1, a2 json.RawMessage) (json.RawMessage, error) {
...
+								queryMetricsArr = queryMetricsArr[:len(queryMetricsArr)-1]
+							}
+
+							// We found and deleted the metric, so we can continue
+							// to the next iteration of the Outer loop.
+							continue Outer
+						}
+					} else {
+						// TODO: What should we do if the assertion fails?
+						fmt.Printf("%s is not a map[string]any", metric)
```
benjaminjb, line 237: If this fails here, then something is wrong with (a) the file that we put in to the operator or (b) part of our marshalling/unmarshalling that wasn't caught when we did it -- is there any other way we could get an error at this point?

If (a), then our metrics collection will probably error out (and what does that look like?).

None of which answers your question, though! I just mean to say: if we've reached this point -- which seems unlikely -- we might have bigger problems. For now, I like the idea of logging the unexpected result but letting the process run on -- and we can revisit this in the future.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/instrumentation_types.go
dsessler7: Yeah so you touched on it; it's that it is the name/key of the configmap, not the queries themselves that makes it feel a little off to me. That being said we've done the same thing in a number of places, so I guess it's fine.

internal/collector/instance.go
```diff
@@ -136,18 +151,22 @@ func AddToPod(
 			}},
 		}
 		container.VolumeMounts = append(container.VolumeMounts, logrotateConfigVolumeMount)
-		outPod.Volumes = append(outPod.Volumes, logrotateConfigVolume)
+		template.Spec.Volumes = append(template.Spec.Volumes, logrotateConfigVolume)
 	}
 
-	if feature.Enabled(ctx, feature.OpenTelemetryMetrics) {
+	if feature.Enabled(ctx, feature.OpenTelemetryMetrics) && includeMetricsPort {
```
benjaminjb, line 157: Do we only want to apply the label and add the config queries if we're enabling this pod for metrics?

internal/naming/labels.go
benjaminjb: Just adding this note after our discussion: in prometheus, we'll add another scrape job to get the OTel collector; and we'll make sure to _only_ add the port (and the label?) to those pods that have metrics exports on

internal/collector/instance.go
dsessler7: Hmm... In my mind the label simply indicates that the Pod has an otel collector sidecar in it; it could feasibly be used by something other than Prometheus.

Not sure what you mean by "config queries"?

internal/collector/postgres_metrics.go
dsessler7: I agree, it does feel like a lot of conversion, but I don't see any better approach.

benjaminjb: Oh, but it seems cleaner to me now that we have a struct

internal/collector/instance.go
benjaminjb: Oh, yeah, "config queries" doesn't make sense. I meant the custom queries files -- right now, if you have custom queries files on the spec, we will add them to every collector, e.g., pgbouncer and pgbackrest pods get these mounted, even though we don't need them. 

(And actually, in pgbouncer's case, we don't want them, right? In pgbackrest's case, they just don't get used. But if we try to run a query meant for postgres on pgbouncer, that might lead to an error.)

dsessler7: Ah, you're talking about the adding of the custom queries configmaps to the config volume as projected sources... Hmm... I don't think that it's a problem that they are added even if the pod doesn't have metrics, but I guess it makes sense to limit where we put them. I'll change that.

dsessler7: This also makes me realize that we'll be projecting these custom query configmaps into pods that will be serving metrics but won't be using the custom queries (like pgbouncer), but I'm not sure that it's worth the effort to prevent that.

internal/collector/instance.go
dsessler7: As indicated in the conversation below, I think the label should go on all Pods that have the collector in place, but I agree that the projecting of the custom queries into the Pod should only happen if metrics is enabled.

internal/collector/instance.go
benjaminjb: Oh, right, and it won't lead to an error in pgbouncer, because we aren't adding the files to pgbouncer_metrics -- and when we do want to open pgbouncer up to custom metrics, we probably will add a `component` tag to api for each file. OK, I've talked myself back down from being worried.

benjaminjb (approved):
Oh, and just double checking -- all the pods/custom query behavior still works, yeah?

dsessler7:
Ha, yup! Everything still works.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/instrumentation_types.go
dsessler7: Tested both having the field with a blank value and deleting the field altogether, and they both result in it defaulting to 5s as expected.

PR #4128, opened 2025-03-07
collector util func

benjaminjb:
A revision to how we check for OTel Feature Gates. (for discussion)


internal/collector/pgadmin.go
```diff
@@ -10,120 +10,120 @@ import (
...
+				LogsBatchProcessor,
+				CompactingProcessor,
+			},
+			Exporters: exporters,
+		}
 
-	otelYAML, err := otelConfig.ToYAML()
-	if err == nil {
-		configmap.Data["collector.yaml"] = otelYAML
+		otelYAML, err := otelConfig.ToYAML()
```
cbandy, line 123: 🤔 Does this shadow `err`?

❓ Is this big hunk just changing indentation? If so, the early return might be preferrable.

internal/collector/pgadmin_test.go
```diff
@@ -31,7 +30,14 @@ func TestEnablePgAdminLogging(t *testing.T) {
 
 		configmap := new(corev1.ConfigMap)
 		initialize.Map(&configmap.Data)
-		err := collector.EnablePgAdminLogging(ctx, nil, configmap)
+		retentionPeriod, err := v1beta1.NewDuration("12 hours")
+		assert.NilError(t, err)
+		instrumentation := v1beta1.InstrumentationSpec{
```
cbandy, line 35: 📝 could also require.Umarshal to avoid juggling err.

internal/collector/pgadmin_test.go
```diff
@@ -44,7 +50,7 @@ collector.yaml: |
   extensions:
     file_storage/pgadmin_data_logs:
       create_directory: false
-      directory: `+pgadmin.LogDirectoryAbsolutePath+`/receiver
+      directory: /var/lib/pgadmin/logs/receiver
```
cbandy, line 53: 👍🏻 I recall assert diffs being harder to read when these long strings aren't constant.

cbandy (approved):

PR #4134, opened 2025-03-17
Modify context path of OTEL transformers

benjaminjb:
**Checklist:**

**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

As of 0.119.0, the parser collector modifies certain paths of transforms. While this is not an error, the resulting log can look a little alarming, so we rewrote some of our transforms to match.

We handle the transforms.

**Other Information**:
Issues: [PGO-2268]


internal/collector/patroni.go
```diff
@@ -69,23 +69,23 @@ func EnablePatroniLogging(ctx context.Context,
 					`set(instrumentation_scope.name, "patroni")`,
 
 					// https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/-/pkg/ottl/ottlfuncs#parsejson
-					`set(cache, ParseJSON(body["original"]))`,
+					`set(log.cache, ParseJSON(log.body["original"]))`,
```
cbandy, line 72: 🤔 Do we still have to set `context=log`? If so, what a strange change they made.

cbandy (approved):

internal/collector/patroni.go
benjaminjb: We do not. I'm not sure of the benefit of this change.

That said, is it expected for almost every emission to be `LogRecord #0`?

cbandy: That sounds like an index, maybe? The batch time is quite small by default, so one-at-a-time sounds plausible.

internal/collector/postgres_logs_transforms.yaml
```diff
@@ -128,7 +126,7 @@
     # https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/-/pkg/ottl/ottlfuncs#time
     # https://git.postgresql.org/gitweb/?p=postgresql.git;hb=REL_10_0;f=src/backend/utils/error/elog.c#l2246
     # https://git.postgresql.org/gitweb/?p=postgresql.git;hb=REL_17_0;f=src/backend/utils/error/elog.c#l2671
-    - set(time, Time(cache["timestamp"], "%F %T.%L %Z"))
+    - set(log.time, Time(log.cache["timestamp"], "%F %T.%L %Z")) where IsString(log.cache["timestamp"])
```
cbandy, line 129: 👍🏻 I don't mind being safe here. Did something warn? I expect Postgres always includes a timestamp.

cbandy (approved):
I couldn't pay attention the whole way through. LGTM

internal/collector/postgres_logs_transforms.yaml
benjaminjb: I can't remember where (pgbouncer, patroni?) but at some point I caused a log to come through that didn't have a timestamp. That might've been when I was trying to reproduce the bug Drew was checking out? But I thought we might want to be safe wherever we're doing a time transform just in case.

PR #4138, opened 2025-03-19
Document how "schemaless" interacts with CEL rules

cbandy:
@tjmoore4 I captured your findings and added a test for something else. 😇


tjmoore4 (approved):
LGTM; test failure seems unrelated.

PR #4157, opened 2025-04-09
Fix issue where the presence of --target-timeline was adding --target…

dsessler7:
…-action.

Adjust tests and add more test cases.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

Currently, if a user adds the `--target-timeline` flag to the datasource options to be used in a restore, the operator will also add `--target-action` which will result in the restore failing. 

Now, if `--target-timeline` is added to the restore options, the operator will not add `--target-action` and the restore should proceed as expected.

**Other Information**:

### DIFF
--- a/internal/controller/postgrescluster/pgbackrest.go
+++ b/internal/controller/postgrescluster/pgbackrest.go
@@ -1166,10 +1166,16 @@ func (r *Reconciler) reconcileRestoreJob(ctx context.Context,
 		"--pg1-path=" + pgdata,
 		"--repo=" + regexRepoIndex.FindString(repoName)}...)
 
+	// Look specifically for the "--target" flag, NOT flags that contain
+	// "--target" (e.g. "--target-timeline")
+	targetRegex, err := regexp.Compile("--target[ =]")
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -1166,10 +1166,15 @@ func (r *Reconciler) reconcileRestoreJob(ctx context.Context,
...
+	targetRegex, err := regexp.Compile("--target[ =]")
+	if err != nil {
+		return err
+	}
+
 	var deltaOptFound, foundTarget bool
 	for _, opt := range opts {
 		switch {
-		case strings.Contains(opt, "--target"):
+		case targetRegex.Match([]byte(opt)):
```
tjmoore4, line 1177: :pick: I think a comment would be good here. Something like, "don't allow the `--target` flag, but allow flags that start with `--target` such as `--target-timeline`" just to help clarify what's going on.

tjmoore4 (approved):

PR #4163, opened 2025-04-24
Fix for PGO-2380:

dsessler7:
Only add logrotate volume mounts to instance pod when backups are enabled. Add kuttl tests to ensure that collector will run on postgres instance when backups are disabled.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

At the moment, we cannot run the collector on a postgres instance pod when backups are disabled as the pod will fail due to it attempting to mount a file that does not exist. 

We only mount logrotate config files for the instance pod when backups are enabled, which allows the collector to run whether backups are enabled or not. 

**Other Information**:


testing/kuttl/e2e/otel-logging-and-metrics/README.md
```diff
@@ -23,6 +23,12 @@ This test assumes that the operator has both OpenTelemetryLogs and OpenTelemetry
...
     1. Ensure that the ConfigMap, Service, and Deployment for the standalone OTel collector come up and that the collector container is running and ready.
     2. Assert that the standalone collector is receiving logs from all of our components (i.e. the standalone collector is getting logs for postgres, patroni, pgbackrest, pgbouncer, pgadmin, and gunicorn).
+5. Create a new cluster with `instrumentation` spec in place, but no `backups` spec to test the OTel features with optional backups.
+    1. Ensure that the cluster comes up and the database and collector containers are running and ready.
+    2. Add a backups spec to the new cluster and ensure that pgbackrest is added to the instance pod, a repo-host pod is created, and the collector runs on both pods.
+    3. Remove the backups spec from the new cluster.
+    4. Annotate the cluster to allow backups to be removed.
+    5. Ensure that the repo-host pod is destroyed, pgbackrest is removed from the instance pod, and the collector continues to run on the instance pod.
```
benjaminjb, line 31: I am curious -- if you skip steps 3 and 4, will 5 fail?

testing/kuttl/e2e/otel-logging-and-metrics/13-assert-instance.yaml
```diff
@@ -0,0 +1,55 @@
...
+# Then, check the collector logs for patroni, and postgres logs.
+# Finally, ensure the monitoring user exists and is configured.
+- script: |
+    retry() { bash -ceu 'printf "$1\nSleeping...\n" && sleep 5' - "$@"; }
+    check_containers_ready() { bash -ceu 'echo "$1" | jq -e ".[] | select(.type==\"ContainersReady\") | .status==\"True\""' - "$@"; }
+    contains() { bash -ceu '[[ "$1" == *"$2"* ]]' - "$@"; }
+
+    pod=$(kubectl get pods -o name -n "${NAMESPACE}" \
+      -l postgres-operator.crunchydata.com/cluster=otel-cluster-no-backups,postgres-operator.crunchydata.com/data=postgres)
+    [ "$pod" = "" ] && retry "Pod not found" && exit 1
```
tony-landreth, line 16: :question: What is the retry mechanism? It looks like the retry function just prints a message sleeps.

benjaminjb: Yeah, "retry" is a bit of a misnomer -- Joseph and I talked about this when he introduced the mechanism. As you point out, the `retry` just print and sleeps; the actual retrying is because 
(a) this exits 1 and
(b) it's a TestAssert
In KUTTL, failed TestAsserts automatically retry a number of times.

tony-landreth: Gotcha. Thanks!

testing/kuttl/e2e/otel-logging-and-metrics/README.md
dsessler7: Not sure I follow... If you don't remove the backups spec and annotate the cluster, pgbackrest will just continue to run (in the instance pod and repo-host)...

benjaminjb: Right -- I'm wondering if step 5 tests what we want it to test.

dsessler7: So, first we create a brand new cluster that does not have backups enabled. And we make sure that the collector runs in that scenario. Then we add backups and ensure that everything works properly with that transition. Then we remove backups and ensure that everything works properly with that transition... What scenario do you think we aren't testing? Or what is it about the test that is insufficient?

benjaminjb: It's been a while since I did a KUTTL test, but I remember running into issues where I asserted that X existed, but it actually ignored what wasn't X. So here, we assert there's a few containers, but not the collector container -- but will it fail if that collector container does exist? Or will it just skip that?

dsessler7: Ahh, I think I understand now... In the assert step we tell it to check the instance pod for specific containers and it will fail if the list is not exactly correct (if we ask about 3 containers, but there are actually 5 containers that include the 3 we ask about, it will fail).

benjaminjb: Ah, it checks if the list (of containers, etc.) is complete still? I think we (and others) were pushing to get it to check only for the presence of the items in the list.

benjaminjb (approved):

tony-landreth (approved):

PR #4170, opened 2025-05-06
Otel metrics change intervals

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Some metric queries are not running at ideal intervals.

A couple metrics queries were moved to more appropriate intervals.
Also, the initial delay for the sqlquery receivers has been increased to avoid attempting to authenticate before sql setup has completed.

**Other Information**:


benjaminjb (commented):
Let me just make sure I've got this right:

- some queries were only running in the wrong time period
- you moved these queries to other / new files
- to help clarify when queries should run, you set a new pattern where every file is either fast or slow
- not every version-based query has both fast & slow query files

I think that all makes sense, but happy to talk it over too.

benjaminjb (approved):

dsessler7:
> Let me just make sure I've got this right:
> 
>     * some queries were only running in the wrong time period
> 
>     * you moved these queries to other / new files
> 
>     * to help clarify when queries should run, you set a new pattern where every file is either fast or slow
> 
>     * not every version-based query has both fast & slow query files
> 
> 
> I think that all makes sense, but happy to talk it over too.

That's exactly correct.

The bump to the `initial_delay` seems to be causing the OTel kuttl test to fail on finding the 5 minute metric. I'm going to move the check for 5 minute metrics to be the last thing we check for to see if that gets it passing again.

benjaminjb (approved):

PR #4172, opened 2025-05-07
Run development tools rather than install them

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

Some Make targets download tools to `hack/tools` while others expect them to be installed already.

Run the tools for contributing code using `go run` and `go tool`. This reduces clutter in the project directory and lets Go cache and share these tools effectively across projects.


Makefile
```diff
@@ -15,9 +15,17 @@ BUILDAH_BUILD ?= buildah bud
 GO ?= go
 GO_BUILD = $(GO) build
 GO_TEST ?= $(GO) test
-KUTTL ?= kubectl-kuttl
+
```
tjmoore4, line 18: :thinking: I like this approach a lot. Did you happen to notice any performance difference using this method? I'm guessing it'd be minimal, but I was curious if you noticed any change.

tjmoore4 (approved):

Makefile
cbandy: The first time I run one of these, it has to download and compile the code, but it's very cached after that. I don't notice a speed difference myself.

PR #4183, opened 2025-05-19
Add per db metrics

benjaminjb:
**Checklist:**

**Type of Changes:**
- [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

No per-db metrics

- Add per-db metrics (based on pgMonitor v5.2.1)
- Allow per-db custom metrics

**Other Information**:
Issues: [PGO-21]


benjaminjb:
Some notes on this PR:
- I based the per db metrics on pgmonitor 5.2.1, which Drew is bumping to elsewhere
- some per db metrics were already included in `internal/collector/gte_pg16_slow_metrics.yaml`/`internal/collector/lt_pg16_slow_metrics.yaml`, so I moved them over to the per db metric list; this let me delete those files and update some of the logic for adding metrics
- I added steps to our OTel KUTTL test (beginning to strain at the bounds of a reasonable test, might break up in the future) to test the different functionality added: targeting certain dbs with the defaults; removing metrics from the defaults; adding custom per db metrics.

config/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml
```diff
@@ -2165,6 +2170,12 @@ spec:
                               type: string
                             type: array
                         type: object
+                      perDBMetricTargets:
```
cbandy, line 2173: 🤔 This and `customQueries` only make sense at/on Postgres. Is there a way to arrange structs so the PGAdmin API doesn't have these fields? Not a blocker; v1beta1.

benjaminjb: I didn't want to make any too big changes with the spec, but I wonder if the pgadmin and postgrescluster versions will diverge even more (or if we'll eventually want to slice up instrumentation by target-type, eg., postgres-instrumentation, pgbouncer-instrumentation, pgadmin-instrumentation).

dsessler7: Hmm yeah good point... Probably makes sense to at least have a pgadmin-instrumentation and a postgrescluster-instrumentation and then pick and choose which structs make sense in each

benjaminjb: Should I add "taking apart the instrumentation struct" to this PR?

benjaminjb: Given the other work due this sprint, I'm going to make a ticket for this topic and not include it in this PR.

benjaminjb: Made that other ticket.

dsessler7:
> @dsessler7 This failed at finding patroni logs -- didn't we run into that before?

It looks like it is still using the 5.8.1 postgres-operator image even though the 5.8.2 image is specified for `RELATED_IMAGE_COLLECTOR` in the kuttl-k3d gh action job... :thinking: We aren't prefetching the postgres-operator image (which we should), but it should still pull the image specified by the related image...

internal/collector/postgres_metrics.go
```diff
@@ -182,29 +183,60 @@ func EnablePostgresMetrics(ctx context.Context, inCluster *v1beta1.PostgresClust
...
+					if querySet.Database != "" {
+						db = querySet.Database
+					}
+					receiverName := "sqlquery/" + querySet.Name
+					config.Receivers[receiverName] = map[string]any{
+						"driver": "postgres",
+						"datasource": fmt.Sprintf(
+							`host=localhost dbname=%s port=5432 user=%s password=${env:PGPASSWORD}`,
+							db,
+							MonitoringUser),
```
dsessler7, line 204: :question: Will the monitoring user always have access to whatever databases the user specifies?

internal/collector/postgres_metrics.go
```diff
@@ -182,29 +183,60 @@ func EnablePostgresMetrics(ctx context.Context, inCluster *v1beta1.PostgresClust
 
 		// Add custom queries if they are defined in the spec
```
dsessler7, line 184:
```suggestion
		// Add custom queries and per-db metrics if they are defined in the spec
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/instrumentation_types.go
```diff
@@ -159,6 +164,12 @@ type InstrumentationCustomQueries struct {
 	// +default="5s"
 	// +optional
 	CollectionInterval *Duration `json:"collectionInterval,omitempty"`
+
+	// The database to target with added custom queries.
+	// Default behavior is to target `postgres`.
+	// ---
+	// +optional
+	Database string `json:"database,omitempty"`
```
dsessler7, line 172: Do we want the user to be able to specify multiple databases for a given batch of custom queries? :thinking:

internal/collector/postgres_metrics.go
benjaminjb: Looking into this -- I got the metrics I expected, but why?

benjaminjb: It might be because we `GRANT pg_monitor to ccp_monitoring;` in the setup.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/instrumentation_types.go
benjaminjb: Want to table this discussion and get feedback from the demo tomorrow?

internal/collector/postgres_metrics.go
```diff
@@ -180,31 +181,65 @@ func EnablePostgresMetrics(ctx context.Context, inCluster *v1beta1.PostgresClust
...
+			if inCluster.Spec.Instrumentation.Metrics.CustomQueries != nil &&
+				inCluster.Spec.Instrumentation.Metrics.CustomQueries.Add != nil {
+
+				for _, querySet := range inCluster.Spec.Instrumentation.Metrics.CustomQueries.Add {
+					// Create a receiver for the query set
+
+					dbs := []string{"postgres"}
+					if len(querySet.Databases) != 0 {
+						dbs = querySet.Databases
+					}
```
dsessler7, line 197: Definitely need to make a note in the documentation that if the user provides any databases for a custom query set that the default "postgres" database will not be included (unless they include it themselves).

PR #4205, opened 2025-07-14
Create cloud repo specific job

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

Currently, backups to cloud repos are done by exec'ing into the pg instance pod's pgbackrest sidecar and running the pgbackrest backup command there. 

Backups to cloud repos will now be done in the backup job itself. If no local volume repo exists, a dedicated repo host will not be created.

**Other Information**:


internal/pgbackrest/util.go
```diff
@@ -10,16 +10,20 @@ import (
...
 
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/rand"
 
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
+var tmpDirSizeLimit = resource.MustParse("16Mi")
```
benjaminjb, line 21: Is this size also copied over from postgrescluster?

internal/pgbackrest/config.go
```diff
@@ -129,8 +131,7 @@ func CreatePGBackRestConfigMapIntent(ctx context.Context, postgresCluster *v1bet
 				postgresCluster.Spec.Backups.PGBackRest.Global,
 			).String()
 
-		if RepoHostVolumeDefined(postgresCluster) &&
-			collector.OpenTelemetryLogsOrMetricsEnabled(ctx, postgresCluster) {
+		if collector.OpenTelemetryLogsOrMetricsEnabled(ctx, postgresCluster) {
```
benjaminjb, line 134: So we're always adding OTel config to the configmap if OTel is enabled on this cluster? (Originally the plan was to only add Otel to repohost, but I guess we might also want to add it to cloud backup job pods?)

internal/pgbackrest/reconcile.go
```diff
@@ -111,17 +112,9 @@ func AddConfigToInstancePod(
 	// volumes stay valid and Kubernetes propagates their contents to those pods.
 	secret := corev1.VolumeProjection{Secret: &corev1.SecretProjection{}}
 	secret.Secret.Name = naming.PGBackRestSecret(cluster).Name
+	secret.Secret.Items = append(secret.Secret.Items, clientCertificates()...)
```
benjaminjb, line 115: Why do we need this here?

internal/controller/postgrescluster/instance.go
```diff
@@ -1510,7 +1508,7 @@ func (r *Reconciler) reconcileInstanceCertificates(
 			root.Certificate, leafCert.Certificate,
 			leafCert.PrivateKey, instanceCerts)
 	}
-	if err == nil {
+	if err == nil && backupsSpecFound {
```
benjaminjb, line 1512: Just double-checking: why don't we need this if using a local volume? Because these certs have to do with talking between repo/instance?

internal/pgbackrest/util.go
dsessler7: yes... don't know if we ever want the size to vary?

internal/pgbackrest/config.go
dsessler7: This section of code changes is a little confusing because there was a redundancy removed... In the old version of the file, we called `RepoHostVolumeDefined` on line 99 and stored the result in `addDedicatedHost`. We then checked that value on line 118 before entering the `if` block. We then redundantly checked for a repo host again on line 132, but by calling `RepoHostVolumeDefined` again. I removed this redundant check...

So we still only check for OTel being enabled (now line 134) if a repo host is defined (checked on line 123).

Collecting logs for cloud backups was out of scope for this effort, so no changes to this yet.

internal/pgbackrest/reconcile.go
dsessler7: We used to only create the pgbackrest server in the instance if a repo host was present. Now we always create the pgbackrest server in the instance (assuming backups are enabled of course) because it is needed for both the repo host and the cloud backups to talk to it.  The server not only needs a server cert and key (which are added in the `AddServerToInstancePod` function), but also requires the client CA cert which is provided via the `clientCertificates` function... So it's a little confusing because the installation of the necessary pieces for the `server` occur in both `AddServerToInstancePod` and `AddConfigToInstancePod`... It probably could be refactored so that the `server` stuff is fully handled in the former function...

internal/controller/postgrescluster/instance.go
dsessler7: Not entirely sure what you are referring to, but we always need the instance certificates now. Both the cloud backup from the backup job and the local volume backup from the repo host need to be able to talk to the pgbackrest server in the instance pod.

internal/pgbackrest/util.go
benjaminjb: I'm not sure if we want the size to vary, or if it's worth commenting on "why" we chose this size.

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -2547,11 +2543,15 @@ func (r *Reconciler) reconcileReplicaCreateBackup(ctx context.Context,
 		replicaRepoReady = (condition.Status == metav1.ConditionTrue)
 	}
 
-	// get pod name and container name as needed to exec into the proper pod and create
-	// the pgBackRest backup
-	_, containerName, err := getPGBackRestExecSelector(postgresCluster, replicaCreateRepo)
-	if err != nil {
-		return errors.WithStack(err)
+	// TODO: Since we now only exec into the repo host when backing up to a local volume and
```
cbandy, line 2546: 👍🏻 Calling this out and leaving it alone sounds good to me.

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -2817,27 +2814,6 @@ func (r *Reconciler) reconcileStanzaCreate(ctx context.Context,
 	return false, nil
 }
 
-// getPGBackRestExecSelector returns a selector and container name that allows the proper
-// Pod (along with a specific container within it) to be found within the Kubernetes
-// cluster as needed to exec into the container and run a pgBackRest command.
-func getPGBackRestExecSelector(postgresCluster *v1beta1.PostgresCluster,
```
cbandy, line 2823: 🎉 🔥 Nice!

internal/controller/postgrescluster/pgbackrest_test.go
```diff
@@ -2648,13 +2602,83 @@ func TestCopyConfigurationResources(t *testing.T) {
 
 func TestGenerateBackupJobIntent(t *testing.T) {
 	ctx := context.Background()
+	cluster := v1beta1.PostgresCluster{}
+	cluster.Name = "hippo-test"
+	cluster.Default()
+
+	// The code assumes that if repo.Volume is nil, then it is a cloud repo backup,
+	// therefore, an "empty" input results in a job spec for a cloud repo backup
```
cbandy: ⛏️ Rather than "assumes," I might say the code "expects" or "interprets" that nil means the backup is to a cloud repo.

internal/pgbackrest/config.go
```diff
@@ -504,6 +517,64 @@ func populateRepoHostConfigurationMap(
...
+	for option, val := range globalConfig {
+		global.Set(option, val)
+	}
+
+	// set the configs for all PG hosts
+	for i, pgHost := range pgHosts {
+		// TODO(cbandy): pass a FQDN in already.
+		pgHostFQDN := pgHost + "-0." +
+			serviceName + "." + serviceNamespace + ".svc." +
+			naming.KubernetesClusterDomain(context.Background())
```
cbandy, line 553: 👍🏻 Agreed!

benjaminjb (approved):
Thanks for the walkthru and now also double-checking: the postgres instance always has the pgbackrest sidecar, so if OTel is enabled, we always want to rotate those, yeah?

dsessler7:
> Thanks for the walkthru and now also double-checking: the postgres instance always has the pgbackrest sidecar, so if OTel is enabled, we always want to rotate those, yeah?

Yes, if backups are enabled at all and OTel Logging is enabled, we want to add log rotate configuration for the pgbackrest logs in the postgres instance pod, and this statement always has been true. However, in the code before we were not checking that backups were explicitly enabled, but were checking for repo host volumes. Up until the recent change that I made, starting in 5.7 we always had a repo host in place if backups were enabled, so checking for repo host was essentially equivalent to checking that backups were enabled. However, now that a repo host is only created if a local volume repo is requested in the spec, backups can be enabled without any repo host volumes, so we must explicitly check for backups being enabled when adding the log rotate config. This is the change I made in the last commit.

internal/pgbackrest/util.go
dsessler7: I'm not sure of the answer, so I added a TODO comment to provide an explanation for the size.

PR #4206, opened 2025-07-14
Simplify development on macOS

cbandy:
Recent macOS changes require a newer pg_query.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

**Other Information**:

Issue: PGO-1046


build/postgres-operator/Dockerfile
```diff
@@ -1,15 +0,0 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal
```
benjaminjb, line 1: Is there anything else in this dir to save?

benjaminjb (approved):
What this looks like to me is
- your previous work on simplifying dev image builds (the Dockerfile, removing old build cruft) (none of which will interfere with our non-dev builds)
- pg_query for newer macs -- but does it work on our older macs still?

build/postgres-operator/Dockerfile
cbandy: This is/was the last file in the directory `build/postgres-operator`, so I don't understand the question, I think. What do you mean?

benjaminjb: Oh, yeah -- GH was saying this file was deleted, but I see on your branch the whole dir is gone, which is what I was hoping would be the case. Just wanted to check.

cbandy: 📝 iirc, Git doesn't track directories at all. When `git checkout` empties a directory, it also removes the directory. 🤔 Maybe `git rm` or `git mv`, too.

PR #4210, opened 2025-07-17
Attach additional volume for postgres,

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Only can add additional volumes to database container through abuse of the tablespace volumes.

Ability to add PVC (either BYO or create from template) to multiple containers in the Postgres pod

**Other Information**:
Issues: [PGO-2556]


internal/controller/postgrescluster/volumes.go
```diff
@@ -906,3 +906,64 @@ func getPVCName(volumes []*corev1.PersistentVolumeClaim, selector labels.Selecto
...
+// reconcileAdditionalVolumes writes the PersistentVolumeClaim for additional volumes.
+func (r *Reconciler) reconcileAdditionalVolumes(
+	ctx context.Context,
+	cluster *v1beta1.PostgresCluster,
+	additionalVolumes []*v1beta1.AdditionalVolume,
+	// instance *appsv1.StatefulSet,
+	// observed *Instance,
+) (additionalVolumePVCs []*corev1.PersistentVolumeClaim, err error) {
+
+	if additionalVolumes == nil || len(additionalVolumes) == 0 {
```
cbandy:
```suggestion
	if len(additionalVolumes) == 0 {
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -531,6 +531,48 @@ type PostgresVolumesSpec struct {
...
+
+	// These CRD xvalidation rules were over the complexity budget
+	// so I turned them off -- what's wrong here
+	// Additional volumes to add to the pod.
+	// ---
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=10
+	// // +kubebuilder:validation:items:XValidation:rule=`[has(self.claimName), has(self.claimTemplate)].filter(x, x == true).size() <= 1`,message=`can only have one of claimName, claimTemplate`
```
cbandy: ⛏️  Might be able to `filter(x, x)` rather than `x == true`. Won't help the cost of PVC template.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -531,6 +531,48 @@ type PostgresVolumesSpec struct {
...
+	Additional []*AdditionalVolume `json:"additional,omitempty"`
+}
+
+type AdditionalVolume struct {
+	// The name of the volume used for mounting path.
+	// Must be unique.
+	// ---
+	// +kubebuilder:validation:Required
+	// // +kubebuilder:validation:XValidation:rule=`self == oldSelf`,message="immutable"
+	Name string `json:"name,omitempty"`
```
cbandy: 🤔 I'm not sure what immutable means on a list-map key. When the key "changes," does that result in one key "removed" and one key "added?"

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -531,6 +531,48 @@ type PostgresVolumesSpec struct {
...
+	// A reference to a preexisting PVC.
+	// ---
+	// +optional
+	// // +kubebuilder:validation:XValidation:rule=`self == oldSelf`,message="immutable"
+	ClaimName string `json:"claimName,omitempty"`
+
+	// A PVC request.
+	// ---
+	// +optional
+	ClaimTemplate *VolumeClaimSpec `json:"claimTemplate,omitempty"`
```
cbandy: 🤔 I remember thinking some field in this section should be immutable, but why, I wonder? Changing which PVC `claimName` is mounted doesn't _destroy_ anything persistent.

Maybe I was thinking that switching _away_ from a template would be bad... In that case, a PVC _would_ be destroyed. That could be expressed on the section, maybe:

```
rule=`has(oldSelf.claimTemplate) != has(self.claimTemplate)`
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -531,6 +531,48 @@ type PostgresVolumesSpec struct {
...
+
+	// The containers to attach this volume to.
+	// ---
+	// +optional
+	// +listType=set
+	Containers []string `json:"containers,omitempty"`
```
cbandy, line 557: 🔧 `atomic` too, I think.

internal/naming/names.go
```diff
@@ -350,6 +350,21 @@ func InstancePostgresWALVolume(instance *appsv1.StatefulSet) metav1.ObjectMeta {
 	}
 }
 
+// AdditionalVolume returns the ObjectMeta for the additional
+// volume for instance.
+func AdditionalVolume(cluster *v1beta1.PostgresCluster,
+	volume *v1beta1.AdditionalVolume) metav1.ObjectMeta {
+	// TODO: What's the name for the PVC if not given?
```
cbandy: 🔧 The `name` field, I imagine. That field should have validation that rejects weird filepath stuff (e.g. slash) and is also valid for K8s API name (subdomain?). We have a few predefined types for names. https://github.com/CrunchyData/postgres-operator/blob/899c4a02f3eff606ae58d5c26ebb71877deae43e/internal/naming/limitations.md?plain=1#L80

https://github.com/CrunchyData/postgres-operator/blob/899c4a02f3eff606ae58d5c26ebb71877deae43e/pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go#L32-L33

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -531,6 +531,44 @@ type PostgresVolumesSpec struct {
...
+type AdditionalVolume struct {
+	// The name of the volume used for mounting path.
+	// Volumes are mounted in the pods at `volumes/<NAME>`
+	// Must be unique.
+	// ---
+	// The `Name` field is a `DNS1123Subdomain` type to enforce
+	// the max length and also allow us to more easily transition
+	// to CPK-provisioned volumes.
+	// +kubebuilder:validation:Required
+	Name DNS1123Subdomain `json:"name,omitempty"`
```
cbandy: ⛏️ Let's use the k8s upstream marker for this.
🔧 Drop the `omitempty` on required fields.


```suggestion
	// +required
	Name DNS1123Subdomain `json:"name"`
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -531,6 +531,44 @@ type PostgresVolumesSpec struct {
...
+	// ---
+	// +kubebuilder:validation:Required
+	ClaimName string `json:"claimName,omitempty"`
+
+	// The containers to attach this volume to.
+	// A blank/unset `Containers` field matches all containers.
+	// ---
+	// +optional
+	// +listType=set
+	// +listType=atomic
```
cbandy: 🤔 Hmm. These markers arrive in the same OpenAPI extension. I'm surprised one can't declare an "atomic set of unique values." If we can only have one, I prefer **atomic** for its effect on managed fields. Controllers can do `sets.New(s...)` easily enough.

- https://kubernetes.io/docs/reference/using-api/server-side-apply#merge-strategy
- There are some [open issues about the extension](https://github.com/kubernetes/kubernetes/issues?q=is%3Aissue%20state%3Aopen%20%22x-kubernetes-list-type%22).

🔧 We should set a max length here, too. Every new list and map should have a max.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -531,6 +531,44 @@ type PostgresVolumesSpec struct {
...
 	// +optional
 	Temp *VolumeClaimSpec `json:"temp,omitempty"`
+
+	// Additional pre-existing volumes to add to the pod.
+	// ---
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=10
+	Additional []*AdditionalVolume `json:"additional,omitempty"`
```
cbandy: 🤔 Is pointer helpful here? Do we want to be doing `nil` checks?


```suggestion
	Additional []AdditionalVolume `json:"additional,omitempty"`
```

pkg/apis/postgres-operator.crunchydata.com/v1/postgrescluster_types.go
```diff
@@ -534,6 +534,14 @@ type PostgresVolumesSpec struct {
...
 	// +optional
 	Temp *v1beta1.VolumeClaimSpec `json:"temp,omitempty"`
+
+	// Additional pre-existing volumes to add to the pod.
+	// ---
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=10
+	Additional []*v1beta1.AdditionalVolume `json:"additional,omitempty"`
```
cbandy: 📝 pointer

internal/controller/postgrescluster/util.go
```diff
@@ -285,3 +287,61 @@ func safeHash32(content func(w io.Writer) error) (string, error) {
...
+	additionalVolumes []*v1beta1.AdditionalVolume) {
+
+	for _, additionalVolumeRequest := range additionalVolumes {
+
+		additionalVolumeMount := AdditionalVolumeMount(
+			additionalVolumeRequest.Name,
+			additionalVolumeRequest.ClaimName)
+
+		additionalVolume := corev1.Volume{
+			Name: additionalVolumeMount.Name,
```
cbandy, line 317: 🤔 Ah, hmm. Volume names must be unique in a pod. What happens If someone uses `postgres-data` for their volume name? Maybe we should prefix with `volume-`?

🤔 🤔 What other (Pod) validation happens on this name in K8s? Maybe a limit on the length?

https://github.com/CrunchyData/postgres-operator/blob/6a30d727f64ab088c32cc520bbae0ce411fb4e56/internal/postgres/reconcile.go#L26-L28

internal/controller/postgrescluster/util.go
```diff
@@ -285,3 +287,61 @@ func safeHash32(content func(w io.Writer) error) (string, error) {
...
+			VolumeSource: corev1.VolumeSource{
+				PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+					// TODO: Using the claim name as the claim name here means this has to be unique
+					// We could add the index if we want someone to mount the same PVC in multiple dirs?
+					ClaimName: additionalVolumeMount.Name,
+					ReadOnly:  additionalVolumeRequest.ReadOnly,
```
cbandy: ❓ I'm getting a little mixed up with which name is where. Should this be `additionalVolumeRequest.claimName` ?

internal/controller/postgrescluster/util.go
```diff
@@ -285,3 +287,61 @@ func safeHash32(content func(w io.Writer) error) (string, error) {
...
+				},
+			},
+		}
+
+		for i := range template.Spec.Containers {
+			if len(additionalVolumeRequest.Containers) == 0 ||
+				slices.Contains(
+					additionalVolumeRequest.Containers,
+					template.Spec.Containers[i].Name) {
```
cbandy: 🔧 Rather than slices.Contains, we can use a [Set.Has](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set.Has) and share it with both loops.

```go
names := sets.New(additionalVolumeRequest.Containers...)

for ..Containers {
  if names.Len() == 0 || names.Has(..Name) { .... }
}

for ..InitContainers { .... }
```

internal/controller/postgrescluster/instance.go
```diff
@@ -1253,6 +1253,11 @@ func (r *Reconciler) reconcileInstance(
 		addDevSHM(&instance.Spec.Template)
 	}
 
+	// mount additional volumes to the Postgres instance containers
+	if err == nil && spec.Volumes != nil && spec.Volumes.Additional != nil {
```
cbandy: ⛏️ 🤷🏻 


```suggestion
	if err == nil && spec.Volumes != nil && len(spec.Volumes.Additional) > 0 {
```

internal/controller/postgrescluster/util.go
```diff
@@ -285,3 +287,61 @@ func safeHash32(content func(w io.Writer) error) (string, error) {
...
+// AdditionalVolumeMount returns the name and mount path of the additional volume.
+func AdditionalVolumeMount(volumeName, claimName string) corev1.VolumeMount {
+	return corev1.VolumeMount{
+		Name:      claimName,
+		MountPath: "/volumes" + "/" + volumeName}
+}
+
+// addAdditionalVolumesToSpecifiedContainers adds additional volumes to the specified
+// containers in the specified pod
+func addAdditionalVolumesToSpecifiedContainers(template *corev1.PodTemplateSpec,
```
cbandy, line 305: 👍🏻 I like that this takes a template.

internal/controller/postgrescluster/util.go
benjaminjb: For the pod.spec.volumes, its says "Must be a DNS_LABEL and unique within the pod." ([cite](https://github.com/kubernetes/api/blob/v0.33.3/core/v1/types.go#L38))

That's what the code comment says, at least.

+1 for a prefix that prevents other problems

internal/controller/postgrescluster/util.go
benjaminjb: This is a little confusing because I'm switching from using the derived Mount spec to our AdditionalVolume spec -- I think I can clean this up by (a) using the request name above, (b) creating the Mount spec with the ReadOnly field, and (c) sticking to the Mount spec here.

benjaminjb: Eh, tackling this and the comment re: naming the volumes, the code is less intuitive than I wanted

internal/controller/postgrescluster/util.go
```diff
@@ -285,3 +287,61 @@ func safeHash32(content func(w io.Writer) error) (string, error) {
...
+			if names.Len() == 0 || names.Has(template.Spec.InitContainers[i].Name) {
+				template.Spec.InitContainers[i].VolumeMounts = append(
+					template.Spec.InitContainers[i].VolumeMounts,
+					additionalVolumeMount)
+			}
+		}
+
+		template.Spec.Volumes = append(
+			template.Spec.Volumes,
+			additionalVolume)
```
dsessler7, line 364: This is probably fine, but just to clarify, it seems we will always add the volume even if we don't mount it to any containers (which would only happen if the user specifies a container(s) but none of the containers specified actually exist in the pod)...

Do we want to warn the user in any way if any of the containers they've specified don't exist? :thinking: Seems unlikely the user would do this on purpose, but if they misspell/typo their container name and don't realize it...

internal/controller/postgrescluster/util_test.go
```diff
@@ -378,3 +379,167 @@ func TestJobFailed(t *testing.T) {
 		})
 	}
 }
+
+func TestAddAdditionalVolumesToSpecifiedContainers(t *testing.T) {
```
dsessler7, line 383: maybe add a test case for when an init container is specified

testing/kuttl/e2e/additional-volumes/files/00-cluster-created.yaml
```diff
@@ -0,0 +1,15 @@
...
+  instances:
+    - name: instance1
+      readyReplicas: 1
+      replicas: 1
+      updatedReplicas: 1
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: additional-vols-primary
```
dsessler7: While the operator is obviously not responsible for creating the PVCs specified in `00-create-cluster.yaml`, shouldn't we still assert that they have indeed been created before moving to the next step in the kuttl test?

internal/controller/postgrescluster/util.go
```diff
@@ -285,3 +287,61 @@ func safeHash32(content func(w io.Writer) error) (string, error) {
...
+
+		for i := range template.Spec.InitContainers {
+			if names.Len() == 0 || names.Has(template.Spec.InitContainers[i].Name) {
+				template.Spec.InitContainers[i].VolumeMounts = append(
+					template.Spec.InitContainers[i].VolumeMounts,
+					additionalVolumeMount)
+			}
+		}
+
+		template.Spec.Volumes = append(
```
dsessler7: If the PVC specified by the user doesn't actually exist, what happens? I'm pretty sure that the Pod will break in some fashion... Do we want to provide any protection from this footgun? :thinking:

benjaminjb: If the PVC doesn't exist, the pod will not be scheduled and we get a warning event  from K8s:

```
5s          Warning   FailedScheduling        pod/hippo-instance1-f8kj-0                          0/1 nodes are available: persistentvolumeclaim "task-pv-claim" not found.
```

It might be nice to check if the PVCs exist, but we don't really do anything for those areas in our spec where we allow a user to specify a preexisting volume (or configmap, secret, etc.).

(I do kind of like the idea of checking the env before accepting a postgrescluster spec, but that's not really what we do.)

internal/controller/postgrescluster/util.go
benjaminjb: Yeah, it's funny, I was just talking with Andrew earlier in the week about how some people might want to _not_ add the volume to any containers through our system, but mount them manually through, say, the sidecar container spec.

But we don't actually let people mount to no containers: we have some (list the containers) and all (a blank list). Should we add a "none" option or special keyword?

But yeah, right now, we have (a) no way to skip mounting and (b) no validation on the containers being present. 

Now if I mispell "database" when trying to mount to the pod, what should our behavior be? I guess we should error out because, likely, if a user put in the wrong container name, and we just warned and created the pod, then they'd find out later, and fix the spec, causing the pod to regenerate. What do you think?

dsessler7: What would be the point of the user adding their PVC to `volumes.additional` if they don't want us to mount it to any containers?

Yeah I think doing something (error/warning event/log) makes sense if the user specifies a container that doesn't exist.

benjaminjb: Yeah, the only potential that I see -- though there might be others -- is that the user has a container definition that already has that volume mounted and it's easier for them (for some reason, potentially CI related) to use that as a custom sidecar (with the volume mounted) rather than define the postgrescluster spec with that volume mounted to that container.

benjaminjb: @andrewlecuyer This question isn't a blocker (as we can refine the solution as we move forwards) but I want to raise it:

If the user supplies a container that doesn't exist, what should our behavior be?

Options: 
1. Ignore that issue, let them figure it out. (Con: not very user friendly to not provide feedback)
2. Issue a warning event, but continue to reconcile the pod (Con: _if_ they see the warning, the user will have to fix the spec, which will cause a pod regeneration of a mis-spec'ed pod)
3. Error, stop reconciling pod, issue event

internal/controller/postgrescluster/util.go
cbandy: Don't check for things before-hand; e.g. Pods can be created before a Secret, before a ConfigMap, before a PVC.

Those things may/can/will come to be on their own schedule separately.

cbandy: Like every other API object, a Pod is a _request_ to run something. It will run when the stars align.

internal/controller/postgrescluster/util.go
cbandy: K8s does a lot of validation in Go. I search these "validation" files for details: https://github.com/kubernetes/kubernetes/blob/790393ae92e97262827d4f1fba24e8ae65bbada0/pkg/apis/core/validation/validation.go#L411

I see only `DNS1123Label` (and unique) here ☝🏻 . Oh! 🔧  "subdomain" is a set of dot-separated "labels," so the two aren't interchangeable.

🔧 Please make a new shared `type DNS1123Label = string` with ~these~ length and regexp validations:

https://github.com/kubernetes/kubernetes/blob/790393ae92e97262827d4f1fba24e8ae65bbada0/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L188

cbandy: ... and if we're prepending something to the volume name, then the max length on our volume name field is less than upstream. (e.g. 63 minus 8 or so)

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
```diff
@@ -32,6 +32,16 @@ type ConfigDataKey = string
 // +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?([.][a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
 type DNS1123Subdomain = string
 
+// ---
+// https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Label
+// https://pkg.go.dev/k8s.io/apiserver/pkg/cel/library#Format
+//
+// +kubebuilder:validation:MinLength=1
+// Max length is less than max 63 to allow prepending `volumes-` to name
+// +kubebuilder:validation:MaxLength=55
```
cbandy: ❓ Can this be 63 here and `AdditionalVolume.Name` override it with 55 there?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
```diff
@@ -32,6 +32,16 @@ type ConfigDataKey = string
 // +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?([.][a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
 type DNS1123Subdomain = string
 
+// ---
```
cbandy, line 35: ⛏️ When possible, I like linking to the user-doc-y stuff, too.

```suggestion
// ---
// https://docs.k8s.io/concepts/overview/working-with-objects/names#dns-label-names
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -531,6 +531,43 @@ type PostgresVolumesSpec struct {
...
+	// +required
+	Name DNS1123Label `json:"name"`
+
+	// A reference to a preexisting PVC.
+	// ---
+	// +required
+	ClaimName string `json:"claimName"`
+
+	// The containers to attach this volume to.
+	// A blank/unset `Containers` field matches all containers.
```
cbandy: ❓ Where did we land on empty-list? Does it mean all or none?

📝 I've used the phrase "when omitted" or "when this is omitted" to explain what happens when a field is missing from the spec.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
```diff
@@ -531,6 +531,43 @@ type PostgresVolumesSpec struct {
...
+	// ---
+	// The `Name` field is a `DNS1123Label` type to enforce
+	// the max length.
+	// +required
+	Name DNS1123Label `json:"name"`
+
+	// A reference to a preexisting PVC.
+	// ---
+	// +required
+	ClaimName string `json:"claimName"`
```
cbandy: 🔧 PVC names are 1123 subdomains, so we can use that type here.

https://github.com/kubernetes/kubernetes/blob/v1.33.0/pkg/apis/core/validation/validation.go#L2253 → `ValidatePersistentVolumeName` → `NameIsDNSSubdomain` → `IsDNS1123Subdomain`

internal/controller/postgrescluster/util.go
```diff
@@ -285,3 +287,77 @@ func safeHash32(content func(w io.Writer) error) (string, error) {
...
+
+		for i := range template.Spec.Containers {
+			if allContainers || names.Has(template.Spec.Containers[i].Name) {
+				template.Spec.Containers[i].VolumeMounts = append(
+					template.Spec.Containers[i].VolumeMounts,
+					additionalVolumeMount)
+
+				if names.Has(template.Spec.Containers[i].Name) {
+					names.Delete(template.Spec.Containers[i].Name)
+				}
```
cbandy: ⛏️ The guard is unnecessary.

```suggestion
				names.Delete(template.Spec.Containers[i].Name)
```

internal/controller/postgrescluster/util.go
```diff
@@ -285,3 +287,77 @@ func safeHash32(content func(w io.Writer) error) (string, error) {
...
+					template.Spec.InitContainers[i].VolumeMounts,
+					additionalVolumeMount)
+
+				if names.Has(template.Spec.InitContainers[i].Name) {
+					names.Delete(template.Spec.InitContainers[i].Name)
+				}
+			}
+		}
+
+		missingContainers = append(missingContainers, names.UnsortedList()...)
```
cbandy, line 360: 💭 Using `Delete` to subtract from `names` relies on init and regular containers having unique names together. That's correct and fine, but I wonder if it is worth calling out.

🤔 Should the resulting slice be a Set?

config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
cbandy, line 1: 🤩 This diff is so small and tidy without the template!

pkg/apis/postgres-operator.crunchydata.com/v1/postgrescluster_types.go
```diff
@@ -534,6 +534,14 @@ type PostgresVolumesSpec struct {
...
 	// +optional
 	Temp *v1beta1.VolumeClaimSpec `json:"temp,omitempty"`
+
+	// Additional pre-existing volumes to add to the pod.
+	// ---
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=10
+	Additional []v1beta1.AdditionalVolume `json:"additional,omitempty"`
```
cbandy, line 538: ⛏️ I like to alphabetize things. Can `Additional` go before/above `Temp`?

testing/kuttl/e2e/additional-volumes/01-assert.yaml
cbandy, line 1: 🤔 This doesn't seem like the kind of change that needs E2E. Is this here because we have more to add in the future? Are there are tests in the `postgrescluster` package more like cluster-spec-in → pod-spec-out ?

internal/controller/postgrescluster/util.go
dsessler7: I was about to raise this concern myself... Is it correct though, when you take sidecars into consideration? Couldn't the user add a sidecar to `Containers` that has the same name as one in `InitContainers`?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
dsessler7: I still don't quite understand the use-case for adding a volume that is not mounted into any containers... But I could see how a zero value (empty list) could mean "none" while leaving the field off entirely would mean "all"

internal/controller/postgrescluster/util.go
dsessler7: Answering my own question: No, the user could not add a sidecar with the same name as an init container. Ben tried it out and the Pod fails to roll out. He also found this in k8s documentation: “The name of each app and init container in a Pod must be unique; a validation error is thrown for any container sharing a name with another.”

internal/controller/postgrescluster/instance.go
```diff
@@ -1253,6 +1253,16 @@ func (r *Reconciler) reconcileInstance(
 		addDevSHM(&instance.Spec.Template)
 	}
 
+	// mount additional volumes to the Postgres instance containers
+	if err == nil && spec.Volumes != nil && len(spec.Volumes.Additional) > 0 {
+		missingContainers := addAdditionalVolumesToSpecifiedContainers(&instance.Spec.Template, spec.Volumes.Additional)
+
+		if len(missingContainers) > 0 {
+			r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "MissingContainers",
+				"The following containers were specified for additional volumes but are missing: %s.", missingContainers)
```
dsessler7: I think what you have is fine, but you expressed dissatisfaction with what you have, so maybe something like this?
```suggestion
			r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "SpecifiedContainerNotFound",
				"The following containers were specified for additional volumes but cannot be found: %s.", missingContainers)
```

internal/controller/postgrescluster/util.go
```diff
@@ -285,3 +287,77 @@ func safeHash32(content func(w io.Writer) error) (string, error) {
...
+				PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+					ClaimName: additionalVolumeRequest.ClaimName,
+					ReadOnly:  additionalVolumeMount.ReadOnly,
+				},
+			},
+		}
+
+		names := sets.New(additionalVolumeRequest.Containers...)
+		allContainers := false
+		if names.Len() == 0 {
```
dsessler7: IF we want omission of containers to mean "All" and an empty slice for containers to mean "None":
```suggestion
		if additionalVolumeRequest.Containers == nil {
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
benjaminjb: In the spec that results in name having the reqs:
```
                                allOf:
                                - maxLength: 63
                                - maxLength: 55
```

internal/controller/postgrescluster/util.go
benjaminjb: OK, for now we are emitting an event with the missing containers listed

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go
benjaminjb: I think the issue is still up for debate, to wit:

- what is the use care for attaching to none? (My squinting suggestion is: you have the volume defined in your sidecar definition and for whatever reason, it's easier to leave it there than to put it into our spec)
- what is the best way to signal none vs. all? I agree that an explicit "[]" looks more like none than leaving the field off, but that's a real thin line.

Ah, what the heck, let's do it.

testing/kuttl/e2e/additional-volumes/01-assert.yaml
benjaminjb: 👍 
This e2e test might've predated my unit testing for the util func, but that util test covers this so.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
cbandy: Interesting!

PR #4215, opened 2025-07-22
Move Postgres log rotation parameters to the postgres package

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

This moves the `log_filename` and `log_rotation` parameters into `internal/postgres`, documents how they're related, and adds some tests. 📝 The `internal/collector` tests don't change.

Issue: PGO-2558

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -75,6 +75,8 @@ linters:
               desc: Use "go.opentelemetry.io/otel/semconv/v1.27.0" instead.
             - pkg: io/ioutil
               desc: Use the "io" and "os" packages instead. See https://go.dev/doc/go1.16#ioutil
+            - pkg: math/rand$
+              desc: Use the "math/rand/v2" package instead. See https://go.dev/doc/go1.22#math_rand_v2
         not-tests:
           files: ['!$test','!**/internal/testing/**']
           list-mode: lax
@@ -139,6 +141,11 @@ linters:
       - legacy
       - std-error-handling
     rules:
+      # It is fine for tests to use "math/rand" packages.
+      - linters: [gosec]
+        path: '(.+)_test[.]go'
+        text: weak random number generator
+
       # This internal package is the one place we want to do API discovery.
       - linters: [depguard]
         path: internal/kubernetes/discovery.go
--- a/internal/postgres/config.go
+++ b/internal/postgres/config.go
@@ -7,9 +7,11 @@ package postgres
 import (
 	"context"
 	"fmt"
+	"math"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/crunchydata/postgres-operator/internal/config"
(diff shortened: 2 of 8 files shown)


internal/postgres/config.go
```diff
@@ -96,6 +98,71 @@ func LogDirectory() string {
 	return fmt.Sprintf("%s/logs/postgres", dataMountPath)
 }
 
+// LogRotation returns parameters that rotate log files while keeping a minimum amount.
+// Postgres truncates and reuses log files after that minimum amount.
+// Log file names start with filePrefix and end with fileSuffix.
+//
+// NOTE: These parameters do *not* enable logging to files.
```
cbandy: ⛏️ 

```suggestion
// NOTE: These parameters do *not* enable logging to files. Set "logging_collector" for that.
```

.golangci.yaml
```diff
@@ -75,6 +75,8 @@ linters:
               desc: Use "go.opentelemetry.io/otel/semconv/v1.27.0" instead.
             - pkg: io/ioutil
               desc: Use the "io" and "os" packages instead. See https://go.dev/doc/go1.16#ioutil
+            - pkg: math/rand$
+              desc: Use the "math/rand/v2" package instead. See https://go.dev/doc/go1.22#math_rand_v2
```
jmckulk, line 79: Any reason behind the switch to v2?

cbandy: No reason to stay _away_ from v1, really. The v2 `rand.N` is generic, so it can do `time.Duration` without casts; that was nice.

internal/postgres/config.go
```diff
@@ -96,6 +98,71 @@ func LogDirectory() string {
...
+//
+// NOTE: These parameters do *not* enable logging to files. Set "logging_collector" for that.
+func LogRotation(minimum metav1.Duration, filePrefix, fileSuffix string) map[string]string {
+	hours := math.Ceil(minimum.Hours())
+
+	// The "log_filename" parameter is interpreted similar to `strftime`;
+	// escape percent U+0025 by doubling it.
+	// - https://www.postgresql.org/docs/current/runtime-config-logging.html#GUC-LOG-FILENAME
+	prefix := strings.ReplaceAll(filePrefix, "%", "%%")
+	suffix := strings.ReplaceAll(fileSuffix, "%", "%%")
```
benjaminjb, line 113: ⛏️ we're doubling `%` to escape in case the prefix/suffix have `%` -- but ours don't, right?
and if someone forgot/missed that we're handling this and doubled the `%` originally, we'd quadruple the `%`?

benjaminjb (approved):

internal/postgres/config.go
cbandy: > ours don't, right?

Right. Escaping occurred to me as I was writing the test. I expect the `postgres` package to know about these details, while other packages don't.

> if someone forgot/missed that we're handling this and doubled the `%` originally, we'd quadruple the `%`?

Yep. If someone thinks to escape before calling, their test would show the wacky number of escapes.

PR #4220, opened 2025-07-25
Share PostgresCluster validation tests across API versions

cbandy:
As I was working a change to v1 validation, I wanted to include the v1beta1 tests so as not to introduce any regressions. I wasn't able to share tests using generics because the tests reach into specific struct fields. I was surprised to find a test failure when using unstructured, but the test was right!

So, unstructured seems like an improvement in validation testing overall:
1. Can quickly set uninitialized fields using `require.UnmarshalIntoField` or `unstructured.SetNestedField`
2. No struct tags that silently affect the payload

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement


internal/testing/validation/postgrescluster/postgres_authentication_test.go
```diff
@@ -0,0 +1,272 @@
...
+	})
+
+	t.Run("LDAP", func(t *testing.T) {
+		t.Run("Required", func(t *testing.T) {
+			cluster := base.DeepCopy()
+			require.UnmarshalIntoField(t, cluster, `{
+				rules: [
+					{ connection: hostssl, method: ldap },
+					{ connection: hostssl, method: ldap, options: {} },
+					{ connection: hostssl, method: ldap, options: { ldapbinddn: any } },
```
cbandy, line 165: The `options: {}` case is what passed with `v1beta1` and `v1` structs but failed with unstructured. I suspect the empty object was silently being omitted, causing this case to match the one above it.

config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
```diff
@@ -136,8 +136,8 @@ spec:
                           "ldapsearchattribute", or "ldapsearchfilter" options with
                           "ldapprefix" or "ldapsuffix" options
                         rule: has(self.hba) || self.method != "ldap" || !has(self.options)
-                          || [["ldapprefix","ldapsuffix"], ["ldapbasedn","ldapbinddn","ldapbindpasswd","ldapsearchattribute","ldapsearchfilter"]].exists_one(a,
-                          a.exists(k, k in self.options))
+                          || 2 > size([["ldapprefix","ldapsuffix"], ["ldapbasedn","ldapbinddn","ldapbindpasswd","ldapsearchattribute","ldapsearchfilter"]].filter(a,
```
cbandy, line 139: The `!has(self.option)` portion was being tested twice when using `v1beta1` and `v1` structs.

The unstructured value emitted `options: {}` which showed that `exists_one` is the wrong behavior:

- With `options: {}` the field "has options" and has _zero_ of ldapprefix, ldapbasedn, etc.
- The correct behavior is "has options" and "at most one" of ldapprefix, ldapbasedn, etc.

internal/testing/validation/postgrescluster/postgres_authentication_test.go
```diff
@@ -0,0 +1,272 @@
...
+	base.Name = "postgres-authentication-rules"
+
+	assert.NilError(t, cc.Create(ctx, base.DeepCopy(), client.DryRunAll),
+		"expected this base cluster to be valid")
+
+	var u unstructured.Unstructured
+	require.UnmarshalInto(t, &u, require.Value(yaml.Marshal(base)))
+	assert.Equal(t, u.GetAPIVersion(), "postgres-operator.crunchydata.com/v1beta1")
+
+	testPostgresAuthenticationCommon(t, cc, u)
```
cbandy, line 52: This is the pattern I used to share tests between v1beta1 and v1. Notice L82-84 of this same file.

internal/testing/validation/postgrescluster/postgres_authentication_test.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+func TestPostgresAuthenticationV1(t *testing.T) {
+	ctx := t.Context()
+	cc := require.KubernetesAtLeast(t, "1.30")
```
cbandy, line 57: These tests pass with v1 on K8s before 1.30, but the plan is for v1 to be 1.30+.

internal/testing/require/encoding_test.go
```diff
@@ -39,3 +43,41 @@ func TestUnmarshalInto(t *testing.T) {
...
+		require.UnmarshalIntoField(t, &u, `asdf`, "spec", "nested", "field")
+
+		if !reflect.DeepEqual(u.Object, map[string]any{
+			"spec": map[string]any{
+				"nested": map[string]any{
+					"field": "asdf",
+				},
+			},
+		}) {
+			t.Fatalf("got %[1]T(%#[1]v)", u.Object)
```
tjmoore4, line 63: :memo: I found these links helpful
- https://pkg.go.dev/fmt#hdr-Explicit_argument_indexes
- https://faun.pub/golangs-fmt-sprintf-and-printf-demystified-4adf6f9722a2

Also, for reference, a simple `%v` would yield
```
map[spec:map[nested:map[field:asdf]]]
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgres_types.go
```diff
@@ -141,12 +141,12 @@ type PostgresHBARule struct {
...
 // https://git.postgresql.org/gitweb/?p=postgresql.git;hb=refs/tags/REL_17_0;f=src/backend/libpq/hba.c#l1886
-// +kubebuilder:validation:XValidation:rule=`has(self.hba) || self.method != "ldap" || (has(self.options) && ["ldapbasedn","ldapprefix","ldapsuffix"].exists(k, k in self.options))`,message=`the "ldap" method requires an "ldapbasedn", "ldapprefix", or "ldapsuffix" option`
-// +kubebuilder:validation:XValidation:rule=`has(self.hba) || self.method != "ldap" || !has(self.options) || [["ldapprefix","ldapsuffix"], ["ldapbasedn","ldapbinddn","ldapbindpasswd","ldapsearchattribute","ldapsearchfilter"]].exists_one(a, a.exists(k, k in self.options))`,message=`cannot use "ldapbasedn", "ldapbinddn", "ldapbindpasswd", "ldapsearchattribute", or "ldapsearchfilter" options with "ldapprefix" or "ldapsuffix" options`
+// +kubebuilder:validation:XValidation:message=`the "ldap" method requires an "ldapbasedn", "ldapprefix", or "ldapsuffix" option`,rule=`has(self.hba) || self.method != "ldap" || (has(self.options) && ["ldapbasedn","ldapprefix","ldapsuffix"].exists(k, k in self.options))`
```
tjmoore4, line 144: :thinking: For this line and line 149, it looks like the only change is that `message` is being defined first. Line 145 has the the same change, plus the adjustment to the logic. I'm curious why the message order is being adjusted. It seems like we've been favoring `message` at the end in other places, right? (not a blocker in any case, mainly curiosity on my part).

cbandy: The message is usually shorter than the rule, and this order has helped me find the line I want.

I'm not certain, but having the English up front might be helping me grok the CEL that follows.

internal/testing/require/encoding_test.go
```diff
@@ -39,3 +43,41 @@ func TestUnmarshalInto(t *testing.T) {
...
+		if !reflect.DeepEqual(u.Object, map[string]any{
+			"one": int64(99),
+			"two": float64(5.7),
+		}) {
+			t.Fatalf("got %[1]T(%#[1]v)", u.Object)
+		}
+	})
+
+	// Correctly fails with: BUG: called without a destination
+	// require.UnmarshalIntoField(t, &u, `true`)
```
benjaminjb, line 82: Leftover?

internal/testing/validation/postgrescluster_test.go
```diff
@@ -6,515 +6,16 @@ package validation
...
-		cluster.Spec.Users = []v1beta1.PostgresUserSpec{
-			{Name: "normal", Options: "CREATEDB valid until '2006-01-02'"},
-			{Name: "very-full", Options: "NOSUPERUSER NOCREATEDB NOCREATEROLE NOINHERIT NOLOGIN NOREPLICATION NOBYPASSRLS CONNECTION LIMIT 5"},
-		}
-
-		assert.NilError(t, cc.Create(ctx, cluster, client.DryRunAll))
-	})
-}
-
 func TestPostgresUserInterfaceAcrossVersions(t *testing.T) {
```
benjaminjb, line 19: Would it be worth renaming this file in the future? I like the specificity of the new files' names.

PR #4221, opened 2025-07-25
Allow user to set an annotation that will specify an existing PVC to …

dsessler7:
…be mounted to cloud backup jobs so that the backup logs can be persisted.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

PGO-2578

Currently, the logs from backups that are being made to cloud repos are not persisted.

A user can now set an annotation, the value of which can point to a PVC that the cloud backups will use to persist logs to.

**Other Information**:


internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -873,6 +874,27 @@ func generateBackupJobSpecIntent(ctx context.Context, postgresCluster *v1beta1.P
...
+		if logVolumeName := postgresCluster.Annotations[naming.PGBackRestCloudLogVolume]; logVolumeName != "" {
+			logVolume := &corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      logVolumeName,
+					Namespace: postgresCluster.GetNamespace(),
+				},
+			}
+			err := errors.WithStack(r.Client.Get(ctx,
+				client.ObjectKeyFromObject(logVolume), logVolume))
+			if err != nil {
```
benjaminjb, line 890: Would we want to differentiate between a NotFound error vs. any other type of Error?

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -873,6 +874,27 @@ func generateBackupJobSpecIntent(ctx context.Context, postgresCluster *v1beta1.P
...
+		// annotation, check for the PVC. If we find it, mount it to the backup job.
+		// Otherwise, create a warning event.
+		if logVolumeName := postgresCluster.Annotations[naming.PGBackRestCloudLogVolume]; logVolumeName != "" {
+			logVolume := &corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      logVolumeName,
+					Namespace: postgresCluster.GetNamespace(),
+				},
+			}
+			err := errors.WithStack(r.Client.Get(ctx,
```
benjaminjb, line 888: What wold happen if we didn't check for the PVC, but just assumed it existed and it didn't? The job pod would get stuck at pending, right?

(I'm asking just to wonder what this code looks like without this change vs. what the user experience of that would be.)

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -2040,8 +2062,31 @@ func (r *Reconciler) reconcilePGBackRestConfig(ctx context.Context,
...
+	// log path. If the user has specified a PVC, but we can't find it, create a warning event.
+	cloudLogPath := ""
+	if logVolumeName := postgresCluster.Annotations[naming.PGBackRestCloudLogVolume]; logVolumeName != "" {
+		logVolume := &corev1.PersistentVolumeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      logVolumeName,
+				Namespace: postgresCluster.GetNamespace(),
+			},
+		}
+		err := errors.WithStack(r.Client.Get(ctx,
```
benjaminjb, line 2076: 🤔 If we check when generating the job spec, can we assume it exists here for generating the configmap?

internal/controller/postgrescluster/pgbackrest_test.go
```diff
@@ -2894,14 +2906,169 @@ volumes:
...
+- emptyDir:
+    sizeLimit: 16Mi
+  name: tmp
+- name: another-pvc
+  persistentVolumeClaim:
+    claimName: another-pvc
+		`))
+
+		// No new events
+		assert.Equal(t, len(recorder.Events), 1)
```
benjaminjb: This may be me being too persnickety, but this particular test case requires the previous test case to run, which makes me want this to be just one test case.

internal/pgbackrest/config.go
```diff
@@ -539,7 +539,12 @@ func populateCloudRepoConfigurationMap(
 		}
 	}
 
-	global.Set("log-level-file", "off")
+	// If we are given a log path, set it in the config. Otherwise, turn off logging to file.
+	if logPath != "" {
+		global.Set("log-path", logPath)
+	} else {
+		global.Set("log-level-file", "off")
```
benjaminjb, line 546: do we just copy the user's options on lines 549-551?

internal/util/volumes_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"gotest.tools/v3/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/crunchydata/postgres-operator/internal/testing/cmp"
+)
+
+func TestAddVolumeAndMountsToPod(t *testing.T) {
+	pod := &corev1.PodSpec{
+		Containers: []corev1.Container{
```
benjaminjb, line 20: 🔧 Might be worth throwing in 2+ initContainers to show that this func adds to those containers as well

internal/controller/postgrescluster/pgbackrest.go
dsessler7: Hmm... My thinking was that either way, the result was the same: we were unable to retrieve the PVC, so we should create a warning event. And the error message is passed into the event message, so the user will see what went wrong there.

internal/controller/postgrescluster/pgbackrest.go
dsessler7: In this spot it would definitely break the job because it would attempt to mount a volume whose source PVC could not be found.

internal/controller/postgrescluster/pgbackrest.go
dsessler7: So, as of right now, if the annotation is in place but the PVC doesn't exist, we still create the backup job, we just don't mount the PVC (and we create a warning event). My thought behind this is that having backups is more important that having logs from backups, so cancelling the backup job just because it can't write the logs seems like overkill. Open to arguments to the contrary though. This does make me wonder though if an event is enough of an alarm to users that something is wrong. Maybe we should log an error too? :thinking:  Either way, the backup will still be created. If we don't check for the PVC here where we reconcile the config, the backup job will attempt to write the logs to a directory that doesn't exist. I can't remember if it will fail or not (I'm guessing it won't fail for the same reason as earlier)... If it doesn't fail I guess we could remove the check for the PVC if we want... :thinking:

internal/controller/postgrescluster/pgbackrest_test.go
dsessler7: Good call. I'll adjust this.

internal/pgbackrest/config.go
dsessler7: yep

internal/util/volumes_test.go
dsessler7: Good call.

internal/controller/postgrescluster/pgbackrest.go
benjaminjb: Doy, I misread. So, going back to basics:

a) if there's an annotation and the PVC exists, create the job with the PVC, and add it to the config
b) if there's an annotation and no PVC, create the job without the logging and issue an event about the missing logging volume -- backups > logging
c) if there's no annotation, create the job

And in that case, any error finding the PVC falls into the category of "backups are more important than logging" (and if there's another underlying problem, like the API is unreachable, we'll hear about it later, e.g., when we try to create the job).

internal/controller/postgrescluster/pgbackrest.go
benjaminjb: OK, after reading all your comments, I understand the "backups are more important than logging" reasoning, and I'm +1 with that now.

internal/controller/postgrescluster/pgbackrest.go
benjaminjb: I do like to skip an API call if I can -- like, I wish we called the func to generate the jobspec before we generated the configmap, because then we could just pass the jobspec in and if we had the volume on the spec that matched the annotation on the cluster, we could assume the PVC was found.

But because we use the configmap for non-job stuff, that doesn't work. I guess we cold log into the ether if the volume wasn't there and skip a step -- but maybe another event will get through to the user. 

(I also don't know the best way to get through to the user for something that's a warning but not an error.)

benjaminjb (approved):

PR #4226, opened 2025-07-30
Set correct permissions on the specified log directory

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Bug fix

Controllers ignore the specified "log_directory" when preparing directories for Postgres. This parameter can be specified when the OpenTelemetryLogs gate is disabled.

Controllers create the logs directory before starting Postgres.

**Other Information**:

Issue: PGO-2558

### DIFF
--- a/internal/postgres/config.go
+++ b/internal/postgres/config.go
@@ -378,14 +378,17 @@ func startupCommand(
 	}
 
 	// Postgres creates "log_directory" but does *not* create any of its parent directories.
-	// Postgres omits the group-write S_IWGRP permission on the directory. Do both here while being
-	// careful to *not* touch "data_directory" contents until after `initdb` or Patroni bootstrap.
+	// Postgres omits group-write S_IWGRP permission when creating the directory.
+	//
+	// Do both here while being careful to *not* touch "data_directory" contents until after
+	// `initdb` or Patroni bootstrap; those abort unless "data_directory" is entirely empty.
 	if path.IsAbs(logDir) && !strings.HasPrefix(logDir, dataDir) {
 		mkdirs = append(mkdirs,
 			`(`+shell.MakeDirectories(dataMountPath, logDir)+`) ||`,
 			`halt "$(permissions `+shell.QuoteWord(logDir)+` ||:)"`,
 		)
 	} else {
+		// Postgres interprets "log_directory" relative to "data_directory" so do the same here.
 		mkdirs = append(mkdirs,
 			`[[ ! -f `+shell.QuoteWord(path.Join(dataDir, "PG_VERSION"))+` ]] ||`,
 			`(`+shell.MakeDirectories(dataDir, logDir)+`) ||`,
--- a/internal/postgres/config.go
+++ b/internal/postgres/config.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"path"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
@@ -121,13 +122,13 @@ func ConfigDirectory(cluster *v1beta1.PostgresCluster) string {
 // DataDirectory returns the absolute path to the "data_directory" of cluster.
 // - https://www.postgresql.org/docs/current/runtime-config-file-locations.html
 func DataDirectory(cluster *v1beta1.PostgresCluster) string {
-	return fmt.Sprintf("%s/pg%d", dataMountPath, cluster.Spec.PostgresVersion)
+	return fmt.Sprintf("%s/pg%d", DataStorage(cluster), cluster.Spec.PostgresVersion)
 }
 
-// LogDirectory returns the absolute path to the "log_directory" of cluster.
-// - https://www.postgresql.org/docs/current/runtime-config-logging.html
-func LogDirectory() string {
-	return fmt.Sprintf("%s/logs/postgres", dataMountPath)
+// DataStorage returns the absolute path to the disk where cluster stores its data.
+// Use [DataDirectory] for the exact directory that Postgres uses.
+func DataStorage(cluster *v1beta1.PostgresCluster) string {
+	return dataMountPath
 }
 
 // LogRotation returns parameters that rotate log files while keeping a minimum amount.
@@ -354,12 +355,16 @@ done
 // PostgreSQL.
 func startupCommand(
 	ctx context.Context,
-	cluster *v1beta1.PostgresCluster, instance *v1beta1.PostgresInstanceSetSpec,
+	cluster *v1beta1.PostgresCluster,
+	instance *v1beta1.PostgresInstanceSetSpec,
+	parameters *ParameterSet,
 ) []string {
 	version := fmt.Sprint(cluster.Spec.PostgresVersion)
+	dataDir := DataDirectory(cluster)
+	logDir := parameters.Value("log_directory")
 	walDir := WALDirectory(cluster, instance)
 
-	mkdirs := make([]string, 0, 7+len(instance.TablespaceVolumes))
+	mkdirs := make([]string, 0, 9+len(instance.TablespaceVolumes))
 	mkdirs = append(mkdirs, `dataDirectory "${postgres_data_directory}" || halt "$(permissions "${postgres_data_directory}" ||:)"`)
 
 	// If the user requests tablespaces, we want to make sure the directories exist with the correct owner and permissions.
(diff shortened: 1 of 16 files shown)


internal/postgres/config.go
```diff
@@ -89,13 +89,13 @@ func ConfigDirectory(cluster *v1beta1.PostgresCluster) string {
...
+	return fmt.Sprintf("%s/pg%d", DataStorage(cluster), cluster.Spec.PostgresVersion)
 }
 
-// LogDirectory returns the absolute path to the "log_directory" of cluster.
-// - https://www.postgresql.org/docs/current/runtime-config-logging.html
-func LogDirectory() string {
-	return fmt.Sprintf("%s/logs/postgres", dataMountPath)
+// DataStorage returns the absolute path to the disk where cluster stores its data.
+// Use [DataDirectory] for the exact directory that Postgres uses.
+func DataStorage(cluster *v1beta1.PostgresCluster) string {
```
benjaminjb, line 130: Why pass in the cluster?

cbandy: Hmm. 😬 Future-proofing, I guess. Something about being a constant does sit right with me, but it absolutely is a constant.

If I remove the arguments, should I also remove the function? Export the constant?

cbandy: 🤔 I think I want things outside this package to _not_ treat it as a constant, even though it is constant right now. Does that English make sense?

internal/postgres/config.go
```diff
@@ -446,8 +449,9 @@ chmod +x /tmp/pg_rewind_tde.sh
 		`halt "$(permissions ` + naming.PGBackRestPGDataLogPath + ` ||:)"`,
 		`(` + shell.MakeDirectories(dataMountPath, naming.PatroniPGDataLogPath) + `) ||`,
 		`halt "$(permissions ` + naming.PatroniPGDataLogPath + ` ||:)"`,
-		`(` + shell.MakeDirectories(dataMountPath, LogDirectory()) + `) ||`,
-		`halt "$(permissions ` + LogDirectory() + ` ||:)"`,
+		`(` + shell.MakeDirectories(DataDirectory(cluster), logDir) + `) ||`,
+		// FIXME: This error prints the wrong directory when logDir is relative (the default).
+		`halt "$(permissions ` + logDir + ` ||:)"`,
```
cbandy: The challenge here is that `initdb` is only happy populating an empty data directory.

It is happy with this:

```
drwxrwxr-x. 2 postgres postgres Jul 31 17:04 pgdata/pg16
```

It is not happy with this:

```
drwxrwxr-x. 3 postgres postgres Jul 31 17:04 pgdata/pg16
drwxrwxr-x. 2 postgres postgres Jul 31 17:04 pgdata/pg16/log
```

benjaminjb: Is now a time to pull the thread on moving logs out of pgdata completely?

cbandy: Very much "almost." I'll think about this tomorrow.

internal/postgres/config.go
```diff
@@ -322,52 +355,53 @@ done
...
-		// and so we add the subdirectory `data` in order to set the permissions.
-		checkInstallRecreateCmd := strings.Join([]string{
-			`if [[ ! -e "${tablespace_dir}" || -O "${tablespace_dir}" ]]; then`,
-			`install --directory --mode=0750 "${tablespace_dir}"`,
-			`elif [[ -w "${tablespace_dir}" && -g "${tablespace_dir}" ]]; then`,
-			`recreate "${tablespace_dir}" '0750'`,
-			`else (halt Permissions!); fi ||`,
-			`halt "$(permissions "${tablespace_dir}" ||:)"`,
-		}, "\n")
+	mkdirs := make([]string, 0, 9+len(instance.TablespaceVolumes))
```
benjaminjb, line 367: I'm missing something -- why 9+? 

I see: 
- postgres_data_directory
- tablespaces (the + in that make)
- logdir
- PatroniPGDataLogPath
- PGBackRestPGDataLogPath

internal/postgres/config.go
```diff
@@ -322,52 +355,53 @@ done
...
-			tablespaceCmd = tablespaceCmd + "\ntablespace_dir=/tablespaces/" + tablespace.Name + "/data" + "\n" +
-				checkInstallRecreateCmd
+			dir := shell.QuoteWord("/tablespaces/" + tablespace.Name + "/data")
+			mkdirs = append(mkdirs, `dataDirectory `+dir+` || halt "$(permissions `+dir+` ||:)"`)
 		}
 	}
 
+	// Postgres creates "log_directory" but does *not* create any of its parent directories.
+	// Postgres omits the group-write S_IWGRP permission on the directory. Do both here while being
+	// careful to *not* touch "data_directory" contents until after `initdb` or Patroni bootstrap.
```
benjaminjb:
```suggestion
	// careful to *not* touch "data_directory" contents until after `initdb` or Patroni bootstrap
	// because `initdb`/Patroni bootstrap will error out if the data directory is not empty
```
?

internal/postgres/config.go
```diff
@@ -322,52 +355,53 @@ done
...
+	if path.IsAbs(logDir) && !strings.HasPrefix(logDir, dataDir) {
+		mkdirs = append(mkdirs,
+			`(`+shell.MakeDirectories(dataMountPath, logDir)+`) ||`,
```
benjaminjb, line 387:
```suggestion
	if path.IsAbs(logDir) && !strings.HasPrefix(logDir, dataDir) {
		mkdirs = append(mkdirs,
			`(`+shell.MakeDirectories(dataDir, logDir)+`) ||`,
```
? `dataDir` is `dataMountPath` right now, yeah?

internal/postgres/config.go
```diff
@@ -322,52 +355,53 @@ done
...
+			`halt "$(permissions `+shell.QuoteWord(logDir)+` ||:)"`,
+		)
+	} else {
+		mkdirs = append(mkdirs,
+			`[[ ! -f `+shell.QuoteWord(path.Join(dataDir, "PG_VERSION"))+` ]] ||`,
+			`(`+shell.MakeDirectories(dataDir, logDir)+`) ||`,
+			`halt "$(permissions `+shell.QuoteWord(path.Join(dataDir, logDir))+` ||:)"`,
```
benjaminjb: I might need you to walk me through this, but 

ll 383-385: if the logdir is absolute and isn't prefixed with datadir, we MakeDirectories with dataMountPath as the base -- but since the logdir abs path won't have the base as part of the path, MakeDirectories kinda ignores the base path/the output will not have any reference to that.

That is: if logdir is outside the datadir, create it with the correct permissions. (Passing "dataDir" into MakeDirectories was throwing me.)

ll 389-392: 
- we have reached this because either the logdir path is relative or includes the datadir
- if the PG_VERSION file exists (and therefore initdb or similar has run?), we can then create our logdir without worrying about it interfering with initdb

Am I missing anything?

So if the version doesn't exist, we haven't run initdb/bootstrap; and then we'll run that at some time, and then re-run this startup command at some time?

internal/postgres/config.go
cbandy: I feel weird when doing this. I have internalize "if you know the size before appending, use make" but I'm not finding a source for that.

I know that `append` one at a time doubles capacity, so from 0 → 9 is five allocations... But who cares, right? gah.

internal/postgres/config.go
cbandy: > ll 383-385: if the logdir is absolute and isn't prefixed with datadir, we MakeDirectories with dataMountPath as the base -- but since the logdir abs path won't have the base as part of the path, MakeDirectories kinda ignores the base path/the output will not have any reference to that.

I'm not following the back half of this. What does `base as part of the path` mean?

We don't pay attention to which volume the directory is on, true. If the absolute path is `/volumes` then it does `mkdir -p` and `chmod` of one directory.

> That is: if logdir is outside the datadir, create it with the correct permissions. (Passing "dataDir" into MakeDirectories was throwing me.)

These lines pass `dataMountPath`.

> Am I missing anything?
> 
> So if the version doesn't exist, we haven't run initdb/bootstrap; and then we'll run that at some time, and then re-run this startup command at some time?

That's right, and `at some time` facts bug me, but all I've got is startup and bash atm.

internal/postgres/config.go
cbandy: No, `dataDir == "/pgdata/pg16"` and `dataMountPath == "/pgdata"`.

benjaminjb: ah, whoops, misremembering DataStorage

internal/postgres/config.go
benjaminjb: Ah, yeah, some of what I said was wrong, partly because I was mixing up `/pgdata` and `/pgdata/pg##`

So:
- if logdir is not in datadir, we create the paths for logdir. And that's true where the log dir is in the pgdata dir or elsewhere. (The real distinction is that the logdir cannot be in `/pgdata/pg##`.)
- and if the logdir is in datadir... yeah, "at some time" is what we've got right now.

benjaminjb (approved):

PR #4228, opened 2025-08-01
Move pgbackrest-restore test to Kyverno Chainsaw

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

**Other Information**:

This test failed in this [other PR](https://github.com/CrunchyData/postgres-operator/pull/4226), and the KUTTL logs did not give me enough information to diagnose the cause. Chainsaw did much better and unblocked me over there.


testing/chainsaw/e2e/pgbackrest-restore/chainsaw-test.yaml
```diff
@@ -0,0 +1,105 @@
...
+            ref:
+              apiVersion: postgres-operator.crunchydata.com/v1beta1
+              kind: PostgresCluster
+              name: clone-two
+
+    - name: 'Lose Data'
+      use: { template: '21-lose-data.yaml' }
+
+    - name: 'Point-In-Time Restore'
+      use: { template: '22-point-in-time-restore.yaml' }
```
benjaminjb: Is it worth it to regularize the file/template names? I always liked the numbered files because you can just read through a directory (well, could, before we started to use the subdir `files`), but if Chainsaw uses this file to sequence the tests, maybe another convention is worth talking about (at a later date!)

.github/workflows/test.yaml
```diff
@@ -144,10 +144,12 @@ jobs:
             --env 'RELATED_IMAGE_COLLECTOR=registry.developers.crunchydata.com/crunchydata/postgres-operator:ubi9-5.8.2-0' \
             --env 'PGO_FEATURE_GATES=TablespaceVolumes=true,OpenTelemetryLogs=true,OpenTelemetryMetrics=true' \
             --name 'postgres-operator' localhost/postgres-operator
-      - name: Install kuttl
-        run: |
-          curl -Lo /usr/local/bin/kubectl-kuttl https://github.com/kudobuilder/kuttl/releases/download/v0.13.0/kubectl-kuttl_0.13.0_linux_x86_64
-          chmod +x /usr/local/bin/kubectl-kuttl
+
```
benjaminjb, line 147: We don't need to install KUTTL any longer?

testing/chainsaw/e2e/pgbackrest-restore/chainsaw-test.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+    - name: 'Verify Primary'
+      description: >
+        Confirm that data was restored to the point-in-time and the cluster is healthy
+      use: { template: '23-verify-primary.yaml' }
+
+    - name: 'Confirm Replica'
```
benjaminjb, line 197: Again: not for today, but I wonder if we could templatize a verify step and reuse it in our 4 verify steps. Note: this is the longest sustained look I've had at chainsaw

testing/chainsaw/e2e/pgbackrest-restore/change-parameters.yaml
```diff
@@ -0,0 +1,62 @@
+apiVersion: chainsaw.kyverno.io/v1alpha1
```
benjaminjb: Is this file getting called?

testing/chainsaw/e2e/pgbackrest-restore/chainsaw-test.yaml
cbandy: Definitely. I think I started with numbers close to the KUTTL ones to ease review/migration, but where I landed is not that!

I like the filenames going in a similar order as the test, too, but now they can get reused so... 🤷🏻 

Maybe something more like "scenario one," "... two," 🤔 🌱

testing/chainsaw/e2e/pgbackrest-restore/change-parameters.yaml
cbandy: It is not. I used it while diagnosing a bug about parameters in the WAL stream a long time ago.

I called this multiple times from the main test to... 🤔 raise and lower `max_connections` maybe. https://github.com/CrunchyData/postgres-operator/commit/44b1847544e099cd825aef64e962855c54916dcc

cbandy: 🔧 If the file stays, it deserves a comment explaining what it does and why it isn't currently used.

cbandy: This is looking a lot like 11 ...

testing/chainsaw/e2e/pgbackrest-restore/chainsaw-test.yaml
cbandy: Hmm. I thought I _was_ reusing verify 02 and 12, but i see each only once.

Should those be like this?
1. Create Data [A] on [one]
2. Verify Data [A] on [one] <-- new
3. Clone [one] to [two]
4. Verify Data [A] on [two]

.github/workflows/test.yaml
cbandy: The Make targets do `go run ...@latest` by default. The binary download is faster, but not as easy to `latest`.

I'm on the fence. Do you have a preference?

testing/chainsaw/e2e/pgbackrest-restore/chainsaw-test.yaml
benjaminjb: Yeah, if the templates are being reused (hopefully!), then putting them in order doesn't make sense. Maybe descriptive titles could make it easier to read?

.github/workflows/test.yaml
benjaminjb: Faster by how much? My guess is not that much, so this seems reasonable.

testing/chainsaw/e2e/pgbackrest-restore/change-parameters.yaml
benjaminjb: Yeah

testing/chainsaw/e2e/pgbackrest-restore/README.md
```diff
@@ -0,0 +1,5 @@
+# pgbackrest-restore
+
+This [chainsaw](https://github.com/kyverno/chainsaw) suite tests that CPK can clone and restore through pgBackRest backups.
+
+...other material here...
```
benjaminjb: What else should we put here to explain this to ourselves when we come back to this?

sfc-gh-jmckulka: Is there anything specific to these tests that wouldn't go in general chainsaw docs? 🤔

testing/chainsaw/e2e/config.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: chainsaw.kyverno.io/v1alpha2
+kind: Configuration
+metadata:
+  name: end-to-end
```
sfc-gh-jmckulka, line 4: Do we need this file? 

I think the info in this file, mainly labels and timeouts, can be defined per test 🤔

testing/chainsaw/e2e/pgbackrest-restore/templates/clone-cluster.yaml
```diff
@@ -0,0 +1,51 @@
...
+kind: StepTemplate
+metadata:
+  name: clone-cluster
+spec:
+  bindings:
+    - name: name
+      value: 'The name of the new PostgresCluster'
+
+  try:
+    -
```
sfc-gh-jmckulka, line 11: What are we doing here? Why is there a newline after this `-` instead of `- description: >` on a single line 😕

sfc-gh-jmckulka: I see that this is a template. Is it related to that?

.github/workflows/test.yaml
```diff
@@ -144,10 +144,12 @@ jobs:
...
+      - run: |
```
sfc-gh-jmckulka, line 148: This step is still named kuttl-k3d... should we make the name more generic?

.github/workflows/test.yaml
benjaminjb: Checking the run times of the `Run make check-kuttl && exit` between this PR and another (with the older style), I see 1 worst case I wouldn't maybe want (abt 2 mins difference), but mostly I see 0-15 secs difference (and with these jobs, not sure where the time is really coming from).

testing/chainsaw/e2e/pgbackrest-restore/README.md
benjaminjb: I can't think of anything right now and the original test didn't have anything, so I'll leave this for now as a placeholder.

testing/chainsaw/e2e/config.yaml
benjaminjb: Yeah, and as needed, I think we might make that change. A lot of this will depend on the second test that gets added/what further development we do to add chainsaw tests.

testing/chainsaw/e2e/pgbackrest-restore/templates/clone-cluster.yaml
benjaminjb: Not sure why Chris started with this pattern, but it goes through all the files.

cbandy: It works around a bug in the GitHub YAML rendering: https://www.github.com/atom/language-yaml/issues/109#issuecomment-645831505

PR #4239, opened 2025-08-13
Refactor Volume Auto Grow to support additional volume types

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

This commit refactors the existing pgData volume auto grow code to better support upcoming feature enhancements relating to auto grow capability for the pgBackRest repository volume and pg_wal volume.

**Other Information**:
Issue: PGO-2606


internal/controller/postgrescluster/util.go
```diff
@@ -285,3 +290,188 @@ func safeHash32(content func(w io.Writer) error) (string, error) {
...
+	}
+}
+
+// getDesiredVolumeSize compares the volume request size to the suggested autogrow
+// size stored in the status and updates the value when the status value is larger.
+func getDesiredVolumeSize(cluster *v1beta1.PostgresCluster,
+	volumeType, instanceSpecName string,
+	volumeRequestSize *resource.Quantity) (string, error) {
+
+	switch volumeType {
```
tjmoore4: :memo: This switch statement will likely be very similar to the one used by `limitIsSet`. The difference is that this function references status objects that don't yet exist whereas `limitIsSet` required nothing new outside of itself, so it made sense to move forward a bit there but not here.

internal/controller/postgrescluster/util.go
tjmoore4, line 1: :thinking: Should the 'auto grow' functions have their own file? I went back and forth on this, but leaned slightly towards just using `util`. Thoughts on this are welcome.

jmckulk: I would prefer having them in their own file. Auto Grow is a major feature and i think it would help navigate the code base if i could see a file specific to auto grow

internal/controller/postgrescluster/util.go
```diff
@@ -285,3 +290,188 @@ func safeHash32(content func(w io.Writer) error) (string, error) {
 	}
 	return rand.SafeEncodeString(fmt.Sprint(hash.Sum32())), nil
 }
+
+// storeDesiredRequest saves the appropriate request value to the PostgresCluster
+// status. If the value has grown, create an Event.
+func (r *Reconciler) storeDesiredRequest(
+	ctx context.Context, cluster *v1beta1.PostgresCluster,
+	volumeType, instanceSetName, desiredRequest, desiredRequestBackup string,
```
jmckulk: Would these be accurate descriptions of these variables? Not necessarily a suggestion... just checking my understanding
```suggestion
	volumeType, instanceSetName, currentRequestString, previousRequestString string,
```

internal/controller/postgrescluster/util.go
```diff
@@ -285,3 +289,172 @@ func safeHash32(content func(w io.Writer) error) (string, error) {
...
+
+	log := logging.FromContext(ctx)
+
+	// Store the limit for this instance set. This value will not change below.
+	volumeLimitFromSpec := pvc.Spec.Resources.Limits.Storage()
+
+	// Capture the largest volume size currently defined for a given instance set.
+	// This value will capture our desired update.
+	volumeRequestSize := pvc.Spec.Resources.Requests.Storage()
+
```
jmckulk: The `volumeLimitFromSpec.IsZero()` check is in both the if/else blocks. We could refactor it out of to simplify the checks
```suggestion
# A limit of 0 is ignored.
if volumeLimitFromSpec.IsZero() {
  return
}
if volumeReqSize.Value() > volumeLimitFromSpec.Value() {
...
} else if feature.Enabled(...) {
...
}
```
i think this would be easier to follow ⛏️

internal/controller/postgrescluster/util.go
tjmoore4: Yeah, that's right. The reason it's referred to as a 'backup' here is that that would be how it would be used. In case there's a parsing issue, etc, this value gives us something to roll back to.

jmckulk: why do we need something to roll back to? wouldn't we typically bail if we hit an issue then assume we the change would pass on the next reconcile?

internal/controller/postgrescluster/autogrow.go
```diff
@@ -0,0 +1,189 @@
...
+// a corresponding boolean value
+func limitIsSet(cluster *v1beta1.PostgresCluster, volumeType, instanceSetName string) bool {
+
+	var limitSet bool
+
+	switch volumeType {
+
+	// Cycle through the instance sets to ensure the correct limit is identified.
+	case "pgData":
+		for _, specInstance := range cluster.Spec.InstanceSets {
```
jmckulk, line 84: if we passed around `instance` instead of `instanceSetName` would that mean we didn't need to loop over each instance ❓ 💭

jmckulk: i wonder if we could make this even more generic. We would use this same function for a pgbackrest repo 🤔 and the repo wouldn't be tied to an instance... right?

jmckulk: Honestly most of these functions pass around some reference to the instance. Does that break when we start thinking about pgbackrest repo ❓

internal/controller/postgrescluster/util.go
tjmoore4: It is also used for comparison purposes by the function, so it's doing a couple of different things. Part of what's happening is we're going from a string to a parsed value, so that has the potential to cause issues. In terms of naming, really 'previous' or 'backup' are both accurate. I guess the difference is a backup is always previous but previous isn't always a backup? Maybe? :man_shrugging:

internal/controller/postgrescluster/autogrow.go
tjmoore4: This function is ultimately being triggered by `observeInstances`, so we don't really have a good way to access the instance object itself if I remember correctly.

We'll need to know the volume type, etc so that we can drill down to the correct portion of the cluster spec, so I don't know if we can simplify this much further.

I think a lot of the logic around ensuring the various instance volumes grow in tandem won't be needed for the repo volume work. There can be more than one instance set and each instance set can have more than one replica. All the volumes for a given instance should grow the same, but each instance set can be different (and have different limits). Repo volumes on the other hand should be (assuming I'm not forgetting anything) 1:1 per repo definition. All that to say, I don't know that it will be quite as complicated even though it might require more code.

For this particular function, I was expecting to just pass in an entry string for the `instanceSetName` and key off of the `volumeType` where that will be `repo1`, `repo2`, etc. That said, we'll have to see how things work out. It might make sense to do things a different way, but I can't say for certain at this point.

internal/util/volumes.go
jmckulk, line 1: would this be in the autgrow file 🤔

internal/postgres/config.go
```diff
@@ -299,9 +284,18 @@ done
...
+	// auto grow annotation scripts
+	finalDone := `done
+`
+
 	// Elide the above script from `ps` and `top` by wrapping it in a function
 	// and calling that.
-	wrapper := `monitor() {` + script + `}; export -f monitor; exec -a "$0" bash -ceu monitor`
+	wrapper := `monitor() {` +
+		script + // main script body
+		util.AutogrowAnnotationScript("pgdata") + // pgdata annotation script
```
jmckulk: what do you think about making this a bash function instead of go?

something like

```bash
check_resize() {
  directory=$1

  // do stuff with $directory
  ...
}

// original cert rotation stuff

check_resize "pgdata"

if [ -d /pgwal ]; then
  check_resize "pgwal"
fi
```

internal/controller/postgrescluster/instance.go
```diff
@@ -360,7 +359,7 @@ func (r *Reconciler) observeInstances(
 		// If autogrow is enabled, get the desired volume size for each instance.
 		if autogrow {
 			for _, instance := range observed.bySet[name] {
-				status.DesiredPGDataVolume[instance.Name] = r.storeDesiredRequest(ctx, cluster,
+				status.DesiredPGDataVolume[instance.Name] = r.storeDesiredRequest(ctx, cluster, "pgData",
 					name, status.DesiredPGDataVolume[instance.Name], previousDesiredRequests[instance.Name])
```
jmckulk, line 366: related to https://github.com/CrunchyData/postgres-operator/pull/4239/files#r2274537730

if we passed instance in here instead of name would we have access to everything we need... ❓ 
```suggestion
				status.DesiredPGDataVolume[instance.Name] = r.storeDesiredRequest(ctx, cluster, "pgData",
					instance, status.DesiredPGDataVolume[instance.Name], previousDesiredRequests[instance.Name])
```

internal/util/volumes.go
tjmoore4: That's used by `config.go` in the `postgres` package. If we moved it to the `autogrow.go` file, we'd have to import `postgrescluster` which I don't think we want to do.

internal/postgres/config.go
tjmoore4: That might be a bit easier to understand, plus it will avoid changes in another package. I'll see what I can get working.

internal/controller/postgrescluster/autogrow.go
```diff
@@ -0,0 +1,189 @@
...
+func (r *Reconciler) setVolumeSize(ctx context.Context, cluster *v1beta1.PostgresCluster,
+	pvc *corev1.PersistentVolumeClaim, volumeType, instanceSpecName string) {
+
+	log := logging.FromContext(ctx)
+
+	// Store the limit for this instance set. This value will not change below.
+	volumeLimitFromSpec := pvc.Spec.Resources.Limits.Storage()
+
+	// Capture the largest volume size currently defined for a given instance set.
+	// This value will capture our desired update.
```
jmckulk: i'm not following this comment. We are just grabbing the Request from the pvc... how does this translate to the "largest volume defined for a set" ❓

internal/controller/postgrescluster/autogrow.go
```diff
@@ -0,0 +1,189 @@
...
+	// a warning event.
+	if volumeRequestSize.Value() > volumeLimitFromSpec.Value() {
+		r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "VolumeRequestOverLimit",
+			"%s volume request (%v) for %s/%s is greater than set limit (%v). Limit value will be used.",
+			volumeType, volumeRequestSize, cluster.Name, instanceSpecName, volumeLimitFromSpec)
+
+		pvc.Spec.Resources.Requests = corev1.ResourceList{
+			corev1.ResourceStorage: *resource.NewQuantity(volumeLimitFromSpec.Value(), resource.BinarySI),
+		}
+		// Otherwise, if the limit is not set or the feature gate is not enabled, do not autogrow.
```
jmckulk: you moved the "limit not set logic" right? I think this comment needs to be updated?

internal/controller/postgrescluster/autogrow.go
tjmoore4: I think this is a leftover comment from a previous refactor (from when the feature was first developed). I'll remove it.

internal/controller/postgrescluster/autogrow.go
tjmoore4: Yep, I'll fix that, thanks.

internal/controller/postgrescluster/autogrow.go
```diff
@@ -0,0 +1,189 @@
...
+	} else if feature.Enabled(ctx, feature.AutoGrowVolumes) {
+
+		// determine the appropriate volume request based on what's set in the status
+		if dpv, err := getDesiredVolumeSize(
```
jmckulk, line 128: what do we think the "P" in `dpv` means?

jmckulk: I'm also not seeing where `dpv` is used in the rest of this function besides when we log and error.
am i missing something 🔧

jmckulk: Oh is see, `volumeRequestSize` is a pointer that is passed in and updated. 

it took me a minute to get here - maybe its worth changing something here ♻️ 

- if we aren't using the return value maybe we don't return it
- or maybe return the value instead of updating the pointer

jmckulk: thinking about the logic in this function

```
	if volumeLimitFromSpec.IsZero() { return }
    if request > limit { // use limit + warn }
    if autogrow {
        dpv = getDesired // this changes value of request
        if requet >= limit { // use limit + warn }
    }
```

if limit isn't set, autogrow isn't enabled so do nothing
if request has exceeded the limit - use the limit
if autogrow is enabled we see if the request is different 
  use getDesiredVolumeSize to override the request
  if value has met or exceeded the limit - use the limit

maybe...

```
	if volumeLimitFromSpec.IsZero() {
		return
	}

    request =  pvc.Spec.Resources.Requests.Storage()
    if feature.Enabled(ctx, feature.AutoGrowVolumes) {
        request = getDesiredVolumeSize()
    }

    if request >= limit {
       // use limit and warn
    }
```

this is assuming we will always use the "desired" volume size if autogrow is enabled and I might be wrong in that assumption 💭

internal/controller/postgrescluster/autogrow.go
jmckulk: what if we just passed in a VolumeClaimSpec 🤔

internal/postgres/config.go
tjmoore4: Changes should be pushed for this, tests are passing as expected with the change in place.

internal/controller/postgrescluster/autogrow.go
tjmoore4: I'm not really following. We'd have to associate the VolumeClaimSpec with the appropriate instance name otherwise we can't associate it with a request.

internal/controller/postgrescluster/instance.go
tjmoore4: I suppose, but we're still getting that from the cluster. I think the idea behind passing just the name is to make sure we're updating the statuses one at a time. `storeDesiredRequest` really only works with a single instance at a time, which is I think what we want to do here.

internal/controller/postgrescluster/autogrow.go
tjmoore4: If we simplified to that logic, I think that eliminates the warning event for a defined request that is higher than the limit. That's a useful warning in my opinion because it helps explains unexpected behavior (due to a misconfiguration on the user's part; you shouldn't even define a request that's higher than the limit you define in the same object).

tjmoore4: For the return by reference, we could definitely return it explicitly. I'm not returning anything that's not used, it's just that the one value is only used in the error log, which feels ok to me.

jmckulk: Which event would we be eliminating? 

dvp doesn't seem to be needed outside of getDesiredVolumeSize beyond logging the error. If we log the error in the function the return isn't necessary. We could pass in ctx to the func and have access to the same logger 📝

tjmoore4: It would eliminate the `VolumeRequestOverLimit` Event if I'm following correctly.

I was trying to keep the pulled out function simple, which also makes the test less complicated. It seemed better to just return the string in that case. :man_shrugging:

jmckulk (approved):
The code looks like it works to me. I still need to dig in to understand how auto grow works but i can do that as part of future changes

PR #4247, opened 2025-08-19
Add Disk Auto Grow for pgBackrest repo host volumes

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

This update adds the ability to automatically grow repo host PVCs. The AutoGrowVolumes feature gate must be enabled and the relevant volumes must include a Limit value.

Once enabled, this feature tracks the current disk utilization and, when utilization reaches 75%, the disk request is updated to 150% of the observed value. At this point and beyond, the requested value will be tracked by CPK.

The volume request can grow up to the configured limit value. Note: This change now treats limit values as authoritative regardless of the feature gate setting. However, the implementation also now allows limits to be updated after being set.

**Other Information**:
Issue: PGO-1427

### DIFF
--- a/internal/controller/postgrescluster/autogrow.go
+++ b/internal/controller/postgrescluster/autogrow.go
@@ -21,7 +21,7 @@ import (
 // status. If the value has grown, create an Event.
 func (r *Reconciler) storeDesiredRequest(
 	ctx context.Context, cluster *v1beta1.PostgresCluster,
-	volumeType, instanceSetName, desiredRequest, desiredRequestBackup string,
+	volumeType, host, desiredRequest, desiredRequestBackup string,
 ) string {
 	var current resource.Quantity
 	var previous resource.Quantity
@@ -33,7 +33,7 @@ func (r *Reconciler) storeDesiredRequest(
 		current, err = resource.ParseQuantity(desiredRequest)
 		if err != nil {
 			log.Error(err, "Unable to parse "+volumeType+" volume request from status ("+
-				desiredRequest+") for "+cluster.Name+"/"+instanceSetName)
+				desiredRequest+") for "+cluster.Name+"/"+host)
 			// If there was an error parsing the value, treat as unset (equivalent to zero).
 			desiredRequest = ""
 			current, _ = resource.ParseQuantity("")
@@ -46,7 +46,7 @@ func (r *Reconciler) storeDesiredRequest(
 		previous, err = resource.ParseQuantity(desiredRequestBackup)
 		if err != nil {
 			log.Error(err, "Unable to parse "+volumeType+" volume request from status backup ("+
-				desiredRequestBackup+") for "+cluster.Name+"/"+instanceSetName)
+				desiredRequestBackup+") for "+cluster.Name+"/"+host)
 			// If there was an error parsing the value, treat as unset (equivalent to zero).
 			desiredRequestBackup = ""
 			previous, _ = resource.ParseQuantity("")
@@ -55,12 +55,12 @@ func (r *Reconciler) storeDesiredRequest(
 	}
 
 	// determine if the appropriate volume limit is set
-	limitSet := limitIsSet(cluster, volumeType, instanceSetName)
+	limitSet := limitIsSet(cluster, volumeType, host)
 
 	if limitSet && current.Value() > previous.Value() {
 		r.Recorder.Eventf(cluster, corev1.EventTypeNormal, "VolumeAutoGrow",
 			"%s volume expansion to %v requested for %s/%s.",
-			volumeType, current.String(), cluster.Name, instanceSetName)
+			volumeType, current.String(), cluster.Name, host)
 	}
 
 	// If the desired size was not observed, update with previously stored value.
--- a/internal/controller/postgrescluster/autogrow.go
+++ b/internal/controller/postgrescluster/autogrow.go
@@ -6,7 +6,9 @@ package postgrescluster
 
 import (
 	"context"
+	"strings"
 
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 
@@ -77,34 +79,44 @@ func limitIsSet(cluster *v1beta1.PostgresCluster, volumeType, instanceSetName st
 
 	var limitSet bool
 
-	switch volumeType {
+	switch {
 
 	// Cycle through the instance sets to ensure the correct limit is identified.
-	case "pgData":
+	case volumeType == "pgData":
 		for _, specInstance := range cluster.Spec.InstanceSets {
 			if specInstance.Name == instanceSetName {
 				limitSet = !specInstance.DataVolumeClaimSpec.Resources.Limits.Storage().IsZero()
 			}
 		}
+
+	// VolumeType for the repository host volumes should be in the form 'repoN'
+	// where N is 1-4. As above, cycle through any defined repositories and ensure
+	// the correct limit is identified.
+	case strings.HasPrefix(volumeType, "repo"):
+		for _, specRepo := range cluster.Spec.Backups.PGBackRest.Repos {
+			if specRepo.Name == volumeType && specRepo.Volume != nil {
+				limitSet = !specRepo.Volume.VolumeClaimSpec.Resources.Limits.Storage().IsZero()
+			}
+		}
 	}
-	// TODO: Add cases for pgWAL and repo volumes
+	// TODO: Add case for pgWAL
 
 	return limitSet
 
 }
 
-// setVolumeSize compares the potential sizes from the instance spec, status
-// and limit and sets the appropriate current value.
+// setVolumeSize compares the potential sizes from the cluster status, volume request
+// and volume limit and sets the appropriate current value.
 func (r *Reconciler) setVolumeSize(ctx context.Context, cluster *v1beta1.PostgresCluster,
-	pvc *corev1.PersistentVolumeClaim, volumeType, instanceSpecName string) {
+	spec *corev1.PersistentVolumeClaimSpec, volumeType, host string) {
 
 	log := logging.FromContext(ctx)
 
 	// Store the limit for this instance set. This value will not change below.
-	volumeLimitFromSpec := pvc.Spec.Resources.Limits.Storage()
+	volumeLimitFromSpec := spec.Resources.Limits.Storage()
 
 	// This value will capture our desired update.
-	volumeRequestSize := pvc.Spec.Resources.Requests.Storage()
+	volumeRequestSize := spec.Resources.Requests.Storage()
 
 	// A limit of 0 is ignorned, so the volume request is used.
 	if volumeLimitFromSpec.IsZero() {
@@ -116,9 +128,9 @@ func (r *Reconciler) setVolumeSize(ctx context.Context, cluster *v1beta1.Postgre
 	if volumeRequestSize.Value() > volumeLimitFromSpec.Value() {
 		r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "VolumeRequestOverLimit",
 			"%s volume request (%v) for %s/%s is greater than set limit (%v). Limit value will be used.",
-			volumeType, volumeRequestSize, cluster.Name, instanceSpecName, volumeLimitFromSpec)
+			volumeType, volumeRequestSize, cluster.Name, host, volumeLimitFromSpec)
 
-		pvc.Spec.Resources.Requests = corev1.ResourceList{
+		spec.Resources.Requests = corev1.ResourceList{
 			corev1.ResourceStorage: *resource.NewQuantity(volumeLimitFromSpec.Value(), resource.BinarySI),
 		}
 		// Otherwise, if the feature gate is not enabled, do not autogrow.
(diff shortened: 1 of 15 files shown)


pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbackrest_types.go
```diff
@@ -427,6 +427,10 @@ type RepoStatus struct {
 	// commands accordingly.
 	// +optional
 	RepoOptionsHash string `json:"repoOptionsHash,omitempty"`
+
+	// Desired Size of the repo volume
+	// +optional
+	DesiredRepoVolume string `json:"desiredRepoVolume,omitempty"`
```
jmckulk, line 433: DesiredRepoVolume doesn't describe a volume size imo ⛏️ 
am i missing some part of what this status is doing?

internal/pgbackrest/config.go
```diff
@@ -622,6 +622,34 @@ func reloadCommand(name string) []string {
...
+# Manage autogrow annotation.
+# Return size in Mebibytes.
+manageAutogrowAnnotation() {
+  local volume=$1
+
+  size=$(df --human-readable --block-size=M "/pgbackrest/${volume}")
+  read -r _ size _ <<< "${size#*$'\n'}"
+  use=$(df --human-readable "/pgbackrest/${volume}")
+  read -r _ _ _ _ use _ <<< "${use#*$'\n'}"
+  sizeInt="${size//M/}"
```
jmckulk: a description/comment of this 🪄 would be appreciated
Maybe not in the script because that is user facing 🤔 
above the script?

internal/pgbackrest/config.go
```diff
@@ -622,6 +622,34 @@ func reloadCommand(name string) []string {
...
+NAMESPACE=$(cat "${SERVICEACCOUNT}"/namespace)
+TOKEN=$(cat "${SERVICEACCOUNT}"/token)
+CACERT=${SERVICEACCOUNT}/ca.crt
+
+# Manage autogrow annotation.
+# Return size in Mebibytes.
+manageAutogrowAnnotation() {
+  local volume=$1
+
+  size=$(df --human-readable --block-size=M "/pgbackrest/${volume}")
```
jmckulk: In my environments the `-h` flag doesn't change the output when used with the block size flag. I don't think we need it ⛏️ 

Do you see something different in your environments?

internal/controller/postgrescluster/autogrow.go
```diff
@@ -77,34 +79,44 @@ func limitIsSet(cluster *v1beta1.PostgresCluster, volumeType, instanceSetName st
...
 
 }
 
-// setVolumeSize compares the potential sizes from the instance spec, status
-// and limit and sets the appropriate current value.
+// setVolumeSize compares the potential sizes from the cluster status, volume request
+// and volume limit and sets the appropriate current value.
 func (r *Reconciler) setVolumeSize(ctx context.Context, cluster *v1beta1.PostgresCluster,
-	pvc *corev1.PersistentVolumeClaim, volumeType, instanceSpecName string) {
+	spec *corev1.PersistentVolumeClaimSpec, volumeType, host string) {
```
jmckulk, line 111: Should we be using our VolumeClaimSpec type here or generally throughout the code base ❓  🤔 
We made the distinction in our api but i'm not sure if we have talked about using one vs the other throughout the code
```suggestion
	spec *v1beta1.VolumeClaimSpec, volumeType, host string) {
```

internal/controller/postgrescluster/autogrow.go
```diff
@@ -116,19 +128,19 @@ func (r *Reconciler) setVolumeSize(ctx context.Context, cluster *v1beta1.Postgre
...
 		r.Recorder.Eventf(cluster, corev1.EventTypeWarning, "VolumeRequestOverLimit",
 			"%s volume request (%v) for %s/%s is greater than set limit (%v). Limit value will be used.",
-			volumeType, volumeRequestSize, cluster.Name, instanceSpecName, volumeLimitFromSpec)
+			volumeType, volumeRequestSize, cluster.Name, host, volumeLimitFromSpec)
 
-		pvc.Spec.Resources.Requests = corev1.ResourceList{
+		spec.Resources.Requests = corev1.ResourceList{
 			corev1.ResourceStorage: *resource.NewQuantity(volumeLimitFromSpec.Value(), resource.BinarySI),
 		}
 		// Otherwise, if the feature gate is not enabled, do not autogrow.
```
jmckulk, line 136: does this comment still make sense here? it is above an "if autogrow enabled" check so its throwing me off

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -2758,12 +2810,45 @@ func (r *Reconciler) reconcileRepos(ctx context.Context,
...
 	return replicaCreateRepo, utilerrors.NewAggregate(errors)
 }
 
+// +kubebuilder:rbac:groups="",resources="pods",verbs={list}
+
+// getRepoHostVolumeRequests gets the pgBackRest repo host volume request annotations
+// from the repo host Pod and stores them in the Postgres Cluster object's status.
+// If there is no repo host or there are no annotations, this function returns
+// without updating the status.
+func (r *Reconciler) getRepoHostVolumeRequests(ctx context.Context,
```
jmckulk: i'd like this function name to say something about status 🤔 ⛏️

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -2758,12 +2810,45 @@ func (r *Reconciler) reconcileRepos(ctx context.Context,
...
+
+	pods := &corev1.PodList{}
+	err := errors.WithStack(
+		r.Client.List(ctx, pods,
+			client.InNamespace(cluster.Namespace),
+			client.MatchingLabelsSelector{
+				Selector: naming.PGBackRestDedicatedLabels(cluster.Name).AsSelector()},
+		))
+
+	if len(pods.Items) > 0 {
```
jmckulk: If we always expect 1 pod should the check be this ❓ 
```suggestion
	if len(pods.Items) == 1 {
```

internal/controller/postgrescluster/autogrow.go
jmckulk: Maybe this is a question to include @cbandy on 🤔

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbackrest_types.go
tjmoore4: It'll hold the value from the annotation, something like "2Gi", etc. This is the same naming convention as what we're using for the pgData volume, it's just handled differently so I'm not using a map:
https://github.com/CrunchyData/postgres-operator/blob/main/pkg/apis/postgres-operator.crunchydata.com/v1beta1/postgrescluster_types.go#L607

internal/controller/postgrescluster/autogrow.go
tjmoore4: I needed to harmonize how we handled repo volumes vs instance volumes in the function. Moving to the spec allowed me to use the same function since I'm only actually working with the underlying spec anyway. I do wonder whether there isn't some broader refactoring that might make sense, but that feels out of scope for the current work.

internal/controller/postgrescluster/autogrow.go
tjmoore4: I think so, mainly because of the way the `if..then` is here, it helps call out the behavior. That said, I'm not totally against removing it if you think it's distracting.

internal/controller/postgrescluster/pgbackrest.go
tjmoore4: :thinking: Yeah, this function evolved over time, so the name isn't as descriptive as it could be. That said, I'm struggling to think of a good name that isn't too long. `storeRepoHostVolumeRequestAnnotationInStatus` is very descriptive but a bit wordy :smile: Any thoughts?

internal/controller/postgrescluster/pgbackrest.go
tjmoore4: That's a fair point. I was initially thinking it would allow it to be a bit more robust, but if there are two repo hosts for some reason, we probably don't want to do anything until the next reconcile anyway. I'll make that change.

internal/pgbackrest/config.go
tjmoore4: I think that'll be fine, that flag was a hold-over from the other script. I'll test it out before pushing the change.

tjmoore4: Pushed.

internal/pgbackrest/config.go
tjmoore4: Sure, I added a comment.

internal/controller/postgrescluster/autogrow.go
jmckulk: fair enough 👍

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -2758,12 +2813,52 @@ func (r *Reconciler) reconcileRepos(ctx context.Context,
...
+
+	pods := &corev1.PodList{}
+	err := errors.WithStack(
+		r.Client.List(ctx, pods,
+			client.InNamespace(cluster.Namespace),
+			client.MatchingLabelsSelector{
+				Selector: naming.PGBackRestDedicatedLabels(cluster.Name).AsSelector()},
+		))
+
+	if len(pods.Items) == 1 {
```
jmckulk: should we have an else case here and create an error? it looks like "if len == 1" we will just return nil
```suggestion
	if err == nil && len(pods.Items) == 1 {
	...
	} else {
	    err = errors.New("blah")
	}
	return err
```

internal/controller/postgrescluster/pgbackrest.go
jmckulk: `writeVolumeSizeRequestStatus`
 maybe the repo is implied by the package 🤔 
I'm not sure if the word annotation adds anything 💭

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -2735,6 +2774,14 @@ func (r *Reconciler) reconcileRepos(ctx context.Context,
 	errMsg := "reconciling repository volume"
 	repoVols := []*corev1.PersistentVolumeClaim{}
 	var replicaCreateRepo v1beta1.PGBackRestRepo
+
+	autogrow := feature.Enabled(ctx, feature.AutoGrowVolumes)
+	if autogrow {
```
jmckulk: why not this? looks like autogrow isn't used elsewhere
```suggestion
	if feature.Enabled(ctx, feature.AutoGrowVolumes) {
```

internal/pgbackrest/config.go
```diff
@@ -621,7 +621,44 @@ func reloadCommand(name string) []string {
...
+
+	// In the manageAutogrowAnnotation function below, df is used to return the
+	// relevant volume size in Mebibytes. The 'read' variable gets the value from
+	// the '1M-blocks' output (second column) and the 'use' variable gets the value
+	// from the 'Use%' column (fifth column). This value is grabbed after stripping
+	// out the column headers (before the '\n') and then getting the respective
+	// value delimited by the white spaces. The percent value is stripped of the
+	// '%' and then used to determine if a expansion should be triggered by setting
+	// the calculated volume size using the 'size' variable.
 	const script = `
```
jmckulk: maybe mention what the `_` is doing?

internal/controller/postgrescluster/autogrow.go
jmckulk: I'm on the fence of if i think it is helpful but it might be for others. so... ⛏️ 
i don't think it confuses things so maybe its fine to leave

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbackrest_types.go
jmckulk: I think the word size would do a lot and probably is the same for pgdata ⛏️ 

but its in the status so hard to change 🤔

tjmoore4: I think I'll leave it for now because I'd like to keep them consistent. We can think about changing it later perhaps (and if so, I'd think all at the same time would be best).

internal/controller/postgrescluster/pgbackrest.go
tjmoore4: Maybe `writeRepoVolumeSizeRequestStatus`? It's still a little long, but it's pretty descriptive. I don't think we need annotation here necessarily.

internal/controller/postgrescluster/pgbackrest.go
tjmoore4: Good call, I was just following the other pattern and didn't think too hard about it. I'll update that.

internal/controller/postgrescluster/pgbackrest.go
tjmoore4: :thinking: I wonder if the `err == nil` is necessary? Is there a scenario where we get an error and the Pod is returned?
I think the else block makes sense, I can add that.

internal/controller/postgrescluster/pgbackrest.go
jmckulk: yeah that sounds good to me 👍

internal/controller/postgrescluster/pgbackrest.go
tjmoore4: I thought about it a bit more and I refactored the function to handle the errors better. Let me know if that works for you.

internal/controller/postgrescluster/pgbackrest.go
tjmoore4: I pushed that name.

jmckulk (approved):

PR #4251, opened 2025-08-20
Log API warnings at a higher verbosity

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature

Warnings about Endpoints deprecation are logged at the info level.

Warnings about Endpoints deprecation are logged at the debug level when `CRUNCHY_DEBUG=2` or higher.

**Other Information**:

### DIFF
--- a/cmd/postgres-operator/main.go
+++ b/cmd/postgres-operator/main.go
@@ -20,6 +20,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
 
@@ -88,6 +89,10 @@ func initLogging() {
 
 	global := logging.FromContext(context.Background())
 	runtime.SetLogger(global)
+
+	// [k8s.io/client-go/tools/leaderelection] logs to the global [klog] instance.
+	// - https://github.com/kubernetes-sigs/controller-runtime/issues/2656
+	klog.SetLoggerWithOptions(global, klog.ContextualLogger(true))
 }
 
 //+kubebuilder:rbac:groups="coordination.k8s.io",resources="leases",verbs={get,create,update,watch}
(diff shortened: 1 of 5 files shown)


cmd/postgres-operator/main.go
```diff
@@ -40,31 +41,66 @@ import (
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
-// assertNoError panics when err is not nil.
-func assertNoError(err error) {
+// must panics when err is not nil.
+func must(err error) { need(0, err) }
```
benjaminjb, line 45: So use `must` when we don't care about the value at all and just want to make sure the error is nil

cmd/postgres-operator/main.go
```diff
@@ -40,31 +41,66 @@ import (
...
+	config, err := runtime.GetConfig()
+
+	if err == nil && userAgent == "" {
+		err = errors.New("call initVersion first")
+	}
+	if err == nil {
+		config.UserAgent = userAgent
+		config.Wrap(otelTransportWrapper())
+
+		// Log Kubernetes API warnings encountered by client-go at a high verbosity.
```
benjaminjb, line 63: Correct me if wrong, but is the magic here that we log at v2 rather than what the native WarningHandler does?

benjaminjb (commented):
Looks like complaining about go mod

cmd/postgres-operator/main.go
cbandy: Yep. These names are used elsewhere, too: https://github.com/CrunchyData/postgres-operator/blob/da337c84f39819ee93871b3c22bebb63760d64ac/internal/collector/generate.go#L54-L55

cmd/postgres-operator/main.go
cbandy: Exactly. When unset, controller-runtime calls only `log.Info`:

- https://github.com/kubernetes-sigs/controller-runtime/blob/v0.21.0/pkg/client/client.go#L115
- https://github.com/kubernetes-sigs/controller-runtime/blob/v0.21.0/pkg/log/warning_handler.go#L64

benjaminjb (approved):

PR #4252, opened 2025-08-21
Validate and use pgbouncer logfile config

benjaminjb:
**Checklist:**

**Type of Changes:**
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Users can set any logfile config for pgbouncer, which might interfere with our OTEL

We now restrict users to only set a logfile with the `.log` suffix; and we get the config to handle the OTEL-related logic

TODO: Restrict the logfile to only certain locations?

**Other Information**:
Issues: [PGO-2565]

### DIFF
--- a/config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
+++ b/config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
@@ -14375,7 +14375,13 @@ spec:
                             description: |-
                               Settings that apply to the entire PgBouncer process.
                               More info: https://www.pgbouncer.org/config.html
+
+                              # Logging
                             type: object
+                            x-kubernetes-map-type: granular
+                            x-kubernetes-validations:
+                            - message: logfile config must end with '.log'
+                              rule: '!has(self.logfile) || self.logfile.endsWith(''.log'')'
                           users:
                             additionalProperties:
                               type: string
@@ -33300,7 +33306,13 @@ spec:
                             description: |-
                               Settings that apply to the entire PgBouncer process.
                               More info: https://www.pgbouncer.org/config.html
+
+                              # Logging
                             type: object
+                            x-kubernetes-map-type: granular
+                            x-kubernetes-validations:
+                            - message: logfile config must end with '.log'
+                              rule: '!has(self.logfile) || self.logfile.endsWith(''.log'')'
                           users:
                             additionalProperties:
                               type: string
--- a/internal/collector/pgbouncer.go
+++ b/internal/collector/pgbouncer.go
@@ -9,6 +9,7 @@ import (
 	_ "embed"
 	"encoding/json"
 	"fmt"
+	"path/filepath"
 	"slices"
 	"strconv"
 
@@ -29,7 +30,7 @@ const PGBouncerPostRotateScript = "pkill -HUP --exact pgbouncer"
 // NewConfigForPgBouncerPod creates a config for the OTel collector container
 // that runs as a sidecar in the pgBouncer Pod
 func NewConfigForPgBouncerPod(
-	ctx context.Context, cluster *v1beta1.PostgresCluster, sqlQueryUsername string,
+	ctx context.Context, cluster *v1beta1.PostgresCluster, sqlQueryUsername, logfile string,
 ) *Config {
 	if cluster.Spec.Proxy == nil || cluster.Spec.Proxy.PGBouncer == nil {
 		// pgBouncer is disabled; return nil
@@ -38,7 +39,7 @@ func NewConfigForPgBouncerPod(
 
 	config := NewConfig(cluster.Spec.Instrumentation)
 
-	EnablePgBouncerLogging(ctx, cluster, config)
+	EnablePgBouncerLogging(ctx, cluster, config, logfile)
 	EnablePgBouncerMetrics(ctx, cluster, config, sqlQueryUsername)
 
 	return config
@@ -49,6 +50,7 @@ func NewConfigForPgBouncerPod(
 func EnablePgBouncerLogging(ctx context.Context,
 	inCluster *v1beta1.PostgresCluster,
 	outConfig *Config,
+	logfile string,
 ) {
 	var spec *v1beta1.InstrumentationLogsSpec
 	if inCluster != nil && inCluster.Spec.Instrumentation != nil {
@@ -56,7 +58,7 @@ func EnablePgBouncerLogging(ctx context.Context,
 	}
 
 	if OpenTelemetryLogsEnabled(ctx, inCluster) {
-		directory := naming.PGBouncerLogPath
+		directory := filepath.Dir(logfile)
 
 		// Keep track of what log records and files have been processed.
 		// Use a subdirectory of the logs directory to stay within the same failure domain.
--- a/internal/controller/postgrescluster/pgbouncer.go
+++ b/internal/controller/postgrescluster/pgbouncer.go
@@ -44,8 +44,9 @@ func (r *Reconciler) reconcilePGBouncer(
 		secret, err = r.reconcilePGBouncerSecret(ctx, cluster, root, service)
 	}
 	if err == nil {
-		config := collector.NewConfigForPgBouncerPod(ctx, cluster, pgbouncer.PostgresqlUser)
-		configmap, err = r.reconcilePGBouncerConfigMap(ctx, cluster, config)
+		logfile := setPGBouncerLogfile(ctx, cluster)
+		config := collector.NewConfigForPgBouncerPod(ctx, cluster, pgbouncer.PostgresqlUser, logfile)
+		configmap, err = r.reconcilePGBouncerConfigMap(ctx, cluster, config, logfile)
 	}
 	if err == nil {
 		err = r.reconcilePGBouncerDeployment(ctx, cluster, primaryCertificate, configmap, secret)
@@ -59,6 +60,20 @@ func (r *Reconciler) reconcilePGBouncer(
 	return err
 }
 
+// setPGBouncerLogfile retrieves the logfile config if present in the user config.
+// If not present, set to the OTEL default.
+// If OTEL is not enabled, we do not use this value.
+// TODO: Check INI config files specified on the cluster
+func setPGBouncerLogfile(ctx context.Context, cluster *v1beta1.PostgresCluster) string {
+	logfile := naming.PGBouncerFullLogPath
+
+	if dest, ok := cluster.Spec.Proxy.PGBouncer.Config.Global["logfile"]; ok {
+		logfile = dest
+	}
+
+	return logfile
+}
+
 // +kubebuilder:rbac:groups="",resources="configmaps",verbs={get}
 // +kubebuilder:rbac:groups="",resources="configmaps",verbs={create,delete,patch}
 
@@ -66,6 +81,7 @@ func (r *Reconciler) reconcilePGBouncer(
 func (r *Reconciler) reconcilePGBouncerConfigMap(
 	ctx context.Context, cluster *v1beta1.PostgresCluster,
 	otelConfig *collector.Config,
+	logfile string,
 ) (*corev1.ConfigMap, error) {
 	configmap := &corev1.ConfigMap{ObjectMeta: naming.ClusterPGBouncer(cluster)}
 	configmap.SetGroupVersionKind(corev1.SchemeGroupVersion.WithKind("ConfigMap"))
@@ -104,7 +120,7 @@ func (r *Reconciler) reconcilePGBouncerConfigMap(
 	// If OTel logging is enabled, add logrotate config
 	if err == nil && collector.OpenTelemetryLogsEnabled(ctx, cluster) {
 		logrotateConfig := collector.LogrotateConfig{
-			LogFiles:         []string{naming.PGBouncerFullLogPath},
+			LogFiles:         []string{logfile},
 			PostrotateScript: collector.PGBouncerPostRotateScript,
 		}
 		collector.AddLogrotateConfigs(ctx, cluster.Spec.Instrumentation, configmap,
--- a/internal/controller/postgrescluster/pgbouncer.go
+++ b/internal/controller/postgrescluster/pgbouncer.go
@@ -25,6 +25,7 @@ import (
 	"github.com/crunchydata/postgres-operator/internal/pgbouncer"
 	"github.com/crunchydata/postgres-operator/internal/pki"
 	"github.com/crunchydata/postgres-operator/internal/postgres"
+	"github.com/crunchydata/postgres-operator/internal/util"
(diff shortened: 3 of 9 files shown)


internal/controller/postgrescluster/pgbouncer.go
```diff
@@ -59,13 +61,28 @@ func (r *Reconciler) reconcilePGBouncer(
...
 }
 
+// setPGBouncerLogfile retrieves the logfile config if present in the user config.
+// If not present, set to the OTEL default.
+// If OTEL is not enabled, we do not use this value.
+// TODO: Check INI config files specified on the cluster
+func setPGBouncerLogfile(ctx context.Context, cluster *v1beta1.PostgresCluster) string {
+	logfile := naming.PGBouncerFullLogPath
+
+	if dest, ok := cluster.Spec.Proxy.PGBouncer.Config.Global["logfile"]; ok {
```
benjaminjb: Oh, meant to double check whether we need more guardrails here.

internal/pgbouncer/config.go
```diff
@@ -127,9 +127,13 @@ func clusterINI(ctx context.Context, cluster *v1beta1.PostgresCluster) string {
 		"unix_socket_dir": "",
 	}
 
+	// Override the above with any specified settings.
+	maps.Copy(global, cluster.Spec.Proxy.PGBouncer.Config.Global)
```
benjaminjb, line 131: I felt better letting the user override some of our settings above

internal/collector/pgbouncer.go
```diff
@@ -49,22 +50,27 @@ func NewConfigForPgBouncerPod(
...
 		spec = inCluster.Spec.Instrumentation.Logs
 	}
 
 	if OpenTelemetryLogsEnabled(ctx, inCluster) {
-		directory := naming.PGBouncerLogPath
+		directory := filepath.Dir(logfile)
+		create_directory := false
+		if directory != "/tmp" {
+			create_directory = true
+		}
```
benjaminjb: I ran into an error when I had set the logfile conf to a non `tmp` location -- our collector startup has 
```
` + shell.MakeDirectories(logDir, path.Join(logDir, "receiver"))
```
which would seem to have solved the issue, but I suppose pgbouncer might start up and error out before this. And the pgbouncer container should manage its own logfile location.

internal/controller/postgrescluster/pgbouncer.go
```diff
@@ -59,13 +61,34 @@ func (r *Reconciler) reconcilePGBouncer(
 	return err
 }
 
+// setPGBouncerLogfile retrieves the logfile config if present in the user config.
+// If not present, set to the OTEL default.
+// If OTEL is not enabled, we do not use this value.
+// TODO: Check INI config files specified on the cluster
+func setPGBouncerLogfile(cluster *v1beta1.PostgresCluster) string {
```
cbandy, line 68: ❓ Where is this called?

internal/util/pod_security.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/crunchydata/postgres-operator/internal/initialize"
+)
+
+// PodSecurityContext returns a v1.PodSecurityContext for cluster that can write
+// to PersistentVolumes.
+func PodSecurityContext(fsgroup int64, supplementalGroups []int64, openshift bool) *corev1.PodSecurityContext {
```
cbandy: 🤔 I think I'd prefer the caller to pass in `fsgroup = 0` when they know its OpenShift, but a single place with this knowledge is good, too.

🤔 🤔 PgBouncer is deployed from a PostgresCluster spec which has an isOpenShift override.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbouncer_types.go
```diff
@@ -26,7 +26,12 @@ type PGBouncerConfiguration struct {
 
 	// Settings that apply to the entire PgBouncer process.
 	// More info: https://www.pgbouncer.org/config.html
+	//
+	// # Logging
+	// +kubebuilder:validation:XValidation:rule=`!has(self.logfile) || self.logfile.endsWith('.log')`,message=`logfile config must end with '.log'`
+	//
 	// +optional
+	// +mapType=granular
```
cbandy, line 35: 👍🏻 I want these granular/set/atomic markers on all the new collections we add to the API.

internal/collector/pgbouncer.go
cbandy: Yeah. I used the `postgres-startup` container for this. This is one of the situations I would love to solve with a Go sidecar.

cbandy: 🤔 In this moment, we could augment the entrypoint.

```go
Command: []string{"sh", "-c", "--",
	mkdirs + `; exec "$@"`, "--",
	"pgbouncer", iniFileAbsolutePath},
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbouncer_types.go
```diff
@@ -26,7 +26,12 @@ type PGBouncerConfiguration struct {
 
 	// Settings that apply to the entire PgBouncer process.
 	// More info: https://www.pgbouncer.org/config.html
+	//
+	// # Logging
+	// +kubebuilder:validation:XValidation:rule=`!has(self.logfile) || self.logfile.endsWith('.log')`,message=`logfile config must end with '.log'`
```
cbandy, line 31: ❓ Where does the file go if there is no directory here? What if there's a relative directory?

internal/collector/pgbouncer.go
benjaminjb: I just re-tested in two scenarios (with OTEL, without OTEL) and even without any special setup of the files, it seemed to work fine. I wonder if the error I saw before was related to the fsGroup not being set.

(Also, when I run `shell.MakeDirectories`, now I get an error about chmod not being allowed.)

cbandy: > chmod not being allowed

What storage do you have? That sounds like NFS, maybe.

benjaminjb: I've been using GKE for testing this, I want to run some more tests to double-check everything

internal/pgbouncer/reconcile.go
```diff
@@ -146,10 +149,15 @@ func Pod(
 		),
 	}
 
+	mkdirCommand := ""
+	if logfile != "" && filepath.Dir(logfile) != "/tmp" {
+		mkdirCommand = shell.MakeDirectories("/tmp", filepath.Dir(logfile)) + "; "
+	}
```
dsessler7: This is confusing me... If the provided `logfile` doesn't have `/tmp` as the base, why would we want to force it to use `/tmp` as the base? :thinking:

benjaminjb: This may be an abuse of `MakeDirectories` but:
- MakeDirectories takes two values, a base and a path
- it creates the dir at path and chmods that dir
- if path includes the base dir, then it chmods the dirs in between
- `filepath.Dir` just takes off the last element -- which in this case, has to be the file name (because of the CEL constraint on this variable / the default we set) 🤞 

```
MakeDirectories("/one", "/one/two/three/four") 
#=> mkdir -p '/one/two/three/four' && { chmod 0775 '/one/two/three/four' '/one/two/three' '/one/two' || :; }
```

```
MakeDirectories("/zero", "/one/two/three/four") 
#=> mkdir -p '/one/two/three/four' && { chmod 0775 '/one/two/three/four' || :; }
```

So I'm only creating + chmoding the final directory here.

Still, this isn't easy to parse -- might be easier to read if I do 
```
MakeDirectories(filepath.Dir(logfile), filepath.Dir(logfile)) + "; "
```

internal/collector/pgbouncer.go
benjaminjb: ok, adding back some dir management to the entrypoint, seems fine -- as long as the location exists

internal/util/pod_security.go
benjaminjb: on PodSecurityContext, fsgroup is a pointer (with omitempty) so we don't want to set it at all, right?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbouncer_types.go
```diff
@@ -26,7 +26,13 @@ type PGBouncerConfiguration struct {
 
 	// Settings that apply to the entire PgBouncer process.
 	// More info: https://www.pgbouncer.org/config.html
+	//
```
cbandy: 🔧 The comments below are getting attached to the field description.

```suggestion
	// ---
```

pkg/apis/postgres-operator.crunchydata.com/v1/postgrescluster_types.go
```diff
@@ -598,6 +598,7 @@ type PostgresInstanceSetStatus struct {
 type PostgresProxySpec struct {
 
 	// Defines a PgBouncer proxy and connection pooler.
+	// +kubebuilder:validation:XValidation:rule=`!has(self.config) || !has(self.config.global) || !has(self.config.global.logfile) || self.config.global.logfile.startsWith('/tmp/logs/pgbouncer/') || self.volumes.additional.exists(x, self.config.global.logfile.startsWith("/volumes/"+x.name))`,message=`logfile destination is restricted to '/tmp/logs/pgbouncer/' or an existing additional volume`
```
cbandy: ❓ can this be on the `PGBouncerPodSpec` godoc?

internal/util/pod_security.go
cbandy: Right. I was picturing L33 as `if fsgroup > 0`

pkg/apis/postgres-operator.crunchydata.com/v1/postgrescluster_types.go
cbandy: 🤔 Is this any more legible? I have no idea.

```cel
self.?config.global.logfile.optMap(f, f.startsWith("/tmp/logs/pgbouncer/") || self.?volumes.additional.optMap(a, a.exists(v, f.startsWith("/volumes/" + v.name))).orValue(true)
```

cbandy: 📝 `a` is only referenced once, so maybe ditching an optional is better.

```diff
-additional.optMap(a, a.exists(v,
+additional.orValue([]).exists(v,
```

```cel
self.?config.global.logfile.optMap(f, f.startsWith("/tmp/logs/pgbouncer/") || self.?volumes.additional.orValue([]).exists(v, f.startsWith("/volumes/" + v.name))).orValue(true)
```

internal/util/pod_security.go
benjaminjb: OK, updated to try it out -- is that clearer/easier to use?

pkg/apis/postgres-operator.crunchydata.com/v1/postgrescluster_types.go
benjaminjb: Funny, I was just lamenting the lack of a ternary in Golang and here's `orValue`

Once I looked up the parts I haven't seen before, I'll say it's certainly shorter. Is it more legible? The more I look at it, the more I like it, but this just raises the need for an internal guide to CEL -- that might fit neatly in PGO as an MD file.

benjaminjb: I think so, that would just require a v1 of the `PGBouncerPodSpec`

benjaminjb: Except now I am getting a cost limit for that

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbouncer_types.go
benjaminjb: pgbouncer won't start if the logfile directory doesn't exist (i.e., pgbouncer creates the file but doesn't handle the directories)

pkg/apis/postgres-operator.crunchydata.com/v1/pgbouncer_types.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
+)
+
+// PGBouncerPodSpec defines the desired state of a PgBouncer connection pooler.
+// +kubebuilder:validation:XValidation:rule=`!has(self.config) || !has(self.config.global) || !has(self.config.global.logfile) || self.config.global.logfile.startsWith('/tmp/logs/pgbouncer/') || self.volumes.additional.exists(x, self.config.global.logfile.startsWith("/volumes/"+x.name))`,message=`logfile destination is restricted to '/tmp/logs/pgbouncer/' or an existing additional volume`
+type PGBouncerPodSpec struct {
+	v1beta1.PGBouncerPodSpec `json:",inline"`
+}
```
cbandy, line 15: 🤔 Some markers work with type aliases. Is `XValidation` one of them?


```suggestion
type PGBouncerPodSpec = v1beta1.PGBouncerPodSpec
```

config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
```diff
@@ -16405,6 +16409,12 @@ spec:
                             x-kubernetes-list-type: map
                         type: object
                     type: object
+                    x-kubernetes-validations:
+                    - message: logfile destination is restricted to '/tmp/logs/pgbouncer/'
+                        or an existing additional volume
+                      rule: '!has(self.config) || !has(self.config.global) || !has(self.config.global.logfile)
+                        || self.config.global.logfile.startsWith(''/tmp/logs/pgbouncer/'')
+                        || self.volumes.additional.exists(x, self.config.global.logfile.startsWith("/volumes/"+x.name))'
```
cbandy, line 16417: ❓ What happens when there is no volumes section?

dsessler7: Just tested this with my pgbackrest stuff, and it does not catch it if you do not have a volumes section.

dsessler7: Nevermind, I was using the wrong API version when testing... If the volumes section is omitted it will reject it

pkg/apis/postgres-operator.crunchydata.com/v1/postgrescluster_types.go
dsessler7: CEL validation has a ternary operator btw

PR #4253, opened 2025-08-21
Separate controller client concerns

cbandy:
This separates the `client.Client` interface in our controllers into three: `Reader`, `Writer`, and `StatusWriter`. The idea is to add a _little_ friction when a Client to a test, so we're more intentional about separating reads from writes. During this refactor, I found a number of tests that didn't need Kubernetes at all. This exercise has increased the amount of code covered by `make check`.

This refactor is one step toward my goal of a single, shared SSA Apply function.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement
 - [x] Other

### DIFF
--- a/internal/bridge/crunchybridgecluster/postgres_test.go
+++ b/internal/bridge/crunchybridgecluster/postgres_test.go
@@ -20,16 +20,10 @@ import (
 )
 
 func TestGeneratePostgresRoleSecret(t *testing.T) {
-	tClient := setupKubernetes(t)
-	require.ParallelCapacity(t, 0)
-
-	reconciler := &CrunchyBridgeClusterReconciler{
-		Client: tClient,
-		Owner:  "crunchybridgecluster-controller",
-	}
+	reconciler := &CrunchyBridgeClusterReconciler{}
 
 	cluster := testCluster()
-	cluster.Namespace = setupNamespace(t, tClient).Name
+	cluster.Namespace = "asdf"
 
 	spec := &v1beta1.CrunchyBridgeClusterRoleSpec{
 		Name:       "application",
@@ -77,8 +71,8 @@ func TestReconcilePostgresRoleSecrets(t *testing.T) {
 	ns := setupNamespace(t, tClient).Name
 
 	reconciler := &CrunchyBridgeClusterReconciler{
-		Client: tClient,
-		Owner:  "crunchybridgecluster-controller",
+		Reader: tClient,
+		Writer: client.WithFieldOwner(tClient, t.Name()),
 	}
 
 	t.Run("DuplicateSecretNameInSpec", func(t *testing.T) {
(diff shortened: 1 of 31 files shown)


internal/controller/standalone_pgadmin/users_test.go
```diff
@@ -316,8 +316,7 @@ func TestWritePGAdminUsers(t *testing.T) {
 		assert.Equal(t, calls, 1, "PodExec should be called once")
 
 		secret := &corev1.Secret{ObjectMeta: naming.StandalonePGAdmin(pgadmin)}
-		assert.NilError(t,
-			reconciler.Get(ctx, client.ObjectKeyFromObject(secret), secret))
+		assert.NilError(t, cc.Get(ctx, client.ObjectKeyFromObject(secret), secret))
```
benjaminjb, line 319: but we could use reconciler.Reader.Get, yeah? Which is more indirection but maybe a reminder of separation of client powers?

benjaminjb: But then -- who cares? This step isn't about what our operator does, this is just a step to get the object. "Getting the object" is not what we're testing here.

internal/bridge/crunchybridgecluster/postgres_test.go
```diff
@@ -20,16 +20,10 @@ import (
...
 
 func TestGeneratePostgresRoleSecret(t *testing.T) {
-	tClient := setupKubernetes(t)
-	require.ParallelCapacity(t, 0)
-
-	reconciler := &CrunchyBridgeClusterReconciler{
-		Client: tClient,
-		Owner:  "crunchybridgecluster-controller",
-	}
+	reconciler := &CrunchyBridgeClusterReconciler{}
```
benjaminjb, line 23: Kind of want to rename `generatePostgresRoleSecret` `generatePostgresRoleSecretIntent` or something

internal/naming/controllers.go
```diff
@@ -5,6 +5,9 @@
 package naming
 
 const (
-	ControllerBridge  = "bridge-controller"
-	ControllerPGAdmin = "pgadmin-controller"
+	ControllerBridge               = "bridge-controller"
```
benjaminjb, line 8: ah, the bridge installation controller already has this split between reader/writer, but no statuswriter -- and we never touch the status

internal/bridge/crunchybridgecluster/postgres_test.go
benjaminjb: Yeah, there's a few funcs here that, from the name, I would've said call kubernetes.

internal/controller/standalone_pgadmin/users_test.go
cbandy: Right. These lines intend to look around the environment, not exercise the Reconciler. Hypothetically, the Reconciler client could have limited permissions, while the test client is superuser with x-ray vision.

PR #4259, opened 2025-08-27
Another test helper for metav1.Status.Details

cbandy:
Every call to "require.StatusError" was followed by reaching into its Details field. This eliminates the check for nil that entailed.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement


benjaminjb (approved):
Saves us from typing a line when we want details, and if we don't, the original func is still there.

PR #4263, opened 2025-09-03
Add LoggingConfiguration struct for providing log path to pgbackrest …

dsessler7:
…processes.

Add ability to send pgbackrest logs to additional volume for repo host processes.

**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Currently, the various pgbackrest processes can only log to the default location that we specify.

Users can now have their pgBackRest processes log to additional volumes.

**Other Information**:


pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbackrest_types.go
```diff
@@ -113,6 +113,10 @@ type PGBackRestArchive struct {
 	// +optional
 	Jobs *BackupJobs `json:"jobs,omitempty"`
 
+	// Logging configuration for pgbackrest processes running in postgres instance pods.
+	// +optional
+	Log *LoggingConfiguration `json:"log,omitempty"`
```
dsessler7, line 118: I am open to other ideas for the placement of this LoggingConfiguration.

pkg/apis/postgres-operator.crunchydata.com/v1/pgbackrest_types.go
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package v1
+
+import (
+	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
+)
+
+// PGBackRestArchive defines a pgBackRest archive configuration
+// +kubebuilder:validation:XValidation:rule=`!has(self.repoHost) || !has(self.repoHost.log) || !has(self.repoHost.log.path) || self.repoHost.volumes.additional.exists(x, self.repoHost.log.path.startsWith("/volumes/"+x.name))`,message=`log path is restricted to an existing additional volume`
```
dsessler7: Right now I just have it set to validate that the `repoHost.log.path` isn't set or if it is, it's set to an additional volume configured in `repoHost`... Not sure if we want to allow other paths or not... Thoughts?

benjaminjb: I believe the repohost (as it now exists) will always have a `/tmp` directory and a `/pgbackrest/repoN` directory. Is that right and if so, would we want to allow logging to those? (Though actually: if we let logging into `/pgbackrest/repoN`, what subdirectories would we allow? Probably `/pgbackrest/repoN/logs/pgbackrest` only?)

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbackrest_types.go
benjaminjb: I do wonder if there would be an even better place to put it, but not sure where, so: Seems reasonable to me for now.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbackrest_types.go
```diff
@@ -215,6 +219,10 @@ type PGBackRestRepoHost struct {
 	// +optional
 	Affinity *corev1.Affinity `json:"affinity,omitempty"`
 
+	// Logging configuration for pgbackrest processes running in the repo host pod.
+	// +optional
+	Log *LoggingConfiguration `json:"log,omitempty"`
```
benjaminjb, line 228: Are we going to add LoggingConfiguration to the Job structs as well?

pkg/apis/postgres-operator.crunchydata.com/v1/pgbackrest_types.go
```diff
@@ -0,0 +1,15 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package v1
+
+import (
+	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
+)
+
+// PGBackRestArchive defines a pgBackRest archive configuration
```
benjaminjb, line 11: Do we also want to add some CEL validation in V1 to disallow setting log path through the `global` map?

pkg/apis/postgres-operator.crunchydata.com/v1/pgbackrest_types.go
benjaminjb: Oh, but `/pgbackrest/repoN/log` (or logs?) is our default, so... I suppose we could allow explicit setting of our default just to let people be explicit.

internal/pgbackrest/config.go
```diff
@@ -145,7 +146,15 @@ func CreatePGBackRestConfigMapIntent(ctx context.Context, postgresCluster *v1bet
...
-						pgBackRestLogPath = fmt.Sprintf(naming.PGBackRestRepoLogPath, repo.Name)
+						// If the user has set a log path in the spec, use it.
+						// Otherwise, default to /pgbackrest/repo#/log
+						if postgresCluster.Spec.Backups.PGBackRest.RepoHost != nil &&
+							postgresCluster.Spec.Backups.PGBackRest.RepoHost.Log != nil &&
+							postgresCluster.Spec.Backups.PGBackRest.RepoHost.Log.Path != "" {
+							pgBackRestLogPath = postgresCluster.Spec.Backups.PGBackRest.RepoHost.Log.Path
+						} else {
+							pgBackRestLogPath = fmt.Sprintf(naming.PGBackRestRepoLogPath, repo.Name)
+						}
```
benjaminjb: Not sure this helps reduce the number of changes but:

If you did this logic up above l. 134, you could have a defined string `pgBackRestLogPath` that you could then pass into `NewConfigForPgBackrestRepoHostPod` as a new var, saving the logic of checking if the log.path value exists (and also not needing to change the signature of the `NewConfigForPgBackrestRepoHostPod` func from instrumentationSpec to cluster)

pkg/apis/postgres-operator.crunchydata.com/v1/pgbackrest_types.go
dsessler7: I guess we could... Obviously can't stop users from setting log path via the `configuration`...

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbackrest_types.go
dsessler7: Yep!

internal/pgbackrest/config.go
dsessler7: I think it does and I like it. Another commit incoming shortly.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/pgbackrest_types.go
dsessler7: Hmm... Theoretically users could have multiple InstanceSets that have different additional volumes mounted, which would mean we'd need to allow different log paths per InstanceSet... So maybe the LoggingConfiguration should go on the PostgresInstanceSetSpec struct? :thinking:

benjaminjb: But we mount the same pgbackrest config to every postgres pod, right?

dsessler7: I was thinking that we had a different `pgbackrest_instance.conf` file per InstanceSet, but it seems you are correct... Only one `pgbackrest_instance.conf` regardless of the number of InstanceSets... Do we move the `Volumes.Additional` out of the `PostgresInstanceSetSpec` and into the `PostgresClusterSpec` and mount them into all instances so the log paths can be consistent amongst instances? :thinking:

benjaminjb: But then we'd be mounting the same PVC to all the different InstanceSet pods and having them all write to the same locations. (Which actually is the situation now if you have a single instanceSet with multiple replicas, so... it's already a fraught situation.)

Since we currently only allow people to bring their own volumes (rather than set up a template), people have to be a little careful about the pods-volumes connections. The safest scenario would probably be something like

- multiple instance sets
- each with only 1 replica
- each with the same **name** of additional volumes but different PVC claims

That way, the config remains the same without any worry about collisions/overwriting between processes (where does each pod log? to the PVC that only that pod is connected)

benjaminjb: In other words, I think this is (for now) going to be a lot of socialization and documentation.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
```diff
@@ -320,3 +320,10 @@ func (in *AdditionalVolume) AsVolume(name string) corev1.Volume {
 
 	return out
 }
+
+// LoggingConfiguration provides logging configuration for various components
+type LoggingConfiguration struct {
+	// +kubebuilder:validation:MaxLength=4096
```
benjaminjb: Why 4096 as maxlength?

internal/pgbackrest/config.go
```diff
@@ -469,11 +462,16 @@ func populateRepoHostConfigurationMap(
...
+			// If the user has set a log path in the spec, use it. Otherwise,
+			// default to /pgbackrest/repo#/log where pgBackRest will log to the
+			// first configured repo volume when commands are run on the pgBackRest
+			// repo host. With our previous check in RepoHostVolumeDefined(),
+			// we've already validated that at least one defined repo has a volume.
+			if repoHost != nil && repoHost.Log != nil && repoHost.Log.Path != "" {
+				global.Set("log-path", repoHost.Log.Path)
+			} else {
+				global.Set("log-path", fmt.Sprintf(naming.PGBackRestRepoLogPath, repo.Name))
+			}
```
benjaminjb: The logic here seems similar to `generateRepoHostLogPath`, though I suppose it might be a little awkward to reuse that func (which loops through repos) here (in a loop of repos). I suppose you could simplify by pulling out the logic from ll 784-790 and reusing it -- but I wonder if there might be a way to keep this legible while reducing even more duplication. I mean: maybe we can pull this out of the repo loop?

pkg/apis/postgres-operator.crunchydata.com/v1/pgbackrest_types.go
```diff
@@ -0,0 +1,17 @@
...
+// +kubebuilder:validation:XValidation:rule=`!has(self.log) || !has(self.log.path) || self.log.path.startsWith("/volumes/")`,message=`pgbackrest sidecar log path is restricted to an existing additional volume`
```
benjaminjb: Could this CEL work?
```suggestion
// +kubebuilder:validation:XValidation:rule=`self.?log.path.startsWith("/volumes/").orValue(true)`,message=`pgbackrest sidecar log path is restricted to an existing additional volume`
```
I'm still fuzzy on the `orValue` part specifically, but I guess my general question is: can we skip the progressive check of the path by using the optional marker `.?`?

benjaminjb: You know, I'm running into some issues there: if I put `orValue` at the end as an escape, I get an error from the `endsWith` func: `found no matching overload for 'endsWith' applied to 'optional_type(string).(string)'")`

So I figure, maybe I should the orValue as part of the string selection, like 

```
self.?log.path.orValue("SOME DEFAULT VALUE HERE")
```

But is that legible?

benjaminjb: It might actually be nice to use the optional `?` marker when the path is multiple steps, but for one step, I might just go with `has(path) && do_something(path)`

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
dsessler7: I figured any limit was better than no limit and I read somewhere that 4096 is usually the character limit for a directory in linux. Can't imagine anyone ever getting close to this.

internal/pgbackrest/config.go
dsessler7: will push a refactor commit in a few

pkg/apis/postgres-operator.crunchydata.com/v1/pgbackrest_types.go
dsessler7: I went with the `!has() || !has() || ...` approach because 1) at the time, it seemed that it was less complex (i.e. lower score on complexity budget) than the optional marker (`.?`) approach and 2) I personally find it more readible... That being said, Chris did say that he found a way to make the optional marker approach less complex (checking that something has a value (`hasValue()`) rather than giving a default empty value (`orValue()`)). The two of you seem more opinionated on how these CEL rules should be written than I am, so I'm fine with whatever.

benjaminjb: I'm pretty open / I think as we learn more we may come back to this, so what you have works for me.

Just to get things on paper, here's some options for the CEL rule:

```
!has(self.log) || !has(self.log.path) || self.log.path.startsWith("/volumes/")
!self.?log.path.hasValue() || self.log.path.startsWith("/volumes/")
self.?log.path.orValue("/volumes/").startsWith("/volumes/")
```

pkg/apis/postgres-operator.crunchydata.com/v1/pgbackrest_types.go
dsessler7: Latest commit has CEL validation to disallow setting log-path via global.

benjaminjb (commented):
LGTM, but still have that question about annotation & additional volumes

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
cbandy: Let's start smaller and go larger if anyone hits it. 🤔 256? 300?

📝 These are multiplied by 4x during cost estimation because UTF-8 characters can be up to 4 bytes.

pkg/apis/postgres-operator.crunchydata.com/v1/pgbackrest_types.go
cbandy: The last form is where I hit costs, but my type was a list `orValue([])`. The compiler might understand the cost of a string constant better than a list.

🤷🏻 I prefer the middle one; mushing the nested `has()` into a single something. All of the options are fine. 👍🏻 
```
!self.?log.path.hasValue() || self.log.path.startsWith("/volumes/")
```

pkg/apis/postgres-operator.crunchydata.com/v1/pgbackrest_types.go
```diff
@@ -0,0 +1,18 @@
...
+// +kubebuilder:validation:XValidation:rule=`!has(self.repoHost) || !has(self.repoHost.log) || !has(self.repoHost.log.path) || self.repoHost.volumes.additional.exists(x, self.repoHost.log.path.startsWith("/volumes/"+x.name))`,message=`repo host log path is restricted to an existing additional volume`
+// +kubebuilder:validation:XValidation:rule=`!has(self.jobs) || !has(self.jobs.log) || !has(self.jobs.log.path) || self.jobs.volumes.additional.exists(x, self.jobs.log.path.startsWith("/volumes/"+x.name))`,message=`backup jobs log path is restricted to an existing additional volume`
+// +kubebuilder:validation:XValidation:rule=`!has(self.global) || !('log-path' in self.global)`,message=`pgbackrest log-path must be set via the various log.path fields in the spec`
```
cbandy: 🤔 This last one might work in that same style, but I'm not certain.


```suggestion
// +kubebuilder:validation:XValidation:rule=`!self.?global["log-path"].hasValue()`,message=`pgbackrest log-path must be set via the various log.path fields in the spec`
```

internal/pgbackrest/config.go
```diff
@@ -771,3 +758,27 @@ func serverConfig(cluster *v1beta1.PostgresCluster) iniSectionSet {
...
+			// If the user has set a log path in the spec, use it.
+			// Otherwise, default to /pgbackrest/repo#/log
+			if cluster.Spec.Backups.PGBackRest.RepoHost != nil &&
+				cluster.Spec.Backups.PGBackRest.RepoHost.Log != nil &&
+				cluster.Spec.Backups.PGBackRest.RepoHost.Log.Path != "" {
+				pgBackRestLogPath = cluster.Spec.Backups.PGBackRest.RepoHost.Log.Path
+			} else {
+				pgBackRestLogPath = fmt.Sprintf(naming.PGBackRestRepoLogPath, repo.Name)
+			}
+			break
```
cbandy: ⛏️ I didn't notice this `break` at first. Early return here is easier for me to recognize as a search.


```suggestion
				return cluster.Spec.Backups.PGBackRest.RepoHost.Log.Path
			}
			return fmt.Sprintf(naming.PGBackRestRepoLogPath, repo.Name)
```

pkg/apis/postgres-operator.crunchydata.com/v1/pgbackrest_types.go
dsessler7: I believe they all (instance for sidecar, repo host, and backup job) have `/tmp`...

dsessler7: Could enforce using `/tmp/pgbackrest` subdir if logging to `/tmp`

pkg/apis/postgres-operator.crunchydata.com/v1/postgrescluster_types.go
```diff
@@ -26,6 +26,13 @@ import (
 //
 // TODO: Also check the above path against volume names: `i.?volumes.additional.hasValue() && i.volumes.additional.exists(directory.startsWith("/volumes/" + volume.name))`
 // https://github.com/kubernetes-sigs/controller-tools/pull/1270#issuecomment-3272211184
+//
+// # pgBackRest Logging
+//
+// +kubebuilder:validation:XValidation:fieldPath=`.backups.pgbackrest.log.path`,message=`all instances need an additional volume for pgbackrest sidecar to log in "/volumes"`,rule=`self.?backups.pgbackrest.log.path.optMap(v, !v.startsWith("/volumes") || self.instances.all(i, i.?volumes.additional.hasValue())).orValue(true)`
+//
+// TODO: Also check the above path against volume names: `i.?volumes.additional.hasValue() && i.volumes.additional.exists(a, v.startsWith("/volumes/" + a.name))`
+// https://github.com/kubernetes-sigs/controller-tools/pull/1270#issuecomment-3272211184
```
cbandy: 📝 I see this matches the TODO I left a few lines up. I was able to make that change in https://github.com/CrunchyData/postgres-operator/pull/4290 by setting max-items on instance sets. 🤔 It probably requires https://github.com/CrunchyData/postgres-operator/pull/4284, too.

benjaminjb (approved):

pkg/apis/postgres-operator.crunchydata.com/v1/pgbackrest_types.go
dsessler7: Spoke with Chris about this. Going to hold off on allowing logging to `/tmp`

PR #4267, opened 2025-09-05
Allow AutoGrow Tuning

jmckulk:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

**Other Information**:


pkg/apis/postgres-operator.crunchydata.com/v1/postgrescluster_types.go
jmckulk, line 1: Should these fields be added to the v1 api?

internal/pgbackrest/config.go
```diff
@@ -605,7 +606,39 @@ func getExternalRepoConfigs(repo v1beta1.PGBackRestRepo) map[string]string {
 // reloadCommand returns an entrypoint that convinces the pgBackRest TLS server
 // to reload its options and certificate files when they change. The process
 // will appear as name in `ps` and `top`.
-func reloadCommand(name string) []string {
+func reloadCommand(name string, repos []v1beta1.PGBackRestRepo) []string {
+	var repo1MaxGrow string
+	var repo2MaxGrow string
+	var repo3MaxGrow string
+	var repo4MaxGrow string
+	repo1Trigger := "75"
```
jmckulk: This default is set in the api so users should never be able to have an empty value if they define the autogrow fields.

However we still need to pass something into the script for cases where they enable autogrow by setting a limit but don't provide the autogrow spec.

I ended up putting the default in the go code but could also see having it as part of the bash script below. Open to opinions

internal/pgbackrest/reconcile.go
```diff
@@ -355,10 +355,11 @@ func addServerContainerAndVolume(
...
 		}
 	}
-
 	reloader := corev1.Container{
-		Name:            naming.ContainerPGBackRestConfig,
-		Command:         reloadCommand(naming.ContainerPGBackRestConfig),
+		Name: naming.ContainerPGBackRestConfig,
+		Command: reloadCommand(
+			naming.ContainerPGBackRestConfig,
+			cluster.Spec.Backups.PGBackRest.Repos),
```
jmckulk, line 362: We use a similar `reloadCommand` for both instance and pgbackrest reloader sidecars. This change introduces new parameters to both functions but they unfortunately don't match. The instance accepts two `VolumeClaimSpecWithAutoGrow` but the pgbackrest pod accepts a list of pgbackrest repos. 

Ideally we would pass the same types to each function because they are doing the same things 🤔 

Open to thoughts here too

internal/util/volumes.go
jmckulk, line 1: Open to other thoughts about where this function should be 💭

internal/controller/postgrescluster/instance_test.go
```diff
@@ -566,9 +568,19 @@ func TestAddPGBackRestToInstancePodSpec(t *testing.T) {
...
         newSize="$(((sizeInt / 2)+sizeInt))"
+        # Only compare with maxGrow if it is set (not empty)
+        if [[ -n "${maxGrow}" ]]; then
+            # check to see how much we would normally grow
+            sizeDiff=$((newSize - sizeInt))
+
+            # Compare the size difference to the maxGrow; if it is greater, cap it to maxGrow
+            if [[ ${sizeDiff} -gt ${maxGrow} ]]; then
+                newSize=$((sizeInt + maxGrow))
+            fi
```
tjmoore4, line 582: 📝 We'll want to add an exception for the `dupword` lint issue unless we can otherwise work around it. Happy to help with that if needed.

jmckulk: Sure - what would that look like?

internal/pgbackrest/config.go
tjmoore4: 🤔 It's probably fine to have this in both places, but perhaps the value should be stored in a constant so if it ever changes only one place will need to be updated.

internal/controller/postgrescluster/instance_test.go
tjmoore4: It will be something like this
https://github.com/CrunchyData/postgres-operator/blob/main/.golangci.yaml#L155-L157
but the path would need to be updated and the text might be something like
`Duplicate words (fi) found`

pkg/apis/postgres-operator.crunchydata.com/v1/postgrescluster_types.go
tjmoore4: 🤔 I think so? I am curious as to why `v1beta1` has this change
```
	if s.DataVolumeClaimSpec.AutoGrow != nil {
		s.DataVolumeClaimSpec.AutoGrow.Default()
	}
```
but `v1` doesn't?

internal/controller/postgrescluster/instance_test.go
jmckulk: I ended up adding a `settings.dupword.ignore` line for `fi`. Think that is sufficient?

pkg/apis/postgres-operator.crunchydata.com/v1/postgrescluster_types.go
jmckulk: Oh its probably a missed change. I'm a little unsure when this is necessary 🤔

internal/pgbackrest/config.go
jmckulk: hmm yeah i can see that 🤔

internal/pgbackrest/config.go
```diff
@@ -668,39 +710,44 @@ until read -r -t 5 -u "${fd}"; do
     pkill -HUP --exact --parent=0 pgbackrest
   then
     exec {fd}>&- && exec {fd}<> <(:||:)
-    stat --dereference --format='Loaded configuration dated %y' "${filename}"
+    stat --dereference --format='Loaded configuration dated %%y' "${filename}"
```
tjmoore4, line 713: ❓ Is this supposed to be here? Seems unrelated to the autogrow changes.

internal/pgbackrest/config.go
```diff
@@ -668,39 +710,44 @@ until read -r -t 5 -u "${fd}"; do
...
   elif
     { [[ "${directory}" -nt "/proc/self/fd/${fd}" ]] ||
       [[ "${authority}" -nt "/proc/self/fd/${fd}" ]]
     } &&
     pkill -HUP --exact --parent=0 pgbackrest
   then
     exec {fd}>&- && exec {fd}<> <(:||:)
-    stat --format='Loaded certificates dated %y' "${directory}"
+    stat --format='Loaded certificates dated %%y' "${directory}"
```
tjmoore4, line 721: ❓ Same here.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
```diff
@@ -149,6 +152,66 @@ func (spec *VolumeClaimSpec) AsPersistentVolumeClaimSpec() corev1.PersistentVolu
...
+	Trigger *int32 `json:"trigger,omitempty"`
+
+	// MaxGrow is the maximum size to which the volume can be automatically
+	// expanded. If not set, the volume will grow by 50% of the original size each
+	// time the Trigger threshold is exceeded.
+	// +optional
+	MaxGrow *resource.Quantity `json:"maxGrow,omitempty"`
+}
+
+func (spec *AutoGrowSpec) Default() {
```
tjmoore4, line 183: 🤔 Do we need this?

internal/pgbackrest/config.go
jmckulk: we moved from using a normal heredoc to using fmt.Sprintf - with that change i had to escape the character

tjmoore4 (approved):

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
jmckulk: I'll make a note to check on this later.

PR #4269, opened 2025-09-05
Refactor additional volume helpers

cbandy:
This moves `AddAdditionalVolumesToSpecifiedContainers` to the same package as `AddVolumeAndMountsToPod` so they can share code. At the same time, I renamed them to clarify their differences:

- `AddAdditionalVolumesToSpecifiedContainers` -> `AddAdditionalVolumesAndMounts`
- `AddVolumeAndMountsToPod` -> `AddCloudLogVolumeToPod`

The last commit is a stretch, but I noticed that the only field used of the PVC passed in was its Name. I switched that to take a string.

🙇🏻 ... and drop the warning per an ongoing conversation.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Other


internal/util/volumes_test.go
```diff
@@ -5,16 +5,268 @@
...
+		expectedInitContainers: `- name: startup
+  resources: {}
+- name: config
+  resources: {}`,
+		expectedVolumes: `- name: volumes-required
+  persistentVolumeClaim:
+    claimName: required`,
+		expectedMissing: []string{},
+	}, {
+		tcName: "multiple additional volumes",
```
dsessler7: I believe this is a duplicate test case

dsessler7 (approved):
Just need to remove that duplicate test case. Otherwise, LGTM

PR #4272, opened 2025-09-07
Allow indented YAML in tests

cbandy:
We use YAML literals to assert the contents of complex Kubernetes DTOs, but it is time-consuming to align those literals correctly with spaces. This PR adds logic to test helpers that allows us to indent YAML and indent it with tabs.

Rather than import any internal packages in pkg/apis, I copied one test helper there.

Existing tests continue to work because their YAML is already standard and valid. I did, however, convert the YAML in `config_types_test.go` to confirm that the helper works and that tab indentation is still readable.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Testing enhancement

### DIFF
--- a/internal/testing/cmp/cmp.go
+++ b/internal/testing/cmp/cmp.go
@@ -80,7 +80,9 @@ func MarshalMatches(actual any, expected string) Comparison {
 	if err != nil {
 		return func() gotest.Result { return gotest.ResultFromError(err) }
 	}
-	return gotest.DeepEqual(string(b), strings.Trim(expected, "\t\n")+"\n")
+	return gotest.DeepEqual(string(b), dedentLines(
+		strings.TrimLeft(strings.TrimRight(expected, "\t\n"), "\n"), 2,
+	))
 }
 
 // Regexp succeeds if value contains any match of the regular expression.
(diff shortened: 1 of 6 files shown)


pkg/apis/postgres-operator.crunchydata.com/v1beta1/config_types_test.go
```diff
@@ -20,14 +18,12 @@ func TestOptionalConfigMapKeyRefAsProjection(t *testing.T) {
...
-- key: two
-  path: three
-name: one
-		`)+"\n")
+		assert.Assert(t, MarshalsTo(out, `
+			items:
+			- key: two
+				path: three
+			name: one
+		`))
```
cbandy, line 26: 📝 Here's a diff between left-justified spaces and tabs indented relevant to nearby Go code.

internal/testing/cmp/cmp.go
```diff
@@ -89,3 +94,68 @@ func MarshalMatches(actual any, expected string) Comparison {
...
+	spaces := bytes.Repeat([]byte{' '}, 20*tabWidth)
+
+	for s := bufio.NewScanner(bytes.NewReader(data)); s.Scan(); {
+		line := slices.Clone(s.Bytes())
+		tabs := leadingTabs.Find(line)
+
+		// Replace any leading tabs with spaces when tabWidth is positive.
+		if need := tabWidth * len(tabs); need > 0 {
+			if need > len(spaces) {
+				spaces = bytes.Repeat(spaces, need/len(spaces)+1)
```
benjaminjb: Could we make spaces = need?

cbandy: `bytes.Repeat` repeats a slice, so this uses the existing slice of spaces to make more spaces. Like,

- (need) 12 / (length) 10 + 1 == 2 ⭢ 20 spaces
- (need) 22 / (length) 10 + 1 == 3 ⭢ 30 spaces

I considered making L106 "big enough," removing this "grow," and letting Go panic when `need` is silly-big.

🤔 L106 is already 20-tabs wide... I'm gonna see if the test suite balks.

cbandy: > Could we make spaces = need?

Yes! I discovered that `strings.Repeat` already does this kind of thing. I switched this function from `[]byte` to `string`, and it got less weird.

cbandy:
squashed

benjaminjb (approved):

PR #4278, opened 2025-09-09
Handle user-set pgBouncer logfile for OTEL, mkdir

benjaminjb:
pgBouncer allows users to set the log destination; when OTEL is enabled, we assume that destination is the default that we set. This PR patches that hole by setting OTEL to use the either the default or the user-set log destination.

This PR also handles creating the destination dir for the logfile.

This is a preliminary step on the way to validating the log config in v1.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [X] Other

Our pgBouncer OTEL settings assume a user hasn't set the logfile config, and won't work right if they do.

This updates OTEL with what the user sets.

**Other Information**:
Issues: [PGO-2565]

### DIFF
--- a/internal/collector/pgbouncer.go
+++ b/internal/collector/pgbouncer.go
@@ -9,6 +9,7 @@ import (
 	_ "embed"
 	"encoding/json"
 	"fmt"
+	"path/filepath"
 	"slices"
 	"strconv"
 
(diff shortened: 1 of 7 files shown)


dsessler7:
There is some overlap between this PR and https://github.com/CrunchyData/postgres-operator/pull/4252

benjaminjb:
> There is some overlap between this PR and #4252

Yeah, I wanted to break down that PR into discrete parts so we could focus on the things that we still have questions about, e.g., CEL validation rules + costs, using the Log config struct

dsessler7 (approved):
LGTM

PR #4281, opened 2025-09-10
Preserve repo volume status when error occurs during apply

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

If an error occurs when applying a repo host volume, any existing status will be lost.

Update to ensure that the repo volume status is preserved in
cases where there is an error when applying the repo volume. When
there is an error, the status is now preserved until the underlying
issue is corrected.

Also, updates certain variables make the behavior of the volume auto-grow code
unclear. This updates variables and adds a comment to better
indicate the behavior in question.

**Other Information**:
Issue: PGO-2654

### DIFF
--- a/internal/controller/postgrescluster/autogrow.go
+++ b/internal/controller/postgrescluster/autogrow.go
@@ -156,12 +156,13 @@ func (r *Reconciler) setVolumeSize(ctx context.Context, cluster *v1beta1.Postgre
 		// Otherwise, if the feature gate is not enabled, do not autogrow.
 	} else if feature.Enabled(ctx, feature.AutoGrowVolumes) {
 
-		// determine the appropriate volume request based on what's set in the status
-		if dpv, err := getDesiredVolumeSize(
+		// Determine the appropriate volume request based on what's set in the status.
+		// Note: request size set by reference.
+		if badDesiredVolumeRequest, err := getDesiredVolumeSize(
 			cluster, volumeType, host, volumeRequestSize,
 		); err != nil {
 			log.Error(err, "For "+cluster.Name+"/"+host+
-				": Unable to parse "+volumeType+" volume request: "+dpv)
+				": Unable to parse "+volumeType+" volume request: "+badDesiredVolumeRequest)
 		}
 
 		// If the volume request size is greater than or equal to the limit and the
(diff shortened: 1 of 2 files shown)


internal/controller/postgrescluster/autogrow.go
```diff
@@ -160,12 +160,13 @@ func (r *Reconciler) setVolumeSize(ctx context.Context, cluster *v1beta1.Postgre
 		// Otherwise, if the feature gate is not enabled, do not autogrow.
 	} else if feature.Enabled(ctx, feature.AutoGrowVolumes) {
 
-		// determine the appropriate volume request based on what's set in the status
-		if dpv, err := getDesiredVolumeSize(
+		// Determine the appropriate volume request based on what's set in the status.
+		// Note: request size set by reference.
+		if badDesiredVolumeRequest, err := getDesiredVolumeSize(
```
jmckulk, line 161: why is the output here always "bad" ❓

tjmoore4: Because that value is only ever returned if `resource.ParseQuantity` was unable to parse it.  It's only use is for the error logging below.

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -2990,19 +2989,26 @@ func getRepoHostStatus(repoHost *appsv1.StatefulSet) *v1beta1.RepoHostStatus {
...
+	// the existing status is preserved.
+	for repoName, rv := range repoVolumes {
 		newRepoVolStatus := true
-		repoName := rv.Labels[naming.LabelPGBackRestRepo]
 		for _, rs := range repoStatus {
+			// Preserve the previous status if it exists and the apply failed.
+			if rs.Name == repoName && rv == nil {
+				updatedRepoStatus = append(updatedRepoStatus, rs)
+				newRepoVolStatus = false
+				continue
```
andrewlecuyer: ⛏️ We could just `break` out of this loop here, right?

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -2990,19 +2989,26 @@ func getRepoHostStatus(repoHost *appsv1.StatefulSet) *v1beta1.RepoHostStatus {
...
-	for _, rv := range repoVolumes {
+	// repository volumes. If there was a problem with the volume when an apply was attempted,
+	// the existing status is preserved.
+	for repoName, rv := range repoVolumes {
 		newRepoVolStatus := true
-		repoName := rv.Labels[naming.LabelPGBackRestRepo]
 		for _, rs := range repoStatus {
+			// Preserve the previous status if it exists and the apply failed.
+			if rs.Name == repoName && rv == nil {
+				updatedRepoStatus = append(updatedRepoStatus, rs)
```
andrewlecuyer, line 2994: ❓Should we be clearing anything out of the status here?

For instanc, when `rv` is `nil` we know the PVC intent didn't apply, so should be remove any status that we set based on a value from `rv`?  It looks like this would specifically include `VolumeName` and `Bound`.

tjmoore4: My thinking was that, if an apply didn't take, we effectively pause updates to the status such that no status changes are made until the apply issues are sorted. Maybe I'm misunderstanding or missing something though?

internal/controller/postgrescluster/pgbackrest.go
tjmoore4: Yes, that seems right. We're checking against the repo name and that only occurs once, so a break would be fine since the rest don't need to be checked. I can make that change, thanks.

PR #4282, opened 2025-09-10
Make the default Postgres log directory outside its data directory

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature

When unspecified, Postgres logs to the `log` directory inside its data directory. This is difficult to coordinate with the OpenTelemetry Collector and can lose log files during replication creation and major upgrades. 

When unspecified, Postgres logs outside the data directory on one of the attached persistent volumes. The allowed values for the `log_directory` parameter have been reduced to a handful of absolute path prefixes to volumes that are also in the spec.

Users that want to continue to log to the data directory can/must set the `log_directory` parameter to `log`. That value is safe to assign even before this change.

**Other Information**:

Issue: PGO-2558

### DIFF
--- a/internal/postgres/validation.go
+++ b/internal/postgres/validation.go
@@ -9,13 +9,17 @@ import (
 	"regexp"
 	"strings"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+
+	"github.com/crunchydata/postgres-operator/internal/text"
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
 
 // SanitizeParameters transforms parameters so they are safe for Postgres in cluster.
-func SanitizeParameters(cluster *v1beta1.PostgresCluster, parameters *ParameterSet) {
+func SanitizeParameters(cluster *v1beta1.PostgresCluster, parameters *ParameterSet, recorder record.EventRecorder) {
 	if v, ok := parameters.Get("log_directory"); ok {
-		parameters.Add("log_directory", sanitizeLogDirectory(cluster, v))
+		parameters.Add("log_directory", sanitizeLogDirectory(cluster, v, recorder))
 	}
 }
 
@@ -52,7 +56,7 @@ var sensitiveRelativePath = regexp.MustCompile(
 // Otherwise, it returns the absolute path to a good "log_directory" value.
 //
 // https://www.postgresql.org/docs/current/runtime-config-logging.html#GUC-LOG-DIRECTORY
-func sanitizeLogDirectory(cluster *v1beta1.PostgresCluster, input string) string {
+func sanitizeLogDirectory(cluster *v1beta1.PostgresCluster, input string, recorder record.EventRecorder) string {
 	directory := path.Clean(input)
 
 	// [path.Clean] leaves leading parent directories. Eliminate these as a security measure.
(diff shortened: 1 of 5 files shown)


pkg/apis/postgres-operator.crunchydata.com/v1/postgres_types.go
```diff
@@ -0,0 +1,234 @@
...
+	//
+	// More info: https://www.postgresql.org/docs/current/auth-pg-hba-conf.html
+	// ---
+	// +kubebuilder:validation:MaxItems=10
+	// +listType=atomic
+	// +optional
+	Rules []PostgresHBARuleSpec `json:"rules,omitempty"`
+}
+
+type PostgresConfigSpec struct {
```
benjaminjb, line 12: We're only using this struct right now; but are you thinking it's easier to copy over all the v1beta1 structs in this file so that we can keep them aligned (up until we drop v1beta1 or let them become unaligned)?

cbandy: I wasn't thinking at all! Which seems better to you presently?

internal/postgres/validation.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+// SanitizeParameters transforms parameters so they are safe for Postgres in cluster.
+func SanitizeParameters(cluster *v1beta1.PostgresCluster, parameters *ParameterSet) {
+	if v, ok := parameters.Get("log_directory"); ok {
+		parameters.Add("log_directory", sanitizeLogDirectory(cluster, v))
+	}
+}
+
+// sensitiveAbsolutePath is used by [sanitizeLogDirectory].
```
benjaminjb: Might've helped me reading at first to understand `sensitiveAbsolutePath` and `sensitiveRelativePath` were disallowed paths.

pkg/apis/postgres-operator.crunchydata.com/v1/postgres_types.go
benjaminjb: At first I was going to say, "bring over everything so we can keep things aligned" and then I realized that the reason to have a v1 is to change things. Here we have a new CEL rule on parameters, and an editing of existing CEL rules (mostly reorganizing messages as the last part, I think) for `PostgresHBARuleSpec` -- but then we're not using the `v1.PostgresHBARuleSpec` anywhere...

So now I lean lightly towards "just bring over the things that have changed."

internal/testing/validation/postgrescluster/postgres_config_test.go
```diff
@@ -82,6 +102,194 @@ func TestPostgresConfigParametersV1(t *testing.T) {
...
+					]`,
+					valid:   false,
+					message: `"/pgwal/logs/postgres"`,
+				},
+
+				// Directories inside /volumes are acceptable, but every instance set needs additional volumes.
+				//
+				// TODO(validation): This could be more precise and check the directory name of each additional
+				// volume, but Kubernetes 1.33 incorrectly estimates the cost of volume.name:
+				// https://github.com/kubernetes-sigs/controller-tools/pull/1270#issuecomment-3272211184
```
cbandy, line 233: 🤔 Is this issue fixed by #4284?

dsessler7 (approved):

benjaminjb (approved):

internal/testing/validation/postgrescluster/postgres_config_test.go
cbandy: I tried it but still over budget.

PR #4284, opened 2025-09-14
Use gojq as a library in Go to adjust CRDs

cbandy:
There are valid schemas that controller-gen markers cannot produce. There are bugs in Kubernetes' interpretation of valid schemas.

This uses `go generate` and a jq filter to find and correct systemic issues throughout our CRDs.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Bug fix
 - [x] Other


.golangci.yaml
```diff
@@ -71,7 +71,7 @@ linters:
     depguard:
       rules:
         everything:
-          files: ['$all']
+          files: ["$all"]
```
benjaminjb: ❓ I did mean to ask what was behind this change?

internal/crd/post-process.go
```diff
@@ -0,0 +1,64 @@
...
+func main() {
+	cwd := need(os.Getwd())
+	dir := filepath.Join(cwd, "..", "..", "config", "crd", "bases")
+	query := "post-process.jq"
+
+	slog.Info("Reading", "file", query)
+	code := need(gojq.Compile(
+		need(gojq.Parse(string(need(os.ReadFile(query))))),
+		gojq.WithModuleLoader(gojq.NewModuleLoader([]string{"$ORIGIN/../lib"})),
+		gojq.WithEnvironLoader(os.Environ),
```
benjaminjb: ⛏️ Can we drop the EnvironLoader?

.golangci.yaml
cbandy: Something in my editor, though I thought I had disabled YAML formatting... I'll revert these lines.

PR #4288, opened 2025-09-15
Add a Make target that runs crd-schema-checker

cbandy:
**Checklist:**

 
---

This tool helped me understand the cost of CEL validation rules as I was writing them. ~~The second and third commits are optional; I followed some of the recommendations emitted by the tool.~~ Edit: Removed these commits.

The output for PGUpgrades looks similar to:

```
go run github.com/openshift/crd-schema-checker/cmd/crd-schema-checker@latest check-manifests --disabled-validators 'NoBools,NoMaps' --new-crd-filename 'config/crd/bases/postgres-operator.crunchydata.com_pgupgrades.yaml' 2>&1 | awk -f hack/check-manifests.awk                                                                                                                          
ERROR: "ListsMustHaveSSATags": crd/pgupgrades.postgres-operator.crunchydata.com version/v1beta1 field/^.spec.imagePullSecrets must set x-kubernetes-list-type                                 
info: "CostBudget": ^.spec: Field has a maximum cardinality of 1.                                                                                                                             
info: "CostBudget": ^.spec: Rule 0 raw cost is 5. Estimated total cost of 5. Rule is 0.00% of allowed budget, 10M.                                                                            
info: "CostBudget": ^.spec: Field has a maximum cardinality of 1.                                                                                                                             
info: "CostBudget": ^.spec: Rule 1 raw cost is 19. Estimated total cost of 19. Rule is 0.00% of allowed budget, 10M.                                                                          
info: "CostBudget": ^.spec: Field has a maximum cardinality of 1.                                                                                                                             
info: "CostBudget": ^.spec: Rule 2 raw cost is 20. Estimated total cost of 20. Rule is 0.00% of allowed budget, 10M.
exit status 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ CONTROLLER ?= $(GO) tool sigs.k8s.io/controller-tools/cmd/controller-gen
 # Run tests using the latest tools.
 CHAINSAW ?= $(GO) run github.com/kyverno/chainsaw@latest
 CHAINSAW_TEST ?= $(CHAINSAW) test
+CRD_CHECKER ?= $(GO) run github.com/openshift/crd-schema-checker/cmd/crd-schema-checker@latest
 ENVTEST ?= $(GO) run sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 KUTTL ?= $(GO) run github.com/kudobuilder/kuttl/cmd/kubectl-kuttl@latest
 KUTTL_TEST ?= $(KUTTL) test
@@ -147,6 +148,12 @@ check: ## Run basic go tests with coverage output
 check: get-pgmonitor
 	QUERIES_CONFIG_DIR="$(CURDIR)/${QUERIES_CONFIG_DIR}" $(GO_TEST) -cover ./...
 
+# Informational only; no criteria to enforce at this time.
+.PHONY: check-crd
+check-crd:
+	$(foreach CRD, $(wildcard config/crd/bases/*.yaml), \
+		$(CRD_CHECKER) check-manifests --new-crd-filename '$(CRD)' 2>&1 | awk -f hack/check-manifests.awk $(newline))
+
 # Available versions: curl -s 'https://storage.googleapis.com/kubebuilder-tools/' | grep -o '<Key>[^<]*</Key>'
 # - KUBEBUILDER_ATTACH_CONTROL_PLANE_OUTPUT=true
 .PHONY: check-envtest
@@ -253,3 +260,9 @@ generate-rbac: ## Generate RBAC
 		) rbac:roleName='postgres-operator' $(\
 		) paths='./cmd/...' paths='./internal/...' $(\
 		) output:dir='config/rbac' # {directory}/role.yaml
+
+# https://www.gnu.org/software/make/manual/make.html#Multi_002dLine
+define newline
+
+
+endef
--- a/hack/check-manifests.awk
+++ b/hack/check-manifests.awk
@@ -1,0 +1,26 @@
+# Copyright 2025 Crunchy Data Solutions, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+## TODO: Exit successfully only when there are no errors.
+#/^ERROR:/ { rc = 1 }
+#END { exit rc }
+
+# Shorten these frequent messages about validation rules.
+/The maximum allowable value is 10000000[.]/ {
+	sub(/ The maximum allowable value is 10000000./, "")
+	sub(/ allowed budget/, "&, 10M")
+}
+
+# These are informational, but "MustNot" sounds like something is wrong.
+/^info: "MustNotExceedCostBudget"/ {
+	sub(/"MustNotExceedCostBudget"/, "\"CostBudget\"")
+}
+
+# Color errors and warnings when attached to a terminal.
+ENVIRON["MAKE_TERMOUT"] != "" {
+	sub(/^ERROR:/, "\033[0;31m&\033[0m")
+	sub(/^Warning:/, "\033[1;33m&\033[0m")
+}
+
+{ print }


Makefile
```diff
@@ -253,3 +262,8 @@ generate-rbac: ## Generate RBAC
 		) rbac:roleName='postgres-operator' $(\
 		) paths='./cmd/...' paths='./internal/...' $(\
 		) output:dir='config/rbac' # {directory}/role.yaml
+
+define NL
+
+
+endef
```
dsessler7: What is this?

benjaminjb: Defining a newline? Does the `awk` on l. 156 need a newline?

config/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml
```diff
@@ -1307,6 +1307,7 @@ spec:
                           type: object
                       type: object
                     type: array
+                    x-kubernetes-list-type: atomic
```
benjaminjb: Why are all these lists getting flagged for atomic? Is that the default?

Makefile
```diff
@@ -147,6 +148,14 @@ check: ## Run basic go tests with coverage output
 check: get-pgmonitor
 	QUERIES_CONFIG_DIR="$(CURDIR)/${QUERIES_CONFIG_DIR}" $(GO_TEST) -cover ./...
 
+# Informational only; no criteria to enforce at this time.
+.PHONY: check-crd
+check-crd:
+	$(foreach CRD, $(wildcard config/crd/bases/*.yaml), \
+		$(CRD_CHECKER) check-manifests --disabled-validators 'NoBools,NoMaps' --new-crd-filename '$(CRD)' 2>&1 | \
```
benjaminjb: Do we need a note about why we're turning off these validators?

benjaminjb (approved):
No blockers

Makefile
cbandy: > What is this?
> Defining a newline?

Yeah, this is a Make variable that contains one newline.

> Does the `awk` on l. 156 need a newline?

Without the newline, Make prints all the `go run` commands at once, then four outputs from the tool come out. With this, each command and its output are separated like this:

```bash
go run github.com/openshift/crd-schema-checker/cmd/crd-schema-checker@latest check-manifests --new-crd-filename 'config/crd/bases/postgresoperator.crunchydata.com_crunchybridgeclusters.yaml' 2>&1 | awk -f hack/check-manifests.awk
##
## wall of log lines
##
go run github.com/openshift/crd-schema-checker/cmd/crd-schema-checker@latest check-manifests --new-crd-filename 'config/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml' 2>&1 | awk -f hack/check-manifests.awk
##
## wall of log lines
##
go run github.com/openshift/crd-schema-checker/cmd/crd-schema-checker@latest check-manifests --new-crd-filename 'config/crd/bases/postgres-operator.crunchydata.com_pgupgrades.yaml' 2>&1 | awk -f hack/check-manifests.awk
##
## wall of log lines
##
go run github.com/openshift/crd-schema-checker/cmd/crd-schema-checker@latest check-manifests --new-crd-filename 'config/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml' 2>&1 | awk -f hack/check-manifests.awk
##
## wall of log lines
##
```

🤔 What's a better way to present / read this stuff?

config/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml
cbandy: > Why are all these lists getting flagged for atomic?

Because the destination for these values is an atomic list in Kubernetes; https://pkg.go.dev/k8s.io/api/core/v1#PodSpec.Tolerations

```go
	// If specified, the pod's tolerations.
	// +optional
	// +listType=atomic
	Tolerations [][Toleration](https://pkg.go.dev/k8s.io/api/core/v1#Toleration)
```

> Is that the default?

Not the way I think you're asking, but also yes, that's my understanding.

https://docs.k8s.io/reference/using-api/server-side-apply#custom-resources-and-server-side-apply

> By default, Server-Side Apply treats custom resources as unstructured data. All keys are treated the same as struct fields, and all lists are considered atomic.

Makefile
cbandy: 🤔 There was a lot of noise from upstream types, but that can be something we decide later. I'll take this arg out.

config/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml
cbandy: > Why are all these lists getting flagged for atomic?

I've added the `+listType=atomic` marker to our Go structs in `/pkg`.

cbandy:
> Just thinking aloud here, but this make target isn't wired up to any other make target, so we might want to write / think about some guidance about when to use it. It's not just for checking CEL rule budget, yeah?

I don't understand the tool enough to give any guidance, really. I used it to check CEL budgets, but I have no idea how accurate that was (or if accuracy even matters).

When we understand the tool better, we can configure it and wire it up as a blocking check.

- I'm curious about the old → new concept. It sounds like it can detect not-good _changes_ to a CRD.
- [This file](https://github.com/openshift/crd-schema-checker/blob/c313b6407231f6d1aabfb602f0823708f8c6b9fb/pkg/defaultcomparators/interfaces.go) has a list of checks and some ideas for others. Are these all enabled?
- Our strategy for v1 relies on validation ratcheting. What are these [calls to `RatchetCompare`](https://github.com/openshift/crd-schema-checker/search?q=RatchetCompare)?

config/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml
cbandy: I've backed out these changes.

cbandy:
I've reverted the CRD changes. This PR adds an easy way to invoke the schema checker.

Makefile
cbandy: 📝 https://www.github.com/openshift/crd-schema-checker/issues/44

benjaminjb (approved):

PR #4294, opened 2025-09-17
Adjust cloud backup volume behavior such that additional volumes take…

dsessler7:
… precedence when there is a naming collision with the pgbackrest-cloud-log-volume annotation. If annotation is present and no log path is given via the spec, use the log path that the annotation logic yields. Add/adjust tests appropriately.

**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Currently, if the cloud backup log volume annotation is present and its value is the same as the name of an additional volume, we attempt to mount both which results in the job not rolling out due to duplicate volume mounts.

Now, if the cloud backup log volume annotation is present and its value is the same as the name of an additional volume, we will not mount the volume specified by the annotation, we will only mount the additional volume. If log.path is not set in the spec for backup jobs and the annotation is present and a cloud backup job is created, it will use the log path determined by the annotation logic.

**Other Information**:

### DIFF
--- a/internal/controller/postgrescluster/pgbackrest.go
+++ b/internal/controller/postgrescluster/pgbackrest.go
@@ -892,9 +892,29 @@ func (r *Reconciler) generateBackupJobSpecIntent(ctx context.Context, postgresCl
 		jobSpec.Template.Spec.SecurityContext = postgres.PodSecurityContext(postgresCluster)
 		pgbackrest.AddConfigToCloudBackupJob(postgresCluster, &jobSpec.Template)
 
-		// Mount the PVC named in the "pgbackrest-cloud-log-volume" annotation, if any.
+		// If the "pgbackrest-cloud-log-volume" annotation has a value, check if it is the
+		// same as any of the additional volume names. If there is a collision of names,
+		// create a warning event. If there is no name collision, mount the volume referenced
+		// by the annotation.
 		if logVolume := postgresCluster.Annotations[naming.PGBackRestCloudLogVolume]; logVolume != "" {
-			util.AddCloudLogVolumeToPod(&jobSpec.Template.Spec, logVolume)
+			var collisionFound bool
+			if jobs != nil && jobs.Volumes != nil {
+				for _, volume := range jobs.Volumes.Additional {
+					if volume.Name == logVolume {
+						collisionFound = true
+						r.Recorder.Event(postgresCluster, corev1.EventTypeWarning,
+							"DuplicateCloudBackupVolume", "The volume name specified in the "+
+								"pgbackrest-cloud-log-volume annotation is the same as one "+
+								"specified in spec.backups.pgbackrest.jobs.volumes.additional. "+
+								"Cannot mount duplicate volume names. Defaulting to the "+
+								"additional volume.")
+						break
+					}
+				}
+			}
+			if !collisionFound {
+				util.AddCloudLogVolumeToPod(&jobSpec.Template.Spec, logVolume)
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -892,9 +892,29 @@ func (r *Reconciler) generateBackupJobSpecIntent(ctx context.Context, postgresCl
...
+			if jobs != nil && jobs.Volumes != nil && len(jobs.Volumes.Additional) > 0 {
+				for _, volume := range jobs.Volumes.Additional {
+					if volume.Name == logVolume {
+						collisionFound = true
+						r.Recorder.Event(postgresCluster, corev1.EventTypeWarning,
+							"DuplicateCloudBackupVolume", "The volume name specified in the "+
+								"pgbackrest-cloud-log-volume annotation is the same as one "+
+								"specified in spec.backups.pgbackrest.jobs.volumes.additional. "+
+								"Cannot mount duplicate volume names. Defaulting to the "+
+								"additional volume.")
```
benjaminjb, line 910:
```suggestion
								"Cannot mount duplicate volume names. Mounting volume once.")
```

I'm not crazy about ^, but seems funny to me to say "I can't mount this identical volume twice, so I'm going to use the one at X" -- when it would be the same if you were mounting the volume at Y.

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -892,9 +892,29 @@ func (r *Reconciler) generateBackupJobSpecIntent(ctx context.Context, postgresCl
...
 
-		// Mount the PVC named in the "pgbackrest-cloud-log-volume" annotation, if any.
+		// If the "pgbackrest-cloud-log-volume" annotation has a value, check if it is the
+		// same as any of the additional volume names. If there is a collision of names,
+		// create a warning event. If there is no name collision, mount the volume referenced
+		// by the annotation.
 		if logVolume := postgresCluster.Annotations[naming.PGBackRestCloudLogVolume]; logVolume != "" {
-			util.AddCloudLogVolumeToPod(&jobSpec.Template.Spec, logVolume)
+			var collisionFound bool
+			if jobs != nil && jobs.Volumes != nil && len(jobs.Volumes.Additional) > 0 {
```
benjaminjb: @cbandy I feel like you've been leaning towards "skip the length check, just range" (ranging over an empty list is the same as skipping the logic). Are there any cases where you wouldn't do that?

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -3340,21 +3360,22 @@ func authorizeBackupRemovalAnnotationPresent(postgresCluster *v1beta1.PostgresCl
...
 func getCloudLogPath(postgresCluster *v1beta1.PostgresCluster) string {
 	cloudLogPath := ""
-	if logVolume := postgresCluster.Annotations[naming.PGBackRestCloudLogVolume]; logVolume != "" {
-		cloudLogPath = "/volumes/" + logVolume
-	} else if postgresCluster.Spec.Backups.PGBackRest.Jobs != nil &&
+	if postgresCluster.Spec.Backups.PGBackRest.Jobs != nil &&
 		postgresCluster.Spec.Backups.PGBackRest.Jobs.Log != nil &&
 		postgresCluster.Spec.Backups.PGBackRest.Jobs.Log.Path != "" {
+		// TODO: I know it should be caught by CEL validation, but is it worthwhile to also
+		// check that Log.Path ~= "/volumes/" + existingAdditionalVolume.name here??
```
benjaminjb, line 3374: That's a good question. I'm usually very defensive ("before we pass the info, let's make sure it's correct -- ok, now that we've passed the info, let's check it again"), but here I feel like it really could go either way, and I'm leaning towards not running the check here:

- if we change the CEL rules ("you can now attach a tmp and log there"), then we change the admission validation rules and don't need to remember all the places where the value is being used.
- if someone gets around the CEL validation, that's good information for us on our validation. (Although one way they can do that is always: just remove the CEL lines before installing the CRD.)

internal/controller/postgrescluster/pgbackrest.go
benjaminjb: ah, wait, you just explained this distinction to me: what would collide is the name / path of the volume, not the actual underlying PVC. So the volumes being discussed might not be the same, so we are preferring the PVC specified in the additional volumes.

benjaminjb: So, disregard my comment above!

internal/controller/postgrescluster/pgbackrest.go
dsessler7: Agreed. Maybe I'll just leave the TODO for now

internal/controller/postgrescluster/pgbackrest.go
dsessler7: It probably makes sense to remove it in this scenario because all we are doing inside the `if` block is ranging over `jobs.Volumes.Additional` and if it is empty that is a zero cost action.

In other places in the code (like just below this block where we mount the additional volumes), it might make sense to keep the length check just to save us the function call (`util.AddAdditionalVolumesAndMounts`) when it's empty.

PR #4296, opened 2025-09-19
Change Reconciler implementations to ObjectReconciler

cbandy:
Each implementation was doing the first fetch of its main object a bit differently. The controller-runtime module can do this since v0.17.0 and Go generics.

**Checklist:**

 
**Type of Changes:**

 
 - [x] Other

### DIFF
--- a/internal/controller/standalone_pgadmin/controller.go
+++ b/internal/controller/standalone_pgadmin/controller.go
@@ -14,10 +14,12 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/record"
-	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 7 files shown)


internal/controller/standalone_pgadmin/controller.go
```diff
@@ -78,39 +80,29 @@ func ManagedReconciler(m ctrl.Manager) error {
...
 		).
 		Watches(
 			&corev1.Secret{},
-			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, secret client.Object) []ctrl.Request {
+			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, secret client.Object) []reconcile.Request {
 				return runtime.Requests(reconciler.findPGAdminsForSecret(ctx, client.ObjectKeyFromObject(secret))...)
 			}),
 		).
-		Complete(reconciler))
+		Complete(reconcile.AsReconciler(kubernetes, reconciler)))
```
benjaminjb, line 93: OK, so `AsReconciler` returns a reconciler from an ObjectReconciler, which `reconciler` is now because our `reconcile` func takes the object rather than a req.

Which is better because it offloads some generic logic from our reconcile to some k8s standard functions. (Also simplifies some tests where we were having to build a request.)

Did I get it? Am I missing something?

cbandy: 👍🏻 That is all of it.

benjaminjb (approved):

PR #4299, opened 2025-09-19
Run Shellcheck on instance reload command

jmckulk:
Shellcheck was being run on the reload command for pgBackRest. During the auto-grow work, we noticed that the reload command for the instance was not being checked. Since we had to update the auto-grow bash logic to pass Shellcheck for pgBackRest, it seemed appropriate to enable the checks for the instance as well.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

**Other Information**:


internal/postgres/config.go
```diff
@@ -313,27 +313,29 @@ manageAutogrowAnnotation() {
   local trigger=$2
   local maxGrow=$3
 
-  size=$(df --block-size=M /"${volume}" | awk 'FNR == 2 {print $2}')
-  use=$(df /"${volume}" | awk 'FNR == 2 {print $5}')
+  size=$(df --block-size=M /"${volume}")
```
tjmoore4, line 327: 🔧 Since this is in a different package, I think it would be worth updating the comments to explain what the script does here. I think it should be more or less a copy/paste though.
See https://github.com/CrunchyData/postgres-operator/blob/main/internal/pgbackrest/config.go#L640-L649

tjmoore4 (changes requested):
Looks good, I think we should just add a comment about the script updates.

PR #4300, opened 2025-09-19
Updates to support changes starting in pgAdmin 9.3

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

**Other Information**:

### DIFF
--- a/internal/controller/standalone_pgadmin/users.go
+++ b/internal/controller/standalone_pgadmin/users.go
@@ -258,6 +258,8 @@ cd $PGADMIN_DIR
 					log.Error(err, "PodExec failed: ")
 					intentUsers = append(intentUsers, existingUser)
 					continue
+				} else if strings.Contains(strings.TrimSpace(stderr.String()), "UserWarning: pkg_resources is deprecated as an API") {
+					log.Info(stderr.String())
 				} else if strings.TrimSpace(stderr.String()) != "" {
 					log.Error(errors.New(stderr.String()), fmt.Sprintf("pgAdmin setup.py error for %s: ",
 						intentUser.Username))
@@ -292,7 +294,9 @@ cd $PGADMIN_DIR
 				log.Error(err, "PodExec failed: ")
 				continue
 			}
-			if strings.TrimSpace(stderr.String()) != "" {
+			if strings.Contains(strings.TrimSpace(stderr.String()), "UserWarning: pkg_resources is deprecated as an API") {
+				log.Info(stderr.String())
+			} else if strings.TrimSpace(stderr.String()) != "" {
 				log.Error(errors.New(stderr.String()), fmt.Sprintf("pgAdmin setup.py error for %s: ",
 					intentUser.Username))
 				continue
--- a/internal/controller/standalone_pgadmin/users.go
+++ b/internal/controller/standalone_pgadmin/users.go
@@ -86,10 +86,20 @@ func (r *PGAdminReconciler) reconcilePGAdminUsers(ctx context.Context, pgadmin *
 	if pgadmin.Status.MajorVersion == 0 || pgadmin.Status.MinorVersion == "" ||
 		pgadmin.Status.ImageSHA != pgAdminImageSha {
 
-		pgadminMinorVersion, err := r.reconcilePGAdminVersion(ctx, podExecutor)
-		if err != nil {
+		// exec into the pgAdmin pod and retrieve the pgAdmin minor version
+		script := fmt.Sprintf(`
+PGADMIN_DIR=%s
+cd $PGADMIN_DIR && python3 -c "import config; print(config.APP_VERSION)"
+`, pgAdminDir)
+
+		var stdin, stdout, stderr bytes.Buffer
+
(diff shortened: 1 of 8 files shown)


internal/controller/standalone_pgadmin/users.go
```diff
@@ -80,40 +80,55 @@ func (r *PGAdminReconciler) reconcilePGAdminUsers(ctx context.Context, pgadmin *
...
 }
 
-// reconcilePGAdminMajorVersion execs into the pgAdmin pod and retrieves the pgAdmin major version
-func (r *PGAdminReconciler) reconcilePGAdminMajorVersion(ctx context.Context, exec Executor) (int, error) {
+// reconcilePGAdminVersion execs into the pgAdmin pod and retrieves the pgAdmin minor version
+func (r *PGAdminReconciler) reconcilePGAdminVersion(ctx context.Context, exec Executor) (string, error) {
 	script := fmt.Sprintf(`
 PGADMIN_DIR=%s
-cd $PGADMIN_DIR && python3 -c "import config; print(config.APP_RELEASE)"
+cd $PGADMIN_DIR && python3 -c "import config; print(config.APP_VERSION)"
```
cbandy: 🤔 It looks like the result of this is only used when reconciling users. Can we move this into the commands there?

📝 It looks like pgAdmin 8.4 and newer have a [version](https://github.com/pgadmin-org/pgadmin4/blob/REL-9_8/web/version.py) module with just these constants.

tjmoore4: I'll look into whether moving the commands makes sense. My guess is that would probably be fine as a refactor.

internal/controller/standalone_pgadmin/users.go
```diff
@@ -171,10 +186,23 @@ cd $PGADMIN_DIR
 	for _, user := range existingUsersArr {
 		existingUsersMap[user.Username] = user
 	}
+
+	var olderThan9_3 bool
+	versionFloat, err := strconv.ParseFloat(pgadmin.Status.MinorVersion, 32)
```
benjaminjb: why bitsize 32 here but 64 when we retrieve the number to put into the status?

internal/controller/standalone_pgadmin/users.go
```diff
@@ -171,10 +186,23 @@ cd $PGADMIN_DIR
...
+	}
+	if versionFloat < 9.3 {
+		olderThan9_3 = true
+	}
+
 	intentUsers := []pgAdminUserForJson{}
 	for _, user := range pgadmin.Spec.Users {
 		var stdin, stdout, stderr bytes.Buffer
-		typeFlag := "--nonadmin"
+		typeFlag := "--role User"
```
benjaminjb, line 195: Comment maybe with a link to >9.3 changes?

internal/controller/standalone_pgadmin/users.go
```diff
@@ -264,7 +294,9 @@ cd $PGADMIN_DIR
 				log.Error(err, "PodExec failed: ")
 				continue
 			}
-			if strings.TrimSpace(stderr.String()) != "" {
+			if strings.Contains(strings.TrimSpace(stderr.String()), "UserWarning: pkg_resources is deprecated as an API") {
+				log.Info(stderr.String())
+			} else if strings.TrimSpace(stderr.String()) != "" {
 				log.Error(errors.New(stderr.String()), fmt.Sprintf("pgAdmin setup.py error for %s: ",
```
benjaminjb: Slightly out of scope for your work, but could you add something to differentiate between the add-user error here and the update-user error on line 264

internal/controller/standalone_pgadmin/users.go
```diff
@@ -230,6 +258,8 @@ cd $PGADMIN_DIR
 					log.Error(err, "PodExec failed: ")
 					intentUsers = append(intentUsers, existingUser)
 					continue
+				} else if strings.Contains(strings.TrimSpace(stderr.String()), "UserWarning: pkg_resources is deprecated as an API") {
```
benjaminjb, line 255: Is it worth commenting on these lines with the deprecation warning, re: what python version we were seeing this error?

internal/controller/standalone_pgadmin/users.go
tjmoore4: Short version: That can probably be a 64, I'll make that change.

Longer version: I originally used 32 because I was intending to store the value directly into the status as a float. However, that's counter-indicated due to issues in float implementation and a string value is recommended instead. When the conversion happens, that function allows for 32 to be set so that the result will be convertible to float32 without changing its value despite returning a type float64 and, at the time, I wanted to use float32 because float64 was overkill for a simple version value.

internal/controller/standalone_pgadmin/users.go
tjmoore4: Updated. The previous functions been removed and one test added to account for the scenario that wasn't being tested.

PR #4302, opened 2025-09-24
Image volume mounts

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

Additional volumes can only use persistent volume claims.

Additional volumes can now use an ImageVolumeSource.

**Other Information**:


pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
```diff
@@ -333,6 +337,13 @@ type AdditionalVolume struct {
 	// +optional
 	Containers []DNS1123Label `json:"containers"`
 
+	// Details for adding an image volume
+	// ---
+	// https://kubernetes.io/docs/concepts/storage/volumes/#image
```
cbandy: ⛏️ I like to shorten these a little.


```suggestion
	// https://docs.k8s.io/concepts/storage/volumes#image
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
```diff
@@ -311,14 +311,18 @@ func (meta *Metadata) GetAnnotationsOrNil() map[string]string {
 // Only one applier should be managing each volume definition.
 // https://docs.k8s.io/reference/using-api/server-side-apply#merge-strategy
 // +structType=atomic
+//
+// +kubebuilder:validation:XValidation:rule=`has(self.claimName) != has(self.image)`,message=`you must set only one of image or claimName`
+// +kubebuilder:validation:XValidation:rule=`!has(self.image) || !has(self.readOnly) || self.readOnly`,message=`readOnly cannot be set false when using an ImageVolumeSource`
+// +kubebuilder:validation:XValidation:rule=`!has(self.image) || (self.?image.reference.hasValue() && self.image.reference.size() > 0)`,message=`if using an ImageVolumeSource, you must set a reference`
```
cbandy, line 317: 😩 So the upstream type lacks `+required` and instead puts `Required:` in the description, eh?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
```diff
@@ -311,14 +311,18 @@ func (meta *Metadata) GetAnnotationsOrNil() map[string]string {
 // Only one applier should be managing each volume definition.
 // https://docs.k8s.io/reference/using-api/server-side-apply#merge-strategy
 // +structType=atomic
+//
+// +kubebuilder:validation:XValidation:rule=`has(self.claimName) != has(self.image)`,message=`you must set only one of image or claimName`
+// +kubebuilder:validation:XValidation:rule=`!has(self.image) || !has(self.readOnly) || self.readOnly`,message=`readOnly cannot be set false when using an ImageVolumeSource`
```
cbandy, line 316: ❓ Can this rule be on the Image field itself?

```
fieldPath=`.readOnly`,rule=`!has(self.readOnly) || self.readOnly`,message=`cannot be false`
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
```diff
@@ -366,6 +377,8 @@ func (in *AdditionalVolume) AsVolume(name string) corev1.Volume {
 			ClaimName: in.ClaimName,
 			ReadOnly:  in.ReadOnly,
 		}
+	case in.Image != nil:
+		out.Image = in.Image
```
cbandy: ❓ The description on this method says it returns a copy. Does `corev1.ImageVolumeSource` have a `DeepCopy` method?

```suggestion
		out.Image = in.Image.DeepCopy()
```

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
benjaminjb, line 317: 🤔 Is there a case where the image.reference would have a value but the size would be 0?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
dsessler7: Not sure I follow... `ReadOnly` is on `AdditionalVolume`, so the rule has to be on `AdditionalVolume` to access both `ReadOnly` and `Image`, right?

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
dsessler7: The original ImageVolumeSource does not have any length requirements on `Reference` so someone could technically put an empty string... I haven't tested it myself, but the comments in the struct say "behaves in the same way as pod.spec.containers[*].image" so I imagine it fails in some fashion if it cannot find an image.

benjaminjb: And the comment describes that it's not required because it might be overwritten (by the spec it's in?), which doesn't quite explain things to me. I mean, I guess it behaves somewhat like a RELATED_ image field in OLM.

benjaminjb: But would it be possible to pass an empty string through on `Reference` (which has `omitempty`)?

dsessler7: The rule checks that the length of the string is greater than 0...

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
cbandy: Whoops! 💯

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
cbandy, line 317: 📝 I'm not finding any practical limit on the size/length of an image reference.

cbandy (approved):

PR #4304, opened 2025-09-26
Look in more directories for upgrade binaries

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature

Images for major upgrades must have binaries at Red Hat paths.

Major upgrades look for binaries in typical paths for multiple distro flavors: Alpine, Debian, Red Hat.

**Other Information**:

Issue: PGO-864

### DIFF
--- a/internal/controller/pgupgrade/jobs.go
+++ b/internal/controller/pgupgrade/jobs.go
@@ -21,6 +21,7 @@ import (
 	"github.com/crunchydata/postgres-operator/internal/feature"
 	"github.com/crunchydata/postgres-operator/internal/initialize"
 	"github.com/crunchydata/postgres-operator/internal/naming"
+	"github.com/crunchydata/postgres-operator/internal/postgres"
 	"github.com/crunchydata/postgres-operator/internal/shell"
 	"github.com/crunchydata/postgres-operator/pkg/apis/postgres-operator.crunchydata.com/v1beta1"
 )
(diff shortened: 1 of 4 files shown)


internal/controller/pgupgrade/jobs.go
```diff
@@ -81,49 +81,52 @@ func upgradeCommand(spec *v1beta1.PGUpgradeSettings, fetchKeyCommand string) []s
...
-		`cp /pgdata/pg"${old_version}"/patroni.dynamic.json /pgdata/pg"${new_version}"`,
-
-		`echo -e "\npg_upgrade Job Complete!"`,
+		// pg_upgrade writes its files in "${new_data}/pg_upgrade_output.d" since PostgreSQL v15.
+		// Change to a writable working directory to be compatible with PostgreSQL v14 and earlier.
+		//
+		// https://www.postgresql.org/docs/release/15#id-1.11.6.20.5.11.3
+		`cd "${data_volume}" || exit`,
+
+		`section 'Initializing new data directory...'`,
```
benjaminjb: Can we keep the step numbers or maybe even better, say n/N steps to give a sense of where in the process we are.

internal/controller/pgupgrade/jobs.go
```diff
@@ -81,49 +81,52 @@ func upgradeCommand(spec *v1beta1.PGUpgradeSettings, fetchKeyCommand string) []s
...
-		// - https://patroni.readthedocs.io/en/latest/existing_data.html#major-upgrade-of-postgresql-version
-		`echo -e "\nStep 7: Copying patroni.dynamic.json...\n"`,
-		`cp /pgdata/pg"${old_version}"/patroni.dynamic.json /pgdata/pg"${new_version}"`,
-
-		`echo -e "\npg_upgrade Job Complete!"`,
+		// pg_upgrade writes its files in "${new_data}/pg_upgrade_output.d" since PostgreSQL v15.
+		// Change to a writable working directory to be compatible with PostgreSQL v14 and earlier.
+		//
+		// https://www.postgresql.org/docs/release/15#id-1.11.6.20.5.11.3
+		`cd "${data_volume}" || exit`,
```
benjaminjb: Do we need to do that if we pass `PGDATA="${data_volume}/pg${new_version}"` to initdb?

benjaminjb: Oh, I guess the `exit` might be important here.

cbandy: This whole script runs with `set -e`, so the `exit` isn't necessary here. I'll make this more explicit.

internal/controller/pgupgrade/jobs.go
```diff
@@ -81,49 +81,52 @@ func upgradeCommand(spec *v1beta1.PGUpgradeSettings, fetchKeyCommand string) []s
 		// - https://cwrap.org/nss_wrapper.html
 		`export LD_PRELOAD='libnss_wrapper.so' NSS_WRAPPER_GROUP NSS_WRAPPER_PASSWD`,
 
+		// Find directories that contain the desired Postgres executables.
```
sfc-gh-jmckulka: Maybe describe what is going on here a little more.

My understanding is that the `postgres.ShellPath` stuff sets the path variable and the `command -v postgres/initdb` returns the full path to the binary

sfc-gh-jmckulka:
With this change do we need the postgres bins on the `PATH` in the image level 🤔

internal/controller/pgupgrade/jobs.go
cbandy: 👍🏻 I've added some description.

That made me realize the PATH built into the image is also searched, so any ol' `postgres` might be found. I added an assertion\* on the `--version` returned by something in `old_bin` and `new_bin`, and learned and doc'd that `pg_upgrade` does a similar check only recently-ish.

\* same check as the postgres init container:
https://github.com/CrunchyData/postgres-operator/blob/e395e2ad815642bc13b16b7fd043ea49b726319f/internal/postgres/config.go#L489-L491

cbandy:
> With this change do we need the postgres bins on the `PATH` in the image level 🤔

`PGUpgrade` (upgrade image) should work without it after this, but `PostgresCluster` (postgres image) still relies on the image PATH.

benjaminjb (approved):

PR #4305, opened 2025-09-29
Github operator build update

benjaminjb:
Adjust GH builds

Issues: [PGO-2695]


hack/extract-licenses.go
```diff
@@ -0,0 +1,236 @@
...
+	//
+	//  dep	github.com/cespare/xxhash/v2	v2.3.0	h1:UL815xU9SqsFlibzuggzjXhog7bL6oX9BbNZnL2UFvs=
+	//  dep	github.com/crunchydata/postgres-operator	v0.0.0-00010101000000-000000000000
+	//  =>	./postgres-operator	(devel)
+	//
+	// [proxy]: https://go.dev/ref/mod#module-proxy
+	// [replace]: https://go.dev/ref/mod#go-mod-file-replace
+	modules = slices.DeleteFunc(modules, func(s string) bool {
+		return strings.HasPrefix(s, "git.crunchydata.com/") ||
+			strings.HasPrefix(s, "github.com/crunchydata/")
```
cbandy: 🔧 Just the latter, please.

cbandy (changes requested):

hack/extract-licenses.go
cbandy, line 1: 🔧 Needs a license header. 🤔 Why doesn't the linter complain?

https://github.com/CrunchyData/postgres-operator/blob/e395e2ad815642bc13b16b7fd043ea49b726319f/.golangci.yaml#L109-L113

hack/extract-licenses.go
```diff
@@ -0,0 +1,235 @@
...
+			if err == nil {
+				err = os.MkdirAll(filepath.Dir(destination), 0o755)
+			}
+			if err == nil {
+				data, err = os.ReadFile(license)
+			}
+			if err == nil {
+				//nolint:gosec // gosec warns on permissions more open than 600
+				// but we need these licenses to be readable by all
+				err = os.WriteFile(destination, data, 0o644)
```
cbandy: 🤔 The Dockerfile does a `COPY --chmod`, can we get away with `0o600` here? If so, a comment/reminder about these permissions not being final might be good.

hack/extract-licenses.go
```diff
@@ -0,0 +1,235 @@
...
+			}
+			if err == nil {
+				//nolint:forbidigo // This is an intentional print to console to inform the user
+				fmt.Println(license, "=>", destination)
```
cbandy: 🤔 The `Fprintln` just below this isn't flagged... Could we quiet this linter by printing to stdout?

```suggestion
				fmt.Fprintln(os.Stdout, license, "=>", destination)
```

hack/extract-licenses.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+}
+
+func downloadModules(ctx context.Context, modules ...string) map[string]string {
+	var stdout bytes.Buffer
+
+	// Download modules and read their details into a series of JSON objects.
+	// - https://go.dev/ref/mod#go-mod-download
+	//nolint:gosec // Suppressing unnecessary warning re: potentially tainted inputs (G204)
+	cmd := exec.CommandContext(ctx, os.Getenv("GO"), append([]string{"mod", "download", "-json"}, modules...)...)
```
cbandy: ⛏️ `gosec` has a [syntax](https://github.com/securego/gosec#annotating-code) for disabling specific rules inline that might work here. (untested)


```suggestion
	//gosec:disable G204 -- Use this environment variable to switch Go versions without touching PATH
	cmd := exec.CommandContext(ctx, os.Getenv("GO"), append([]string{"mod", "download", "-json"}, modules...)...)
```

cbandy (changes requested):
I took a closer look at the linting. At the very least, this needs a license header.

hack/extract-licenses.go
benjaminjb: Curious that other linters hit this file, but not header.

If I remove the go:build directive (and the header), it complains

cbandy: The main module is now 1.24, so we can drop the troublesome `go:build` constraint.

(I have no idea what in this requires 1.21)

cbandy (approved):

PR #4306, opened 2025-09-29
Proceed with major upgrade when data directory lacks checksums

cbandy:
**Checklist:**

 
Data directories created by PostgresCluster have checksums enabled, so existing tests exercise the logic that invokes `initdb` with the `--data-checksums` argument.

**Type of Changes:**

 
 - [x] New feature
 - [x] Bug fix

A data directory without checksums enabled fails during `pg_upgrade`.

PGUpgrade can upgrade a data directory that does not have checksums enabled.

**Other Information**:

This can happen when a data directory is imported from an external system.

Issue: PGO-619

### DIFF
--- a/internal/controller/pgupgrade/jobs.go
+++ b/internal/controller/pgupgrade/jobs.go
@@ -117,8 +117,28 @@ func upgradeCommand(spec *v1beta1.PGUpgradeSettings, fetchKeyCommand string) []s
 		// steps used and command flag specifics can be found in the documentation:
 		// - https://www.postgresql.org/docs/current/pgupgrade.html
 
+		// Examine the old data directory.
+		`control=$(LC_ALL=C PGDATA="${old_data}" "${old_bin}/pg_controldata")`,
+		`read -r checksums <<< "${control##*page checksum version:}"`,
+
+		// Data checksums on the old and new data directories must match.
+		// Configuring these checksums depends on the version of initdb:
+		//
+		// - PostgreSQL v17 and earlier: disabled by default, enable with "--data-checksums"
+		// - PostgreSQL v18: enabled by default, enable with "--data-checksums", disable with "--no-data-checksums"
+		//
+		// https://www.postgresql.org/docs/release/18#RELEASE-18-MIGRATION
+		//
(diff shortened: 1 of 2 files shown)


internal/controller/pgupgrade/jobs.go
```diff
@@ -117,8 +117,18 @@ func upgradeCommand(spec *v1beta1.PGUpgradeSettings, fetchKeyCommand string) []s
...
+		// Data page checksum version zero means checksums are disabled; track that as blank.
+		//
+		// https://git.postgresql.org/gitweb/?p=postgresql.git;hb=refs/tags/REL_11_0;f=src/bin/pg_verify_checksums/pg_verify_checksums.c#l303
+		// https://git.postgresql.org/gitweb/?p=postgresql.git;hb=refs/tags/REL_12_0;f=src/bin/pg_checksums/pg_checksums.c#l523
+		// https://git.postgresql.org/gitweb/?p=postgresql.git;hb=refs/tags/REL_18_0;f=src/bin/pg_checksums/pg_checksums.c#l571
+		`read -r checksums <<< "${control##*page checksum version:}" && checksums="${checksums#0}"`,
+
 		`section 'Step 3 of 7: Initializing new data directory...'`,
-		`PGDATA="${new_data}" "${new_bin}/initdb" --allow-group-access --data-checksums` + argEncryptionKeyCommand,
+		`PGDATA="${new_data}" "${new_bin}/initdb" --allow-group-access ${checksums:+--data-checksums}` + argEncryptionKeyCommand,
```
benjaminjb: Ah, ok, if checksum == 0, we drop the `--data-checksums` flag.

Looks like this flag is enabled by default in PG 18: https://www.postgresql.org/docs/18/app-initdb.html#APP-INITDB-DATA-CHECKSUMS (but not in PG 17 or earlier -- according to the docs)

cbandy: 🔧 Good find. I'll update this for PG 18.

cbandy: This grew into an `if` statement, but it also seems more legible than string manipulation and parameter expansion.

PR #4307, opened 2025-09-30
Look in more directories for Postgres binaries

cbandy:
This is similar to #4304, but for PostgresCluster and Postgres images.

**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [x] Other

**Other Information**:

The last piece missing from the upgrade image is the `pgbackrest` command.

Issue: PGO-864

### DIFF
--- a/internal/patroni/reconcile.go
+++ b/internal/patroni/reconcile.go
@@ -105,7 +105,14 @@ func InstancePod(ctx context.Context,
 		}
 	}
 
-	container.Command = []string{"patroni", configDirectory}
+	// Start Patroni in an environment that prioritizes executables of the specified Postgres version.
+	// NOTE: Patroni ignores PATH when "postgresql.bin_dir" is set.
+	//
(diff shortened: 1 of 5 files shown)


cbandy (commented):
🔧 Restore jobs also call Postgres tools now.

internal/patroni/reconcile.go
```diff
@@ -105,7 +105,14 @@ func InstancePod(ctx context.Context,
 		}
 	}
 
-	container.Command = []string{"patroni", configDirectory}
+	// Start Patroni in an environment that prioritizes executables of the specified Postgres version.
+	// NOTE: Patroni ignores PATH when "postgresql.bin_dir" is set.
```
jmckulk, line 109: Could users set this and break something?

cbandy: No. Patroni reads this setting from YAML files or environment variables, neither of which is configurable.

jmckulk (approved):

PR #4308, opened 2025-09-30
Use OpenAPI to validation ImageVolumeSource.Reference

cbandy:
**Checklist:**

 
**Type of Changes:**

 
 - [x] Other

We cannot add validation markers inside upstream types, so we have to use CEL validation from outside the type.

We can use the `+kubebuilder:title` marker to add identifiers to CRD schemas, and manipulate those after `controller-gen` runs.


go.mod
```diff
@@ -133,9 +133,11 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	k8s.io/apiextensions-apiserver v0.33.0 // indirect
 	k8s.io/apiserver v0.33.0 // indirect
+	k8s.io/code-generator v0.33.0 // indirect
+	k8s.io/gengo/v2 v2.0.0-20250207200755-1244d31929d7 // indirect
 	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738 // indirect
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.31.2 // indirect
-	sigs.k8s.io/controller-tools v0.17.3 // indirect
+	sigs.k8s.io/controller-tools v0.18.0 // indirect
```
cbandy, line 140: 🤔 v0.18 requires k8s.io/.. v0.33, but it wasn't bundled in the k8s.io updates last month: https://github.com/CrunchyData/postgres-operator/pull/4248

I don't know why!

internal/crd/post-process.jq
```diff
@@ -3,13 +3,26 @@
...
+# Use the controller-gen "+kubebuilder:title" marker to identify schemas that need special manipulation.
 #
 # [jq]: https://jqlang.org
 
 # merge recursively combines a stream of objects.
 # https://jqlang.org/manual#multiplication-division-modulo
 def merge(stream): reduce stream as $i ({}; . * $i);
 
+# https://pkg.go.dev/k8s.io/api/core/v1#ImageVolumeSource
+reduce paths(try .title == "$corev1.ImageVolumeSource") as $path (.;
```
cbandy, line 15: 📝 This may very well be "too clever." I'm curious what others think.

benjaminjb: Setting a `title` field on our CRD to add some validation rules and then removing the `title` field -- is that the part you're thinking might be too clever?

Seems reasonable to me, with the caveat that having the CRD be assembled from kubebuilder annotations and then post-process functions might make it a little harder to know easily where some element is coming from. Maybe that's a documentation issue, e.g., when we add the title in the struct, we could link to or reference this jq script at this point?

config/crd/bases/postgres-operator.crunchydata.com_crunchybridgeclusters.yaml
```diff
@@ -1,9 +1,8 @@
 ---
+# controller-gen.kubebuilder.io/version: v0.18.0
```
benjaminjb, line 2: I'm curious to hear your thinking behind this change

internal/crd/post-process.go
```diff
@@ -44,8 +45,12 @@ func main() {
 				panic(err)
 			}
 
+			// Turn top-level strings that start with octothorpe U+0023 into YAML comments by removing their quotes.
+			yamlData := need(yaml.Marshal(v))
+			yamlData = regexp.MustCompile(`(?m)^'(#[^']*)'(.*)$`).ReplaceAll(yamlData, []byte("$1$2"))
```
benjaminjb, line 50: And we can't just capture the whole line as a single capture group because you want to exclude the quotes, yeah?

internal/crd/post-process.jq
cbandy: > Setting a `title` field on our CRD to add some validation rules and then removing the `title` field -- is that the part you're thinking might be too clever?

Yeah. Smuggling information through a marker that does something else.

> Seems reasonable to me, with the caveat that having the CRD be assembled from kubebuilder annotations and then post-process functions might make it a little harder to know easily where some element is coming from. Maybe that's a documentation issue, e.g., when we add the title in the struct, we could link to or reference this jq script at this point?

A comment on the marker is better than none. I'll add one.

config/crd/bases/postgres-operator.crunchydata.com_crunchybridgeclusters.yaml
cbandy: This information is minimally interesting during development, but AFAIK has no use after that. As an annotation, it is deployed with the CRD but serves no purpose.

My impression has been that the generator is using `metadata.annotations` to push "stuff" through code that can only handle K8s objects.

internal/crd/post-process.go
cbandy: 🤔 It doesn't seem like I need the second capture... The line should end just after the quoted string, right? 🦆

I'll check this tomorrow.

cbandy: > And we can't just capture the whole line as a single capture group because you want to exclude the quotes, yeah?

Right. This finds lines starting with `'#` _and_ removes a pair of `'`. This is the transformation:

```diff
-'# controller-gen.kubebuilder.io/version': v0.18.0
+# controller-gen.kubebuilder.io/version: v0.18.0
```

> 🤔 It doesn't seem like I need the second capture...

I forgot that the quotes are around only the key of the flow mapping. The second capture is for the remainder of the line, if any.

internal/crd/post-process.jq
cbandy: > A comment on the marker is better than none. I'll add one.

Done.

pkg/apis/postgres-operator.crunchydata.com/v1beta1/shared_types.go
```diff
@@ -337,9 +336,11 @@ type AdditionalVolume struct {
 	// +optional
 	Containers []DNS1123Label `json:"containers"`
 
-	// Details for adding an image volume
+	// Reference to an image or OCI artifact.
+	// More info: https://kubernetes.io/docs/concepts/storage/volumes#image
 	// ---
-	// https://docs.k8s.io/concepts/storage/volumes#image
+	// Use "title" to add more validation in [internal/crd/post-process.jq].
```
benjaminjb, line 342: 👍

benjaminjb (approved):

PR #4310, opened 2025-10-02
Look in more directories for restore binaries

cbandy:
This is similar to https://github.com/CrunchyData/postgres-operator/pull/4304, but for restore jobs in PostgresCluster and Postgres images.

**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature

**Other Information**:

Issue: PGO-864


internal/pgbackrest/config.go
```diff
@@ -212,10 +212,9 @@ func MakePGBackrestLogDir(template *corev1.PodTemplateSpec,
 //   - Renames the data directory as needed to bootstrap the cluster using the restored database.
 //     This ensures compatibility with the "existing" bootstrap method that is included in the
 //     Patroni config when bootstrapping a cluster using an existing data directory.
-func RestoreCommand(pgdata, hugePagesSetting, fetchKeyCommand string, _ []*corev1.PersistentVolumeClaim, args ...string) []string {
-	ps := postgres.NewParameterSet()
+func RestoreCommand(postgresVersion int32, pgdata string, params *postgres.ParameterSet, args ...string) []string {
```
benjaminjb, line 215: 👍

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -1190,63 +1191,51 @@ func (r *Reconciler) reconcileRestoreJob(ctx context.Context,
 	pgdataVolume, pgwalVolume *corev1.PersistentVolumeClaim,
 	pgtablespaceVolumes []*corev1.PersistentVolumeClaim,
 	dataSource *v1beta1.PostgresClusterDataSource,
-	instanceName, instanceSetName, configHash, stanzaName string) error {
-
+	instanceName, instanceSetName, configHash, stanzaName string,
+) error {
+	flagName := regexp.MustCompile(`--([^ =]+)`)
```
benjaminjb: any flag of the form `--something`, breaking on spaces and equal signs? Because we don't care about value (if there is one), just the option flag they're trying to set, ok, got it.

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -1190,63 +1191,51 @@ func (r *Reconciler) reconcileRestoreJob(ctx context.Context,
...
+		}
+	}
 
 	// ensure options are properly set
 	// TODO (andrewlecuyer): move validation logic to a webhook
-	for _, opt := range options {
+	{
 		var msg string
 		switch {
 		// Since '--repo' can be set with or without an equals ('=') sign, we check for both
```
benjaminjb: comment unnecessary/no longer applicable?

internal/controller/postgrescluster/pgbackrest.go
```diff
@@ -1190,63 +1191,51 @@ func (r *Reconciler) reconcileRestoreJob(ctx context.Context,
...
+	for _, input := range dataSource.Options {
+		for _, matches := range flagName.FindAllStringSubmatch(input, -1) {
+			hasOption[matches[1]] = true
+		}
+	}
 
 	// ensure options are properly set
 	// TODO (andrewlecuyer): move validation logic to a webhook
-	for _, opt := range options {
+	{
```
benjaminjb, line 1208: Do we want to check all the flags and emit all the errors at once?

cbandy: I don't want to change any more of this behavior right now. 🌱 The comment above suggests CRD validation.

benjaminjb: Should we be ranging over hasOptions?

internal/controller/postgrescluster/pgbackrest.go
cbandy: I dropped the submatch to simplify the regular expression and shrink the diff.

<details><summary>before / after</summary>

```diff
-	flagName := regexp.MustCompile(`--([^ =]+)`)
+	matchFlag := regexp.MustCompile(`--[^ =]+`)
…
	for … {
		for … {
-			hasOption[matches[1]] = true
+			hasFlag[match] = true
		}
	}
…
		switch {
-		case hasOption["repo"]:
+		case hasFlag["--repo"]:
…
```

</details>

internal/controller/postgrescluster/pgbackrest.go
cbandy: Not that I know of. What would a range do here?

benjaminjb: Ah, I see where I was going wrong: in the original code, we loop over the options, and if one of them is wrong for some reason, we emit an event and return.

(As opposed to, say, telling the user every things that's wrong with their input in one pass.)

benjaminjb (approved):

PR #4329, opened 2025-11-05
Allow OTel metrics to work with postgres versions greater than 17. If…

dsessler7:
… postgres_exporter is used with postgres versions greater than 17, create a warning event and set the setup.sql blank.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

Currently, if you are using either metrics exporter (OTel or `postgres_exporter`) with a postgres version greater than 17, you will get a reconciler error and metrics will not work.

You can now get metrics with pg > 17 if you use OTel as your exporter. We will not support `postgres_exporter` with pg versions greater than 17. If you attempt to use `postgres_exporter` with pg > 17, you will see a warning event and the metrics will essentially be turned off. 

**Other Information**:

### DIFF
--- a/internal/controller/postgrescluster/pgmonitor.go
+++ b/internal/controller/postgrescluster/pgmonitor.go
@@ -43,6 +43,7 @@ func (r *Reconciler) reconcilePGMonitorExporter(ctx context.Context,
 	monitoringSecret *corev1.Secret) error {
 
 	var (
+		err              error
 		writableInstance *Instance
 		writablePod      *corev1.Pod
 		setup            string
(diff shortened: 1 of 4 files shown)


internal/controller/postgrescluster/pgmonitor.go
```diff
@@ -145,6 +134,40 @@ func (r *Reconciler) reconcilePGMonitorExporter(ctx context.Context,
 	return err
 }
 
+// reconcileExporterSqlSetup generates the setup.sql string based on
+// whether the OTel metrics feature is enabled or not and the postgres
+// version being used. This function assumes that at least one of
+// OTel metrics or postgres_exporter are enabled.
+func (r *Reconciler) reconcileExporterSqlSetup(ctx context.Context,
+	cluster *v1beta1.PostgresCluster) (string, error) {
+	if !collector.OpenTelemetryMetricsEnabled(ctx, cluster) {
```
jmckulk: I think `if exporter` would make more sense here since we are only talking about exporter specific things (comments, warnings, setup) ⛏️

jmckulk (approved):
One naming ⛏️but functionally looks good

internal/controller/postgrescluster/pgmonitor.go
dsessler7: Not sure I follow... In this function the "Exporter" in `reconcileExporterSqlSetup` refers to both/either OTel and/or `postgres_exporter`. If the OTel metrics feature is enabled, it takes precedence (even if `postgres_exporter` is enabled). If OTel metrics isn't enabled, we get the setup for `postgres_exporter`.

jmckulk: That makes sense. I should have said `if postgres_exporter`... something like this:

```
if pgmonitor.ExporterEnabled(ctx, cluster) {
  # do some postgres_exporter specific things
  return metrics for postgres_exporter
}

# default to otel
return metrics for otel
```

It feels weird to me to have 

```
if !otel {
  # do postgres_exporter things
}
```

dsessler7: Well, that's just the thing, if OTel metrics is enabled it will always be used, even if `postgres_exporter` is also enabled. Also, `pgmonitor.ExporterEnabled` checks that at least one of OTel metrics or `postgres_exporter` are enabled (not only `postgres_exporter`).

I guess I could change it to more of a guard clause for OTel and make `postgres_exporter` the "default":
```
if otel {
  return metricsSetupForOTelCollector, nil
}

# do postgres_exporter stuff
```
But then we're kinda splitting hairs here

sfc-gh-jmckulka: Definitely splitting hairs 👍

PR #4338, opened 2025-11-19
Github Action Debugging

jmckulk:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

**Other Information**:

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -100,7 +100,6 @@ jobs:
           prefetch-images: |
             registry.developers.crunchydata.com/crunchydata/crunchy-pgbackrest:ubi9-2.56.0-2542
             registry.developers.crunchydata.com/crunchydata/crunchy-postgres:ubi9-17.6-2542
-            registry.developers.crunchydata.com/crunchydata/crunchy-postgres:ubi9-16.10-2542
 
       - name: Get pgMonitor files.
         run: make get-pgmonitor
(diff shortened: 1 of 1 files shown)


jmckulk:
The two biggest improvements here were reducing the number of prefetch images and splitting chainsaw and kuttl tests into separate jobs. I think we were hitting issues with disk pressure and these two changes help with that. 

For the prefetch images change, we aren't testing GIS here so there is no point in fetching them. In addition to removing these extra images, this change tries to ensure we are using the images that we fetch. This does mean we might need to add images, specifically to chainsaw, as we test more functionality.

For the split change, this feels like a stopgap fix. We are really just giving the tests more space to run by splitting them up. We should continue to consider ways reduce disk space usage. One option may be to refactor the chainsaw pgbackrest-restore test to only use one cluster.

Other changes in this pr are mostly either timeout increases or adding extra logging or checks. The change to reduce volume request size likely doesn't have much impact on actual disk usage but feels like a safe change.

ValClarkson (approved):

.github/workflows/test.yaml
```diff
@@ -134,14 +124,7 @@ jobs:
             --env 'RELATED_IMAGE_PGEXPORTER=registry.developers.crunchydata.com/crunchydata/crunchy-postgres-exporter:ubi9-0.17.1-2542' \
             --env 'RELATED_IMAGE_PGUPGRADE=registry.developers.crunchydata.com/crunchydata/crunchy-upgrade:ubi9-18.0-2542' \
```
dsessler7: I guess these (and PGBOUNCER) could be removed as well

.github/workflows/test.yaml
```diff
@@ -134,14 +124,7 @@ jobs:
...
             --env 'RELATED_IMAGE_POSTGRES_17=registry.developers.crunchydata.com/crunchydata/crunchy-postgres:ubi9-17.6-2542' \
-            --env 'RELATED_IMAGE_POSTGRES_17_GIS_3.4=registry.developers.crunchydata.com/crunchydata/crunchy-postgres-gis:ubi9-17.6-3.4-2542' \
-            --env 'RELATED_IMAGE_POSTGRES_17_GIS_3.5=registry.developers.crunchydata.com/crunchydata/crunchy-postgres-gis:ubi9-17.6-3.5-2542' \
-            --env 'RELATED_IMAGE_POSTGRES_17_GIS_3.6=registry.developers.crunchydata.com/crunchydata/crunchy-postgres-gis:ubi9-17.6-3.6-2542' \
-            --env 'RELATED_IMAGE_POSTGRES_18=registry.developers.crunchydata.com/crunchydata/crunchy-postgres:ubi9-18.0-2542' \
-            --env 'RELATED_IMAGE_POSTGRES_18_GIS_3.6=registry.developers.crunchydata.com/crunchydata/crunchy-postgres-gis:ubi9-18.0-3.6-2542' \
             --env 'RELATED_IMAGE_STANDALONE_PGADMIN=registry.developers.crunchydata.com/crunchydata/crunchy-pgadmin4:ubi9-9.8-2542' \
             --env 'RELATED_IMAGE_COLLECTOR=registry.developers.crunchydata.com/crunchydata/postgres-operator:ubi9-5.8.4-0' \
```
dsessler7: These can be removed too

.github/workflows/test.yaml
jmckulk: I was thinking leave related images. That way if an image isn't pre-fetched it could be pulled.

Or we could fail when we don't have all the images we need?

PR #4364, opened 2025-12-11
Remove optional types from ssl_groups validation rule since they are …

dsessler7:
…not supported in k8s versions earlier than 1.29

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

There is a validation rule that uses CEL optional types which is only supported in k8s 1.29 and later. We support earlier k8s versions (via OpenShift support) for 5.8, so users running these earlier versions are unable to upgrade to 5.8.5.

The rule has been changed to not use optional types.

**Other Information**:


tjmoore4 (approved):

Gnarfoz:
Excellent, thanks for the reply! 👍🏻

PR #4359, opened 2025-12-05
Copy Installers from postgres-operator-examples

jmckulk:
We are moving our developer installers from the
postgres-operator-examples repo to the postgres-operator.

Note: These images might not exist, check the Releases page for current images.

https://github.com/CrunchyData/postgres-operator/releases

**Checklist:**

 
 - [ ] Have you added an explanation of what your changes do and why you'd like them to be included?
 - [ ] Have you updated or added documentation for the change, as applicable?
 - [ ] Have you tested your changes on all related environments with successful results, as applicable?
   - [ ] Have you added automated tests?

**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

**What is the current behavior (link to any open issues here)?**

**What is the new behavior (if this is a feature change)?**
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

**Other Information**:

PR #4373, opened 2025-12-21
Update PGO installer to 5.8.6

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

Older installer before we moved the official installer to the postgres-operator repo

Latest installer with newer images and general installer structure.

**Other Information**:


config/LICENSE.txt
tjmoore4, line 1: ❓ I'm a bit unclear as to what's happening with this license versus what we're doing with [this one](https://github.com/CrunchyData/postgres-operator/blob/REL_5_8/LICENSE.md). Do we need both?

config/manager/manager.yaml
```diff
@@ -8,57 +8,74 @@ spec:
...
+      serviceAccountName: pgo
       containers:
       - name: operator
         image: postgres-operator
         env:
         - name: PGO_INSTALLER
           value: kustomize
         - name: PGO_INSTALLER_ORIGIN
-          value: postgres-operator-repo
+          value: examples-repo
```
tjmoore4: 🔧 This value should probably be reverted.

config/LICENSE.txt
jmckulk: The installer on main doesn't have this file but I do see it as a removed file from the examples repo. Maybe it should be added to `main` as well?

tjmoore4: 🤔 As mentioned on the [PR](https://github.com/CrunchyData/postgres-operator/pull/4359) for `main`, assuming this license should be here, I think this update is good so I will approve. Might still be worth verifying though.

config/dev/rbac/cluster/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
+- role_binding.yaml
+- role.yaml
+- service_account.yaml
```
tjmoore4: ⛏️ Missing newline here

config/dev/rbac/namespace/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
+- ../cluster
+components:
+- ../../../components/single-namespace
```
tjmoore4: ⛏️ Missing newline here

tjmoore4 (approved):

dsessler7:
We only release one community installer for the latest version per release, and that is now 6.0, so we can close this.

PR #4377, opened 2026-01-07
Update 6.0 installer CRDs, rbac, and pgbouncer image

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [x] Other

The current installer isn't using our "official" CRDs. It still has crunchybridgecluster rbac. It has the wrong pgbouncer image.

Updated CRDs, removes unneeded rbac, fixes pgbouncer version.

**Other Information**:


config/rbac/kustomization.yaml
tjmoore4, line 1: 📝 Just to note, these files were used prior to the [recent refactor](https://github.com/CrunchyData/postgres-operator/pull/4359) that now points to `../rbac/cluster` instead of `../rbac` for RBAC resources.

config/LICENSE.txt
tjmoore4, line 1: 🤔 I'm still not sure about this file, but if it's needed to keep installers consistent with our existing pattern, this isn't a blocker.

dsessler7: I'm not sure either, but the LICENSE.txt was included with the installer when in postgres-operator-examples and is different than the LICENSE.md in the root directory. Seems like the former is specific to using our released installer whereas the latter is for using the code in the repo.

dsessler7: Andrew indicated this can safely be removed.

PR #4385, opened 2026-01-21
Adjust pgAdmin version comparision logic

tjmoore4:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

Now that the current version of pgAdmin is greater than 9.9, the version comparison logic has to be updated to properly compare the values (i.e. we can no longer use a simple x < y to compare floats).

This update makes the necessary adjustments so our code works as expected for newer versions of pgAdmin.

**Other Information**:
Issue: PGO-2838

### DIFF
--- a/internal/controller/standalone_pgadmin/users.go
+++ b/internal/controller/standalone_pgadmin/users.go
@@ -179,11 +179,19 @@ cd $PGADMIN_DIR
 	}
 
 	var olderThan9_3 bool
-	versionFloat, err := strconv.ParseFloat(pgadmin.Status.MinorVersion, 64)
-	if err != nil {
-		return err
+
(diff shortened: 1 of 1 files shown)


dsessler7:
> > I imagine that the idiomatic way to do this in `go` would be to use the following package: https://pkg.go.dev/golang.org/x/mod/semver
> > But I'm fine with what you've done here. LGTM.
> 
> 🤔 If I understand correctly, I think that package requires the version string to begin with a `v`? The current string is stored in the status without one, for example "9.11". In any case, hopefully this will be robust enough without requiring a new import.

Yeah, I'd just slap a "v" on the front... But again, I think what you've got here is fine.

PR #4392, opened 2026-01-26
Fix issue with resourcedetection processor failing with no detectors …

dsessler7:
…set.

Minor refactoring of some collector code.

**Checklist:**

 
**Type of Changes:**

 
 - [x] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

We want to move to the newest version of the [opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main), however, there was a change with the `resourcedetection` processor where having zero detectors set results in zero "successes" which results in a failure, killing the collector:
https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/50cd2e8433cee1e292e7b7afac9758365f3a1298

We therefore need to change our code so that we only add the `resourcedetection` processor if we have detectors set.

We only add the `resourcedetection` processor to collector pipelines if there are detectors set in the `postgrescluster`/`pgadmin` manifests. Allows us to move to otel v0.144.0.

**Other Information**:

### DIFF
--- a/components/image-collector/Dockerfile
+++ b/components/image-collector/Dockerfile
@@ -4,7 +4,7 @@
 
 # Rather than build the binary, retrieve the already-built binary from
 # the OpenTelemetry image
-FROM otel/opentelemetry-collector-contrib:0.139.0 AS collector
+FROM otel/opentelemetry-collector-contrib:0.144.0 AS collector
 
 # Aggregate the collector licenses from binary
 # and from root of the PGO repo


internal/collector/patroni.go
```diff
@@ -17,12 +17,8 @@ func EnablePatroniLogging(ctx context.Context,
 	inCluster *v1beta1.PostgresCluster,
 	outConfig *Config,
 ) {
-	var spec *v1beta1.InstrumentationLogsSpec
-	if inCluster != nil && inCluster.Spec.Instrumentation != nil {
```
tjmoore4, line 21: ❓ Did something change so that these checks are no longer necessary?

internal/collector/pgbouncer.go
```diff
@@ -52,12 +52,8 @@ func EnablePgBouncerLogging(ctx context.Context,
 	outConfig *Config,
 	logfile string,
 ) {
-	var spec *v1beta1.InstrumentationLogsSpec
-	if inCluster != nil && inCluster.Spec.Instrumentation != nil {
```
tjmoore4, line 56: ❓ Same question as above.

internal/collector/postgres.go
```diff
@@ -134,12 +134,8 @@ func EnablePostgresLogging(
 	inParameters *postgres.ParameterSet,
 	outConfig *Config,
 ) {
-	var spec *v1beta1.InstrumentationLogsSpec
-	if inCluster != nil && inCluster.Spec.Instrumentation != nil {
```
tjmoore4, line 138: ❓ Same here.

internal/collector/patroni.go
dsessler7: So, we checked them to ensure that accessing `inCluster.Spec.Instrumentation.Logs` wouldn't result in a nil pointer error, but the `inCluster.Spec.Instrumentation != nil` check happens in the `OpenTelemetryLogsEnabled` function call below anyways, so we can get rid of them and move the `spec` assignment inside the `if` block... That being said, `OpenTelemetryLogsEnabled` doesn't check that `inCluster != nil`, so we actually theoretically could have had nil pointer exceptions even before my change, although I don't really know if this is possible since I can't think of a reason why we'd be passing around a nil pointer to a `v1beta1.PostgresCluster`... Either way, probably a good idea to stick a check in there.

PR #4401, opened 2026-02-02
Dependabot/go modules/kubernetes 1b4ea1c485 with updates

tjmoore4:
Adds CRD and lint fix updates to the dependabot PR (#4381)

Issue: PGO-2848

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -184,6 +184,11 @@ linters:
         text: >-
           client.Apply is deprecated
 
+      - linters: [staticcheck]
+        path: internal/controller/(postgrescluster|pgupgrade|standalone_pgadmin)
+        text: >-
+           GetEventRecorderFor is deprecated
+
 # https://golangci-lint.run/usage/formatters
 formatters:
   enable:


go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/crunchydata/postgres-operator
 
 // If this is changing when you don't want it to, see hack/go-get.sh
-go 1.24.0
+go 1.25.0
```
tjmoore4, line 4: 📝 Note that these dependencies required Go 1.25

internal/controller/runtime/apply_test.go
```diff
@@ -54,15 +54,15 @@ func TestServerSideApply(t *testing.T) {
 
 		// Create the object.
 		before := constructor()
-		assert.NilError(t, cc.Patch(ctx, before, client.Apply))
+		assert.NilError(t, cc.Patch(ctx, before, client.Apply)) //nolint:staticcheck // Testing deprecated API behavior
```
dsessler7: ❓ Can you please explain what prompted this comment?

tjmoore4: Sure. The initial lint issue was
```
SA1019: client.Apply is deprecated: Use client.Client.Apply() and client.Client.SubResource("subrsource").Apply() instead.
```
so the most basic adjustment is to no long use `client.Apply`. That is why in `internal/controller/runtime/apply.go`, the update is to just get `ApplyPatchType` instead of invoking the `client.Apply` constant to get to the type we actually need. However, the purpose of these tests is to verify the apply behavior itself, so we need to use the deprecated behavior here for the test to check the patch itself.

Based on the versions [we seem to be concerned with](https://github.com/CrunchyData/postgres-operator/blob/main/internal/controller/runtime/apply_test.go#L68-L78), maybe a lot of this is no longer necessary, but I'm leaning on the side of not impacting any existing functionality.

dsessler7: Yeah, my interpretation of the deprecation warning is "if you are doing a patch with a type of `Apply`, you should just be doing an apply with `client.Client.Apply()`," however, changing the apply stuff in our `runtime` package would likely be a significant refactor for us...

dsessler7: The [comment explaining this is actually more detailed in v0.22.4](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.22.4/pkg/client#Patch) before `client.Apply` was technically deprecated than it is in [the deprecated comment in v0.23.1](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Patch) (latest).

tjmoore4: 👍 Agreed a major refactor for this doesn't seem like the right approach. The positive is that the only place we're leaving the deprecated implementation is in the tests, so that feels more reasonable.

dsessler7: Well, no, we're actually still using it everywhere because it's baked into our `runtime.Apply` function. By replacing `client.Apply` with `types.ApplyPatchType` you're just swapping out one constant for another; they both result in the same function call. It does make me wonder why they put a deprecation notice in the `client/patch.go` file and not in the `types/patch.go` file though... 🤔

tjmoore4: I don't believe that's quite right. As I understand it, since [SA1019](https://staticcheck.dev/docs/checks#SA1019) is simply indicating the code is using a deprecated variable, we simply want to avoid using the deprecated `Apply` variable since it will presumably be removed in a later version. On the other hand, there isn't any indication that the `ApplyPatchType` type is planned for removal, so its continued use should be fine.

Both before and after the change, I agree that the relevant function calls don't change. We're still using `client.RawPatch` when making the patch and `cc.Patch` when sending it. In this case though, those functions should be fine to leave in place.

Hopefully that makes sense. Happy to discuss further to make sure I'm not missing anything.

dsessler7: Well, first off, the linter is saying SA1019 because that's its interpretation of the deprecation comment since it is above the package variable. The linter doesn't attempt to understand the guidance in the comment, look at comments in different versions of the package, look at related github issues, etc. Besides, what would be the point in deprecating the variable if not for deprecating the feature that it is used for?

The comment says: "Deprecated: Use client.Client.Apply() and client.Client.SubResource("subrsource").Apply() instead." To me, this is pretty clearly saying, "if you want to do server side apply, you should use the client.Client.Apply() function," not "just go use a different constant to tell the client.Client.Patch() function to do SSA."

The comment in 0.22.4 is more detailed: "This should now only be used to patch sub resources, e.g. with client.Client.Status().Patch(). Use client.Client.Apply() instead of client.Client.Patch(..., client.Apply, ...). This will be deprecated once the Apply method has been added for sub resources. See the following issue for more details: https://github.com/kubernetes-sigs/controller-runtime/issues/3183". If you look at the github issue, it's clearly about improving the `client.Client.Apply()` function so that it can handle server-side apply stuff that people historically used `client.Client.Patch()` for.

Lastly, I looked at the `types` constants and those are used all over the place by various HTTP calls. Basically, they don't want to deprecate all instances of SSA patches, they're just deprecating it for the Client, hence why the `client.Apply` variable is deprecated and the `ApplyPatchType` isn't.

This all being said, it's deprecated, not removed, so whether it's worth the refactor on our end or not is certainly worth discussing.

tjmoore4: Just for traceability, after discussion with @dsessler7 I've decided to look into the level of effort involved in updating the SSA behavior instead of the current approach given the potential risks involved with a "hidden" deprecation. Ideally those should be in place before moving to the new package versions.

tjmoore4: Story is created in the backlog to address this issue more comprehensively, for now I've added a linter configuration update.

PR #4430, opened 2026-02-17
Add test for collector package AddToPod function

dsessler7:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [ ] Bug fix
 - [ ] Documentation
 - [x] Testing enhancement
 - [ ] Other

No testing of collector.AddToPod function.

Go test for collector.AddToPod function which gives us more collector package test coverage and ensures the Instrumentation configuration fields environmentVariables and files get tested.

**Other Information**:

### DIFF
--- a/internal/collector/instance_test.go
+++ b/internal/collector/instance_test.go
@@ -1,0 +1,524 @@
+// Copyright 2024 - 2026 Crunchy Data Solutions, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package collector
+
+import (
+	"context"
+	"testing"
+
+	"gotest.tools/v3/assert"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/crunchydata/postgres-operator/internal/feature"
+	"github.com/crunchydata/postgres-operator/internal/naming"
+	"github.com/crunchydata/postgres-operator/internal/testing/cmp"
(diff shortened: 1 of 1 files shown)


internal/collector/instance_test.go
```diff
@@ -0,0 +1,524 @@
...
+
+		AddToPod(ctx, spec, corev1.PullIfNotPresent, newConfigMap(),
+			template, nil, "my-password", nil, false, false)
+
+		container := findContainer(template)
+		assert.Assert(t, container != nil, "expected collector container")
+
+		// Verify the three default env vars are present
+		assert.Assert(t, cmp.Len(container.Env, 3), "expected 3 environment variables")
+		assert.Equal(t, container.Env[0].Name, "K8S_POD_NAMESPACE")
```
tjmoore4, line 118: ❓ Is order guaranteed for the environment variables here and below?
I think the answer is probably yes since you can have [dependent env vars](https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/) but I wanted to see if this had come up in testing at all.

dsessler7: Yeah, the default env vars are set [here](https://github.com/CrunchyData/postgres-operator/blob/main/internal/collector/instance.go#L95-L112) where the order is hardcoded, and then any user provided env vars are appended [here](https://github.com/CrunchyData/postgres-operator/blob/main/internal/collector/instance.go#L118-L121).

Issue #3293, opened 2022-07-07
pgbackrest_info OOM + 100% CPU


lingwooc:
## Overview

There is an issue around the script /opt/crunchy/bin/postgres/pgbackrest_info.sh. I think this is related to monitoring. It is using over 3gb of ram and getting oomkilled. Suprisingly it is bash that is killed not pgbackrest, and pgbackrest does not seem to be started (although I may be mistaken).

The issue seems to be that:

```echo $(echo -n "$conf|" | tr '/' '_'; pgbackrest --output=json ${cmd_args[*]} info | tr -d '\n')```

Causes bash to use a biblical amount of ram and CPU while not really doing anything.

Running the commands manually in the containers bash console gives some insight
These commands work:
```pgbackrest --output=json ${cmd_args[*]} info```
```pgbackrest --output=json ${cmd_args[*]} info | tr -d '\n'```
```echo -n "$conf|" | tr '/' '_'; pgbackrest --output=json ${cmd_args[*]} info | tr -d '\n'```

This command is oomkilled (and uses 100% cpu).
```echo $(echo -n "$conf|" | tr '/' '_'; pgbackrest --output=json ${cmd_args[*]} info | tr -d '\n')```

Maybe i'm missing something but I don't see the value in the wrapping echo. I also can't see why this would be such an impactful thing to do.

The troublesome environments do have quite a few incremental and and differential backups (up to 690 diffs with 3 times as many incremental) but its not pgbackrest that is the problem.

I should note I'm not seeing this issue on _all_ our environments, but that could be because there is enough headroom in ram by dumb luck.

I'm goign to try and patch this script and see where I end up.

## Environment

Please provide the following details:

- Platform: Kubernetes
- Platform Version: 1.21.12-gke.1500
- PGO Image Tag: registry.developers.crunchydata.com/crunchydata/crunchy-postgres:ubi8-13.7-0
- Postgres Version 13
- Storage: google ssd

lingwooc:
Just linking to the troublesome file. https://github.com/CrunchyData/crunchy-containers/blob/master/bin/postgres_common/postgres/pgbackrest_info.sh
Also it references `/tmp/pgbackrest_env.sh` which does not exist.

andrewlecuyer:
@lingwooc thanks for submitting this issue.

Looking at the PGO code where we enable the exporter using pgMonitor, you can actually see that this is something we'd like to revisit:

[...]

PR #4482, opened 2026-05-13
fix: propagate pgbackrest info failures

immanuwell:
Small fix for `pgbackrest_info.sh`.

Right now the script prints `|`, runs `pgbackrest info`, then always reaches `echo`. So if `pgbackrest` exits non-zero, the helper can still look green. sneaky little footgun.

Repro:

```sh
printf '#!/bin/sh\nexit 42\n' > /tmp/pgbackrest
chmod +x /tmp/pgbackrest
PATH=/tmp /bin/bash components/image-postgres/bin/pgbackrest-info.sh
```

Before: exits `0`.
After: exits `42`.

This is reachable in real clusters: `pgbackrest info` can fail from config/stanza issues, repo storage or network errors, missing binaries in a bad image, or an OOM/signal. No cloud quota edge-case needed here.

Tested:

```sh
go test ./components/... -count=1
```

Related: #3293

### DIFF
--- a/components/image-postgres/bin/pgbackrest-info.sh
+++ b/components/image-postgres/bin/pgbackrest-info.sh
@@ -4,6 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+set -e
+
 printf '|'
 pgbackrest info --output=json --log-level-console=info --log-level-stderr=warn
 echo
--- a/components/image-postgres/bin/pgbackrest-info_test.go
+++ b/components/image-postgres/bin/pgbackrest-info_test.go
@@ -1,0 +1,46 @@
+// Copyright 2026 Crunchy Data Solutions, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package bin_test
+
+import (
+	"os"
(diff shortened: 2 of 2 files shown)


ValClarkson:
Hi @immanuwell,
Thanks for this contributions but this PR needs rebased and updated.

CI failures look mostly mechanical:

vulnerabilities (govulncheck), e2e-k3d-chainsaw, and e2e-k3d-kuttl will all clear with a rebase onto current main — they're all from changes that landed since this PR was opened (x/net v0.54 in #4481, chainsaw/kuttl Go pin revert in #4488).
The golangci-lint failure has 2 real gosec G306 hits on the new test file — os.WriteFile with 0o700 perms (gosec wants ≤0o600). Easiest fix is to write 0o600 then os.Chmod(path, 0o700) so the executable bit is set after creation. That keeps the lint config clean.
Once you rebase + apply the chmod fix, this should be all green.

ValClarkson (changes requested):
See comment

immanuwell:
@ValClarkson done, addressed all required suggestions, PTAL

PR #4535, opened 2026-08-11
fix(collector): honor custom Postgres and PgBouncer ports in OTel metrics configs

andrewlecuyer:
Replaces hardcoded `port=5432` in the generated OpenTelemetry collector receivers with the actual configured ports from the cluster spec for both Postgres and PgBouncer metrics. Also updates the `server` static attribute in Postgres metric queries to reflect the configured Postgres port, and adds tests covering custom ports.

**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

OTel metrics collection does not honor custom PG or PgBouncer ports.

OTel metrics collection honors custom PG or PgBouncer ports.

**Other Information**:

### DIFF
--- a/internal/collector/pgbouncer.go
+++ b/internal/collector/pgbouncer.go
@@ -191,8 +191,8 @@ func EnablePgBouncerMetrics(ctx context.Context, inCluster *v1beta1.PostgresClus
 		config.Receivers[SqlQuery] = map[string]any{
 			"driver": "postgres",
 			"datasource": fmt.Sprintf(
-				`host=localhost dbname=pgbouncer port=5432 user=%s password=${env:PGPASSWORD}`,
-				sqlQueryUsername),
+				`host=localhost dbname=pgbouncer port=%d user=%s password=${env:PGPASSWORD}`,
+				*inCluster.Spec.Proxy.PGBouncer.Port, sqlQueryUsername),
 			"queries": slices.Clone(pgBouncerMetricsQueries),
 		}
 
--- a/internal/collector/pgbouncer_test.go
+++ b/internal/collector/pgbouncer_test.go
@@ -11,7 +11,9 @@ import (
 	"gotest.tools/v3/assert"
 
 	"github.com/crunchydata/postgres-operator/internal/feature"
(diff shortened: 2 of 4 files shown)


ValClarkson:
Nice, focused fix. The four hardcoded port=5432 values and the localhost:5432 in Postgres metric static_attributes.server are correctly threaded through from spec.port / spec.proxy.pgBouncer.port, and the added updateServerAttribute runs after appendToJSONArray, so the version-specific fast-metrics files get rewritten too. Tests pass locally.

A few notes:

* The nil-dereferences of *Spec.Port and *Spec.Proxy.PGBouncer.Port are safe on the reconcile path (cluster.Default() runs first, and NewConfigForPgBouncerPod guards Proxy/PGBouncer), and it matches how we deref Spec.Port everywhere else. Fine.
* updateServerAttribute now runs on every reconcile even without customizations — small perf hit, not worth conditionalizing.
* If any of the three json.Unmarshal calls fail, *MetricsArr stays nil and we end up marshaling null into queries. Same shape as before, so not a regression, but worth a follow-up.
* Test coverage: CustomPort only asserts on sqlquery/5s and the pgBouncer sqlquery. Would be good to also cover sqlquery/300s, a custom-query add, and a per-DB target since those receivers are all touched.
cmp.Contains(datasource, "port=5433") would match port=54331 — tightening to port=5433 user= is more precise.

Nits: the // Sets Port to 5432 comments in tests will go stale if defaults change; and updateServerAttribute could note the port arg is spec.port.
The three red CI checks are unrelated: vulnerabilities is x/net@v0.55.0 / x/crypto/openpgp, and the kuttl failures are the usual pgBackRest stanza-create / namespace-teardown flakes — nothing OTel or port related.

Optional follow-up: the generated JSON still ships "server": "localhost:5432", which reads misleadingly now that it's always rewritten. Could either use a placeholder or just inject the attribute in updateServerAttribute.

ValClarkson (approved):
See comment

PR #4536, opened 2026-08-12
Throttle pgbackrest metrics collection frequency

### LINKED ISSUE #4532: Sharp increase in S3 egress fees with OpenTelemetry metrics enabled
## Overview

Since upgrading to PGO 6.0 and pg18, I've had to switch to OpenTelemetry metrics because the old Exporter is no longer available. It works fine except that I've seen a very steep increase in AWS S3 data egress fees from the pgbackrest buckets I'm using for backup. It's not proportionate to the size of the dataset being backed-up: a very small database is costing around as much in egress as a much larger one. Removing the `instrumentation: {}` spec stops metrics and the egress fees drop back down to the expected near-zero level.

It's too expensive to sustain, so I've had to disable metrics on the pg18 clusters until this problem can be addressed.

Perhaps the `instrumentation: {}` config in a `PostgresCluster` spec could be expanded to allow the user to set a level of backup monitoring which works with the constraints of their egress budget?

## Environment

Please provide the following details:

- Platform: `Kubernetes`
- Platform Version: `1.35.3`
- PGO Image Tag: `ubi9-6.0.2-0`
- Postgres Version: `18`
- Storage: AWS S3

### REPRO

Provide steps to get to the error condition:

1. Add `instrumentation: {}` to a `PostgresCluster` spec which uses postgres 18 and a pgbackrest backup to an AWS S3 bucket.
1. Ensure the PGO has the `OpenTelemetryMetrics=true` feature flag set.
1. Monitor AWS egress cost.

### EXPECTED

1. For data egress and request fees to run at near-zero cost.

### ACTUAL

1. Egress fees from buckets run at several dollars per bucket, per day, which adds up to a significant monthly bill over multiple buckets.

## Logs

The jump in July corresponds with rollout to a subset of our staging pg clusters:

Image

Note the lack of variation in the cost of storage over the same period.

## Additional Information

I can provide any more detail you require on request in a non-public channel.

bendilley:
**Checklist:**

 
**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [x] Documentation
 - [x] Testing enhancement
 - [ ] Other

As documented in CrunchyData/postgres-operator#4532, Open Telemetry monitoring of a pgbackrest S3 store generates steep egress fees. I think the default should have a low impact on egress, and it seems the legacy exporter did implement a throttle for this reason.

For Open Telemetry, implement a similar throttling technique to the legacy exporter, with the same default (10 minutes). The 2nd commit includes a configuration option.

Fixes CrunchyData/postgres-operator#4532

**Other Information**:
I'm not a Go/PGSQL developer so I've used github copilot to vibe-code this change. I hope it helps. Please let me know if there are further changes I can make.

### DIFF
--- a/config/dev/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml
+++ b/config/dev/crd/bases/postgres-operator.crunchydata.com_pgadmins.yaml
@@ -2572,6 +2572,15 @@ spec:
                         items:
                           type: string
                         type: array
+                      pgBackRestInfoThrottleMinutes:
+                        default: 10
+                        description: |-
+                          Minimum number of minutes between pgBackRest info collection runs when
+                          OpenTelemetry metrics are enabled. Lower values update backup metrics more
+                          frequently but can increase cloud egress from object storage-backed repos.
+                        format: int32
+                        minimum: 0
+                        type: integer
                     type: object
                   resources:
                     description: Resources holds the resource requirements for the
--- a/config/dev/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
+++ b/config/dev/crd/bases/postgres-operator.crunchydata.com_postgresclusters.yaml
@@ -13218,6 +13218,15 @@ spec:
                         items:
                           type: string
                         type: array
+                      pgBackRestInfoThrottleMinutes:
+                        default: 10
+                        description: |-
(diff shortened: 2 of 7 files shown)


bendilley:
I believe the first commit would be sufficient on its own btw. The 2nd commit introduces a lot more, but only for the sake of making the throttle configurable.

PR #4538, opened 2026-08-17
fix(collector): add explicit column aliases in replication slot queries

andrewlecuyer:
Adds missing `AS` aliases for the `database`, `conflicting`, `failover`, and `synced` columns in the `pg_replication_slots` queries across all supported PostgreSQL version metric configs and their generated JSON outputs.

**Checklist:**

 
 - [x] Have you added an explanation of what your changes do and why you'd like them to be included?
 - [x] Have you updated or added documentation for the change, as applicable?
 - [ ] Have you tested your changes on all related environments with successful results, as applicable?
   - [ ] Have you added automated tests?

**Type of Changes:**

 
 - [ ] New feature
 - [x] Bug fix
 - [ ] Documentation
 - [ ] Testing enhancement
 - [ ] Other

**What is the current behavior (link to any open issues here)?**

**What is the new behavior (if this is a feature change)?**
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

**Other Information**:

PR #4540, opened 2026-08-17
fix(collector): add explicit column aliases in replication slot queries (REL_5_8)

andrewlecuyer:
Backpatch https://github.com/CrunchyData/postgres-operator/pull/4538 to `REL_5_8`.


tjmoore4 (approved):
Plan to verify fix later, should be low risk due to backpatch.

